PR #33, opened 2014-11-18
PYYaml dependency missing from setup

nickpack:
...dependent on ansible installation method this package may not be installed


nickpack:
@willthames Sorry, only just got to this. Version bumped and commits squashed as requested.

willthames:
It's showing up as two commits here so the squash may not have worked quite right.

`git rebase -i HEAD^^` and change the second pick to a fixup (or squash if you prefer to edit the message) should do the trick, then `git push -f` the result.

nickpack:
@willthames Sorry, was half asleep when I did it - all fixed :)

willthames:
I forgot to ask you to rebase it against latest (which was at 2.0.2) - I changed version to 2.0.3 while merging your commit so no harm done.

I'll update pypi too.

PR #44, opened 2015-01-13
Exclude apt-get build-dep from commands instead of modules test

nickpack:
There is currently no ansible equivalent of apt-get build-dep, so this PR excludes it from the test until this feature appears.


willthames:
I know there are false positives, but I don't really want to add special cases for each one. (You could make your own copy of the modules and use them instead).

I recommend just using the fix detailed in #40 to ignore the false positive.

Issue #47, opened 2015-02-19
Add minimum requirement info


widyono:
Hi, usage of "contains_vars" (in utils.py line 89) means one must have at least ansible 1.8.x.

Could you please insert a note in README.md (for example) stating such minimum requirement? Or, optionally, allow use with ansible 1.7.x by replacing line 89 in lib/ansiblelint/utils.py with:

if "$" in child['path'] or "{{" in child['path']:

willthames:
Fixed in #48

PR #48, opened 2015-02-20
Make backward compatible with ansible 1.7.x

widyono:
See issue #47. To test, use ansible 1.7.x and try to run ansible-lint on examples/include.yml. It will fail due to lack of contains_vars() function definition. This commit simply inlines the singular usage of this function such that one can use ansible-lint with ansible 1.7.x.


willthames:
This seems reasonable to me.

PR #43, opened 2015-01-12
Add Parseable output format

JeanFred:
Add a 'parseable' output which outputs lines as
```
"<filename>:<linenumber>: <rule.id> <message>"
```

- Add class ParseableFormatter
- Add commandline option -p --parseable

Closes #41

Issue #41, opened 2014-12-19
Parseable format


JeanFred:
The quiet output is `[<ERROR_ID>] <FILE>:<LINE>`

I may be mistaken, but I was under the impression that the format `<FILE>:<LINE>: <ERROR_ID>` was more commonly used − eg pep8, pylint or cpplint.

Could the output be changed, or a `parseable` output implemented?

My use case being: I would like to track ansible-lint violations in a Jenkins job, and I would love to *not have* to add a Parser to some plugin. :-)

JeanFred:
(Fixed formatting of first post, with unreadable bits)

willthames:
How about some kind of option that allows output format using various keys - e.g.
```--format "%{file}:%{line}: %{code} %{message}"```
and then passing that to the output. I'm not really sure what other people are doing with output parsing so I wouldn't like to change the default (even if it is more widely used elsewhere)

Note that line number is not generally accurate (or set) except in the `matchline` task. But you probably know that already.

JeanFred:
(Sorry for the noise with the commits and pull-requests)

PR #49, opened 2015-02-23
Add brackets around rule ID in parseable format.

JeanFred:
This formatter was supposed to model after the PEP8 output format,
but was incorrect. The actualy format is:
`"<filename>:<linenumber>: [<rule.id>] <message>"`

Follow-up to #41 and #43.

### DIFF
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -20,7 +20,7 @@ class QuietFormatter:
 class ParseableFormatter:
 
     def format(self, match):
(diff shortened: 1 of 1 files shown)


JeanFred:
(Sorry about this − should have checked my syntax before the first PR >_<)

willthames:
No problem - you're the only user for that format (I'm sure other people will use it in time though).

willthames:
Which reminds me, I should up the version number and push it out to pip

JeanFred:
Thanks!

PR #51, opened 2015-03-05
Remove cp from CommandsInsteadOfModulesRule

danrue:
Ansible's copy module is not analagous to the cp command.

Copy's description is:

    The copy module copies a file on the local box to remote locations. Use
    the fetch module to copy files from remote locations to the local box.

Running 'command: cp', however, is used to copy a file within a remote
system, which copy does not actually support.

I don't think there is a valid case to suggest using 'copy' in lieu of
'command: cp'.


willthames:
Can you put the `unzip` key back on the same line as its value? The split makes it harder to read.

danrue:
Sure thing. I also ran pylint against this file and fixed two minor issues.

willthames:
Great, thanks.

PR #52, opened 2015-03-05
Do not suggest a command when using the raw module

danrue:
According to http://docs.ansible.com/raw_module.html, the raw module has
two purposes:

    - Installing python-simplejson in old python installations
    - Installing python or using ansible without python

Consider the following valid task (which currently fails lint):

    - name: Install python-simplejson
      raw: yum install python-simplejson -y

For these reasons, the raw module should probably be ignored.


willthames:
Good point, well made - thanks.

PR #53, opened 2015-03-06
Fix unicode text in a task causing an AssertionError while parsing

underyx:
I had a non-breaking space in one of my tasks. Ironically, it broke ansible-lint.

As it turns out, the yaml parser used returns a unicode object if there's any unicode characters in a key or a value, not a str one. Pretty easy to fix.

Note that my non-breaking space was there erroneously (in fact, I had 8 spread over my playbook, the OS X keyboard layout makes it too easy to accidentally type one after braces or similar special characters.) It might be worth it to check for unicode characters in the playbook as part of the linting process?

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -186,7 +186,7 @@ def normalize_task(task):
             else:
                 result[k] = v
         else:
-            if isinstance(v, str):
+            if isinstance(v, basestring):
                 v = _kv_to_dict(k + ' ' + v)
             elif not v:
                 v = dict(module=k)


willthames:
I think allowing unicode characters should be a feature, rather than a bug (i.e. there should be no real reason to disallow non breaking spaces or accented characters etc.) I'm surprised this hasn't been spotted earlier. I'd think your editor should be the best place to highlight unexpected unicode characters. 

I'll do some testing of this and it should be merged shortly

underyx:
Alrighty then, I might be a very special case with my clumsy fingers, and the editor highlighting these is a perfectly valid point, too (I actually had that on for a while, but then it broke and I haven't remembered to look into why.)

Thanks for the response!

willthames:
This change is added, along with a test.

Thanks for the fix.

PR #55, opened 2015-03-13
Fix searching the wrong path when encountering dependent roles

underyx:
Code should be pretty self-explanatory. Though, before merging, please confirm that `roles/meta/main.yml` is not a possible case here at all, I haven't heard of any such directory structure, but I might be wrong of course.


MiLk:
:+1: LGTM

willthames:
Thanks for the :+1: @MiLk and thanks for the fix @underyx

PR #56, opened 2015-03-17
Ensure role_path is defined in _look_for_role_files

JeanFred:
The _look_for_role_files method looks for Ansible files
under a possible role_path, as returned by _rolepath().

But this method may return None in some circumstances.

This caused ansible-lint to crash with:
```
  File "ansiblelint/utils.py", line 166, in _look_for_role_files
    thpath = os.path.join(_rolepath(basedir, role), th, 'main.yml')
  File "/usr/lib/python2.7/posixpath.py", line 68, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```
Now we proceed with the files lookup only if role_path is defined.

I hesitated between this and catching the exception, but I decided that this was more explicit. Happy to change either way though.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -31,6 +31,7 @@ from yaml.constructor import Constructor
 
 LINE_NUMBER_KEY = '__line__'
 
+
 def load_plugins(directory):
     result = []
     fh = None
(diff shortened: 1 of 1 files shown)


willthames:
Hi @JeanFred, I think your approach looks fine, but given that you assign the results of _rolepath, it would be worth reusing that assignment in os.path.join too.

JeanFred:
Right @willthames, good catch. Update the PR. (Also fixed two small PEP8 issues in that file).

JeanFred:
Hi @willthames, do you have more concerns on this ? :)

willthames:
No, no further concerns :)

JeanFred:
Thanks for merging! :)

PR #57, opened 2015-04-02
Allow for continuation characters

joefis:
Ansible allows | at the end of a line to indicate that the next
line is a continuation. Current checks for matching brackets
break if you use this because lint considers each line
separately.

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -105,6 +105,7 @@ class RulesCollection(object):
         matches = list()
         with open(playbookfile['path'], 'Ur') as f:
             text = f.read()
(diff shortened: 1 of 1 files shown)


willthames:
This seems fairly straightforward a change, and thus a good candidate for merging - would you be able to add a test for this?

shaleh:
You are probably better off using a regex so you can include arbitrary whitespace between the pipe char and the newline.

willthames:
Matching brackets rule is no longer in ansible-lint so guess this can be closed.

There may still be a need for this, but probably worth reopening with new use cases (probably don't want to use matchline on lines that wrap)

eemz:
ok no worries, thanks folks. eemz.

PR #61, opened 2015-05-27
Add 'legacy' tag for tests

dguerri:
'legacy' tag marks rules that have no longer sense in recent Ansible
code. I.e. using constructs checked by the rule is syntactically wrong
in recent version of Ansible.

'legacy' is like a regular tags, but it is enabled by default.
If needed it can be enable using the -t option.

ANSIBLE0001 (DeprecatedTemplateBracketsRule) is tagged 'legacy'.

### DIFF
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -72,6 +72,9 @@ def main(args):
         parser.print_help(file=sys.stderr)
         return 1
 
+    if 'legacy' not in options.tags:
+        options.skip_tags.append('legacy')
+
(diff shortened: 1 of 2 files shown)


willthames:
To be honest, I'd accept you removing the check altogether. Ansible handles this check better than ansible-lint does.

willthames:
I've removed ANSIBLE0001

willthames:
I've considered your code change for ignoring legacy rules, but I think I'd rather just let people use -x legacy if we have further legacy rules in future.

PR #68, opened 2015-08-14
Tasks/py cleanup

shaleh:
Runs on Py3.4. 
flake8/pep8 cleanups. Allowed for longer lines.
General Python cleanups. [] v. list() etc.

Removed code like the following

```
def foo(arg1, arg2=set()):
    pass
```

leads to nasty bugs. http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments has details.

This is like @kovacsbalu's pull request but I think mine has more changes.

### DIFF
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -20,14 +20,17 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-import ansiblelint.utils as utils
-import ansiblelint.formatters as formatters
-from ansiblelint import RulesCollection
-import ansiblelint
+from __future__ import print_function
+
 import errno
+import optparse
 import os
 import sys
-import optparse
+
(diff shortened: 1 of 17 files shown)


willthames:
I don't really want widespread formatting changes only (I'll accept pylint/pep8/pyflakes improvements as part of features or bug fixes). I do state this in https://github.com/willthames/ansible-lint/blob/master/CONTRIBUTING.md#standards

Certainly improvements such as d882157 would be accepted on their own.

willthames:
5d32e65 might be accepted if I had some context as to what it was breaking.

shaleh:
Your code is not following standard Python idioms. Yes it is your code but it makes it harder for anyone else to contribute.

No offense intended but I took the code as "my first Python app" so I submitted updates to help you use more standard Python. You are welcome to ignore the suggestion. No harm, no foul.

shaleh:
I am making a new pull request to that just does the flake8/py34 update.

PR #78, opened 2015-12-02
modify tokenizer to not split when inside quote or jinja block

careyrp:
Modify the tokenizer to not split and keep the whitespace when inside quotes for jinja blocks.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -75,6 +75,11 @@ def load_plugins(directory):
     return result
 
 
+def contains_quote(str1, list1):
+    for item in list1:
+        if item in str1:
+            return item
+
 def tokenize(line):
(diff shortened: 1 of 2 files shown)


willthames:
Apologies for not reviewing earlier.

This change could use a test case or two, and perhaps further motivation in the commit log.

careyrp:
I'll add a test case later. The problem that this is trying to fix is with tasks that use nest jinja code block with whitespace in the block. Here is an example:

- yum: name={{ item }} state=present
  with_items: my_package_list

The current tokenizer thinks the value for parameter "name" is "{{", and it should be "{{ item }}"

jzafran:
@careyrp what's the status on this?  Are you going to resolve the merge conflicts or should I (and open a new PR)?

PR #79, opened 2015-12-02
Ansible 2.0 Support

careyrp:
Discovered three issues when moving to Ansible 2.0:
1. function parse_yaml_from_file has been replaced by method on DataLoader class
2. function path_dwim has been replaced by method on DataLoader class
3. list VALID_KEYS has been removed

For the functions, added try-catch block on the import. If 1.x ansible import fails, import ansible 2.0, and create parse_yaml_from_file and path_dwim functions that use the DataLoader class
For VALID_KEYS, I copied the definition from older ansible release to ansiblelint/utils.py.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -21,13 +21,21 @@
 import os
 import glob
 import imp
-import ansible.utils
+from ansible.utils import path_dwim
 from ansible.playbook.task import Task
 import ansible.constants as C
 from ansible.module_utils.splitter import split_args
 import yaml
 from yaml.composer import Composer
 from yaml.constructor import Constructor
+
+try:
+    from ansible.utils import parse_yaml_from_file
+except ImportError:
+    from ansible.parsing import DataLoader
+    def parse_yaml_from_file(filepath):
+        dl = DataLoader()
+        dl.load_from_file(filepath)
 
 LINE_NUMBER_KEY = '__line__'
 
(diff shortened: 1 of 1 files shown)


willthames:
I meant to reply to this earlier. I'm just a little worried about the magic VALID_TASKS - that it might become out of date. I had a quick look at v2 ansible and couldn't trivially determine how it's being done any more, but we should probably use that mechanism rather than keep using what will be a legacy set of task arguments.

dmsimard:
Just figured I'd link the related issue this pull request is fixing: https://github.com/willthames/ansible-lint/issues/75

lib/ansiblelint/utils.py
```diff
@@ -21,16 +21,36 @@
...
 import ansible.constants as C
 from ansible.module_utils.splitter import split_args
 import yaml
 from yaml.composer import Composer
 from yaml.constructor import Constructor
 
+try:
+    from ansible.utils import parse_yaml_from_file, path_dwim
+except ImportError:
+    from ansible.parsing import DataLoader
```
pabelanger: this should be:

ansible.parsing.dataloader import DataLoader

pabelanger:
Just tested this locally and with my comments it works with ansible 1.9.4 and 2.0. Any chance we can merge and get a release?

willthames:
Merged with @pabelanger's fix.

Will upload the new module to pypi shortly

willthames:
Thanks for all the hard work!

willthames:
ansible-lint 2.3.0 is now available on pypi

lib/ansiblelint/utils.py
plesiv: `from ansible.parsing.dataloader import DataLoader` works from ansible commit [dcdcd9e9c518eafd533842664fe3c1545793a524](https://github.com/ansible/ansible/commit/dcdcd9e9c518eafd533842664fe3c1545793a524). If someone has odd version of ansible installed (from git).

PR #81, opened 2015-12-04
Add --exclude-paths to ansible-lint

### LINKED ISSUE #80: Feature Request - Add exclude-paths option
I work on a project that is slowly integrating ansible-lint into the standards. We're working on doing this on a per-role basis and what we'd like to do is have a list of paths that we don't check and slowly remove them from that list until we have no more left.

Would the project be open to a command-line option that would allow us to do something like

```
$ ansible-lint --exclude-paths=roles/roleA,roles/roleB,roles/roleC *.yml
```

Perhaps instead something like

```
$ ansible-lint --exclude-roles=roleA,roleB,roleC *.yml
```

I'd be happy to work on implementing this.

sigmavirus24:
This allows users to specify a list of paths they want to exclude from
being checked.

Example usage:

```
$ ansible-lint --exclude-paths examples/lots_of_warnings.yml \
    examples/*.yml
```

Closes #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,22 @@ Usage
 Usage: ansible-lint playbook.yml
 
 Options:
-  --version     show program's version number and exit
-  -h, --help    show this help message and exit
-  -L            list all the rules
-  -q            quieter, although not silent output
-  -r RULESDIR   location of rules directory
-  -t TAGS       only check rules tagged with these values
-  -T            list all the tags
-  -x SKIP_LIST  only check rules whose id/tags do not match these values
+  --version             show program's version number and exit
+  -h, --help            show this help message and exit
+  -L                    list all the rules
+  -q                    quieter, although not silent output
+  -p                    parseable output in the format of pep8
+  -r RULESDIR           add one or more rules directories using one or more -r
+                        arguments. Defaults to ['/Users/icordasc/sandbox
+                        /ansible-lint/lib/ansiblelint/rules'] but any -r flags
+                        completely override this.
+  -t TAGS               only check rules tagged with these values
+  -T                    list all the tags
+  -x SKIP_LIST          only check rules whose id/tags do not match these
+                        values
+  --exclude=EXCLUDE_PATHS
+                        path to directories or files to skip. This option is
+                        repeatable.
 ```
 
 Rules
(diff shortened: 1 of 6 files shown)


cloudnull:
:+1: good change and would be very helpful.

willthames:
I think this would be better if it used `store=append` and then handled the resulting list. 
`--exclude-paths examples/lots_of_warnings.yml --exclude_paths another_file_to_exclude.yml`

willthames:
Feel free to keep the comma separated list as an alternative approach - not sure about the space separated example you gave!

sigmavirus24:
@willthames updated this. The difference is that since each usage is a single path, I changed the name to ``--exclude-path`` instead, e.g., 

```
$ ansible-lint --exclude-path path/to/playbooks/roles/roleA \
               --exclude-path path/to/playbooks/roles/roleB \
               path/to/playbooks/*.yml
```

Make sense?

willthames:
@sigmavirus24 it does make sense, and I wonder if it's worth going a step further and just calling it `--exclude`.

sigmavirus24:
I have no opinion on `--exclude` versus `--exclude-path`. I suppose shorter is easier for people, but they're the same to me.

sigmavirus24:
@willthames since I have no opinion, I leave the decision to you. If you'd rather `--exclude` that will take me 5 seconds.

willthames:
If you change it to `--exclude`, I will merge it :)

sigmavirus24:
@willthames done

willthames:
Thanks @sigmavirus24 
I'll add a couple of other tweaks and push out a new package to pypi shortly

sigmavirus24:
Thanks @willthames!

sigmavirus24:
@willthames when you do get the chance to push the package to PyPI, would you mind pinging me here? I'm not in a rush (so please don't think I'm trying to nag you), I'd just like to remove some of the hacks that we have in place to work around this feature missing as soon as the next ansible-lint is released.

Cheers! :cake:

willthames:
@sigmavirus24 thanks for the friendly reminder. 

2.2.0 is now available from pypi

sigmavirus24:
Woot! Thanks @willthames! Sorry for bugging you on the weekend.

willthames:
@sigmavirus24 you're alright, it was Monday morning here.

Issue #36, opened 2014-12-05
Multi-line open and close curly braces cause [ANSIBLE0003] Mismatched { and } 


maxamillion:
When breaking open and close curly braces over multiple lines like the following:

```yaml
- name: "Set Origin specific facts on localhost (for later use)"
  hosts: localhost
  gather_facts: no
  tasks:
    - name: Setting oo_minion_ips fact on localhost
      set_fact:
        oo_minion_ips: "{{ hostvars
            | oo_select_keys(groups['tag_env-host-type-' + oo_env + '-openshift-minion'])
            | oo_collect(attribute='ansible_eth0.ipv4.address') }}"
      when: groups['tag_env-host-type-' + oo_env + '-openshift-minion'] is defined
```
and
```yaml
 roles:
    - ../../../roles/base_os
    - ../../../roles/repos
    - {
        role: ../../../roles/openshift_master,
        oo_minion_ips: "{{ hostvars['localhost'].oo_minion_ips | default(['']) }}",
        oo_bind_ip: "{{ hostvars[inventory_hostname].ansible_eth0.ipv4.address | default(['']) }}"
      }
    - ../../../roles/pods
```
(From https://github.com/openshift/openshift-online-ansible/blob/master/playbooks/gce/openshift-master/config.yml)

ansible-lint complains that the { and } are mis-matched, it appears to just be a multi-line parse issue.
```
$ ansible-lint playbooks/gce/openshift-master/launch.yml 
[ANSIBLE0003] Mismatched { and }
/home/admiller/src/dev/openshift-online-ansible/playbooks/gce/openshift-master/config.yml:21
        oo_minion_ips: "{{ hostvars

[ANSIBLE0003] Mismatched { and }
/home/admiller/src/dev/openshift-online-ansible/playbooks/gce/openshift-master/config.yml:23
            | oo_collect(attribute='ansible_eth0.ipv4.address') }}"

[ANSIBLE0003] Mismatched { and }
/home/admiller/src/dev/openshift-online-ansible/playbooks/gce/openshift-master/config.yml:35
    - {

[ANSIBLE0003] Mismatched { and }
/home/admiller/src/dev/openshift-online-ansible/playbooks/gce/openshift-master/config.yml:39
      }
```

JeanFred:
Confirmed. In my case multiline curly braces are so common that I had to disable this rule in the linting.

wybczu:
The only solution which comes to my mind (and looks kinda dirty...) is the following:

[...]

PR #84, opened 2015-12-16
Refactor MismatchedBracketRule to inspect parsed YAML instead of raw …

shaleh:
…strings

The check is whether the underlying Jinja2 variables can be parsed not whether the YAML
is valid so use the YAML parser to simplify the test.

Added 2 new text files for testing as well as a unit test specifically targeting
MismatchedBraceRule.


shaleh:
Resolves #36

willthames:
Merged manually with a version update, some file renames (all `.txt` files in `test` have been renamed `.yml`) while retaining attribution

Issue #88, opened 2016-01-18
'become: yes' yields 'Was not expecting value True'


moon-musick:
Hi,

thank you for all your work on `ansible-lint`, it certainly makes our lives a bit easier :)

We recently upgraded our Ansible to 2.0 and started fixing stuff in playbooks / roles due to deprecation warnings.

We are using `ansible-lint` 2.3.1 now.

Apparently the following task:

    - name: clone content repository
      git:
        repo: '{{ archive_services_repo_url }}'
        dest: '/home/www'
        accept_hostkey: yes
        version: master
        update: no
      become: yes
      become_user: nobody
      notify:
      - restart apache2

yields errors like this one:

    Was not expecting value True of type <type 'bool'> for key become
    Task: {'git': {'repo': '{{ archive_services_repo_url }}', 'version': 'master', 'module': 'git', 'dest': '/home/www', 'accept_hostkey': True, 'update': False, '__line__': 62, 'module_arguments': []}, 'name': 'clone content repository', 'become_user': 'nobody', 'become': True, '__line__': 61, 'notify': ['restart apache2']}

The task in question seems to be syntactically valid and works well.

We are using Ubuntu Trusty x86_64 and Python 2.7.6 (standard Ubuntu repo version).

`pip show ansible-lint` says:

    ---
    Name: ansible-lint
    Version: 2.3.1
    Location: /usr/local/lib/python2.7/dist-packages
    Requires: ansible, pyyaml

Is there any more information needed to help resolving this issue?

willthames:
If you have the opportunity to test #89, please do (it should work, as I used your example as a test case). 

Otherwise I can just push 2.3.2 to pypi

spk:
Same problem, PR #89 fix the problem for me.

moon-musick:
If the test procedure is simply to checkout the fix branch and run `ansible-lint troublesome_playbook.yml` then indeed it works now. Thank you :)

willthames:
Thanks @spk and @moon-musick 

I've merged the fix and pushed 2.3.2 to pypi.

spk:
Thanks \o/

PR #89, opened 2016-01-18
Quick and dirty become fix

willthames:
Pragmatic fix for #88.

Proper fix expected within weeks.


willthames:
This change is now live

I've pushed 2.3.2, which contains this fix, to pypi.

PR #91, opened 2016-01-22
Use ansible parsing for module arguments

willthames:
Use ModuleArgumentParser to parse tasks rather than VALID_TASKS

Also update to work on just roles
Some minor coding standards improvements.


willthames:
@moon-musick, @spk this is the proper fix to #88 
Are you able to test that this still meets your use cases. 

@shaleh this is the return to 100 character line length too.

spk:
Hi @willthames, yes this still fix on my case.

careyrp:
This only works with Ansible 2.0+. The ModuleArgumentParser doesn't exist in 1.9.4.

shaleh:
1.9.x is important. 2.0 is too weird for many of us to migrate to just yet.

willthames:
@careyrp @shaleh I'll have to see what I can do to make the fix work with 1.9.

For now it will remain unmerged.

willthames:
@spk @shaleh @careyrp 

This fix now retains the old Ansible 1.9 code path - the new code path is only used for Ansible 2.0+

lib/ansiblelint/rules/MismatchedBracketRule.py
```diff
@@ -1,54 +0,0 @@
...
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-# THE SOFTWARE.
-
-from ansiblelint import AnsibleLintRule
-
-
-class MismatchedBracketRule(AnsibleLintRule):
```
careyrp: Why has MismatchedBracketRule been removed? Does the rule not work with the Ansible 2.0 parser?

willthames: Yeah, that's pretty much exactly it. If you try and use ansible's libraries to read a playbook with a mismatched bracket, it blows up. Given that ansible-playbook won't read it, ansible-lint no longer tries either.

Playbooks passing `ansible-playbook --syntax-check` is an undocumented requirement of ansible-lint (i.e. I don't try and fix bugs that the syntax checker doesn't like)

PR #95, opened 2016-01-27
Dev hack 3599

careyrp:
ANSIBLE1001: pip is not a recommended tool for installing
ANSIBLE1002: easy_install is not a recommended tool for installing
ANSIBLE1003: Yum installing package without explicit version
ANSIBLE1004: Task should have a name

### DIFF
--- a/lib/ansiblelint/rules/EasyInstallRule.py
+++ b/lib/ansiblelint/rules/EasyInstallRule.py
@@ -1,0 +1,13 @@
+import ansiblelint.utils
+from ansiblelint import AnsibleLintRule
+
+
+class EasyInstallRule(AnsibleLintRule):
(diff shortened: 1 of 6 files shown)


careyrp:
Note: Issue with 1003 yum package installed without explicit version. This rule doesn't pickup everything. Ansible-Lint doesn't process with_items or resolve variables, so this rule skips over yum tasks where {{variable}} is used in name attribute.

willthames:
@careyrp why did you close this?

Most of these seem eminently sensible.

willthames:
Some of them are a bit RHEL/CentOS/Fedora specific - can we recommend Debian/Ubuntu alternatives for pip/easy_install. We might also want a similar rule for dnf for modern Fedora and future RHEL, although that could just be the same rule.

PR #96, opened 2016-01-27
add sudo rule

careyrp:
sudo/sudo_user are deprecated commands as of ansible 1.9. Instead playbooks should be using become/become_user. This rule looks for uses of sudo and sudo_user.

### DIFF
--- a/lib/ansiblelint/rules/SudoRule.py
+++ b/lib/ansiblelint/rules/SudoRule.py
@@ -1,0 +1,30 @@
+from ansiblelint import AnsibleLintRule
+
+
+class SudoRule(AnsibleLintRule):
+    id = 'ANSIBLE0008'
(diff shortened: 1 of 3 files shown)


willthames:
This looks good - can we have an associated test for it?

willthames:
In particular, there seem to be two main code paths, where sudo is in a list, and where it's in a dict - it would be good to exercise both paths.

careyrp:
@willthames I created some unit tests for the Sudo rule.

siddharthist:
+1

willthames:
This currently fails flake8. Once it passes, I will merge this, thanks.

careyrp:
flake8 warnings have been fixed

willthames:
Merged, thanks

PR #104, opened 2016-02-04
Add rpm_key as an alternative module to rpm

dmsimard:
The documented way of adding a key is with rpm --import <key.asc>.
This is provided by the rpm_key ansible module and not with yum.


willthames:
Merged, thanks!

willthames:
Pushed as 2.3.5 to pypi

PR #108, opened 2016-02-06
Check if value is iterable before extend

achaussier:
Hello,
Without this test, raises an exception in this example :

    ...
    pre_tasks:
    roles:
      ...

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -272,7 +272,11 @@ def extract_from_list(blocks, candidates):
     for block in blocks:
         for candidate in candidates:
             if candidate in block:
-                results.extend(block[candidate])
+                if isinstance(block[candidate], list):
+                    results.extend(block[candidate])
+                elif block[candidate] is not None:
+                    raise RuntimeError(
+                            "Key '%s' defined, but bad value: '%s'" % (candidate, str(block[candidate])))
     return results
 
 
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/utils.py
```diff
@@ -272,7 +272,11 @@ def extract_from_list(blocks, candidates):
     for block in blocks:
         for candidate in candidates:
             if candidate in block:
-                results.extend(block[candidate])
+                if hasattr(block[candidate], '__iter__'):
```
willthames: presumably `if isinstance(block[candidate], list)` would work equally well - do we expect anything other than a (possibly empty) list?

achaussier: Hello,

Yes, you're right. I update the PR.

achaussier:
Following your comment, I think It's better to return an error if key exists without a list value.

phemmer:
Will this be customizable via a `-x` option? Having an empty list is syntactically valid.

willthames:
@phemmer, I'd be happy to just ignore empty lists rather than raise an error. i.e. treat empty `pre_tasks` the same as no `pre_tasks`.

achaussier:
I've updated to not raise an error if empty value

willthames:
Can you squash the commits into one and then this one will be ready to go.

achaussier:
Done

willthames:
Version 2.3.6 includes this change.

PR #110, opened 2016-02-09
Allow octal permissions of length 4

### LINKED ISSUE #109: False positive detecting OctalPermissionsRule
When trying to setgid permissions

```
[ANSIBLE0008] Octal file permissions must contain leading zero
mode: 02775
```

Looks like the regex requires exactly 3 digits, which is not always correct.

```
# At least an indent, "mode:", optional whitespace, any digits, EOL
mode_regex = re.compile(r'^\s+mode:\s*[0-9]+\s*$')
# Same as above, but with a leading zero before three digits
valid_mode_regex = re.compile(r'^\s+mode:\s*0[0-7]{3}\s*$')
```

siddharthist:
Fixes #109
Improves test coverage and efficiency considerably

### DIFF
--- a/test/octalpermissions-success.yml
+++ b/test/octalpermissions-success.yml
@@ -17,3 +17,13 @@
       file:
         path: foo
         mode: 0123
+
+    - name: octal permissions test success (04000)
+      file:
+        path: baz
+        mode: 04000
+
+    - name: octal permissions test success (02000)
+      file:
+        path: bar
+        mode: 02000
--- a/lib/ansiblelint/rules/OctalPermissionsRule.py
+++ b/lib/ansiblelint/rules/OctalPermissionsRule.py
@@ -33,7 +33,7 @@ class OctalPermissionsRule(AnsibleLintRule):
     # At least an indent, "mode:", optional whitespace, any digits, EOL
     mode_regex = re.compile(r'^\s+mode:\s*[0-9]+\s*$')
     # Same as above, but with a leading zero before three digits
-    valid_mode_regex = re.compile(r'^\s+mode:\s*0[0-7]{3}\s*$')
+    valid_mode_regex = re.compile(r'^\s+mode:\s*0[0-7]{3,4}\s*$')
 
     def match(self, file, line):
         if re.match(self.mode_regex, line):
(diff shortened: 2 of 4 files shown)


test/octalpermissions-success.yml
```diff
@@ -17,3 +17,13 @@
       file:
         path: foo
         mode: 0123
+
+    - name: octal permissions test success (4000)
+      file:
+        path: baz
+        mode: 4000
```
willthames: Shouldn't this be 04000?

siddharthist: @cblakkan Should it be? Maybe I didn't understand the problem fully... Do four digit permissions also need a leading zero?

cblakkan: Yes, it should be 04000. 4000 should trigger a warning. The leading 0 is always necessary to indicate octal format to the parser. 04000 in octal is 2048 in decimal, if a user wrote 4000, it may get interpreted as decimal and end up setting permission to be 4000 (which is bogus), when they intended it to be 2048.

siddharthist: Gotcha, I'll fix the regexp and tests and update this PR. Thanks for explaining it to me!

willthames:
I'm not sure what value TestOctalPermissions is providing

There are more efficient ways to generate all octal numbers
```[str(oct(x)) for x in range(4096)]```
but I think as long as

```
0644
0700
02775
```
are valid and
```
644
755
123
0800
```
aren't we don't need an overwhelming test coverage. The playbook check is much clearer in intent.

siddharthist:
@willthames Is that blocking the merge? I think it's pretty clear what it's doing, I'm happy to add more comments or something. I just think there's no reason not to test every case when it is very feasible (and I already wrote the code).

cblakkan:
I think the YAML tests provide sufficient coverage and are very easy to understand. The less code added (including in tests) the better.

siddharthist:
I've removed test cases, as requested.

cblakkan:
Looks good to me.

willthames:
@siddharthist could you squash the commits for this please. Then we'll be good to merge.

siddharthist:
@willthames Done

willthames:
Ah sorry @siddharthist, the commit summary is not reflective of the actual commit (Honestly not trying to make you do pointless work)

siddharthist:
Fixed again. Thought I had edited that out.

willthames:
Merged, thanks

PR #114, opened 2016-02-24
Unique ID for OctalPermissionsRule

tomdev:
According to the project rules the ID should be a unique identifier. SudoRule already uses `ANSIBLE0008` as its id so I updated the ID of OctalPermissionsRule which was added later on.


mschuchard:
Updating the atom-linter package soon and will be making the assumption in the uipdate that this PR is approved.

willthames:
Thanks for this. I'll get the version bumped too and a new package released.

PR #117, opened 2016-03-03
altering regular expression to also include 'file: ... mode=octal'

mschuchard:
This allows the octal permissions rule to capture

file: ... mode=octal

in addition to the already captured

mode: octal

### DIFF
--- a/lib/ansiblelint/rules/OctalPermissionsRule.py
+++ b/lib/ansiblelint/rules/OctalPermissionsRule.py
@@ -31,9 +31,9 @@ class OctalPermissionsRule(AnsibleLintRule):
     tags = ['formatting']
 
     # At least an indent, "mode:", optional whitespace, any digits, EOL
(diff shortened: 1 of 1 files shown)


willthames:
Actually this is an indication that this rule isn't as well written as it could be.

Really it should use `match_task`, and check for `module=file` and check whether the mode is correctly set in `module_arguments['mode']`. 

Otherwise we'll get some false positives if other modules use a non-octal `mode` argument.

mschuchard:
Did you want to take care of this then?  I can cancel the PR.
I think the mode argument also applies to other modules such as copy, fetch, unarchive, etc. though.

mschuchard:
I took another look at this and I think I understand what you mean about removing false positives in conjunction with getting this check working.

I will change this commit in the near future and re-push.

mschuchard:
Redoing this PR.

PR #129, opened 2016-03-24
altering regular expression to also include 'module: ... mode=octal'

mschuchard:
I am taking another stab at getting the octal permissions rule to function for all relevant situations.  This time I am also rewriting some of the original code to remove false positives as requested.

### DIFF
--- a/lib/ansiblelint/rules/OctalPermissionsRule.py
+++ b/lib/ansiblelint/rules/OctalPermissionsRule.py
@@ -30,11 +30,15 @@ class OctalPermissionsRule(AnsibleLintRule):
         'http://docs.ansible.com/ansible/file_module.html'
     tags = ['formatting']
 
-    # At least an indent, "mode:", optional whitespace, any digits, EOL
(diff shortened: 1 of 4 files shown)


mschuchard:
Ok taking a break from this for now.  If it is obvious to you how to get the Rule 9 octal permissions working then please point it out.

willthames:
Unfortunately `match_task` only gets the already converted number (i.e. it's an integer by the time match_task sees it), whereas `match` operates on the raw lines. 

It's possible to get quite close with `match_task` if you assume that for each of the last three sets of three bits, if the `r` bit is not set but `w` and `x` are, then it's probably wrong. But `4000` decimal converts to `07640` octal, and is therefore one example (of perhaps many) of a false positive.

willthames:
#133 works for all sensible file permissions but someone might want nonsense file permissions.

mschuchard:
Thanks for figuring it out.  I don't know if I would have thought of that.

PR #133, opened 2016-03-29
Octal match task

willthames:
Alternative to #129 - doesn't work for all file permissions but works for all sensible ones

### DIFF
--- a/test/octalpermissions-success.yml
+++ b/test/octalpermissions-success.yml
@@ -27,3 +27,6 @@
       file:
         path: bar
         mode: 02000
+
+    - name: octal permissions test success (0777)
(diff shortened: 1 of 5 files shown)


test/octalpermissions-success.yml
```diff
@@ -13,17 +13,13 @@
...
-
     - name: octal permissions test success (02000)
       file:
         path: bar
         mode: 02000
+
+    - name: octal permissions test success (0777)
+      file: path=baz mode=0777
+
+    - name:  permissions test success (0777)
```
mschuchard: good test idea

mschuchard:
Thanks for fixing my commit.  This is clever.

Issue #118, opened 2016-03-03
Odd line number output behavior from 'MismatchedBracketCheck' rule


mschuchard:
Executing ansible-lint 2.3.8 on example.yml in the examples directory yields:

```
[ANSIBLE0003] mismatched braces
example.yml:2
debug oops a missing {{bracket}
```

Which is many lines above the actual infraction.

Interestingly enough, you can edit the example.yml to discover that the check always returns the line number of the final top-level item.

examples -

```
---
- hosts: webservers







  vars:
    oldskool: "1.2.3"
    bracket: "and close bracket"

  tasks:
  - name: bobbins
    action: command echo ${oldskool}

  - name: blah
    action: debug oops a missing {{bracket}
```

returns line 2

```
---
- hosts: webservers

  - name: blah
    action: debug oops a missing {{bracket}
```

returns line 2

```
---
- hosts: webservers
- foo: foo
- bar: bar
- baz: baz

  - name: blah
    action: debug oops a missing {{bracket}
```

returns line 5

I am unsure what part of the code causes this error, but I have verified the line number returned for this rule is always the final top level item.  This can be verfieid by editing the example.yml in this repository in a similar fashion as the above examples.

mschuchard:
Update to this: ansible now throws an error for mismatched brackets which may render this check pointless.  Note the following output from ansible-lint for the mismatched bracket check in the examples:

```
Traceback (most recent call last):
  File "/usr/local/bin/ansible-lint", line 116, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/usr/local/bin/ansible-lint", line 103, in main
    matches = runner.run()
  File "/usr/local/lib/python2.7/dist-packages/ansiblelint/__init__.py", line 201, in run
    for file in ansiblelint.utils.find_children(arg):
  File "/usr/local/lib/python2.7/dist-packages/ansiblelint/utils.py", line 122, in find_children
    for child in play_children(basedir, item, playbook[1]):
  File "/usr/local/lib/python2.7/dist-packages/ansiblelint/utils.py", line 151, in play_children
    v = template(basedir, v, dict(playbook_dir=os.path.abspath(basedir)), fail_on_undefined=False)
  File "/usr/local/lib/python2.7/dist-packages/ansiblelint/utils.py", line 53, in template
    return templar.template(varname, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/ansible/template/__init__.py", line 349, in template
[...]

PR #107, opened 2016-02-06
Renaming module parameters into __ansible_module__

Yannig:
Avoid collision with action taking module arguments.

Fix for https://github.com/willthames/ansible-lint/issues/106

PR #137, opened 2016-03-31
New rules

willthames:
These commits will form the basis of ansible-lint v3 and include three new default rules

* All tasks should have a name
* Packages do not use `state=latest`
* Commands check whether stuff should/will change using `changes/removes/changed_when`

Let me know if any of these are show stoppers, or if you have feedback

Note that v3 will also include #134, #135 and #136 but they'll probably get into v2.5.x or similar

Notifying common contributors ansible-lint for heads up

@pabelanger 
@mschuchard 
@MiLk 
@siddharthist 
@dmsimard 
@shaleh 
@msabramo


MiLk:
If you want to make a rule throw an error when `state=latest` is used, you might want to throw an error when `state=present` is used without any version specified, because your cluster might end up with different versions of software if you don't provision all the nodes at the same time.

You might want to be able to use `state=latest` for things like `openssl` which should always be in the latest version. So that you don't need to release a new version of your role every time they fix a major issue.

willthames:
@MiLk, you're quite right about `state=present`, the only difficulty will be checking that the packages have a version, particularly with `with_items`, but I'll give it a go.

With openssl, I guess that's a good time to use `skip_ansible_latest` tag.

willthames:
We do have some roles where the specific version of a minor software dependency isn't that important, and look a little like

In particular with tasks that are of the form:

```
- yum: state=present name={{ item }}
  with_items:
  - application-{{ application_version }}
  - minor-dependency{{ "-" ~ minor_dependency_version if minor_dependency_version is defined else "" }}
```
I am not at all sure how we'd check the former has a version, let alone the latter!

MiLk:
Sure. Let's say it's ok and you can use `skip_ansible_lint` for `openssl`.

willthames:
There are now some breaking changes in 3.0.0 as a result of #107 

If you have your own custom modules and use `matchtask`, the keys for the `task['action']` have changed to avoid clashes.

mschuchard:
#118 seems to have already made it into HEAD.

willthames:
These changes have been merged and are in the released version 3.

PR #144, opened 2016-04-06
initial commit

pcastellazzi:
The pull request ensure CommandsInsteadOfArgumentsRule and CommandsInsteadOfModulesRule do not fail on Ansible 1.9.x where the parsing of shell like variable assignation is not processed correctly.

If you input a playbook like this:

```yaml
- host: all
  gather_facts: no
  tasks:
    - shell: PATH=$PATH:/tmp
```

the task object is constructed as follows:

```python
# for ansible 1.9
{
  'action': {
    'PATH': '$PATH:/tmp',
    'module_arguments': [],
    'module': 'shell'
  },
  '__line__': 4
}

# for ansible 2.0
{
  'delegate_to': None,
  'action': {
    '_uses_shell': True,
    'module_arguments': [
      u'PATH=$PATH:/tmp'
    ],
    'module': 'command'
  },
  'shell': 'PATH=$PATH:/tmp',
  '__line__': 4
}
```

Both rules assumed modules_arguments is not empty, but that's not always the case.

### DIFF
--- a/lib/ansiblelint/rules/CommandsInsteadOfArgumentsRule.py
+++ b/lib/ansiblelint/rules/CommandsInsteadOfArgumentsRule.py
@@ -40,7 +40,7 @@ class CommandsInsteadOfArgumentsRule(AnsibleLintRule):
                   'rmdir': 'state=absent', 'rm': 'state=absent'}
 
     def matchtask(self, file, task):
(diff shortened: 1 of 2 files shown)


willthames:
Thanks for the fix.

Issue #145, opened 2016-04-06
ansible-lint fails when using role with embeded modules


mgagne:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible==2.0.1.0
ansible-lint==2.5.0
```

# Desired Behaviour

ansible-lint versions before 2.5.0 used to work fine with role with [embedded modules](http://docs.ansible.com/ansible/playbooks_roles.html#embedding-modules-in-roles).

# Actual Behaviour

ansible-lint fails when ran against a playbook using a role with embedded modules.

```
Traceback (most recent call last):
  File "/Users/mgagne/venv/ansible/bin/ansible-lint", line 125, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/Users/mgagne/venv/ansible/bin/ansible-lint", line 112, in main
    matches = runner.run()
  File "/Users/mgagne/venv/ansible/lib/python2.7/site-packages/ansiblelint/__init__.py", line 222, in run
    skip_list=set(self.skip_list)))
  File "/Users/mgagne/venv/ansible/lib/python2.7/site-packages/ansiblelint/__init__.py", line 134, in run
    matches.extend(rule.matchtasks(playbookfile, text))
  File "/Users/mgagne/venv/ansible/lib/python2.7/site-packages/ansiblelint/__init__.py", line 62, in matchtasks
    for task in ansiblelint.utils.get_action_tasks(yaml, file):
  File "/Users/mgagne/venv/ansible/lib/python2.7/site-packages/ansiblelint/utils.py", line 368, in get_action_tasks
    if 'include' not in task.keys()]
  File "/Users/mgagne/venv/ansible/lib/python2.7/site-packages/ansiblelint/utils.py", line 326, in normalize_task
    return normalize_task_v2(task)
  File "/Users/mgagne/venv/ansible/lib/python2.7/site-packages/ansiblelint/utils.py", line 258, in normalize_task_v2
    action, arguments, result['delegate_to'] = mod_arg_parser.parse()
  File "/Users/mgagne/venv/ansible/lib/python2.7/site-packages/ansible/parsing/mod_args.py", line 304, in parse
    raise AnsibleParserError("no action detected in task. This often indicates a misspelled module name, or incorrect module path.", obj=self._task_ds)
ansible.errors.AnsibleParserError: no action detected in task. This often indicates a misspelled module name, or incorrect module path.
```

cloudnull:
[...]

PR #146, opened 2016-04-06
Fix when roles have embeded modules

cloudnull:
This change updates the v2 task to try and use the module parse method
from ansible 2 and will fall back to the v1 normalization method if it
fails due because of an AnsibleParserError.

Closes-Bug: #145
Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -23,7 +23,7 @@ import imp
 import os
 
 import ansible.constants as C
-from ansible.errors import AnsibleError
+from ansible import errors
 from ansible.module_utils.splitter import split_args
 import yaml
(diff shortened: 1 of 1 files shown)


mgagne:
Works for me now! Thanks!

willthames:
I'll merge this if I can't work out how Ansible detects embedded modules when doing its own parsing (really I want to use Ansible's libraries to do as much as I can)

willthames:
See #145. I propose closing this if it's been fixed with latest v2.7 and v3.0 releases

cloudnull:
Everything looks like its working now. So this PR may no longer be needed. Thank you for everything.

Issue #148, opened 2016-04-07
No module named ansiblelint.__main__; 'ansiblelint' is a package and cannot be directly executed


ssbarnea:
```
No module named ansiblelint.__main__; 'ansiblelint' is a package and cannot be directly executed
```

Python modules that act as utility scripts are supposed to also be loaded using the module approach like ```
ansible-lint --version

python2 -m ansiblelint --version

```
```

ssbarnea:
These are fixed by PR#149

willthames:
Fixed in v3.4.8

ssbarnea:
Thank you!

PR #149, opened 2016-04-07
Implemented console script creation via pip approach, now `ansible-li…

ssbarnea:
Solving #148


willthames:
I'd rather large PRs were discussed before submission. 

The README is not updated to say how to run from source (as `bin/ansible-lint` will no longer work).

This will need rebasing onto the new v3.0 branch, as that'll be the next release. 

Might be worth doing `git mv bin/ansible-lint lib/ansiblelint/__main__.py` as one commit and then making changes afterwards (otherwise I suspect the history will be harder to follow)

willthames:
I think I've now implemented the console_scripts stuff separately. If there are any outstanding improvements remaining in this PR, please rebase them on current master, otherwise I'll close this one.

ssbarnea:
Yep, this PR become obsolete and I will have to create another one if I find some minor bugs that were not fixed yet.

I really hate that github is not able to keep forks in sync with upstream.

PR #167, opened 2016-06-28
Adding colored output formatter

Lowess:
Hi @willthames,

First of all thank you for ansible-lint!

I have worked on adding a colored output (really similar to the one used in the ansible project).

The default remains uncolored by adding ```-c``` will print colored rules.

Feel free to comment this PR, I will make the changes if needed!

### DIFF
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -83,9 +83,6 @@ def main(args):
     if options.parseable:
         formatter = formatters.ParseableFormatter()
 
-    if options.colored:
-        formatter = formatters.ColoredFormatter()
-
     if len(args) == 0 and not (options.listrules or options.listtags):
         parser.print_help(file=sys.stderr)
         return 1
@@ -121,7 +118,7 @@ def main(args):
         matches.extend(runner.run())
 
     for match in matches:
-        print(formatter.format(match))
+        print(formatter.format(match, options.colored))
 
     if len(matches):
         return 2
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -48,9 +48,6 @@ def main(args):
     parser.add_option('-p', dest='parseable',
                       default=False, action='store_true',
                       help="parseable output in the format of pep8")
-    parser.add_option('-c', dest='colored',
-                      default=False, action='store_true',
-                      help="colored output using ANSI escape sequences")
     parser.add_option('-r', action='append', dest='rulesdir',
                       default=[], type='str',
                       help="specify one or more rules directories using "
@@ -72,6 +69,10 @@ def main(args):
     parser.add_option('-x', dest='skip_list', default=[],
                       help="only check rules whose id/tags do not " +
                       "match these values")
+    parser.add_option('--nocolor', dest='colored',
+                      default=hasattr(sys.stdout, 'isatty') and sys.stdout.isatty(),
+                      action='store_false',
+                      help="disable colored output")
     parser.add_option('--exclude', dest='exclude_paths', action='append',
                       help='path to directories or files to skip. This option'
                            ' is repeatable.')
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -48,6 +48,9 @@ def main(args):
     parser.add_option('-p', dest='parseable',
                       default=False, action='store_true',
                       help="parseable output in the format of pep8")
+    parser.add_option('-c', dest='colored',
+                      default=False, action='store_true',
+                      help="colored output using ANSI escape sequences")
     parser.add_option('-r', action='append', dest='rulesdir',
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/utils.py
```diff
@@ -461,3 +461,34 @@ def construct_mapping(node, deep=False):
...
+    'cyan':      u'0;36', 'bright green':   u'1;32',
+    'red':       u'0;31', 'bright cyan':    u'1;36',
+    'purple':    u'0;35', 'bright red':     u'1;31',
+    'yellow':    u'0;33', 'bright purple':  u'1;35',
+    'dark gray': u'1;30', 'bright yellow':  u'1;33',
+    'magenta':   u'0;35', 'bright magenta': u'1;35',
+    'normal':    u'0'
+}
+
+def stringc(text, color):
```
willthames: I would just use ansible's stringc routine (you might need to import it differently for ansible 1.9 and ansible 2.0+)

lib/ansiblelint/formatters/__init__.py
```diff
@@ -26,3 +26,16 @@ def format(self, match):
                                 "E" + match.rule.id,
                                 match.message,
                                 )
+
+from ansiblelint.utils import stringc
```
willthames: as mentioned below, use ansible's stringc rather than reimplementing it here.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -26,3 +26,16 @@ def format(self, match):
...
+
+class ColoredFormatter(object):
```
willthames: Rather than using a new formatter, perhaps update the existing formatters to handle colour if it's switched on.

lib/ansiblelint/formatters/__init__.py
Lowess: Yeah, makes sense indeed. I pushed my changes

lib/ansiblelint/formatters/__init__.py
```diff
@@ -26,3 +26,22 @@ def format(self, match):
...
                                 match.message,
                                 )
+
+
+try:
+    from ansible.color import stringc
+    ANSIBLE_VERSION = 1
+except ImportError:
+    from ansible.utils.color import stringc
+    ANSIBLE_VERSION = 2
```
willthames: don't really need this in this file (it's needed in utils.py due to checks in other methods)

Lowess: Right, so I should group my import of stringc with others in utils.py ?

willthames: Sorry, I realise my response wasn't clear. Just delete the two lines that set ANSIBLE_VERSION.

I mentioned utils.py because that file does need to set ANSIBLE_VERSION. We don't need to here.

willthames:
This PR is failing flake8 tests at the moment. Shouldn't be too hard to tidy them up (you can either run `tox` and run all the tests or `flake8 lib` to just run the flake8 tests - you may need to pip install flake8 if you don't have it already)

lib/ansiblelint/formatters/__init__.py
```diff
@@ -26,3 +26,22 @@ def format(self, match):
...
+
+class ColoredFormatter(object):
```
willthames: Rather than using a new formatter, perhaps update the existing formatters to handle colour if it's switched on.

bin/ansible-lint
```diff
@@ -48,6 +48,9 @@ def main(args):
     parser.add_option('-p', dest='parseable',
                       default=False, action='store_true',
                       help="parseable output in the format of pep8")
+    parser.add_option('-c', dest='colored',
+                      default=False, action='store_true',
```
willthames: I'd favour making the default whether the terminal can handle colour codes perhaps using something like 

```
default=hasattr(sys.stdout, 'isatty') and sys.stdout.isatty()
```

Solution from: http://stackoverflow.com/a/22254892

And then change this to `--nocolor/--nocolour` to switch colour off (I'll assume that the above detection is good enough that no one will want to switch colour on if the autodetection doesn't work)

Lowess:
@willthames , I think I am done with the changes, whenever you have a minute to review feel free to do it

willthames:
Merged, thanks very much.

Lowess:
For you information the colored output is using ansible color utils. 

If you run ansible-lint from a Jenkins freestyle job or a pipeline you can add `export ANSIBLE_FORCE_COLOR=true` in order to keep the log coloration.

zentavr:
@Lowess `export ANSIBLE_FORCE_COLOR=true` didn't work for me, so I pushed #175 PR into the project.

PR #168, opened 2016-06-28
Adding UsingBareVariablesIsDeprecated rule and tests

Lowess:
Hi @willthames it's me again :) ,

I have been working on implementing a new rule for the Ansible WARNING we usually get when we use vars like this in loops:

```{yaml}
---
- hosts: localhost
  become: no
  vars:
    my_list:
      - foo
      - bar

  tasks:
    - name: with_items loop using bare variable
      debug:
        msg: "{{ item }}"
      with_items: my_list
```

Which gives the following output
```{shell}
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks so that the environment value uses the full variable syntax ('{{my_list}}').
This 
feature will be removed in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

I am hoping this new rule can detect this kind of warning.

### DIFF
--- a/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
+++ b/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2013-2014 Will Thames <will@thames.id.au>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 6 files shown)


willthames:
This is great, thank you so much for this. I'm not really sure how the detection of bare variables is working, but your thorough test cases convince me that it is!

willthames:
ansible-lint 3.1.0 released with this change

Lowess:
Wonderful, thanks!

willthames:
Rule ID 0014 was already in use, so I've updated this to 0015 now and released 3.1.1.

I also added a test so this would be spotted at PR time, rather than later!

PR #173, opened 2016-07-08
additional package managers to PackageIsNotLatest and a test for that rule

mschuchard:
More OS and language package managers for the rule and a test for the rule.
[list of ansible package modules](http://docs.ansible.com/ansible/list_of_packaging_modules.html)

I don't think the new test was added to tox based on the TravisCI output. What do I need to do for that?

### DIFF
--- a/lib/ansiblelint/rules/PackageIsNotLatestRule.py
+++ b/lib/ansiblelint/rules/PackageIsNotLatestRule.py
@@ -28,6 +28,10 @@ class PackageIsNotLatestRule(AnsibleLintRule):
                   'with or without a version'
     tags = ['repeatability']
 
+    _package_managers = ['yum', 'apt', 'dnf', 'homebrew', 'pacman', 'openbsd_package', 'pkg5',
+                         'portage', 'pkgutil', 'slackpkg', 'swdepot', 'zypper', 'bundler', 'pip',
+                         'pear', 'npm', 'gem', 'easy_install', 'bower', 'package']
+
     def matchtask(self, file, task):
-        return (task['action']['__ansible_module__'] in ['yum', 'apt', 'package'] and
+        return (task['action']['__ansible_module__'] in self._package_managers and
                 task['action'].get('state') == 'latest')
--- a/test/package-check-failure.yml
+++ b/test/package-check-failure.yml
@@ -1,0 +1,13 @@
+- tasks:
+  - name: install ansible
+    yum: name=ansible state=latest
+
+  - name: install ansible-lint
+    pip: name=ansible-lint
+    args:
+      state: latest
+
+  - name: install some-package
+    package:
+      name: some-package
+      state: latest
--- a/test/package-check-success.yml
+++ b/test/package-check-success.yml
@@ -1,0 +1,14 @@
+- tasks:
+  - name: install ansible
+    yum: name=ansible-2.1.0.0 state=present
+
+  - name: install ansible-lint
+    pip: name=ansible-lint
+    args:
+      state: present
+      version: 3.1.2
+
+  - name: install some-package
+    package:
+      name: some-package
+      state: present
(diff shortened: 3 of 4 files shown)


test/package-check-failure.yml
```diff
@@ -0,0 +1,8 @@
+- tasks:
+  - name: install ansible
+    yum: name=ansible state=latest
+
+  - name: install ansible-lint
+    pip: name=ansible-lint
+    args:
+      state: latest
```
willthames: not sure why state: latest is a key of args.

If you want the full YAML form, and certainly it's preferred these days:

```
  - name: install ansible-lint
    pip:
      name: ansible-lint
      state: latest
```

willthames:
The tests are definitely showing in the Travis output:

```
test_command_changes_negative (test.TestPackageIsNotLatest.TestPackageIsNotLatestRule) ... ok
test_command_changes_positive (test.TestPackageIsNotLatest.TestPackageIsNotLatestRule) ... ok
```

although that does show that the test method names should be changed to something like `test_package_not_latest_negative`

lib/ansiblelint/rules/PackageIsNotLatestRule.py
```diff
@@ -29,5 +29,10 @@ class PackageIsNotLatestRule(AnsibleLintRule):
     tags = ['repeatability']
 
     def matchtask(self, file, task):
-        return (task['action']['__ansible_module__'] in ['yum', 'apt', 'package'] and
-                task['action'].get('state') == 'latest')
+        return (task['action']['__ansible_module__'] in ['yum', 'apt', 'dnf', 'homebrew', 'pacman',
```
willthames: for readability, I'd define a list of package manager modules in a variable and then do the test against that list. You'll have far fewer line breaks.

lib/ansiblelint/rules/PackageIsNotLatestRule.py
```diff
@@ -29,5 +29,10 @@ class PackageIsNotLatestRule(AnsibleLintRule):
...
+                                                         'homebrew_cask', 'openbsd_package',
```
willthames: We need to confirm that all of these package managers actually take a version parameter.

I checked homebrew_cask as an example, and it doesn't take version:
http://docs.ansible.com/ansible/homebrew_cask_module.html

mschuchard: I was checking each of them for a `state` parameter and forgot that I need to verify that each `state` parameter needs to be able to take `latest`.

lib/ansiblelint/rules/PackageIsNotLatestRule.py
mschuchard: Will do.

test/package-check-failure.yml
mschuchard: I was doing this for total testing coverage of any use cases so we know that ansible-lint is picking up any kind of formatting people are doing. I can add a third test for the situation you mentioned.

mschuchard:
New push.

test/package-check-success.yml
```diff
@@ -0,0 +1,13 @@
+- tasks:
+  - name: install ansible
+    yum: name=ansible state=2.1.0.0
+
+  - name: install ansible-lint
+    pip: name=ansible-lint
+    args:
+      state: 3.1.2
```
willthames: this wouldn't work in reality - it should be

```
  pip: name=ansible-lint-3.1.2
```

state can only usually be `present` (typically default) or `latest` (pip has more options but it's latest that is bad)

mschuchard: So this is where my being a polyglot is bad, because I had thought all automation/CM tools could specify versions in that parameter and I forgot Ansible does not do that.

Funny thing is ansible-lint still flags that.

mschuchard:
New push.

test/package-check-success.yml
willthames: Sure - I don't disagree: ansible/ansible-modules-core#2717

mschuchard:
soft ping

willthames:
Merged, thanks (and nice reminder :) )

willthames:
Updated version to 3.2.0 - applied tag and pushed to pypi.

PR #175, opened 2016-07-12
Adding --force-color command line parameter for ansible-lint

zentavr:
We use **ansible-lint** with Jenkins and is does not provide TTY to stdout.
Detecting TTY is being used when you deciding whether or not you want to have colored output.

In order to have colored output, ansible-lint used the `ansible.utils.color::stringc()` which tries to detect the ability to have colored output itself. You can force doing that with setting up
```
export ANSIBLE_NOCOLOR=false
export ANSIBLE_FORCE_COLOR=true
```


bin/ansible-lint
```diff
@@ -73,6 +73,10 @@ def main(args):
                       default=hasattr(sys.stdout, 'isatty') and sys.stdout.isatty(),
                       action='store_false',
                       help="disable colored output")
+    parser.add_option('--force-color', dest='force_colored',
```
willthames: Use `dest=colored` and `action=store_true`, then you don't need the check at lines 91-96.

bin/ansible-lint
```diff
@@ -119,7 +129,7 @@ def main(args):
         matches.extend(runner.run())
 
     for match in matches:
-        print(formatter.format(match, options.colored))
+        print(formatter.format(match, colored_output))
```
willthames: not needed if `--force-color` uses `colored`

lib/ansiblelint/version.py
```diff
@@ -1 +1 @@
-__version__ = '3.1.2'
+__version__ = '3.1.2.1'
```
willthames: Should be `3.1.3`

bin/ansible-lint
zentavr: Ok, can we avoid that users can put both `--nocolor` and `--force-color`

zentavr:
@willthames: did some changes.

bin/ansible-lint
```diff
@@ -69,10 +69,16 @@ def main(args):
     parser.add_option('-x', dest='skip_list', default=[],
                       help="only check rules whose id/tags do not " +
                       "match these values")
+    # We must keep '--nocolor' and '--force-color' in that order, because they store the value in the same destination
+    # and the latest 'default' decides :)
```
willthames: default should remain as before

bin/ansible-lint
```diff
@@ -78,7 +78,7 @@ def main(args):
     parser.add_option('--force-color', dest='colored',
                       default=True,
                       action='store_true',
-                      help="Try force colored output")
+                      help="Try force colored output [default]")
```
willthames: force color should not the default. The default is whatever the running process supports - force color is only needed if the running process (e.g. Jenkins) is not a TTY but does support colour.

zentavr:
@willthames: I reverted the defaults for the colors and now `--nocolor` parameter detects the default value.

willthames:
Merged, thanks - v3.1.3 tagged and released

PR #183, opened 2016-07-29
add block recursion

pcastellazzi:
Tasks nested inside blocks, and nested blocks where not supported. This PR attempts to fix it.


willthames:
Looks good, I'll need a little time to check this out properly, but I'll aim for it in the next week or so.

willthames:
This is available for pre-testing in 3.3.0rc1

PR #184, opened 2016-08-02
implement a check for duplicate keys

larsks:
this will cause parsing to fail when encountering something like:

```
- hosts: localhost
  tasks:
    - debug:
        msg: task 1

  hosts: localhost
  tasks:
    - debug:
        msg: task 2
```

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -40,6 +40,7 @@ class AnsibleLintRule(object):
     match = None
     matchtask = None
     matchplay = None
+    duplicate_keys = 'ignore'
 
     def matchlines(self, file, text):
         matches = []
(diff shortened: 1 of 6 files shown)


larsks:
By way of additional explanation, we were recently bitten by exactly this problem in https://review.openstack.org/#/c/349564/, and it turns out that at the moment it is largely an un-catchable error: neither `ansible` itself nor `ansible-lint` provides a way of flagging duplicate keys as a problem.

There are clear situations in which you *want* duplicate keys, for example, ansible/ansible#13603, but there are other situations where duplicate keys are a clear sign of an error.

conorsch:
Having also been bitten by this (slipped in due to a typo; it should have been two separate plays), i'm in favor of merging!

larsks:
I am cleaning up my stale pull requests. It looks as if this one wasn't of interest.

PR #185, opened 2016-08-03
Don't fail when entering in an except block with Python 3

ramnes:


willthames:
I've merged this into the py3 branch, and rebased py3 on top of latest master (this PR definitely helped).

The py3 branch is now failing on two tests, both of which are due to Ansible not currently supporting python3 (I know that python3 support for Ansible is on the roadmap)

willthames:
Related to #208, looking at getting this into the next release

willthames:
If I can't get #208 included due to incompatibility with ansible 1.9, I'll still try and get this in.

willthames:
Now released with v3.4.1

Issue #174, opened 2016-07-09
ansible-lint exits when using newer role dependency dict structure


sjmiller:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version: 2.1.0.0
ansible-lint --version: 3.1.2
```

- ansible installation method: homebrew
- ansible-lint installation method: pip

# Desired Behaviour

Command should not exit with error when role dependency is in newer dictionary format

# Actual Behaviour (Bug report only)

When a role dependency in `meta/main.yml` is defined like so:
```
dependencies:
  - name: my-role
    src: https://example.com/my-role.git
    scm: git
```
Errors with with:
`role dict {u'src': u'https://example.com/my-role.git', u'name': u'my-role', u'scm': u'git'} does not contain a "role" key`

Defining a role dependency providing the rolespec format in the `role` key appears to be deprecated in ansible 2.1.0.0.

azenk:
I'm seeing the same bug.  This is the ansible code that seems to be used for parsing such declarations.  https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/role/requirement.py#L122  It looks like the logic around parsing these is complicated.   Would it make sense to use the ansible RoleRequirement parser here if it's available?  Obviously that would only work if ansible 2.x were installed.

willthames:
@azenk, `RoleRequirement` is used for the requirements.yml file when installing roles through ansible-galaxy. `RoleDefinition`'s `_load_role_name` is all we need here, and the current code just needs an equivalent to https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/role/definition.py#L116 to work.

willthames:
Pushed test release as 3.3.0rc1

willthames:
Fixed in released 3.3.0

bbaassssiiee:
REGRESSION

Successfully installed MarkupSafe-1.0 ansible-2.3.0.0 ansible-lint-3.4.12 asn1crypto-0.22.0 cffi-1.10.0 cryptography-1.8.1 idna-2.5 jinja2-2.9.6 paramiko-2.1.2 pyasn1-0.2.3 setuptools-35.0.2
demo:buildserver bas$ ansible-lint provision.yml 
role dict {u'src': u'base_java8'} does not contain a "role" or "name" key

PR #188, opened 2016-08-25
Fix handling of new-style deps.

faulkner:
It isn't really documented anywhere, but new-style deps raise a
deprecation warning if you use "role", so it doesn't make sense to raise
a linting exception in this case.

See https://github.com/ansible/ansible/blob/5d865ec1efd529db7eb7d02aefbf60b453123d48/lib/ansible/playbook/role/requirement.py#L147.

I originally had a simpler solution for this but had to flatten it out
with `continue` statements simply to get tox to pass the linting tests.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -42,6 +42,7 @@ except ImportError:
     from ansible.parsing.mod_args import ModuleArgsParser
     from ansible.plugins import module_loader
     from ansible.errors import AnsibleParserError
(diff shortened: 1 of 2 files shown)


willthames:
Should probably add Fixes #174 to the commit

willthames:
There already exists a fix - see #188, but waiting someone to actually test 3.3.0rc2.

willthames:
Closing, fix in released 3.3.0

PR #191, opened 2016-09-01
Bugfix in UsingBareVariablesIsDeprecatedRule for variable with default empty dictionary

jzafran:
Fixes https://github.com/willthames/ansible-lint/issues/190

# What?

Changes `BareVariablesIsDeprecatedRule` to check for bare variables the [same way Ansible 2 does it](https://github.com/ansible/ansible/commit/8716bf8021800a18cb8d6cfea3f296ba4f834692), by checking for  `{%` or `{{`.

# Why?

`BareVariablesIsDeprecatedRule` raises an error when a task has `with_items` with a default empty dictionary:  
```yml
- name: configure
  ini_file:
    option: "{{ item.key }}"
    value: "{{ item.value }}"
  with_dict: "{{ ini_options | default({}) }}"
```

### DIFF
--- a/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
+++ b/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
@@ -19,6 +19,9 @@
 # THE SOFTWARE.
 
 import re
+
+from jinja2.defaults import VARIABLE_START_STRING, BLOCK_START_STRING
+
 from ansiblelint import AnsibleLintRule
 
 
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
```diff
@@ -31,13 +34,17 @@ class UsingBareVariablesIsDeprecatedRule(AnsibleLintRule):
...
 
         if loop_type and isinstance(task[loop_type], basestring):
-                if not self._jinja.match(task[loop_type]):
-                    message = "Found a bare variable '{0}' used in a '{1}' loop." + \
-                        " You should use the full variable syntax ('{{{{{0}}}}}')"
-                    return message.format(task[loop_type], loop_type)
+            # Check if the text after "with_" has a '{%' or '{{' to determine if it
+            # is wrapped correctly for Ansible 2. This is a similar approach to how
+            # it's done in Ansible 2.
+            # see https://github.com/ansible/ansible/commit/8716bf8021800a18cb8d6cfea3f296ba4f834692
```
willthames: That commit seems to bear little resemblance to this change

lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
```diff
@@ -19,6 +19,9 @@
 # THE SOFTWARE.
 
 import re
+
+from jinja2.defaults import VARIABLE_START_STRING, BLOCK_START_STRING
```
willthames: I don't really expect to see a jinja block inside a `with_` loop

lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
```diff
@@ -31,13 +34,17 @@ class UsingBareVariablesIsDeprecatedRule(AnsibleLintRule):
     tags = ['formatting']
 
     _loops = re.compile(r'^with_.*$')
-    _jinja = re.compile("\{\{[^\}]*\}\}")
```
willthames: A much simpler fix is to change `[\}]*` to `.*`

willthames:
I propose splitting this commit into two - one for the tests, one for the change. I'll merge your tests but I'll implement my fix.

willthames:
v3.3.0rc4 now contains the contents of #192 which implements my proposed alternative fix.

If you're able to test, please do

lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
jzafran: Ahh, my bad.  The test actually happens in [`Templar. _contains_vars`](https://github.com/ansible/ansible/blob/e396d5d508ecb9ca7699dc1e7c9f554f51667c8e/lib/ansible/template/__init__.py#L370-L377)  where it checks if `{{`, `{%`, or `{#` is in the string.

lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
jzafran: Good point, I guess we could remove that part of the check.

willthames:
Fixed in 3.3.0

PR #195, opened 2016-09-06
Update installation command

ludovicc:
We need to specify the version of pip to use because on new systems like Ubuntu 16.03, Python3 is the default and 'pip install' will install ansible-lint for Python 3, then a dumb user like myself will report random failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Setup
 
 Using pip:
 ```
-pip install ansible-lint
+pip2.7 install ansible-lint
 ```
 
 From source:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ Setup
 
 Using pip:
 ```
-pip install ansible-lint
+pip2.7 install ansible-lint
```
willthames: Can we just make this `pip2` rather than `pip2.7`. I think that should have the same result.

Just in case anyone is running it on RHEL6.

willthames:
Thanks very much!

PR #197, opened 2016-09-16
Treat NEWLINE as a shell special character

no2a:
It allows us to write a shell task having multiple commands
using literal style. Like this:

```
  - name: collecting and compressing static assets
    shell: |
      set -e
      cd /usr/share/openstack-dashboard
      python manage.py collectstatic --noinput
      python manage.py compress --force
```


willthames:
I'm not sure I've ever wanted to do that, but I also don't have any strong objections to it.

I'd probably write the above

```
- name: collecting and compressing static assets
   shell: python manage.py collectstatic --noinput && python manage.py compress --force
   args:
      chdir: /usr/share/openstack-dashboard
```
but I'm sure there are literal examples that might be easier.

willthames:
Merged, thanks!

PR #199, opened 2016-09-22
New rule: shell task uses pipeline without pipefail

stpierre:
Without the pipefail shell option, errors from commands early in the
pipeline can fail but the pipeline as a whole returns 0, masking the
failures.


willthames:
Is there a bug with ansible for this? This seems like the wrong place to be solving this problem (but, given that it is currently a problem, we should probably alert on it)

stpierre:
Looks like there's https://github.com/ansible/ansible/issues/12903

willthames:
I've given this some thought, and I'm reluctant to include this for now as it would effectively break a lot of people's lint jobs if they upgraded.

I would support adding an 'optional' rules directory that people can point their lint at. I have a few rules myself that I would be reluctant to enforce on others but people might choose to use.

As a side note, [ansible-review](https://github.com/willthames/ansible-review) is far better for this kind of thing precisely because of the flexibility of choosing which rules are warnings, which rules are errors etc.

willthames:
For such an optional rules directory, currently we have `lib/ansiblelint/rules` - perhaps have `lib/ansiblelint/optional` or something that people could choose to point at?

stpierre:
It seems to me that it *should* break people's lint jobs; it's a real bug in loads of shell tasks, and its effects can be pretty serious.

If it's really a concern, perhaps rather than a separate optional rules directory we could just add a new class variable for checks that are disabled by default.

willthames:
I'm not 100% convinced this is always a problem

For example if I do

```
shell: rpm -ql doesnotexist | grep filename
```

it really doesn't matter to me if it fails in the rpm bit or the grep bit - if the rpm bit fails, the grep will fail anyway.

JeanFred:
I don’t have much opinion on this particular rule, but wanted to comment on this:
> I've given this some thought, and I'm reluctant to include this for now as it would effectively break a lot of people's lint jobs if they upgraded.

I think this is normal behaviour − one expects releases to include new rules that they may not follow, no?
(For example just yesterday I upgraded from 2.7 to 3.3, which makes 5 additional rules. I had violations all over the place for one of them that I could not fix right away, so I muted it in my tox job and that’s it − no big deal :)

stpierre:
There are indeed cases where `pipefail` isn't necessary, but they're fairly marginal and easy to break. Add another command to the end of that pipeline, or change it to `grep -v`, and suddenly `pipefail` is necessary.

I think we can trust the user who really doesn't want to `set -o pipefail` in that situation to recognize it and tag the task `skip_ansible_lint`. (Personally, I think that user is doing some pretty fiddly microoptimizations that will be counterproductive in the future, but they do have the power they need to skip it.)

gundalow:
@willthames Something for v4?

awcrosby:
I resolved merge conflicts on rebase, then did minor refactor of the rule and consolidated test case into one.

PR #201, opened 2016-09-28
Improve bare variable detection

stpierre:
This avoids false negatives in all of the loop types that took
multiple loops, and false positives in several other loop types as
well.


willthames:
Merged, thanks. I'll get this into 3.3.1

PR #202, opened 2016-09-29
Fix missing filename in exceptions

stpierre:
Without this, exceptions raised when parsing tasks uniformly report
"Unknown" as their filename and line number.


willthames:
Do you have an example? I'd have thought this would have been noticed earlier!

stpierre:
https://gist.github.com/stpierre/4f9891b521e7e61b74dc01e24107e103

willthames:
Weirdly, the Ansible 1.9 tests now fail as a result of this change.

willthames:
Fixed the Ansible 1.9 with an additional commit.

willthames:
v3.3.3 has been released to include this change

PR #203, opened 2016-09-29
fix regression in TestUsingBareVariablesIsDeprecated

jzafran:
Fix regression introduced in https://github.com/willthames/ansible-lint/pull/201 that occurs when `with_together` is used with a single variable (instead of a list):

```yml
    - name: with_flattened loop with a variable
      debug:
        msg: "{{ item }}"
      with_flattened: "{{ some_list }}"
```

This was failing ansible-lint with:

```
ANSIBLE0015 Found a bare variable '{{ some_list }}' used in a 'with_flattened' loop. You should use the full variable syntax ('{{{{ some_list }}}}')
```

### DIFF
--- a/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
+++ b/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
@@ -33,12 +33,18 @@ class UsingBareVariablesIsDeprecatedRule(AnsibleLintRule):
     _jinja = re.compile("\{\{.*\}\}")
 
     def matchtask(self, file, task):
-        loop_type = next((key for key in task.keys()
+        loop_type = next((key for key in task
                           if key.startswith("with_")), None)
-
         if loop_type:
             if loop_type in ["with_nested", "with_together", "with_flattened"]:
-                for var in task[loop_type]:
+                # These loops can either take a list defined directly in the task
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
```diff
@@ -33,17 +33,23 @@ class UsingBareVariablesIsDeprecatedRule(AnsibleLintRule):
...
+                items = task[loop_type]
+                if not isinstance(items, (list, tuple)):
+                    items = [items]
+                for var in items:
                     return self._matchvar(var, task, loop_type)
             elif loop_type == "with_subelements":
                 return self._matchvar(task[loop_type][0], task, loop_type)
-            elif loop_type in ["with_sequence", "with_ini",
-                               "with_inventory_hostnames"]:
+            elif loop_type in ("with_sequence", "with_ini",
```
willthames: Is there a reason for this change? This seems as much personal preference as anything.

jzafran: reverted

willthames:
Merged, thanks

willthames:
v3.3.3 released to include this fix.

PR #208, opened 2016-10-18
Fix Python 3 compatibility

bochecha:
With this, I can run the py3 branch just fine in a Python 3 virtual environment, and check my ansible stuff.

### DIFF
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -24,13 +24,11 @@ from __future__ import print_function
 
 import errno
 import optparse
-import os
 import sys
 
 import ansiblelint
 import ansiblelint.formatters as formatters
 from ansiblelint import RulesCollection
-import ansiblelint.utils as utils
 from ansiblelint.version import __version__
 
 
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -28,6 +28,7 @@ import sys
 
 import ansiblelint
 import ansiblelint.formatters as formatters
+import six
 from ansiblelint import RulesCollection
 from ansiblelint.version import __version__
 
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -25,7 +25,15 @@ import os
 
 import ansible.constants as C
 from ansible.errors import AnsibleError
-from ansible.module_utils.splitter import split_args
+
+try:
+    # Try to import the Ansible 2 module first, it's the future-proof one
+    from ansible.parsing.splitter import split_args
+
+except ImportError:
+    # Fallback on the Ansible 1.9 module
+    from ansible.module_utils.splitter import split_args
+
 import yaml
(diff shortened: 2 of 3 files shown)


bochecha:
One thing I forgot to note, is that this works with the devel branch of ansible, as the latest release doesn't work with Python 3 at all.

bochecha:
I just tested it, and it also works with the (very) recent ansible 2.2.0.0-0.2.rc2 release.

willthames:
Why are the travis tests failing?

lib/ansiblelint/utils.py
```diff
@@ -25,7 +25,7 @@
 
 import ansible.constants as C
 from ansible.errors import AnsibleError
-from ansible.module_utils.splitter import split_args
+from ansible.parsing.splitter import split_args
```
willthames: Does this work with ansible 1.9? Not sure I'm totally ready to abandon ansible 1.9 yet (unless Ansible themselves have)

bochecha:
> Not sure I'm totally ready to abandon ansible 1.9 yet

Ansible 1.9 isn't compatible with Python 3.

Ansible 2.2 will pretty much be required to get ansible-lint working with Python 3.

I guess we could wrap the import into a `try`/`except` block, but that means the fallback case (`module_utils.splitter`) for ansible 1.9 would only work on Python 2.

It might not be a problem, since people who, like me, want to use ansible and ansible-lint with Python 3 would probably use the latest version of ansible anyway?

willthames:
That sounds like a sensible approach. I'm fine with python 3 only working
with latest ansible (you can't really do any more than that) and allowing
ansible-lint to still work with ansible 1.9 and python 2. There are a lot
of try/import/except hacks already to allow compatibility, one more won't
hurt!

bochecha:
One more thing to note: `ansible.parsing.splitter` is GPLv3, which might not be what you want for ansible-lint. (I'd be happy with it, but that's not for me to decide)

I'm not sure how upstream is going to resolve this though, see https://github.com/ansible/ansible/pull/18100

willthames:
The license issue is an interesting one. Put it this way, using ansible.parsing.splitter will make things no worse (as we already use ansible.parsing.mod_args for example).

willthames:
Now released with v3.4.1

bin/ansible-lint
```diff
@@ -24,13 +24,12 @@ from __future__ import print_function
 
 import errno
 import optparse
-import os
 import sys
 
 import ansiblelint
 import ansiblelint.formatters as formatters
+import six
```
willthames: what do we actually do with this?

willthames:
I've just realised I've added some more python2 things (`if isinstance(x, basestring)`) so will need the six module.

Once ansible 2.2 comes out I will see about adding a python3 test to catch this stuff automatically

PR #209, opened 2016-10-19
Use the code in ansible.parsing rather than module_utils

mscherer:
The code in module_utils seems to not be python 3 compliant ( Bochecha on irc showed me traceback ) and is duplicated between the 2 modules. So I am trying to clean that upstream in ansible.


bochecha:
Note that I had done that in #208 (among other things)

mscherer:
Yeah, I will close it and comment on yours

PR #211, opened 2016-10-26
Only trigger modules-instead-of-commands for specific VCS subcommands

stpierre:
git, hg, and svn do lots and lots of things, but their modules can
only do some of those things. Only trigger the
CommandsInsteadOfModules rule when those are being used to clone/check
out.

This does introduce some false negatives, e.g. where someone does 'git
--no-pager clone ...', but false negatives in this case are preferable
to false positives.

### DIFF
--- a/lib/ansiblelint/rules/CommandsInsteadOfModulesRule.py
+++ b/lib/ansiblelint/rules/CommandsInsteadOfModulesRule.py
@@ -35,8 +35,9 @@ class CommandsInsteadOfModulesRule(AnsibleLintRule):
     tags = ['resources']
 
     _commands = ['command', 'shell']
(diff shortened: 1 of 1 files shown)


willthames:
We have `warn: no` in the `command` module for this reason. I don't want to
overcomplicate this rule with special cases

stpierre:
These aren't special cases, though -- they're the only cases this module should care about. The vast majority of `git` (and `svn`, and so on) subcommand simply aren't relevant here. It doesn't make sense to continue producing false positives for the sake of avoiding the "overcomplication" of a 45-character change.

stpierre:
Okay, after looking at this a bit more I see your point. We don't want to just disable warnings in ansible-lint for these commands, but also in Ansible. It'd sure be nice if Ansible's warnings were smarter than they are, but they aren't, so we need a solution -- even if it's manual and dumb -- that works in both, and mine doesn't do that.

PR #212, opened 2016-10-26
Handle when: conditions that are not strings

stpierre:
It may be silly to specify a boolean as a when condition, but it's
legal, and we shouldn't raise an exception when it happens.


willthames:
This prompts me that the code doesn't work right for `when` lists either -
should probably do more cases based on type.

willthames:
Now released with v3.4.1

PR #216, opened 2016-11-02
Add globbing, generator characters to shell check

stpierre:
The shell globbing (*?[]) and generator ({}) characters need to be
added to UseCommandInsteadOfShell so that we don't flag actions that
use those shell functions as needing to be converted to command.


willthames:
Nice, like the obvious-in-hindsight change from character array to string for the iterator!

PR #218, opened 2016-11-05
Fix Python 3 compatibility

bochecha:
Here's one more Python 3 compatibility pull request.

Note that I'm basing this on master, which reports to be 3.4.0, while the latest release is 3.4.1. Did you forget to push something? :smile: 

Let me know if I should rebase this.

I tested this with ansible 2.2.0.0 which itself is compatible with Python 3.

### DIFF
--- a/bin/ansible-lint
+++ b/bin/ansible-lint
@@ -107,9 +107,9 @@ def main(args):
         print(rules.listtags())
         return 0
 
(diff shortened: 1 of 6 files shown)


willthames:
I've rebased this on top of 3.4.2 and added python3/ansible2.2 tests to Travis. 

This is now released as the main change in 3.4.3.

Thanks!

Issue #165, opened 2016-06-27
[ANSIBLE0012] Handlers should be allows to run shell/command entries without when


robinro:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

ansible --version: 2.1
ansible-lint --version: 3.0.1

- ansible installation method: source
- ansible-lint installation method: pip

# Desired Behaviour

ANSIBLE0012 checks for usage of `shell` or `command` without `when`, `changed_when`, ...
While this represent good practice for tasks, it is not needed for handlers, which have an implicit `when: original_task|changed`.

Thus handlers should be excluded from this check.

# Actual Behaviour (Bug report only)

[ANSIBLE0012] Commands should not change things if nothing needs doing
.../roles/foo/handlers/main.yml:3
Task/Handler: run shell command

willthames:
Yes, you're quite right - expect a patch in the next day or so.

robinro:
Thanks.

btw: ansible-lint is a great tool! We started running it in our CI setup and it gave some good hints already.

willthames:
3.1.0rc1 released for testing purposes

willthames:
If you like ansible-lint, you may also like ansible-review (I spotted this bug in ansible-review a while back but because you can tell ansible-review what types of files to run a lint check against, I could avoid the bug at the ansible-review end - I didn't think to fix it here at that time)

robinro:
Thanks for the fix!

PR #221, opened 2016-11-22
fixup for #165 - unconditional commands in handlers

MartinNowak:
- get correct type of included tasks from playbook
  section containing the include

Also see #165 and d9a6aa08d347a8b48418f379ea5a42078608be08.


MartinNowak:
Seems like this was cherry-picked as ae8a8d3bef17b2cad888517b38b16f29b0273521, weird.

PR #231, opened 2017-01-11
Add deprecated tag to ANSIBLE0015

anthcourtney:
ANSIBLE0015 is currently tagged as ```formatting``` however it also reflects identification of a deprecated syntax.

I believe it should also be tagged as ```deprecated```, given that its error:

```
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks so that the environment 
value uses the full variable syntax ('{{people}}').
This feature will be removed in a future release. 
Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

is similar to the error for ANSIBLE0008 which is tagged as ```deprecated```.

```
[DEPRECATION WARNING]: Instead of sudo/sudo_user, use become/become_user and make sure become_method is 
'sudo' (default).
This feature will be removed in a future release. Deprecation warnings can be disabled 
by setting deprecation_warnings=False in ansible.cfg.
```


willthames:
Merged, thanks!

Issue #220, opened 2016-11-15
always_run is deprecated in Ansible 2.2


jbscare:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

- ansible 2.2.0.0
- ansible-lint 3.4.3
- ansible installation method: source
- ansible-lint installation method: source

# Desired Behaviour

always_run is deprecated in Ansible 2.2, so it'd be nice if ansible-lint caught it and complained.

willthames:
This is quite a simple rule to write if you'd like to write it - you're basically checking for

```
'always_run' in task
```
using the `matchtask` check - not hugely different to `TaskHasNameRule.py`

willthames:
v3.4.10 is now available with this fix.

PR #232, opened 2017-01-12
Add rule for always_run which is deprecated in ansible v2.2

anthcourtney:
As per #220, ```always_run``` is deprecated in ansible v2.2 and has been replaced by ```check_mode```.

### DIFF
--- a/lib/ansiblelint/rules/AlwaysRunRule.py
+++ b/lib/ansiblelint/rules/AlwaysRunRule.py
@@ -1,0 +1,31 @@
+# Copyright (c) 2017 Anth Courtney <anthcourtney@gmail.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


lib/ansiblelint/utils.py
```diff
@@ -76,8 +76,8 @@ def ansible_template(basedir, varname, templatevars, **kwargs):
     'name', 'action', 'when', 'async', 'poll', 'notify',
     'first_available_file', 'include', 'tags', 'register', 'ignore_errors',
     'delegate_to', 'local_action', 'transport', 'remote_user', 'sudo', 'sudo_user',
-    'sudo_pass', 'when', 'connection', 'environment', 'args',
-    'any_errors_fatal', 'changed_when', 'failed_when', 'always_run', 'delay', 'retries', 'until',
+    'sudo_pass', 'when', 'connection', 'environment', 'args', 'always_run',
+    'any_errors_fatal', 'changed_when', 'failed_when', 'check_mode', 'delay', 'retries', 'until',
```
willthames: No need for this change, as this is only used for pre 2.0 code (2.0 onwards uses ModuleArgsParser so that I can rely on ansible's handling)

lib/ansiblelint/rules/AlwaysRunRule.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2016 Will Thames <will@thames.id.au>
```
willthames: I didn't write, this, you did (thanks!)

lib/ansiblelint/utils.py
willthames: Actually, you still want to know if your code base is deprecated, even if running on an old version of ansible. Ignore this review item!

willthames:
This is all good, just fix the copyright message and I'll merge.

lib/ansiblelint/rules/AlwaysRunRule.py
anthcourtney: Updated.

willthames:
Merged, thanks!

PR #248, opened 2017-03-03
Move skip_ansible_lint check into get_normalized_tasks()

pabelanger:
It is possible we to skip invalid tasks, becase they are custom and
the local environment may not have the plugins installed properly. As
a result, move the skip check into get_normalized_tasks() and skip the
norimalize process if skip_ansible_lint exists.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>


willthames:
3.4.12 tagged and pushed to pypi

pabelanger:
Thanks, it seems pypi didn't update properly. https://pypi.python.org/pypi/ansible-lint/3.4.12

willthames:
My new ~/.pypirc wasn't correct - it was pointing at www.python.org/pypi and telling me it was all 200 OK - when I changed it to pypi.python.org/pypi, the upload seems to have taken.

Issue #250, opened 2017-03-18
[ANSIBLE0015][UsingBareVariablesIsDeprecatedRule] False positive when multiline jinja template


gzm55:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version
ansible 2.2.1.0

ansible-lint --version
ansible-lint 3.4.11
```

- ansible installation method: OS package
- ansible-lint installation method: pip

# Desired Behaviour

For multiline template in `with_flattened` and left newline chars in the multi line parsing result, rule ANSIBLE0015 is always not happy:

```pl.yml
---
- hosts: all
  tasks:
  - name: test
    debug:
    with_flattened: >
      {{  'a',
          'b' }}
```

and runing `ansible-lint pl.yml` should be successful.

# Actual Behaviour (Bug report only)

```
[ANSIBLE0015] Found a bare variable '{{  'a',
    'b' }}
' used in a 'with_flattened' loop. You should use the full variable syntax ('{{{{  'a',
    'b' }}
}}')

```

PR #251, opened 2017-03-18
UsingBareVariablesIsDeprecatedRule supports multiline templates

gzm55:
should fix issue #250


gundalow:
@willthames What are you thoughts on this?

gzm55:
@willthames @gundalow 
fix the conflicts. is there anyone have time to review this pr?

tux21b:
Please merge, we are hitting the same bug (issue #250) after updating to ansible-lint 4.1 and there seems to be currently no way to make the linter happy:

If we avoid newlines, we will trigger the maximum line length check and if we do not, we will trigger this check because it does not support newlines.

awcrosby:
We'll plan on getting this in the next release

PR #253, opened 2017-03-22
Run UseHandlerRatherThanWhenChangedRule only for tasks

vegai:
The unit test is false positive at the moment for some reason I don't fully understand: it succeeded also before the actual code change, and the actual code change fixed the actual issue I was having.


vegai:
Ah, figured it out. I had forgotten to register the rule under test in my test. Seems quite ok now.

willthames:
Can we give test/252/main.yml a more meaningful name?

vegai:
Now like this:

```
$ tree test/role-with-handler
test/role-with-handler
├── a-role
│   └── handlers
│       └── main.yml
└── main.yml

2 directories, 2 files
```

willthames:
Sorry, but this is going to need far better justification.

`matchtask` absolutely should run on handlers as for most linting purposes, tasks and handlers should be linted equivalently. Obviously there are some rules that make sense only for tasks (e.g. checking changed status) but there is code in those rules to enforce that (e.g. https://github.com/willthames/ansible-lint/blob/master/lib/ansiblelint/rules/CommandHasChangesCheckRule.py#L36) - if there are further similar issues, that is likely how they should be addressed.

vegai:
Thanks for taking the time to review this. I obviously didn't have a good view of the whole. 

I'll change the implementation to be less drastic.

vegai:
Removed the bad change from __init__ and put a similar one in UseHandlerRatherThanWhenChangedRule instead.

willthames:
Sorry for leaving this so long after you fixed the initial problem, this change looks great, thank you.

Issue #257, opened 2017-04-12
Instructions for installing from source


udondan:
Just wanted to install ansible-lint from source and noticed the instructions in the README are not accurate:

```
git clone https://github.com/willthames/ansible-lint
export PYTHONPATH=$PYTHONPATH:`pwd`/ansible-lint/lib
export PATH=$PATH:`pwd`/ansible-lint/bin
```
There is no `bin` directory. I now just symlinked `ansible-lint/lib/ansiblelint/main/__init__.py` into my users bin dir.

    ln -s ansible-lint/lib/ansiblelint/main/__init__.py $HOME/bin/ansible-lint

udondan:
Which of course doesn't work. :) What would be the correct way of installing from source (without using easy_install)?

vegai:
Have you tried 

```
python setup.py install --user
```

willthames:
See also #254 and #259

udondan:
@vegai I'm fine with the symlink approach. Just docs need to be updated which will be covered with #259. Thanks

PR #259, opened 2017-04-24
Documentation update for running ansible-lint out of a clone

### LINKED ISSUE #234: since update to version 3.4.8 there is no bin/ansible-lint script in the repo
# Issue Type
- Bug report

```
ansible --version
    ansible 2.2.1.0 (detached HEAD 5362910000) last updated 2017/01/17 05:43:47 (GMT +200)
ansible-lint --version
    3.4.10
```

- ansible installation method: source
- ansible-lint installation method: source

# Desired Behaviour

Please provide bin/ansible-lint again.

since Update to version 3.4.8 there is no bin/ansible-lint script in the repo.
From source installation doesn't work anymore.

### LINKED ISSUE #254: Please provide binary to run ansible-lint from clone directly
# Issue Type
- Bug report

# Desired Behaviour

From the README.md, your document how to use bin/ansible-lint direct from checkout. That is a pretty useful way to call it.

However, in 828925978cc6e1e53a499fdd2a8c3f4993e2b102, you deleted that tool so it is not possible anymore to call it without poluting the own system with custom python stuff.

(For the records, "pip" is very, very bad think for my oppinion.)

jbscare:
Fixes #234.
Fixes #254.
(Well, works around them anyway.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ From source:
 git clone https://github.com/willthames/ansible-lint
 export PYTHONPATH=$PYTHONPATH:`pwd`/ansible-lint/lib
 export PATH=$PATH:`pwd`/ansible-lint/bin
+cd ansible-lint
+mkdir bin
+ln -s ../lib/ansiblelint/main/__init__.py bin/ansible-lint
 ```
 
 Usage


README.md
```diff
@@ -20,6 +20,9 @@ From source:
 git clone https://github.com/willthames/ansible-lint
 export PYTHONPATH=$PYTHONPATH:`pwd`/ansible-lint/lib
 export PATH=$PATH:`pwd`/ansible-lint/bin
+cd ansible-lint
+mkdir bin
+ln -s ../lib/ansiblelint/main/__init__.py bin/ansible-lint
```
pabelanger: IMO, this is the wrong approach, has it now makes your git repo dirty.  I think the proper way to do this is install ansible-lint into a virtualenv:

```
pip install virtualenv
virtualenv ansible-lint-venv
./ansible-lint-venv/pip install ./ansible-lint
./ansible-lint-venv/bin/ansible-lint
```

more commands, but now you have compiled version of ansible-lint installed into sandboxed virtualenv

jbscare: True, although if you're fundamentally averse to pip, you won't want to get virtualenv that way. (But you might be able to get it another way, e.g. Amazon Linux has a python27-virtualenv package.)

It probably depends what you're trying to do. If you really don't want to touch anything whatsoever on your system other than a git clone in your home directory (maybe you're using Ansible as a non-privileged user, who knows), a symlink is a way to run it out of a clone without touching anything else. I agree that a virtualenv is probably better for a lot of people; and, as @willthames suggested in #254, if you like package managers and don't like pip, you may be able to install it via yum or apt or whatnot.

I don't know how much @willthames wants to clutter the docs with alternatives; because there's *currently* instructions for running directly out of a clone, and at least two people were using it that way before, that seems like a reasonable thing to keep instructions for. But I don't personally mind if folks want to include virtualenv, yum, apt, etc, as additional options.

willthames:
Also addresses #257 

However, I think my preferred fix is to just provide a minimal bin/ansible-lint. The reason bin/ansible-lint was removed was to improve pip packaging, but if bin/ansible-lint isn't referred to in setup.py, we should get away with a bin/ansible-lint along the lines of

```
import ansiblelint
import sys

sys.exit(ansiblelint.main(sys.argv))
```

or something (improvements welcome)

jbscare:
I think bd7518dc7cf32718993ff7bfea4636a0e4672adf obsoletes this PR, so I'm closing it.

PR #284, opened 2017-09-28
Handle the 'cmd' argument in command/shell

j2sol:
At some point, ansible grew a hidden argument to shell/command tasks,
the 'cmd' argument. History found at
https://github.com/ansible/ansible/issues/12856

Our project is making use of the cmd argument so that our shell bits do
not get jinjad, so that we can use a here doc in the command.
Unfortunately this was tripping up various tests in ansible-lint, and we
had to ignore the files, which let other errors creep in. Thus this
change adds the type of syntax we use to the test yaml files and updates
the tests to propery handle them.

Signed-off-by: Jesse Keating <jkeating@j2solutions.net>

### DIFF
--- a/lib/ansiblelint/rules/CommandsInsteadOfArgumentsRule.py
+++ b/lib/ansiblelint/rules/CommandsInsteadOfArgumentsRule.py
@@ -47,9 +47,14 @@ class CommandsInsteadOfArgumentsRule(AnsibleLintRule):
                   'rmdir': 'state=absent', 'rm': 'state=absent'}
 
     def matchtask(self, file, task):
-        if task["action"]["__ansible_module__"] in self._commands and \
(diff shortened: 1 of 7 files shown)


willthames:
Thanks @j2sol, this looks great! I had no idea about the `cmd` argument!

willthames:
I'll look at releasing this this morning.

j2sol:
Thanks for the quick merge! The `cmd` thing is odd, because it's not really documented outside of that issue that gets linked to occasionally. Maybe I can convince upstream to document it.

willthames:
If you write the PR for upstream, I'll happily merge that too :)

willthames:
I actually think the `cmd` argument should be the better practice, tempted to add a rule for `ansible-review` to enforce it (but not ansible-lint, as not using it isn't a terrible thing)

Issue #285, opened 2017-09-29
Included tasks via include_tasks or import_tasks will not be checked


MatrixCrawler:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible 2.4.1.0 (stable-2.4 adc3bedaa2) last updated 2017/09/29 11:28:30 (GMT +200)
ansible-lint 3.4.15
```

- ansible installation method: source
- ansible-lint installation method: pip

# Desired Behaviour

If i iclude a subtask with import_tasks or include_tasks i expect that ansible-lint would check those files.

# Actual Behaviour (Bug report only)

When i am using the _include_tasks_ or _import_tasks_ directive instead of _include_, ansible-lint will not check the included tasks.
Here is my main.yml that is working, but deprecated sind ansible 2.4
```
- name: Deploying configuration maps
  include: configuration-maps.yml
  with_items:
    - { name: someItem.yml }
  loop_control:
    loop_var: configuration_value
```

This version will not include the configuration-maps.yml
```
- name: Deploying configuration maps
  include_tasks: configuration-maps.yml
  with_items:
    - { name: someItem.yml }
  loop_control:
    loop_var: configuration_value
```

willthames:
Fixed with #287

willthames:
Thanks for the report and the fix, much appreciated!

Stepych:
The same bug with `include_role` - ansible-lint ignores tasks, included that way! For example, `include_role: name=myrole tasks_from=service`.

PR #287, opened 2017-10-02
Fix for #285

MatrixCrawler:
Added import_tasks and include_tasks checks as the include task is deprecated with ansible 2.4+

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -43,7 +43,6 @@ try:
     from ansible.utils import path_dwim
     from ansible.utils.template import template as ansible_template
     from ansible.utils import module_finder
-
     module_loader = module_finder
     ANSIBLE_VERSION = 1
 except ImportError:
@@ -51,7 +50,6 @@ except ImportError:
     from ansible.template import Templar
     from ansible.parsing.mod_args import ModuleArgsParser
     from ansible.errors import AnsibleParserError
-
     ANSIBLE_VERSION = 2
 
     def parse_yaml_from_file(filepath):
@@ -68,7 +66,6 @@ except ImportError:
         dl.set_basedir(basedir)
         templar = Templar(dl, variables=templatevars)
         return templar.template(varname, **kwargs)
-
     try:
         from ansible.plugins import module_loader
     except ImportError:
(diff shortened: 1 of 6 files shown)


lib/ansiblelint/utils.py
```diff
@@ -236,6 +240,30 @@ def _taskshandlers_children(basedir, k, v, parent_type):
                 'path': path_dwim(basedir, th['include']),
                 'type': playbook_section
             })
+        elif 'include_tasks' in th:
```
willthames: the sections for include, include_tasks and import_tasks are identical other than the path - this can be abstracted to a method or a loop

lib/ansiblelint/utils.py
```diff
@@ -44,13 +43,15 @@
...
     from ansible.utils import module_finder
+
     module_loader = module_finder
     ANSIBLE_VERSION = 1
 except ImportError:
     from ansible.parsing.dataloader import DataLoader
     from ansible.template import Templar
     from ansible.parsing.mod_args import ModuleArgsParser
     from ansible.errors import AnsibleParserError
+
```
willthames: please don't add non-functional whitespace for code that isn't currently violating PEP8

lib/ansiblelint/utils.py
```diff
@@ -355,7 +383,7 @@ def normalize_task_v2(task):
     # denormalize shell -> command conversion
     if '_uses_shell' in arguments:
         action = 'shell'
-        del(arguments['_uses_shell'])
+        del (arguments['_uses_shell'])
```
willthames: Is there any reason for this change? Seems like it is completely against normal function calling conventions

lib/ansiblelint/utils.py
```diff
@@ -483,7 +511,9 @@ def get_action_tasks(yaml, file):
     block_rescue_always = ('block', 'rescue', 'always')
     tasks[:] = [task for task in tasks if all(k not in task for k in block_rescue_always)]
 
-    return [task for task in tasks if 'include' not in task.keys()]
+    return [task for task in tasks if
+            ('include' or 'include_tasks' or 'import_tasks')
```
willthames:
```
{'include', 'include_tasks', 'import_tasks'}.isdisjoint(task.keys())
```
might be simpler, and actually work:

```
>>> ('d' or 'b') not in v
True
>>> {'d', 'b'}.isdisjoint(v)
False
```

willthames: I'd like there to be a test that would have caught this bug

lib/ansiblelint/utils.py
MatrixCrawler: Sorry, my IDE just "optimized" the code this way. It slipped through. Shouldn't have commited that.

lib/ansiblelint/utils.py
MatrixCrawler: What bug exactly? 
The wrong return for get_action_tasks?

willthames:
Merged, thank you!

willthames:
Released as 3.4.17 (really should have done this at the time!)

Issue #283, opened 2017-09-27
Feature request - configuration file


zigarn:
# Issue Type
- Feature request

# Ansible and Ansible Lint details

```
$ ansible --version
ansible 2.4.0.0
$ ansible-lint --version
ansible-lint 3.4.15
```

- ansible installation method: OS package
- ansible-lint installation method: pip

# Desired Behaviour

Be able to set configuration in a config file, with a default like `.ansible-lint` in which we can set some options like the skip list or exclude list.
This way, it's not necessary to put those options on the command line each time we call `ansible-lint`.

njenan:
@zigarn I opened a PR to add this feature #290 please take a look.

brandonkal:
Looks like this can be closed

PR #290, opened 2017-10-09
Issue #283: Configuration file for ansible-lint

njenan:
I'm trying my hand at adding the configuration file mentioned in issue #283.  I'm a newbie to Python so please go easy on me.  All code was written with TDD.  I wrote the tests to invoke ansible-lint as a subprocess and compare the output of the CLI arguments to its equivalent config file.  I wasn't sure if there was a better way to write them but I'm open to suggestions.

Please let me know any concerns or suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,32 @@ As of version 2.4.0, ansible-lint now works just on roles (this is useful
 for CI of roles)
 
 
+Configuration File
+==================
+
+Ansible-lint supports local configuration via a `.ansible-lint` configuration file.  Ansible-lint checks the working directory for the presence of this file and applies any configuration found there.
+
+The following values are supported and function identically to their CLI counterparts.
+
+```yaml
+exclude:
+  - ./my/excluded/directory/
+  - ./my/other/excluded/directory/
+  - ./last/excluded/directory/
+parseable: true
+quiet: true
+rulesdir:
+  - ./rule/directory/
+skip_list:
+  - skip_this_tag
+  - and_this_one_too
+tags:
+  - run_this_tag
+use_default_rules: true
+verbosity: 1
+```
+
+
 Pre-commit
 ==========
 
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ Options:
                         repeatable.
   --force-color         Try force colored output (relying on ansible's code)
   --nocolor             disable colored output
+  -F F                  Specify configuration file to use.  Defaults to
+                          ".ansible-lint"
 ```
 
 False positives
@@ -225,7 +227,7 @@ for CI of roles)
 Configuration File
 ==================
 
-Ansible-lint supports local configuration via a `.ansible-lint` configuration file.  Ansible-lint checks the working directory for the presence of this file and applies any configuration found there.
+Ansible-lint supports local configuration via a `.ansible-lint` configuration file.  Ansible-lint checks the working directory for the presence of this file and applies any configuration found there.  The configuration file location can also be overridden via the `-F path/to/file` CLI flag.
 
 The following values are supported and function identically to their CLI counterparts.
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Options:
                         repeatable.
   --force-color         Try force colored output (relying on ansible's code)
   --nocolor             disable colored output
-  -F /path/to/file      Specify configuration file to use.  Defaults to
+  -c /path/to/file      Specify configuration file to use.  Defaults to
                           ".ansible-lint"
 ```
 
@@ -227,7 +227,7 @@ for CI of roles)
 Configuration File
 ==================
 
-Ansible-lint supports local configuration via a `.ansible-lint` configuration file.  Ansible-lint checks the working directory for the presence of this file and applies any configuration found there.  The configuration file location can also be overridden via the `-F path/to/file` CLI flag.
+Ansible-lint supports local configuration via a `.ansible-lint` configuration file.  Ansible-lint checks the working directory for the presence of this file and applies any configuration found there.  The configuration file location can also be overridden via the `-c path/to/file` CLI flag.
 
 The following values are supported and function identically to their CLI counterparts.
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Options:
                         repeatable.
   --force-color         Try force colored output (relying on ansible's code)
   --nocolor             disable colored output
-  -F F                  Specify configuration file to use.  Defaults to
+  -F /path/to/file      Specify configuration file to use.  Defaults to
                           ".ansible-lint"
 ```
 
--- a/README.md
+++ b/README.md
@@ -231,8 +231,10 @@ Ansible-lint supports local configuration via a `.ansible-lint` configuration fi
 
 The following values are supported and function identically to their CLI counterparts.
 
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/__main__.py
```diff
@@ -31,6 +31,25 @@
 import six
 from ansiblelint import RulesCollection
 from ansiblelint.version import __version__
+import yaml
+import os
+
+config = None
+
+if os.path.exists(".ansible-lint"):
```
willthames: This should be in a function called by `main` called `read_config_file` or something so we don't need executable code in the body.

It's probably also worth making the filename a parameter, in case people want a system default or to pass in a configuration file name to override.

In order of precedence:
* A command line option (-F maybe)
* The cwd .ansible-lint file
* A file in the result of `AppDirs("ansible-lint", "com.github.willthames").user_config_dir` (this becomes ~/.config/ansible-review/ on linux)
* System file /etc/ansible-lint/ansible-lint.yml or something

willthames:
@njenan this is great, thank you for doing this! In terms of next steps, don't feel you need to implement all of the different config files, but I would like the reading of the configuration file to be parameterised and allow for different future options (and avoid the `config` global)

njenan:
Sounds good, working on the command line switch for the config file right now.  Because the configuration values are passed as the default values of the CLI flags (so it can be overridden from the command line), the config file needs to be parsed and loaded before the other flags are evaluated.  I'll probably have an update sent out tomorrow afternoon because it is getting late here.

njenan:
Ok, I pushed some changes.  I parse the `-F` argument directly.  The reason for this is because I can't pull the `-F` value from the optparse because I need the configuration file to populate the default CLI args.  I tried using getopt, but it had problems with certain ordering of CLI args.  So... I don't have a better solution than this.  Let me know if you have any ideas.

willthames:
I'm actually ok with the defaults shown with just `ansible-lint --version` being the standard defaults and ignoring the configuration files (otherwise they're not really defaults)

However, if you're keen to show the *current* defaults, maybe try the default configuration file locations first, and use them to populate the default CLI args - I don't think people need to see the defaults to be the values they push in with a file described with `-F` which will pretty much override any defaults.

njenan:
If I understand the way opt parse works, the problem is this:

```python
import optparse

parser = optparse.OptionParser()
parser.add_option("-a", dest="a", default="default_value")

opts, _ = parser.parse_args()
print(opts.a)
```

So if I invoke the script
```
python file.py
>> default_value
```
But if I add the -a arg:
```
python file.py -a override
>> override
```

So in terms of the current code, if I do the following:

```
def main():

     load_config_file()  # <- This can't be the first line because I don't know the location of the config file yet
...
    parser.add_option('-F', help='Specify configuration file to use.  Defaults to ".ansible-lint"')
...
    parser.add_option('-q', dest='quiet',
                      default=config.config_or_default("quiet", False),  # <- But I can't define these parser options without having the configuration files and the default values
                      action='store_true',
                      help="quieter, although not silent output")
...
```

So basically I need to load the configuration file, but I can't load the file until I parse the -F arg, but I can't parse the -F arg with opt parse without defining the default values, and I can't define the default values until I load the configuration file.

Hopefully I am explaining this adequately, maybe there is an obvious solution I am not seeing.

njenan:
Actually, I thought about it, and I think I know what another option would be.  Load the configuration file, and then merge the args from the configuration file with the args from optparse.  There would not be a cyclical execution dependency with this approach.

The disadvantage of this approach would be that if someone set a config in a configuration file, then they would _not_ be able to override it from the command line.  Because I wouldn't be able to tell what options were default or user specified when merging the config and CLI args.

willthames:
> The disadvantage of this approach would be that if someone set a config in a configuration file, then they would not be able to override it from the command line. Because I wouldn't be able to tell what options were default or user specified when merging the config and CLI args.

Yeah, you're right, this is harder than it looks!

It's not clear to me what should be expected when you add more append values  (e.g. rules directories) on the command line when using a config file anyway - should it replace or augment them?

A lot of command line arguments might work as people expect anyway if they're store_append or increase/decrease a value.

If we can get to a reasonable, documented result then we should be good - even if the documentation says "If you have a configuration file and wish to override values, please use -F with a new configuration file" that will be better than we currently have

njenan:
Ok, please take another look.  I added logic to merge the config and CLI options.  In the case of True/False values, the True value is preferred, and in the case of lists they are simply merged.  I think this will produce the least surprising behavior for the user.

njenan:
Ok, branch should be working now.

lib/ansiblelint/__main__.py
```diff
@@ -77,9 +99,40 @@ def main():
...
+
+        if 'use_default_rules' in config:
+            options.use_default_rules = \
+                options.use_default_rules or config['use_default_rules']
+
+        if 'verbosity' in config:
+            options.verbosity = options.verbosity + config['verbosity']
+
+        if 'exclude_paths' in config:
+            options.exclude_paths = \
```
willthames: `max-line-length` is set to 100 in setup.cfg so we don't have to do things like this :)

README.md
```diff
@@ -51,6 +51,8 @@ Options:
                         repeatable.
   --force-color         Try force colored output (relying on ansible's code)
   --nocolor             disable colored output
+  -F F                  Specify configuration file to use.  Defaults to
```
willthames: Can you make the second F here /path/to/file (or similar if /path/to/file won't fit)

willthames:
Two relatively minor changes and we should be good to go. 

Might check with the owners of issues regarding a configuration file if this works for them before merge though.

zigarn:
Great! Thanks.

Some small feedback:
- is `-F` a good option? It's for `File` which can be confusing with the program argument (files to check). Wouldn't be `-c` and/or `--config[-file]` better? (cf. [yamllint](https://github.com/adrienverge/yamllint/blob/c8fc170ff0525e639940e4687a3c39cb65ca491d/yamllint/cli.py#L83), [pycodestyle](https://github.com/PyCQA/pycodestyle/blob/c7448b4f8c50904eb71855b6559a9c3650bf510c/pycodestyle.py#L2318))
- maybe all the options are not meant to be in conf file like `verbosity` or `quiet`
- I'm not sure about the merging/adding, can be weird to use `-v` on the command-line but ending with `-vvv` because `verbosity: 2` in config file. The typical behavior is having CLI args overriding config file
- May be nice to avoid listing all options in merging strategy as new CLI args may be missed in the future (add in the argparse but not in the merging).

njenan:
1. I'll update it to `-c`
2.  I can remove those, I figured it was better to include it rather than not, I'll defer to @willthames judgement on this
3.  For verbosity specifically, as far as I can tell, values 1 and over function identically (based on my quick search of the code base and a few experiments), I don't think this will be an issue.  For overriding args, there are a couple issues with overriding the default args, namely I don't have a good way of telling if an argument is default or not, so I decided to go with a merging strategy (see my first commit).    I was not able to figure out how to solve this issues, but I'm open to suggestions.
4.  I had considered doing some sort of dynamic combination of the args and config file, but my concern was there might be bugs with unexpected values passed in by the user.  Do you have any ideas how I could make sure my code covered all cases?

strangeman:
Looks great! Waiting for this feature so much.

willthames:
Ok, I'm going to merge this and one other PR and release it as 2.5.0-rc1 for wider testing

PR #292, opened 2017-10-09
ANSIBLE0012: also allow register with command.

rhpvorderman:
In my playbook I need to gather some system values by running custom commands and registering them into variables. 
I can not use ansible-facts for this.
Example use case:
```YAML
- name: Get ansible-lint version
  command: ansible-lint --version
  register: ansible_lint_version_output

- name: Register ansible-lint version
  set_fact:
    ansible_lint_version: ansible_lint_version_output.stdout
```
Ansible-lint sees this as improper use of the command statement. So I have to add the skip_ansible_lint tag to these tasks currently.

With this pull request these tasks are also allowed.

### DIFF
--- a/lib/ansiblelint/rules/CommandHasChangesCheckRule.py
+++ b/lib/ansiblelint/rules/CommandHasChangesCheckRule.py
@@ -37,5 +37,6 @@ class CommandHasChangesCheckRule(AnsibleLintRule):
             if task["action"]["__ansible_module__"] in self._commands:
                 return 'changed_when' not in task and \
                     'when' not in task and \
+                    'register' not in task and \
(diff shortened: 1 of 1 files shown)


willthames:
This has nothing to do with register - it's to do with the fact that command always returns `changed: True` unless you work to prevent that. 

In this case, the easy fix is to do:

```
- name: Get ansible-lint version
  command: ansible-lint --version
  changed_when: false
  register: ansible_lint_version_output
```

as running `ansible-lint --version` does not change anything.

rhpvorderman:
Thanks for the information! I will try that. Did not realize I could do that.

PR #294, opened 2017-10-12
Added vars and defaults to directories to check

msuringa:
# Test setup
testRole/tasks/main.yml
```
---

- name: Print variable
  debug: var=myVar

...
```
testRole/vars/main.yml
```
---

# Below line has some trailing whitespaces
myVar: myString    

...
```
# Current Behaviour
```
ansible-lint testRole
```
Returns OK

# Behaviour after change
```
ansible-lint testRole
[ANSIBLE0002] Trailing whitespace
/home/msuringa/test/vars/main.yml:3
myVar: myString    
```

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -315,7 +315,7 @@ def _look_for_role_files(basedir, role):
 
     results = []
 
-    for th in ['tasks', 'handlers', 'meta']:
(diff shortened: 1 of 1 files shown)


willthames:
This is likely to cause too many false positives. 

[ansible-review](https://github.com/willthames/ansible-review) does a much better job of applying specific rules to specific types of files.

PR #298, opened 2017-10-24
added test rule to check for key=value format

pescobar:
I wrote this quick&dirty rule to check for `key=value` format to try to enforce native yaml syntax. More details about this in this issue https://github.com/willthames/ansible-lint/issues/297

I have written this in 15min and this is my first time writing an ansible-lint rule so most probably this code is not the right way to do it but it seems to do the trick for me. these are same sample outputs:

```
Examining tasks/main.yml of type playbook
[ANSIBLE0003] Dont use key=value format for tasks
tasks/main.yml:8
    timezone: name="{{ ntp_timezone }}"

[ANSIBLE0003] Dont use key=value format for tasks
/tasks/main.yml:11
    yum: name=ntp state=present
```

when processing this test playbook:

```
  - name: set timezone
    timezone: name="{{ ntp_timezone }}"

  - name: Install ntp
    yum: name=ntp state=present

  - name: Copy ntp conf file
    template:
      src: etc/ntp.conf.j2
      dest: /etc/ntp.conf
    notify: restart ntp
    tags: configs_sync
    when: ntp_timezone == 'lelele'

  - name: Start and enable the ntp service
    service:
      name: ntpd
      state: started
      enabled: true
```

Any suggestion is appreciated. 

Any chance to get this merged once it's properly done?

### DIFF
--- a/lib/ansiblelint/rules/NoKeyValueFormat.py
+++ b/lib/ansiblelint/rules/NoKeyValueFormat.py
@@ -1,0 +1,38 @@
+# Copyright (c) 2013-2014 Will Thames <will@thames.id.au>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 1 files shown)


pescobar:
another possible implementation

```python
import re


class NoKeyValueFormat(AnsibleLintRule):
    id = 'ANSIBLE0003'
    shortdesc = 'Dont use key=value format for tasks'
    description = 'You should use native YAML syntax instead of key=value format'
    tags = ['formatting']

    def match(self, file, line):
        pattern_name = re.compile('^[\s|-].*name:')
        pattern_when = re.compile('^[\s].*when:')
        pattern_key_value = re.compile('[a-zA-Z0-9_.-]=["{a-zA-Z0-9_.-]')
        if pattern_name.findall(line):
            return False  #  if this is the name of the task skip check
        if pattern_when.findall(line):
            return False  #  if this is a conditional skip check
        if pattern_key_value.findall(line):
            return True  #  search for key=value format in any other line
```

willthames:
It's too quick and dirty. There's no way this is becoming an enforced rule in ansible-lint - there would be far too many false positives. You are welcome to add it to your list of custom rules. 

There are ways to do this check (clearly, as ansible-review does it properly), and you might find this works for you but I'm not merging it. To be honest, I'm not even comfortable with making the rule that's in ansible-review enforced through ansible-lint - using the YAML form is preferred, but the key=value form isn't massively dangerous.

As I've said before, I recommend ansible-review for letting you pick and choose what standards you wish to enforce.

PR #305, opened 2017-11-29
Fix for Python 2.6

larsweiler:
Python 2.6 does not offer the non-empty set shortcut statement with {}, which was introduced in Python 2.7: "As of Python 2.7, non-empty sets (not frozensets) can be created by placing a comma-separated list of elements within braces, for example: {'jack', 'sjoerd'}, in addition to the set constructor."

Unfortunately, Python 2.6 is still the default on RHEL6 systems.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -492,7 +492,7 @@ def get_action_tasks(yaml, file):
     tasks[:] = [task for task in tasks if all(k not in task for k in block_rescue_always)]
 
     return [task for task in tasks if
-            {'include', 'include_tasks', 'import_tasks'}.isdisjoint(task.keys())]
+            set(['include', 'include_tasks', 'import_tasks']).isdisjoint(task.keys())]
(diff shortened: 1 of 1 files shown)


willthames:
Ideally we'd catch this bug automatically by updating tox.ini and .travis.yml to test on python 2.6.

If you fancy adding the automation too, https://github.com/willthames/ansible-inventory-grapher/blob/master/.travis.yml (and tox.ini in the same directory) show what I did to get python 3.6 working and 2.6 would be very similar, but I'm happy enough to merge this as is, just let me know.

larsweiler:
Sorry, I'm not experienced with Travis CI, so I don't know what I have to configure in detail. It would be more like try and error. Probably I can work on it in the future.

For the moment, I think a merge would be fine.

larsweiler:
I just want to ask, if there is still something to do from my side?

willthames:
Sorry, no, this is all fine :)

PR #317, opened 2018-01-28
OctalPermissionsRule: Fix false positive for octal directory permissions with just execute bit set

### LINKED ISSUE #235: Octal Permissions Rule, false positive ?
Hello !

# Issue Type
- Bug report

```
ansible 2.2.0.0
ansible-lint 3.4.9
```

# Desired Behaviour
- name: Ensure destination directories exist and have proper permissions

```yml
  file:
    name: /etc/ssl/private
    state: directory
    owner: root
    group: ssl-cert
    mode: 0710
```

No warning should be triggered

# Actual Behaviour

This task trigger the rule "ANSIBLE0009" because "file permission modes don't have write or execute bit set when read bit is not set", which isn't true for a folder.

On another subject I don't understand why string and int are treated differently:

```python
if isinstance(mode, six.string_types) and self.mode_regex.match(mode):
if isinstance(mode, int):
```
IMHO, we should convert string mode into int and use the same check for both.

I can make a PR if you agree with this bug report.

Regards,

ribbons:
It is relatively common to want to set just the execute bit for group or world for directories (e.g. `0711` or `0751`) to allow user access to paths below without being able to list the directory contents.  This however
causes a task to be flagged as violating `ANSIBLE0009` even if a valid octal permissions value was used.

Bypass the group and world execute only permission checks just when `state=directory` is used with the file module.

Resolves #235.

### DIFF
--- a/lib/ansiblelint/rules/OctalPermissionsRule.py
+++ b/lib/ansiblelint/rules/OctalPermissionsRule.py
@@ -43,16 +43,20 @@ class OctalPermissionsRule(AnsibleLintRule):
             if isinstance(mode, six.string_types) and self.mode_regex.match(mode):
                 return not self.valid_mode_regex.match(mode)
             if isinstance(mode, int):
-                # sensible file permission modes don't
+                # sensible file (but not dir g/w) permission modes don't
                 # have write or execute bit set when read bit is
                 # not set
                 # also, user permissions are more generous than
                 # group permissions and user and group permissions
                 # are more generous than world permissions
 
-                result = (mode % 8 and mode % 8 < 4 or
-                          (mode >> 3) % 8 and (mode >> 3) % 8 < 4 or
-                          (mode >> 6) % 8 and (mode >> 6) % 8 < 4 or
+                if (task["action"]["__ansible_module__"] != 'file' or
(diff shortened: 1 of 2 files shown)


ribbons:
I've checked the Travis CI failure and it looks like the same issue as the last build of `master` so I don't think it is my doing :smile:

willthames:
@ribbons I've updated the CI test suite now so tests should pass again

ribbons:
> @ribbons I've updated the CI test suite now so tests should pass again

Perfect, thanks.  Have rebased and all of the tests are now passing :smile:

willthames:
The trouble with this fix is that it's specific to the file module. It would be good to have a more generic solution. Of course, if coming up with a halfway decent solution was easy, we'd have done it by now.

I've had another look at this and opened #318 as an alternative implementation.

willthames:
Merged #318 instead so closing this - thanks for raising the problem and coming up with a fix.

cafuego:
Even with this fix applied, directory mode 02751 seems to still report as invalid; it shouldn't.

The use case is a web site asset directory. The owner can write with the GID enforced. Group members able to list and access the contents, but others are only able to access complete file paths.

PR #318, opened 2018-01-29
Allow directories to have execute but not read bit set

### LINKED ISSUE #235: Octal Permissions Rule, false positive ?

willthames:
Fixes #235

Alternative to #317

### DIFF
--- a/lib/ansiblelint/rules/OctalPermissionsRule.py
+++ b/lib/ansiblelint/rules/OctalPermissionsRule.py
@@ -44,15 +44,16 @@ class OctalPermissionsRule(AnsibleLintRule):
                 return not self.valid_mode_regex.match(mode)
             if isinstance(mode, int):
                 # sensible file permission modes don't
-                # have write or execute bit set when read bit is
-                # not set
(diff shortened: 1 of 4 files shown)


ribbons:
That seems like a cleaner way of doing it than mine and addresses my particular directory case, so gets a :+1: from me.

PR #326, opened 2018-03-06
WIP: Octal permissions

twouters:
`mode=0755`, `mode: "0755"` and `mode: 0755` appear to be treated differently.
Convert the string to its octal notation if it's numerical to treat them equally and check it for sensible file permission modes too.

This breaks the current tests (on my setup) so either the conversion from string to octal is wrong, or the sensible file permission check is wrong.

Also, would it make sense to move the check if the permissions are sensible to a separate rule? (This would make it possible to check for leading zeros but ignore less sensible permissions.)

### DIFF
--- a/lib/ansiblelint/rules/OctalPermissionsRule.py
+++ b/lib/ansiblelint/rules/OctalPermissionsRule.py
@@ -41,7 +41,9 @@ class OctalPermissionsRule(AnsibleLintRule):
         if task["action"]["__ansible_module__"] in self._modules:
             mode = task['action'].get('mode', None)
             if isinstance(mode, six.string_types) and self.mode_regex.match(mode):
-                return not self.valid_mode_regex.match(mode)
+                if not self.valid_mode_regex.match(mode):
+                    return True
+                mode = int(mode.strip(), 8)
             if isinstance(mode, int):
                 # sensible file permission modes don't
                 # have write bit set when read bit is
--- a/test/octalpermissions-success.yml
+++ b/test/octalpermissions-success.yml
@@ -21,11 +21,16 @@
     - name: octal permissions test success (0777)
       file: path=baz mode=0777
 
-    - name: octal permissions test success (0711)
+    - name: octal permissions test success [1/2] (0711)
       file: path=baz mode=0711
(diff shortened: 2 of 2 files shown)


willthames:
I stand by the tests in the absence of alternative theories.

Because this rule takes effect after the YAML has been parsed, any number has already been converted - we don't know if it ever had a leading zero. We can only really check a proxy, that the permissions aren't ridiculous. So `mode: 755` is really `mode: 01363`, which we can rule out because `-wxrw-wx` is not a sensible permission (because of write without read, and user being more restrictive than group)

willthames:
You can see how (although I can't trivially see why) your tests are failing

```
$ export PYTHONPATH=lib
$ bin/ansible-lint test/octalpermissions-success.yml
[ANSIBLE0009] Octal file permissions must contain leading zero
test/octalpermissions-success.yml:24
Task/Handler: octal permissions test success [1/2] (0711)

[ANSIBLE0009] Octal file permissions must contain leading zero
test/octalpermissions-success.yml:27
Task/Handler: octal permissions test success [2/2] (0711)

[ANSIBLE0009] Octal file permissions must contain leading zero
test/octalpermissions-success.yml:32
Task/Handler: octal permissions test success (0710)
```

twouters:
I've added the second test before I made any other changes because ansible-lint returned ANSIBLE0009 when it shouldn't have (according to the commit that added the previous tests).
I think the tests passed because `mode=0711` is treated as a string and `mode: 0711` is treated as an int/oct.
Which made me (wrongly) conclude that the string based check shouldn't just accept the mode when it starts with a 0.

I understand now that the permission sensibility check is there as an extra safeguard to detect int<->octal conversions and not to warn the user that they're setting non-sensible permissions (e.g. `mode=01363`) in their playbook. Which kind of makes sense.

I'll try to figure out why `0711` isn't a sensible mode according to the test tomorrow.

willthames:
Ah, I already have a fix for 0711 in my more_octal_fix branch - it's ready to be merged according to tests.

willthames:
Merged the fix for `711` permissions

PR #330, opened 2018-03-08
Fix TaskHasName check for tasks with an empty name field

ekeih:
ansible-lint failed to detect tasks with an empty name field. This
commit fixes the issue and adds some tests for it.


ekeih:
@willthames do you have any thoughts about this PR? Does it need changes or is reviewing/merging just a time issue?
Let me know if I can do anything to help :slightly_smiling_face:

lib/ansiblelint/rules/TaskHasNameRule.py
```diff
@@ -31,5 +31,7 @@ class TaskHasNameRule(AnsibleLintRule):
     _nameless_tasks = ['meta', 'debug']
 
     def matchtask(self, file, task):
-        return task.get('name', '') == '' and \
```
willthames: any reason we can't just change this to

```
    return (task.get('name') and 
            task["action"]["__ansible_module__"] not in self._nameless_tasks)
```

ekeih: In my opinion it is much easier to read the code if we check each condition in a separate step.
By combining the different conditions with and/or operators it becomes complicated to understand the logic.

willthames: Sorry, but the step through with a missing condition failing through and implicitly returning None is much less readable in my opinion. I'd agree with you if it were a complicated logical expression but it's effectively `return A && !B`

PR #348, opened 2018-06-17
Fix comma-separated skip_list arguments

willthames:
skip.add only takes one element - skip.update allows for a set
to be updated with multiple elements

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -163,7 +163,7 @@ def main():
 
     skip = set()
     for s in options.skip_list:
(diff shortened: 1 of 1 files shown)


ianw:
I ran our existing tox test [1] with this and it works.  Prior to this it failed per [2]

[1] https://git.openstack.org/cgit/openstack-infra/system-config/tree/tox.ini#n11
[2] http://logs.openstack.org/21/575921/3/check/tox-linters/c0fd272/job-output.txt.gz#_2018-06-17_16_10_19_439285

PR #365, opened 2018-08-27
Fix task files included in roles not parsed

zeot:
Task files included using `include_tasks` or `import_tasks` in roles are not parsed.
They will now be parsed using the _include_children method.


willthames:
This looks great, thanks @zeot, great first contribution!

PR #368, opened 2018-09-04
Proposal: Category tagging and id edits for documentation

awcrosby:
In the near future the ansible galaxy dev team would like to use the default ansible-lint rules (lib/ansiblelint/rules) as part of the galaxy import process. This PR is a proposal to organize the default rules to help with display, documentation, and adding future rules. We are looking for community feedback.

Proposal:
- Have each rule be in a category, and have the id relate to the category
- Use the first tag in the rule `tags` list to denote the category
- If that seems acceptable, I took a pass at editing the tags and ids, including shortening the ids. The result matches the [pep8 error code setup](https://pep8.readthedocs.io/en/latest/intro.html#error-codes), and enables auto-generated documentation: RULE_DOCS.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,8 @@ rulesdir:
 skip_list:
   - skip_this_tag
   - and_this_one_too
+  - skip_this_id
+  - '401'
 tags:
   - run_this_tag
 use_default_rules: true
--- a/RULE_DOCS.md
+++ b/RULE_DOCS.md
@@ -1,0 +1,28 @@
+| id                                                       | sample message                                           |
+|----------------------------------------------------------|----------------------------------------------------------|
+| **E1**                                                   | *deprecated*                                             |
+| E101                                                     | Deprecated always_run                                    |
+| E102                                                     | No Jinja2 in when                                        |
+| E103                                                     | Deprecated sudo                                          |
+| E104                                                     | Using bare variables is deprecated                       |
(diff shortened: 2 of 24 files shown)


willthames:
I am all in favour of this change, other than that I would like to retain a mapping of historic names to new IDs for backward compatibility purposes so that people using `-t`, `-x` or `skip_list` don't have to update their usage. 

I would be fine with a deprecation cycle of a couple of releases

tima:
Understand what you are saying @willthames. The way this PR is done, the output would change in that these new python codestyle IDs would display and not the historic ones being used today. Would this be too much of a break? Is using that ID when parsing results a common use case?

willthames:
@tima I'm not too worried about the display names as long as existing ansible-lint invocations work in the same way 

We do have a ParseableFormatter so that jenkins junit style reports work correctly - it's worth checking that these new names behave correctly with the ParseableFormatter - `ansible-lint -p`. Looking at the code, the ParseableFormatter prepends `E` to the rule id, so you'll end up with EE101 when using `-p`.

awcrosby:
I updated the PR based on this conversation:
- added the historic id as a tag for each rule
- added a test case to check running and skipping both by rule id and tag
- added example to README `skip_list`

willthames:
This can go into 3.5.0

PR #370, opened 2018-09-19
adding support for include_tasks tags

jielee:


willthames:
Thanks for this, it reminded me that someone had already contributed something along these lines, that also catered for import_tasks, so I merged #365. I don't think this one is needed now #365 is merged, sorry to have caused duplicated effort

PR #380, opened 2018-10-19
enhancement: add -F/--format option and JSON output formatter

ssato:
This PR adds a feature to support JSON formatted output and -F/--format option to chose output format along with misc. changes and refactorings to ansible-lint.

Here is an example session log w/ this changes:
```
ssato@localhost% PYTHONPATH=./lib python3 bin/ansible-lint -vv test/always-run-failure.yml
Examining test/always-run-failure.yml of type playbook
[101] Deprecated always_run
test/always-run-failure.yml:4
Task/Handler: always_run is deprecated

ssato@localhost% PYTHONPATH=./lib python3 bin/ansible-lint -vv -q test/always-run-failure.yml
Examining test/always-run-failure.yml of type playbook
[101] test/always-run-failure.yml:4
ssato@localhost% PYTHONPATH=./lib python3 bin/ansible-lint -vv -p test/always-run-failure.yml
Examining test/always-run-failure.yml of type playbook
test/always-run-failure.yml:4: [E101] Deprecated always_run
ssato@localhost% for f in default quiet parseable json; do
for> PYTHONPATH=./lib python3 bin/ansible-lint -vv -F $f test/always-run-failure.yml
for> done
Examining test/always-run-failure.yml of type playbook
[101] Deprecated always_run
test/always-run-failure.yml:4
Task/Handler: always_run is deprecated

Examining test/always-run-failure.yml of type playbook
[101] test/always-run-failure.yml:4
Examining test/always-run-failure.yml of type playbook
test/always-run-failure.yml:4: [E101] Deprecated always_run
Examining test/always-run-failure.yml of type playbook
[
  {
    "rule_id": "101",
    "message": "Deprecated always_run",
    "filename": "test/always-run-failure.yml",
    "linenumber": 4,
    "line": "Task/Handler: always_run is deprecated"
  }
]
ssato@localhost%
```

P.S. I made the commits separated into primitive ones as much as I could and it should be easy to cherry-pick some of the changes in this PR if entire changes are not accepted, or if some additional works are needed to merge them, please let me know.

PR #381, opened 2018-10-19
enhancement: add an experimental JUnit XML format option

ssato:
This is depends on #380.

This PR adds a very simple JUnit XML output format support to ansible-lint based on #380.

Here is an example session log if the changes of this PR were applied:
```
ssato@localhost% PYTHONPATH=./lib python3 bin/ansible-lint -h | grep -A 3 format
  -p                    parseable output in the format of pep8
  -r RULESDIR           specify one or more rules directories using one or
                        more -r arguments. Any -r flags override the default
                        rules in
--
  -F FORMAT, --format=FORMAT
                        Specify the output format. Choices are: default, json,
                        junitxml, parseable, quiet [default]
  -c C                  Specify configuration file to use.  Defaults to
                        ".ansible-lint"
ssato@localhost% PYTHONPATH=./lib python3 bin/ansible-lint -vv -F junitxml test/always-run-failure.yml
Examining test/always-run-failure.yml of type playbook
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
    <testsuite name="ansible-lint 3.4.23" timestamp="2018-10-20T04:59:01.757025" id="0"
               tests="16" time="0" errors="1"
               skipped="0">
      <properties>
        <property name="tool" value="ansible-lint"/>
        <property name="version" value="3.4.23"/>
      </properties>
      <testcase id="101" name="Deprecated always_run" classname="test/always-run-failure.yml">
        <failure type="ERROR" message="Deprecated always_run">
            rule_id=101
            linenumber=4
            line=Task/Handler: always_run is deprecated
        </failure>
      </testcase>
</testsuite>
<testsuites>

ssato@localhost%
```

P.S. This is based on #380, extra works might be necessary to merge this RP if that PR was merged.
Please let me know if you decided to merge this after that PR because some rework might be necessary to merge this PR (I'll do that if so).


gundalow:
Closing and re-opening to trigger CI.

webknjaz:
Ref #396

lib/ansiblelint/formatters/junitxml.py
```diff
@@ -0,0 +1,58 @@
...
+_JUNIT_XML_TESTCASE = """\
+<testcase id="{rule_id}" name="{shortdesc}" classname="{filename}">
+        <failure type="ERROR" message="{message}">
+            rule_id={rule_id}
+            linenumber={linenumber}
+            line={line}
+        </failure>
+      </testcase>
+"""
+_JUNIT_XML = """\
```
webknjaz: I think an actual XML generator is needed here

ignatenkobrain:
@awcrosby any chance to review this?

webknjaz:
FTR I'm observing folks in the Python community talking about nunit format being superior over junit. Should ansible-lint consider using it?

lib/ansiblelint/formatters/junitxml.py
ssato: (very sorry, I missed your review comment.)

Do you mean that XML template string should be built with element tree or something like that?
It should be better but I feel that a little too much because it's very small XML snippet.

ssato:
> FTR I'm observing folks in the Python community talking about nunit format being superior over junit. Should ansible-lint consider using it?

If you need that and it's important I can/will add such function w/o much efforts also I think but my motive of this PR is that if ansible-lint can geneate reports in JUnit XML format, jenkins users should be able to use it by default. I don't think the NUnit plugin is much popular than the JUnit plugin so I want to keep this as it is now and focus on the support of the JUnit XML format in this PR, and will open another PR for NUnit XML support as needed.

lib/ansiblelint/formatters/junitxml.py
webknjaz: Are you sure that you properly escape **each** variable contents before interpolation, then?

lib/ansiblelint/formatters/junitxml.py
```diff
@@ -0,0 +1,58 @@
...
+    def format(self, match, colored=False):
+        """
+        :param match: :class:`~ansiblelint.Match` object
+        :param colored: It's completely ignored in this formatter
+        """
+        return _JUNIT_XML_TESTCASE.format(**match.as_dict())
+
+    def formats(self, matches, colored=False, **kwargs):
+        cases = super(JUnitXmlFormatter, self).formats(matches)
+        info = dict(version=VERSION, timestamp=_TIMESTAMP,
```
webknjaz: Please extend your tests with more interesting cases, like:

```suggestion
        info = dict(version='" sdfh lK ☃', timestamp=_TIMESTAMP,
```

lib/ansiblelint/formatters/junitxml.py
```diff
@@ -0,0 +1,58 @@
...
+The XML Schema of JUnit XML format:
+
+- https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd
+  [http://windyroad.com.au/2011/02/07/apache-ant-junit-xml-schema/]
+"""
+from __future__ import absolute_import
+import datetime
+
+from .base import BaseFormatter
+from ..version import __version__ as VERSION
```
webknjaz: Add this boilerplate for defaulting to new-style classes automatically:
```suggestion
from ..version import __version__ as VERSION


__metadata__ = type
```

lib/ansiblelint/formatters/base.py
```diff
@@ -0,0 +1,29 @@
...
+        """
+        :param matches: A list of :class:`ansiblelint.Match` objects or []
+        :param colored: Colored output will be returned if True
+        :return: Formatted string for matches
+        """
+        return os.linesep.join(self.format(m, colored=colored, **kwargs) for m
+                               in matches)
+
+    def format(self, match, colored=False, **kwargs):
+        """
```
webknjaz: Could you please add an actual docstring here?

lib/ansiblelint/formatters/base.py
```diff
@@ -0,0 +1,29 @@
...
+
+class BaseFormatter(object):
+
+    @classmethod
+    def name(cls):
+        return getattr(cls, "_name",
+                       cls.__name__.replace("Formatter", "").lower())
+
+    def formats(self, matches, colored=False, **kwargs):
+        """
```
webknjaz: Add a docstring plz. https://www.python.org/dev/peps/pep-0257/

lib/ansiblelint/formatters/junitxml.py
ssato: oh, i see. I understand what you mean. I'll check that further.

CHANGELOG.md
```diff
@@ -0,0 +1,28 @@
+# 3.5.1
```
webknjaz: (unrelated)

README.md
```diff
@@ -0,0 +1,279 @@
+Ansible-lint
```
webknjaz: (unrelated)

lib/ansiblelint/__init__.py
```diff
@@ -214,6 +214,11 @@ def __repr__(self):
         return formatstr.format(self.rule.id, self.message,
                                 self.filename, self.linenumber, self.line)
 
+    def as_dict(self):
+        return dict(rule_id=self.rule.id, message=self.message,
```
webknjaz: plz use a dict literal instead of the constructor.

lib/ansiblelint/__main__.py
```diff
@@ -35,6 +35,14 @@
 import os
 
 
+__FORMATTERS = (formatters.Formatter,
+                formatters.QuietFormatter,
+                formatters.ParseableFormatter,
+                formatters.JsonFormatter,
+                formatters.JUnitXmlFormatter)
+_FORMATTERS = dict((fmtr.name(), fmtr) for fmtr in __FORMATTERS)
```
webknjaz: plz use a dict comprehension instead of a dict constructor combined with a generator expression

lib/ansiblelint/formatters/base.py
```diff
@@ -0,0 +1,33 @@
+r"""Abstract formatters
```
webknjaz:
```suggestion
"""Abstract formatters."""
```

lib/ansiblelint/formatters/base.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+from __future__ import absolute_import
+import os
+
+
+class BaseFormatter(object):
```
webknjaz:
```suggestion
__metaclass__ = type


class BaseFormatter:
```

ssato:
Although I disagree with you at the point that it brings "too much complexity" but anyway, I guess the situation and some backgrounds, and honestly, I don't have strong motivation about this, so that I close this. Later, maybe someone interested in it might re-open this.

ssbarnea:
@ssato Thanks for understanding it. The reality is that we have far more important issue to address first. As you said, if the need arises we can always reopen them. I saw projects adopting a pluggable approach in order to avoid increasing complexity for features like this.

webknjaz:
@ssbarnea I do see supporting machine-readable output as an important feature. Back when I was playing with creating a GitHub integration based on the ansible-review tool, this is what I was truly missing. I also believe that the CIs that are capable of showing the summary of the tests (AppVeyor, Azure Pipelines, Shippable, GitHub Actions /w extensions etc.) would really benefit from this.

@ssato thanks for the initial effort anyway! This would be useful as prior art if I (or somebody else) ever get to implementing it.

P.S. Regarding the complexity, it's always subjective but there's some truth in it. I'd offer y'all to read this article, it's quite insightful: https://sobolevn.me/2019/10/complexity-waterfall.

greenscar:
@ssbarnea : I see yall also closed https://github.com/ansible/ansible-lint/pull/380 so I'm commenting here.

I am running a GitLab CI job, which utilizes Code Quality formatting for its reports. Therefore, I need to save my output to a [JSON formatted file](https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool). 

Is this / 380 just being closed as "Will not fix", or is there a solution?

ssbarnea:
Does flake8, pylint, pre-commit, yamllint,... support output in the same format? If answer is not, I would likely be against implementing it.

If that output format would prove to be a future standard adopted my multiple tools, I would support it.

At this moment, that unofficial standard is the “pep8” format, also known as parsable in ansible-lint.

Basically I am against

webknjaz:
OTOH tools like pytest implement junit output just fine.

ssbarnea:
To be clear, I am not against the idea of having various output formats. I have serious concerns regarding which formats should be supported by core functionality or not. I am very aware of the problems pytest-cov had due to the multiple ways of reporting the results, which impacted the project maintenance. 

That is why they decided to move the formatting out of the tool in longer term. I do not want us to go the same way. remember any format we add to core will make any attempt to improve the output harder.

This is why I am inclined to support one of the following approaches:
- a plugin implementation that would allow someone to write output filters
- a filter: `ansible-lint -f parsable | parsable-to-foo`

I would support adding a new format to core only if I see it already implemented by other mainstream linters, like flake8.

I guess that implementing formatters as a plugin would be very easy, a decent example being how molecule externalized its providers. If I have time I may be able to propose a change.

PR #384, opened 2018-10-25
Additional default rules for ansible-lint

awcrosby:
This PR contains proposed additions to the default ansible-lint rules. Feedback is requested.

The galaxy import process will use ansible-lint and these rules to provide feedback to the author and determine a quality score. If the ansible-lint cli output, and the galaxy import lint output are the same… it should benefit developer workflow. We tried to avoid any rules that would yield output for users who do not import into galaxy.

The sources of these rules are: tsukinowasha/ansible-lint-rules and willthames/ansible-review, along with additional checks for deprecation and metadata.

Note:
- Given that these or any new rules will add new cli output and cause some CI pipelines to break, if we merge these we'll likely bump the major version number.
- Rules would get test cases before merging since most existing rules have them.


lib/ansiblelint/rules/LineTooLongRule.py
```diff
@@ -0,0 +1,15 @@
...
+
+class LineTooLongRule(AnsibleLintRule):
+    id = '204'
+    shortdesc = 'Lines should be no longer than 160 chars'
+    description = 'Long lines make code harder to read and ' \
+                  'code review more difficult'
+    tags = ['formatting']
+
+    def match(self, file, line):
+        return len(line) > 160
```
willthames: Should probably be nearer 120

lib/ansiblelint/rules/ShellAltPatch.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2016, Tsukinowa Inc. <info@tsukinowa.jp>
+# Copyright (c) 2018, Ansible Project
+
+from ansiblelint import AnsibleLintRule
+
+
+class ShellAltPatch(AnsibleLintRule):
```
willthames: This seems like an extension to CommandsInsteadOfModules rather than a new rule

lib/ansiblelint/rules/TaskManyArgs.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2016, Tsukinowa Inc. <info@tsukinowa.jp>
+# Copyright (c) 2018, Ansible Project
+
+from ansiblelint import AnsibleLintRule
+
+
+class TaskManyArgs(AnsibleLintRule):
+    id = '206'
+    shortdesc = 'Use ":" YAML format when arguments are over 4'
```
willthames: I'd favour using YAML format in all cases.

ansible-review has a better solution to this, but relies on tests run before canonicalization.

lib/ansiblelint/rules/TaskNoLocalAction.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2016, Tsukinowa Inc. <info@tsukinowa.jp>
+# Copyright (c) 2018, Ansible Project
+
+from ansiblelint import AnsibleLintRule
+
+
+class TaskNoLocalAction(AnsibleLintRule):
+    id = '504'
+    shortdesc = 'Do not use local_action. use delegate_to: localhost instead'
```
willthames: Or `connection: local`

lib/ansiblelint/rules/ModuleTemplateExt.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2016, Tsukinowa Inc. <info@tsukinowa.jp>
```
willthames: I'm heavily against this because syntax highlighting tends to be based on the extension, and so you don't get proper syntax highlighting for .j2 files

Please remove this and move it to an optional rule.

awcrosby: This was removed

lib/ansiblelint/rules/TaskNoLocalAction.py
awcrosby: Added this to the descriptions

webknjaz:
Hey @awcrosby, could you please sign-off your commits?
https://github.com/ansible/ansible-lint/pull/384/checks?check_run_id=27231145

lib/ansiblelint/rules/TaskManyArgs.py
awcrosby: Given that changing this from 4 to 0 would have false positives with any `=` in a text string, and the ansible-review solution does exist, I'm leaning towards removing this rule.

awcrosby: Removed TaskManyArgs for now

lib/ansiblelint/rules/MetaTagValidRule.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class MetaTagValidRule(AnsibleLintRule):
+    id = '702'
+    shortdesc = 'Tags must contain lowercase letters and digits only'
+    description = ("Tags must contain lowercase letters and digits only, "
+                   "and 'galaxy_tags' is expected to be a list")
+    tags = ['metadata']
+
+    TAG_REGEXP = re.compile('^[a-z0-9]+$')
```
privateip: valid tag characters should include `_` (underscore) imho

lib/ansiblelint/rules/ComparisonToLiteralBoolRule.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2016, Will Thames and contributors
+# Copyright (c) 2018, Ansible Project
+
+from ansiblelint import AnsibleLintRule
+import re
+
+
+class ComparisonToLiteralBoolRule(AnsibleLintRule):
+    id = '601'
+    shortdesc = "Don't compare to literal True/False"
```
privateip: there are use cases where comparing to `False` is required if the var is `null`.

lib/ansiblelint/rules/PlaybookExtension.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright (c) 2018, Ansible Project
+
+from ansiblelint import AnsibleLintRule
+
+import os
+
+
+class PlaybookExtension(AnsibleLintRule):
+    id = '205'
+    shortdesc = 'Playbooks should have the ".yml" extension'
```
privateip: should be `.yml` and/or `.yaml` imho

lib/ansiblelint/rules/LineTooLongRule.py
privateip: I agree with @willthames here.  playbooks / roles / task lists are not code, they are meant to be read as documents.  I think the shorter character width makes more sense in this case

lib/ansiblelint/rules/ComparisonToLiteralBoolRule.py
willthames: @privateip I guess `var != False` is more elegant than `var is not None and not var` or `not var|default(True)`

Perhaps it should just be ComparisonToTrue

lib/ansiblelint/rules/LineTooLongRule.py
awcrosby: Updated to 120

lib/ansiblelint/rules/ComparisonToLiteralBoolRule.py
webknjaz: @willthames `!=` comparisons with bools are highly discouraged in Python. So to Pythonistas it looks just as ugly as Java :D

webknjaz: Also, identify comparison is not exactly the same as comparison by value.

lib/ansiblelint/rules/MetaTagValidRule.py
awcrosby: This is the current validation check that galaxy uses on import. If any edits are made in the galaxy importer, we'll update this rule so it matches.

awcrosby:
- Test cases have been added.
- Also removed the suggestion of `local_action` to `connection: local` since there are differences
- Added .yaml in the extension check
- Removed 'company' as required meta field

test/__init__.py
```diff
@@ -32,3 +32,14 @@ def run_role_tasks_main(self, tasks_main_text):
...
         shutil.rmtree(role_path)
         return results
+
+    def run_role_meta_main(self, meta_main_text):
+        role_path = tempfile.mkdtemp(prefix='role_')
+        meta_path = os.path.join(role_path, 'meta')
+        os.makedirs(meta_path)
+        with open(os.path.join(meta_path, 'main.yml'), 'w') as fp:
+            fp.write(meta_main_text)
+            fp.seek(0)
```
webknjaz: This is probably not needed.

awcrosby: I tried using without `fp.seek(0)` but wasn't getting cli output for the test cases

webknjaz: That's odd: with-block closes the file descriptor upon exit.

test/__init__.py
```diff
@@ -32,3 +32,14 @@ def run_role_tasks_main(self, tasks_main_text):
...
+            results = self._call_runner(role_path)
```
webknjaz: Should this be inside `with`?

awcrosby: Ah yes, thanks for the catch. Fixed this in #401

petemounce:
Apologies to post to a PR long merged - from the [current error message](https://github.com/ansible/ansible-lint/commit/a02e6669cd73e0057a846854190c6b9fb207d5cf#diff-c2387cd8745df70a1872ee8c5a61feb8) for `504 Do not use 'local_action', use 'delegate_to: localhost'`, I don't know _why_ `delegate_to` is the preference?

The [Ansible Delegation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_delegation.html#delegation) documentation indicates that `local_action` is just a shorthand - it would be handy to know why this rule exists?

PR #389, opened 2018-10-30
[WIP] Packaging: consistency and automation

### LINKED ISSUE #376: Packaging problems: broken PyPI metadata and related suggestions
This is similar to what I've proposed @ willthames/ansible-review#70.
cc @gundalow

# Issue Type
- Bug report
- Feature request
- (I think it kinda combines both)

#  Ansible Lint details

https://pypi.org/project/ansible-lint/#description:
```
Project description
The author of this package has not provided a project description
```

# Desired Behaviour

1. README contents present on PyPI.
2. Project version not hardcoded in Python, but is sourced from a Git tag
3. `project_urls` metadata on PyPI
4. Metadata in declarative `setup.cfg` (with `setup.py` only containing a shim for old `setuptools`)
5. Both wheel and sdist published
6. Building/publishing automated

None of above.

webknjaz:
Resolves #376.


setup.cfg
```diff
@@ -1,6 +1,88 @@
...
+url = https://github.com/ansible/ansible-lint
+project_urls =
+    Bug Tracker = https://github.com/ansible/ansible-lint/issues
+    CI: Travis = https://travis-ci.org/ansible/ansible-lint
+    Code of Conduct = https://docs.ansible.com/ansible/latest/community/code_of_conduct.html
+    Documentation = https://github.com/ansible/ansible-lint/blob/master/RULE_DOCS.md
+    Mailing lists = https://docs.ansible.com/ansible/latest/community/communication.html#mailing-list-information
+    Source Code = https://github.com/ansible/ansible-lint
+description = Checks playbooks for practices and behaviour that could potentially be improved
+long_description = file: README.md  ### FIXME: README.rst or mime type
```
webknjaz: @gundalow we need to decide whether we want to keep md or switch to rst. rst is native to PyPI and has reacher feature set, so it'd be better.
If not, I'll need to add a mime type option to support md.

.travis.yml
```diff
@@ -70,8 +72,25 @@ matrix:
...
       env: TOXENV=py36-flake8
 
+    - stage: Deploy
+      name: Publishing current Git tagged version of dist to PyPI
+      if: tag IS present
+      install: skip
+      script: skip
+      deploy:
+        provider: pypi
+        user: any-maintainer-on-pypi  # I recommend having a separate account
```
webknjaz: @gundalow this will require creating a "service" account on PyPI just for pushing releases.

.travis.yml
```diff
@@ -70,8 +72,25 @@ matrix:
...
+        password:
+          secure: >  # needs real data
+            PLACE_FOR_AN_ENCRYPTED_STRING
```
webknjaz: @gundalow this will also need to be filled in.

setup.cfg
gundalow: Personally I prefer RST. We've standardized on that in Ansible. I'll convert the `README`

webknjaz:
Demo 2: https://travis-ci.com/webknjaz/ansible-lint/builds/89727322

.travis.yml
TomasTomecek: I hated that I couldn't set this value easily: I literally had to download the travis gem and use it to encrypt the value. Obviously it also implies that you are sharing your PyPI password with travis.

gundalow: @TomasTomecek Do you know if there is a better way of doing this?

webknjaz: Other ways are harder to maintain: more infra management on our side. As for the password, you can add it as a `PYPI_PASSWORD` secret var in Travis CI UI if you don't like embedding it into the config, but personally, I prefer using config.

.travis.yml
```diff
@@ -70,8 +72,30 @@ matrix:
       python: "3.6"
       env: TOXENV=py36-flake8
 
+    - stage: Deploy
+      name: Publishing current Git tagged version of dist to PyPI
+      if: tag IS present
```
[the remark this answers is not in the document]
webknjaz: In general, you might want to push without `v` part or with some a/b/dev/rc suffices, so it would require a way more complex regexp. I think since there's only a few people doing releases it's an unneeded complication — all of those probably know which tags to use. Also, setuptools-scm itself validates tags against PEP 440.

.travis.yml
TomasTomecek: We developed a bot which does this for us. The down side is as @webknjaz says: you need to run it somewhere on your infra. https://github.com/user-cont/release-bot

.travis.yml
```diff
@@ -70,8 +72,30 @@ matrix:
...
+        user: any-maintainer-on-pypi  # I recommend having a separate account
+        password:
+          secure: >  # needs real data
+            PLACE_FOR_AN_ENCRYPTED_STRING
+        distributions: dists
+        skip-cleanup: true
+        on:
+          tags: true
+          all_branches: true
+
```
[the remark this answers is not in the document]
webknjaz: Done.

.travis.yml
```diff
@@ -70,8 +72,37 @@ matrix:
...
+    - <<: *deploy-job
+      if: type != pull_request  # Always run, except if PR
+      deploy:
+        <<: *deploy-step
+        server: https://test.pypi.org/legacy/
```
webknjaz: @gundalow ^

.travis.yml
webknjaz: Also:
1) and doesn't match the workflow
2) will probably require adjustments
3) config is separate so that it's unobvious to follow just looking at the CI pipelines one
4) is not reusable (I mean you cannot just hit install on GitHub, bound to actual user account)

webknjaz: P.S. I recognize that Travis CI has some flaws. However, my experience is probably also different from yours. I've been contributing to their deployment tool, for example.

TomasTomecek: I wasn't implying you should use our release-bot. It's definitely easier for you to use travis. I just wanted to share a story how we tackled the issue of automating releases.

PR #398, opened 2018-11-27
Without this ansible-lint will fail if meta/main.yml has dependent ro…

cinek810:
…le without name. The approach is probable incorrect, but if you'll give me a hint I can fix it properly.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -299,6 +299,9 @@ def _load_library_if_exists(path):
 def _rolepath(basedir, role):
     role_path = None
 
+    if role is None:
+        raise SystemExit('role name cannot be empty')
+
     possible_paths = [
         # if included from a playbook
         path_dwim(basedir, os.path.join('roles', role)),


lib/ansiblelint/utils.py
```diff
@@ -276,9 +276,13 @@ def append_children(taskhandler, basedir, k, parent_type, results):
 
 def _roles_children(basedir, k, v, parent_type, main='main'):
     results = []
+    print v
     for role in v:
         if isinstance(role, dict):
             if 'role' in role or 'name' in role:
+                role=role.get('role',role.get('name'))
+                if role == None:
```
webknjaz: Never do equality comparison with `None` use `is None` or `is not None`.

lib/ansiblelint/utils.py
```diff
@@ -276,9 +276,13 @@ def append_children(taskhandler, basedir, k, parent_type, results):
 
 def _roles_children(basedir, k, v, parent_type, main='main'):
     results = []
+    print v
     for role in v:
         if isinstance(role, dict):
             if 'role' in role or 'name' in role:
+                role=role.get('role',role.get('name'))
```
webknjaz: Always put whitespaces around assignment operator.

lib/ansiblelint/utils.py
```diff
@@ -276,9 +276,13 @@ def append_children(taskhandler, basedir, k, parent_type, results):
 
 def _roles_children(basedir, k, v, parent_type, main='main'):
     results = []
+    print v
```
webknjaz: This is python2-only syntax which is completely incompatible with modern Python interpreter versions. Use print-function instead of a statement if you really need it.

lib/ansiblelint/utils.py
webknjaz: And a space after comma.

lib/ansiblelint/utils.py
```diff
@@ -276,9 +276,13 @@ def append_children(taskhandler, basedir, k, parent_type, results):
...
     for role in v:
         if isinstance(role, dict):
             if 'role' in role or 'name' in role:
```
webknjaz: This'd probably be smth like
```suggestion
             if role.get('role', role.get('name')) is not None:
```

cinek810:
@webknjaz Could you please check current version I believe it's better approach.

ssbarnea:
@webknjaz I would close this one. No reply in almost and year, no clear description or bug attached to it.

Issue #307, opened 2017-12-06
My custom rules cannot be skipped using tag "skip_ansible_lint"


pescobar:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible 2.4.2.0
ansible-lint 3.4.17
```

- ansible installation method: pip
- ansible-lint installation method: pip

# Desired Behaviour

Being able to use the tag `skip_ansible_lint` for my custom rules

# Actual Behaviour (Bug report only)

I have this custom rule which I use doing `ansible-lint -r`

```
from ansiblelint import AnsibleLintRule
import re


class NoKeyValueFormat(AnsibleLintRule):
    id = 'ANSIBLE0003'
    shortdesc = 'Dont use key=value format for tasks'
    description = 'You should use native YAML syntax instead of key=value format'
    tags = ['formatting']

    def match(self, file, line):
        pattern_name = re.compile('^[\s|-].*name:')
        pattern_when = re.compile('^[\s].*when:')
        pattern_key_value = re.compile('[a-zA-Z0-9_.-]=["{a-zA-Z0-9_.-]')
        if pattern_name.findall(line):
            return False  # if this is the name of the task skip check
        if pattern_when.findall(line):
            return False  # if this is a conditional skip check
        if pattern_key_value.findall(line):
            return True  # search for key=value format in any other lines
```

I have a false positive in one of my tasks and I would like to use `skip_ansible_lint` for this task but it doesn't work.  The custom rule is always applied no matter which tags I apply to the task. 

Is this an error in my custom rule or is it a bug in ansible-lint?

willthames:
you're matching against lines, and skip_ansible_lint works at the task level.

ansible-review has a check for this. It would be possible to convert the check to be used by ansible-lint, I believe, and handle skip_ansible_lint, but I've made no effort to do so

https://github.com/willthames/ansible-review/blob/master/lib/ansiblereview/tasks.py

pescobar:
thanks for the clarification @willthames . I understand now.

Is there any easy way to apply my rule at the task level so I can use `skip_ansible_lint`. Something like this pseudococe?

```
for task in playbook:
    applyMyrule()
```

I think this way I would keep the same functionality and I would be able to use `skip_ansible_lint` but I am not sure if this is easy to do with ansible-lint

willthames:
There are three match functions, one of which matches at the task level.
Have a look at the built-in rules, such as UseBecomeInsteadOfSudo (not at a
[...]

Issue #279, opened 2017-09-07
ansible-lint ignores -x parameters except the last one


KellerFuchs:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

ansible-2.3.2.0 & ansible-lint-3.4.15, installed via pip

# Desired Behaviour

`-x foo,bar -x baz` should result in all of `foo`, `bar` and `baz` being ignored.

# Actual Behaviour (Bug report only)

`-x foo,bar -x baz` results in only `baz` being ignored.

See [this build](https://travis-ci.org/hashbang/admin-tools/builds/272718491#L493) vs. [this one](https://travis-ci.org/hashbang/admin-tools/builds/272720279#L493)

retr0h:
ping

KellerFuchs:
@retr0h I submitted a patch and @willthames merged it; please confirm that fixes your issue  :)

PR #399, opened 2018-11-28
Added Native Yaml rule

rvben:
Hi,

I'm working on creating a rule for Native Yaml syntax. As per https://www.ansible.com/blog/ansible-best-practices-essentials point 3, Native YAML is best practice. If so, let's implement it as a rule.

**Is that a good idea? Feedback is welcome.**

I've added a first version in the PR. Used the following tasks to test it on:
```
- name: install telegraf
  yum:
    name: telegraf-{{ telegraf_version }} state=present update_cache=yes disable_gpg_check=yes enablerepo=telegraf
  notify: restart telegraf

- name: configure telegraf
  template: src=telegraf.conf.j2 dest=/etc/telegraf/telegraf.conf
  notify: restart telegraf

- name: start telegraf
  service: name={{ item }} state=started enabled=yes
  with_items:
    - telegraf

- name: start telegraf-helper
  service: name="{{ item }}" state=started enabled=yes
  with_items:
    - telegraf-helper

- name: debug message
  debug:
    msg: "Hello world=awesome"

- name: Replace line in file
  lineinfile:
    regexp: "^Hello"

    line: "Hello=World"
    path: /hello/world
```
It results in:
49006046-7664a400-f168-11e8-88dd-1c482fd4b56e

Note: This PR was first made at https://github.com/ansible/galaxy-lint-rules/pull/20, and was moved here.


willthames:
I honestly don't believe this rule is possibly using a regex - it's likely that there would be too many false positives.

See #279 and #307 for all the discussion around this rule. If we're going to do it, it needs a proper hook to do it properly.

willthames:
An example false positive here would be:

```
- pip:
     name: ansible==2.7.2
```

rvben:
@willthames, thanks for the great feedback. I noticed other tries, and have seen the ansible-review project that is referred to. Would be nice to have it in ansilbe-lint as well.

Thanks for the false positive, I updated the regex to prevent this false positive. Any other false positives you can think of?

willthames:
```
- apt:
    name: package={{ package_version }}
```

Having said that, the point of my comment isn't to start a game of whack-a-mole, but to attempt to convince you that, like using regexes to validate email addresses or parse HTML, it's a bad idea.

willthames:
Also:

```
- apt:
    name: 'ansible>=2.7.1'

- apt:
    name: >
      ansible>=2.7.1
```

YAML parsing is hard, use a YAML parser! (which is how ansible-review does it - it compares the canonicalized version with the input version)

rvben:
@willthames, Thanks again. I follow you. There were two general questions in my effort here; One was to found out if it is possible, one is to see if it's viable/a good idea. It's clear that regex is not the most elegant solution here.

It would be nice to have this as an ansible-lint rule. Is that doable? Or is ansible-review doing things radically different in general?

webknjaz:
@rvben you can start from here: https://github.com/willthames/ansible-review/blob/c55c8f1/lib/ansiblereview/tasks.py#L24-L26

PR #404, opened 2018-12-03
Docsite tweaks

awcrosby:


docs/docsite/rst/configuring/configuring.rst
```diff
@@ -39,3 +39,17 @@ The following values are supported, and function identically to their CLI counte
       - run_this_tag
     use_default_rules: true
     verbosity: 1
+
+
+Pre-commit
```
webknjaz: @awcrosby you can include chunks from readme instead of maintaining copy-pasted paragraphs :)

docs/docsite/Makefile
```diff
@@ -35,8 +35,7 @@ docs: clean htmldocs
 generate_rst: staticmin
 
 htmldocs: generate_rst
-	cd ../../lib/ansiblelint; python ./generate_docs.py; cd ../../docs/docsite
-	cp ../../RULE_DOCS.rst rst/rules/rule_docs.rst
```
webknjaz: (could be solved via `include`)

webknjaz: https://github.com/ansible/ansible-lint/pull/408

Issue #405, opened 2018-12-03
Rule 405 (PackageHasRetryRule) false positive on apt update_cache


robertdebock:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version
ansible 2.7.1
  config file = None
  configured module search path = [u'/home/robertdb/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.15 (default, Oct 15 2018, 15:26:09) [GCC 8.2.1 20180801 (Red Hat 8.2.1-2)]

ansible-lint --version
ansible-lint 3.4.23
```

- ansible installation method: pip
- ansible-lint installation method: pip (Took the rules from this repository: ff0770eedf8cf68f3747e923fc7f78cbc9359535)

# Desired Behaviour
Rule 405 checks of modules to install packages have retries configured.

The apt module can do more than install packages:

```
- name: remove software
  package:
    name: some_package
    state: absent
```

In the above example, no network communication is required, so a retry does not make sense.

# Actual Behaviour (Bug report only)
Linting failes:

```
[405] Remote package tasks should have a retry
tasks/main.yml:3
Task/Handler: remove software
```

robertdebock:
Another false positive situation, when the state is absent.

```
- name: remove software
  package:
    name: some_package
    state: absent
```

robertdebock:
In case of local installations, retries won't help:

```
- name: install software from a local source
  package:
    name: "/tmp/some_package.rpm"
    state: present
```

awcrosby:
The `state: absent` issue, and local installation issue, were addressed in the referenced PRs

PR #409, opened 2018-12-05
Exclude packages that have `state` set to `absent`.

robertdebock:
Signed-off-by: Robert de Bock <robert@meinit.nl>

This PR solved #405 by excluding tasks that have `state` set to `absent`. Tests are included in pass locally using `tox`.

### DIFF
--- a/lib/ansiblelint/rules/PackageHasRetryRule.py
+++ b/lib/ansiblelint/rules/PackageHasRetryRule.py
@@ -18,7 +18,7 @@ class PackageHasRetryRule(AnsibleLintRule):
     #   | sort | awk -F '/' \
     #   '/__|dpkg|_repo|_facts|_sub|_chan/{next} {split($NF, words, ".");
     #   print "\""words[1]"\","}'
-    package_modules = [
+    _package_modules = [
         "apk",
         "apt_key",
         "apt",
@@ -67,6 +67,16 @@ class PackageHasRetryRule(AnsibleLintRule):
         "zypper",
     ]
 
+    _module_ignore_states = [
+        "absent",
+    ]
+
     def matchtask(self, file, task):
-        return (task['action']['__ansible_module__'] in self.package_modules
-                and 'until' not in task)
+        module = task["action"]["__ansible_module__"]
+        if module in self._package_modules:
+            if 'until' not in task:
+                # Exclude tasks where state is `absent`.
+                if task['action'].get('state') not in self._module_ignore_states:
+                    message = '{0} {1}'
+                    return message.format(self.shortdesc, module)
+                return False
--- a/test/TestPackageHasRetry.py
+++ b/test/TestPackageHasRetry.py
@@ -13,6 +13,11 @@ SUCCESS = '''
         pkg: foo
       register: result
       until: result|success
+
+    - name: remove software
+      package:
+        name: some_package
+        state: absent
 '''
 
 FAILURE = '''


lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -67,6 +67,16 @@ class PackageHasRetryRule(AnsibleLintRule):
...
     def matchtask(self, file, task):
-        return (task['action']['__ansible_module__'] in self.package_modules
-                and 'until' not in task)
+        module = task["action"]["__ansible_module__"]
+        if module in self._package_modules:
+            if 'until' not in task:
+                # Exclude tasks where state is `absent`.
+                if task['action'].get('state') not in self._module_ignore_states:
+                    message = '{0} {1}'
+                    return message.format(self.shortdesc, module)
```
webknjaz: How about
```suggestion
                    return ' '.join((self.shortdesc, module))
```

webknjaz: Oh, the return value should be bool. It's strongly discouraged to mix different return types in a function.

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -18,7 +18,7 @@ class PackageHasRetryRule(AnsibleLintRule):
     #   | sort | awk -F '/' \
     #   '/__|dpkg|_repo|_facts|_sub|_chan/{next} {split($NF, words, ".");
     #   print "\""words[1]"\","}'
-    package_modules = [
+    _package_modules = [
```
webknjaz: Why do you think this change is related to the PR purpose? I'd not want to see non-related changes making PR non-atomic and addressing multiple things.

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -67,6 +67,16 @@ class PackageHasRetryRule(AnsibleLintRule):
...
+    ]
+
     def matchtask(self, file, task):
-        return (task['action']['__ansible_module__'] in self.package_modules
-                and 'until' not in task)
+        module = task["action"]["__ansible_module__"]
+        if module in self._package_modules:
+            if 'until' not in task:
+                # Exclude tasks where state is `absent`.
+                if task['action'].get('state') not in self._module_ignore_states:
```
webknjaz: It's 3 nesting levels. Better use "guard expressions" style to keep it flat.

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -67,6 +67,16 @@ class PackageHasRetryRule(AnsibleLintRule):
...
 
+    _module_ignore_states = [
+        "absent",
+    ]
+
     def matchtask(self, file, task):
-        return (task['action']['__ansible_module__'] in self.package_modules
-                and 'until' not in task)
+        module = task["action"]["__ansible_module__"]
+        if module in self._package_modules:
```
webknjaz:
```suggestion
        if module not in self._package_modules or 'until' in task:
            return False
```
will help keep 2 indentation levels less and will hugely improve readability.

lib/ansiblelint/rules/PackageHasRetryRule.py
webknjaz: This will collapse to
```suggestion
        return task['action'].get('state') not in self._module_ignore_states
```
after the suggestions above.

lib/ansiblelint/rules/PackageHasRetryRule.py
webknjaz: Another approach to improve readability would be to have "named conditions":
```suggestion
        is_module_package = module in self._package_modules
        is_task_retryable = 'until' in task
        is_state_whitelisted = task['action'].get('state') in self._module_ignore_states
        return is_module_package and not is_task_retryable and is_state_whitelisted
```
Flat and simple.

lib/ansiblelint/rules/PackageHasRetryRule.py
awcrosby: For better or worse, currently ansible-lint rules use mixed return types, returning True ansible-lint cli outputs the `shortdesc` field, returning text ansible-lint cli outputs the text. In this case True might be descriptive enough for this rule.

awcrosby:
I think these are good edits and we should get them in for 4.0.0. I agree with the indenting suggestions.

awcrosby:
Merging this as-is, I’ll address the suggestions in a separate PR. Thanks for the contribution @robertdebock!

PR #411, opened 2018-12-07
Improve rule descriptions and display in documentation

awcrosby:


docs/docsite/rst/rules/v4_default_rules.rst
```diff
@@ -0,0 +1,40 @@
+
```
gundalow: How come this is in it's versioned file?

Should version be a column in table?

awcrosby: The thought is, this would be a one-time file to show what changed in v4.0.0 since there are many rules here and since the `default_rules.rst` table is auto-generated. If we want an "added in version" column in `default_rules.rst` we can add a property to the new rules and display in the column... it would be blank for rules older than v4.0.0 - I can add that if it sounds good

awcrosby: Sounds good, added.

PR #426, opened 2018-12-12
Refactor retry rule and ignore local packages

awcrosby:
Refactor via suggestion comments in #409. Also this addresses some false positives for local package files, as noted in #405, where retries would not be necessary.

### DIFF
--- a/lib/ansiblelint/rules/PackageHasRetryRule.py
+++ b/lib/ansiblelint/rules/PackageHasRetryRule.py
@@ -83,6 +83,14 @@ class PackageHasRetryRule(AnsibleLintRule):
         "deb",
     ]
 
+    # attempt to find package name
+    def get_package_name(self, action):
+        for key in self._package_name_keys:
+            found_package_name = action.get(key)
+            if found_package_name:
+                break
+        return found_package_name
+
     def matchtask(self, file, task):
         module = task["action"]["__ansible_module__"]
 
@@ -97,16 +105,14 @@ class PackageHasRetryRule(AnsibleLintRule):
         if is_state_whitelisted:
             return False
 
-        # attempt to check if this is a local package file
-        for key in self._package_name_keys:
-            found_package_name = task['action'].get(key)
-            if found_package_name:
-                break
-        if found_package_name:
-            is_package_file = '.' in found_package_name
-            is_package_html = '://' in found_package_name
-            is_local_package_file = is_package_file and not is_package_html
-            if is_local_package_file:
-                return False
+        found_package_name = self.get_package_name(task['action'])
+        if not found_package_name:
+            return True
+
+        is_package_file = '.' in found_package_name
+        is_package_html = '://' in found_package_name
+        is_local_package_file = is_package_file and not is_package_html
+        if is_local_package_file:
+            return False
 
         return True
--- a/lib/ansiblelint/rules/PackageHasRetryRule.py
+++ b/lib/ansiblelint/rules/PackageHasRetryRule.py
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
         "absent",
     ]
 
+    _package_name_keys = [
+        "name",
+        "package",
+        "pkg",
+        "deb",
+    ]
+
     def matchtask(self, file, task):
         module = task["action"]["__ansible_module__"]
-        if module in self._package_modules:
-            if 'until' not in task:
-                # Exclude tasks where state is `absent`.
-                if task['action'].get('state') not in self._module_ignore_states:
-                    message = '{0} {1}'
-                    return message.format(self.shortdesc, module)
-                return False
+
+        if module not in self._package_modules or 'until' in task:
+            return False
+
+        if task['action'].get('state') in self._module_ignore_states:
+            return False
+
+        for key in self._package_name_keys:
+            package_name = task['action'].get(key)
+            if package_name:
+                break
+            package_name = ''
+        is_package_file = '.' in package_name
+        is_package_html = '://' in package_name
+        is_local_package_file = is_package_file and not is_package_html
+        if is_local_package_file:
+            return False
+
+        return True
--- a/lib/ansiblelint/rules/PackageHasRetryRule.py
+++ b/lib/ansiblelint/rules/PackageHasRetryRule.py
@@ -86,21 +86,27 @@ class PackageHasRetryRule(AnsibleLintRule):
     def matchtask(self, file, task):
         module = task["action"]["__ansible_module__"]
 
-        if module not in self._package_modules or 'until' in task:
+        if module not in self._package_modules:
             return False
 
-        if task['action'].get('state') in self._module_ignore_states:
+        is_task_retryable = 'until' in task
+        if is_task_retryable:
             return False
 
+        is_state_whitelisted = task['action'].get('state') in self._module_ignore_states
+        if is_state_whitelisted:
+            return False
+
+        # attempt to check if this is a local package file
         for key in self._package_name_keys:
-            package_name = task['action'].get(key)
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
...
         module = task["action"]["__ansible_module__"]
-        if module in self._package_modules:
-            if 'until' not in task:
-                # Exclude tasks where state is `absent`.
-                if task['action'].get('state') not in self._module_ignore_states:
-                    message = '{0} {1}'
-                    return message.format(self.shortdesc, module)
-                return False
+
+        if module not in self._package_modules or 'until' in task:
```
webknjaz: I still insist on naming checks. Here I see just a list of some comparisons/checks. But it doesn't carry any human-understandable context.
Now it reads as "if some_var is missing from some struct or there's no until key in another struct".

If you name those checks with reasonable labels, it'd look different: "if module is a python package and the task is not retryable".
This makes more sense to the reader, isn't it?
Code is being read much more than being written. This is why we need to maintain readability and it's also one of the core coding principles when it comes to coding in Python.

```suggestion
        is_module_package = module in self._package_modules
        is_task_retryable = 'until' in task
        is_state_whitelisted = task['action'].get('state') in self._module_ignore_states
        
        is_retryable_whitelisted_package = (
            is_module_package
            and not is_task_retryable
            and is_state_whitelisted
        )
        if not is_retryable_whitelisted_package:
```

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
...
+            return False
+
+        if task['action'].get('state') in self._module_ignore_states:
```
webknjaz:
```suggestion
```

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
...
-            if 'until' not in task:
-                # Exclude tasks where state is `absent`.
-                if task['action'].get('state') not in self._module_ignore_states:
-                    message = '{0} {1}'
-                    return message.format(self.shortdesc, module)
-                return False
+
+        if module not in self._package_modules or 'until' in task:
+            return False
+
```
webknjaz:
```suggestion
```

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
...
+        if task['action'].get('state') in self._module_ignore_states:
+            return False
```
webknjaz:
```suggestion
```

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
...
+
+        for key in self._package_name_keys:
+            package_name = task['action'].get(key)
+            if package_name:
+                break
+            package_name = ''
```
webknjaz: I've had to reread this multiple times to understand why would you do this...
Let's try to refactor it. See comment above.

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
...
-                    return message.format(self.shortdesc, module)
-                return False
+
+        if module not in self._package_modules or 'until' in task:
+            return False
+
+        if task['action'].get('state') in self._module_ignore_states:
+            return False
+
+        for key in self._package_name_keys:
```
webknjaz: Let's start with a generator expression for extracted package to have it lazy evaluated and stop. Now, let's have a generator for filtering empty modules away. Finally, let's retrieve :
```python
packages_gen = (
    task['action'].get(key) for key in self._package_name_keys
)
non_empty_packages_gen = (
    package_name for package_name in packages_gen
    if package_name
)
try:
    package_name = next(non_empty_packages_gen)
except StopIteration:
    return True  # there's no non-empty packages
```

First two steps could be
```python
packages_gen = map(task['action'].get, self._package_name_keys)
non_empty_packages_gen = filter(bool, packages_gen)
```
if you prefer FP style but most of Pythonistas agree that it's non-pythonic.

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,31 @@ class PackageHasRetryRule(AnsibleLintRule):
...
+            package_name = task['action'].get(key)
+            if package_name:
+                break
+            package_name = ''
+        is_package_file = '.' in package_name
+        is_package_html = '://' in package_name
+        is_local_package_file = is_package_file and not is_package_html
+        if is_local_package_file:
```
webknjaz: This is probably better as:
```suggestion
        return not is_local_package_file
```

lib/ansiblelint/rules/PackageHasRetryRule.py
awcrosby: I agree it's not readable enough as-is, I'm thinking this approach might be best for future readers:
```
        for key in self._package_name_keys:
            package_name = task['action'].get(key)
            if package_name:
                break
        if package_name:
            is_package_file = '.' in package_name
            is_package_html = '://' in package_name
            is_local_package_file = is_package_file and not is_package_html
            if is_local_package_file:
                return False
```

webknjaz: It's not a good idea to "side-effect" vars outside of for-block.
If you want to keep it as a loop block, better put it into a function returning that package name then.

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,37 @@ class PackageHasRetryRule(AnsibleLintRule):
...
+        is_state_whitelisted = task['action'].get('state') in self._module_ignore_states
+        if is_state_whitelisted:
+            return False
+
+        # attempt to check if this is a local package file
+        for key in self._package_name_keys:
+            found_package_name = task['action'].get(key)
+            if found_package_name:
+                break
+        if found_package_name:
```
webknjaz: Flip this to
```suggestion
        if not found_package_name:
            return True
```
to maintain low nesting.

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,43 @@ class PackageHasRetryRule(AnsibleLintRule):
...
+        "pkg",
+        "deb",
+    ]
+
+    # attempt to find package name
+    def get_package_name(self, action):
+        for key in self._package_name_keys:
+            found_package_name = action.get(key)
+            if found_package_name:
+                break
```
webknjaz: @awcrosby Better return right away. It's safer
```suggestion
                return found_package_name
```

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,12 +76,43 @@ class PackageHasRetryRule(AnsibleLintRule):
...
     ]
 
+    _package_name_keys = [
+        "name",
+        "package",
+        "pkg",
+        "deb",
+    ]
+
+    # attempt to find package name
```
webknjaz: This should go to docstring

PR #428, opened 2018-12-12
Fixing remediation message since the filtered success syntax is depre…

ahuffman:
…cated in Ansible 2.9

Full Ansible warning message if using `my_result | success` is as follows:
`[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using `result|success` use `result is success`. This feature will be 
removed in version 2.9. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.`

### DIFF
--- a/lib/ansiblelint/rules/PackageHasRetryRule.py
+++ b/lib/ansiblelint/rules/PackageHasRetryRule.py
@@ -12,7 +12,7 @@ class PackageHasRetryRule(AnsibleLintRule):
         'network communication and the availability of remote '
         'servers. To mitigate the potential problems, retries '
         'should be used via '
-        '``register: my_result`` and ``until: my_result | success``'
+        '``register: my_result`` and ``until: my_result is succeeded``'
     )
     severity = 'LOW'
     tags = ['module', 'reliability']


webknjaz:
@ahuffman could you please sign off https://github.com/ansible/ansible-lint/pull/428/checks?check_run_id=40528277 ?

lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -12,7 +12,7 @@ class PackageHasRetryRule(AnsibleLintRule):
         'network communication and the availability of remote '
         'servers. To mitigate the potential problems, retries '
         'should be used via '
-        '``register: my_result`` and ``until: my_result | success``'
+        '``register: my_result`` and ``until: my_result is success``'
```
willthames: Should be `is successful` (I know `is success` works but it's not grammatically correct)

ahuffman: @willthames is it succeeded or successful?  Ansible 2.7 docs has it `succeeded`.  https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html.

Ansible's deprecation warning needs to be updated as well to the newer syntax: `[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using result|success use result is success. `

willthames: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/test/core.py#L170-L177 suggests all will work. I'd argue `is successful` is more readable than `is succeeded`. 

The [2.5 porting guide](https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/porting_guides/porting_guide_2.5.rst#jinja-tests-used-as-filters) suggests `successful` as the way forward but I don't disagree that it's inconsistent in both documentation and in deprecation messages.

awcrosby:
Merging this as-is... since `succeeded` and similar like `failed` and `skipped` have similar word structure, and 2.7 is the latter docs reference... if other direction comes out we can update accordingly.

PR #432, opened 2018-12-13
Fix checkdocs rst warning duplicate implicit target

awcrosby:
Local testing vs. jenkins gave different output for `python -m setup checkdocs check --metadata --restructuredtext --strict --verbose`. This rst edit attempts to allow jenkins deployment checks to pass, for our rst to work in pypi.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -281,8 +281,8 @@ The following values are supported, and function identically to their CLI counte
     verbosity: 1
 
 
-Pre-commit
-----------
+Pre-commit Setup
+----------------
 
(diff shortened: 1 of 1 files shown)


webknjaz:
@awcrosby GitHub nor PyPI support sphinx extensions like including other documents. That's why it didn't work. And also you might've had env difference thing if you tried to run it without tox.

webknjaz:
Oh, and current tox envs rely on what `python` command points to where you run it. In Travis CI it's `python3.7`.

PR #436, opened 2018-12-17
Refactor PackageHasRetryRule, avoid false positive with apt_key data

### LINKED ISSUE #434: false positive on apt_key with data field
# Issue Type
- Bug report

```
$ ansible --version
ansible 2.7.4
  config file = /home/lae/.ansible.cfg
  configured module search path = ['/home/lae/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.7.1 (default, Oct 22 2018, 10:41:28) [GCC 8.2.1 20180831]
$ ansible-lint --version
ansible-lint 4.0.0a1
```

- ansible installation method: OS

# Desired Behaviour

Rule 405 is meant for remote connections but using `apt_key` with the `data` field doesn't require network connectivity (not sure if there's an appropriate network lookup, but if so, that would be an exception).

# Actual Behaviour

```yaml
- name: Trust Proxmox' packaging key
  apt_key:
    data: "{{ lookup('file', pve_release_key) }}"
    id: "{{ pve_release_key_id }}"
    state: present
```

The above results in the following.

```
[405] Remote package tasks should have a retry
/home/lae/src/ansible-role-proxmox/tasks/main.yml:47
Task/Handler: Trust Proxmox' packaging key
```

awcrosby:
Resolves #434

### DIFF
--- a/lib/ansiblelint/rules/PackageHasRetryRule.py
+++ b/lib/ansiblelint/rules/PackageHasRetryRule.py
@@ -85,6 +85,7 @@ class PackageHasRetryRule(AnsibleLintRule):
         "package",
         "pkg",
         "deb",
+        "key",
     ]
 
     def get_package_name(self, action):


lib/ansiblelint/rules/PackageHasRetryRule.py
```diff
@@ -76,19 +76,24 @@ class PackageHasRetryRule(AnsibleLintRule):
...
 
-    # attempt to find package name
     def get_package_name(self, action):
+        """Attempt to find package name."""
         for key in self._package_name_keys:
             found_package_name = action.get(key)
             if found_package_name:
-                break
+                return found_package_name
         return found_package_name
```
webknjaz: This statement will always return `None` so it's unneeded.

PR #444, opened 2018-12-20
Rule 405 now ignore case of 'yum: list=package'

zigarn:
https://docs.ansible.com/ansible/latest/modules/yum_module.html
`list` is for retrieving available packages, not sure it needs the `until` loop.

### DIFF
--- a/lib/ansiblelint/rules/PackageHasRetryRule.py
+++ b/lib/ansiblelint/rules/PackageHasRetryRule.py
@@ -76,9 +76,14 @@ class PackageHasRetryRule(AnsibleLintRule):
         "absent",
     ]
 
-    _module_ignore_parameters = [
-        "data",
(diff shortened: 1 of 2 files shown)


awcrosby:
I think this is a reasonable addition. @zigarn can you [signoff](https://github.com/ansible/ansible-lint/pull/444/checks?check_run_id=43248243) your commit?

zigarn:
Signed-off

alfredopalhares:
This rule is not on `405`, is not yet on https://docs.ansible.com/ansible-lint/rules/default_rules.html

zigarn:
> This rule is not on `405`, is not yet on https://docs.ansible.com/ansible-lint/rules/default_rules.html

@alfredopalhares: Is not _anymore_: #465

Issue #449, opened 2018-12-27
ansible-lint 4.0.0 does not support py35 or py34


ssbarnea:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version
ansible-lint --version
```

- ansible installation method: one of source, pip, OS package
- ansible-lint installation method: one of source, pip, OS package

# Desired Behaviour

ansible-lint should run with any Python version that is currently used on production systems, especially on the Red Hat distributions. 

Today I had a big surprise to see that installation failed because current python was py35. CentOS 7 has an optional **python34** and **python35**, python34 being the default python3 interpreter! (installable via epel repository).

# Actual Behaviour (Bug report only)
```
ansible-lint requires Python '>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*' but the running Python is 3.5.2
```
Example: http://logs.openstack.org/45/627545/1/check/openstack-tox-linters/e290e9d/job-output.txt.gz

So as long CentOS 7 and RHEL 7 are under normal or extended support, ansible-lint should keep compatibility with them, which should not be a big deal.

awcrosby:
The supported versions of ansible allow python35 so ansible-lint should allow that. We don’t test  ansible-lint with python34 since ansible doesn’t support it, but we could still allow ansible-lint to be installed. PR #452 allows ansible-lint installation with python35 and python34. It also adds tox testing for python35. @webknjaz any issues with adding this back in? We can do a bug fix release by the end of the week.

awcrosby:
All good points. I will add py35 in our tox testing and allow it in the `python_requires`. I will hold off adding py34 since we cannot test with the supported `ansible` versions.

PR #452, opened 2019-01-01
Allow install with python35 and add to tox testing

awcrosby:
Resolves: #449 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,11 @@ jobs:
     env:
       ANSIBLE_VERSION: "27"
 
+  - python: "3.5"
+    <<: *if-cron-or-manual-run-or-tagged
+    env:
+      ANSIBLE_VERSION: "27"
+
   - python: "2.7"
     env:
       ANSIBLE_VERSION: "27"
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.5.3
-envlist = flake8,py{37,36,27}-ansible{27,26,25,devel}
+envlist = flake8,py{37,36,35,27}-ansible{27,26,25,devel}
```
webknjaz: travis config would also have to be modified to test this env if we go this road.

.travis.yml
```diff
@@ -84,6 +99,11 @@ jobs:
     env:
       ANSIBLE_VERSION: devel
 
+  - python: "3.5"
+    <<: *if-cron-or-manual-run-or-tagged
+    env:
+      ANSIBLE_VERSION: devel
+
   - python: "2.7"
     <<: *if-cron-or-manual-run-or-tagged
```
awcrosby: @webknjaz should we remove these mixtures so each PR raise/update has all environment combinations tested? It looks like all were tested by travis prior to 4.0.0

webknjaz: This was done to avoid overloading usage of Travis CI which has limits and queues not only per repo but also per org. That's why I've set such if-clauses. Keep in mind that we also have Molecule which already consumes a tremendous amount of its resources despite a lot more limitations which I've applied there.

PR #460, opened 2019-01-10
Support skipping specific rule(s) for a specific task

### LINKED ISSUE #364: Feature request - Skip specific rule(s) for a specific task
# Issue Type
- Feature request

```
ansible 2.4.0.0
  config file = /root/.ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.5 (default, Nov  6 2016, 00:28:07) [GCC 4.8.5 20150623 (Red Hat 4.8.5-11)]

ansible-lint 3.4.23
```

# Desired Behaviour

Currently, we can skip linting on a task by adding `skip_ansible_lint` as a tag to the task. But I'm looking for something more granular than that. I want to be able to skip/ignore a specific rule (ie. `ANSIBLE00XX`) on a task-level basis. For example, something like:

```yaml
- hosts: ...
  tasks:
    - name: Install stuff
      yum:
        name: ...
        state: latest
      tags:
        - skip_ansible_lint_rule_ANSIBLE0010
```

Syntax is obviously tentative and subject to change.

In this example, I may indeed want to always install the `latest` version of a package, for whatever reason. I only want to ignore ANSIBLE0010 ("Package installs should not use latest"), but keep other linting like trailing whitespace.

awcrosby:
Closes #364


README.rst
```diff
@@ -351,35 +351,43 @@ It's also possible to skip specific rules by passing the rule ID. For example, t
...
+-------------------------------
 
 Some rules are a bit of a rule of thumb. Advanced *git*, *yum* or *apt* usage, for example, is typically difficult to achieve through the modules. In this case, you should mark the task so that warnings aren't produced.
 
-There are two mechanisms for this - one works with all tasks, the other works with the command checking modules.
+To skip a specific rule for a specific task, inside your ansible yaml add ``# noqa [rule_id]`` at the end of the line. If the rule is task-based (most are), add at the end of any line in the task. You can skip multiple rules via a space-separated list.
 
-Use the ``warn`` parameter with the *command* or *shell* module.
-
-Use ``skip_ansible_lint`` tag with any task that should be skipped.
```
Rickkwa: All documentation for skip_ansible_lint is being removed? To my understanding, the feature is still there, no?

awcrosby: It is not being removed (but may be a good candidate for deprecation in the future). I will update the docs to say it is a less-preferred method to skip.

lib/ansiblelint/__init__.py
```diff
@@ -50,53 +50,76 @@ def matchlines(self, file, text):
...
+            if not result:
+                continue
+            message = None
+            if isinstance(result, str):
+                message = result
+            matches.append(Match(prev_line_no+1, line,
+                           file['path'], self, message))
         return matches
 
     def matchtasks(self, file, text):
```
webknjaz: Not strictly related to the current change, but `file` should be `file_` or something that doesn't shadow builtins.

lib/ansiblelint/__init__.py
```diff
@@ -50,53 +50,76 @@ def matchlines(self, file, text):
...
-                        if isinstance(result, six.string_types):
-                            message = result
-                        taskstr = "Task/Handler: " + ansiblelint.utils.task_to_str(task)
-                        matches.append(Match(task[ansiblelint.utils.LINE_NUMBER_KEY], taskstr,
-                                       file['path'], self, message))
+        if not yaml:
+            return matches
+
+        for task in ansiblelint.utils.get_normalized_tasks(yaml, file):
+            if 'skipped_rules' in task and self.id in task['skipped_rules']:
```
webknjaz:
```suggestion
            if self.id in task.get('skipped_rules', ()):
```

lib/ansiblelint/__init__.py
```diff
@@ -50,53 +50,76 @@ def matchlines(self, file, text):
...
-            if result:
-                message = None
-                if isinstance(result, str):
-                    message = result
-                matches.append(Match(prev_line_no+1, line,
-                               file['path'], self, message))
+            if not result:
+                continue
+            message = None
+            if isinstance(result, str):
```
webknjaz: This thing is inconsistent with similar check below where `six.string_types` is used. Needs python compat check.

lib/ansiblelint/__init__.py
```diff
@@ -50,53 +50,76 @@ def matchlines(self, file, text):
...
+            if 'action' not in task:
+                continue
+            result = self.matchtask(file, task)
+            if not result:
+                continue
+
+            message = None
+            if isinstance(result, six.string_types):
+                message = result
+            taskstr = "Task/Handler: " + ansiblelint.utils.task_to_str(task)
```
webknjaz: snake_case plz
```suggestion
            task_msg = "Task/Handler: " + ansiblelint.utils.task_to_str(task)
```

lib/ansiblelint/__init__.py
```diff
@@ -50,53 +50,76 @@ def matchlines(self, file, text):
...
-                        matches.append(Match(play[ansiblelint.utils.LINE_NUMBER_KEY],
-                                             section, file['path'], self, message))
+        if not yaml:
+            return matches
+
+        if isinstance(yaml, dict):
+            yaml = [yaml]
+
+        for play in yaml:
+            if 'skipped_rules' in play and self.id in play['skipped_rules']:
```
webknjaz: Same as above: use `.get()`

lib/ansiblelint/__init__.py
```diff
@@ -50,53 +50,76 @@ def matchlines(self, file, text):
...
+
+            result = self.matchplay(file, play)
+            if not result:
+                continue
+
+            if isinstance(result, tuple):
+                result = [result]
+
+            if not isinstance(result, list):
+                raise Exception("{} is not a list".format(result))
```
webknjaz: We need to be more precise for exceptions to be useful:
```suggestion
                raise TypeError("{} is not a list".format(result))
```

lib/ansiblelint/utils.py
```diff
@@ -573,3 +575,82 @@ def construct_mapping(node, deep=False):
...
+        skipped_rules = _get_rule_skips_from_task(task)
+        if skipped_rules:
+            task['skipped_rules'] = skipped_rules
+            any_skipped_rules = True
+
+    if not any_skipped_rules:
+        return orig_file_text
+
+    # dump to file with new skipped_rules list(s), read file and return text
+    with tempfile.NamedTemporaryFile('w') as tf:
```
webknjaz: Use in-memory file-like buffers instead.

webknjaz: You can rely on `six.StringIO` (or `six.BytesIO`?)

It should look like:
```python
import six
with six.StringIO() as in_memory_file:
    yaml.dump(data, in_memory_file)
    in_memory_file.seek(0)
    new_file_text = in_memory_file.read()
```

(needs testing)

lib/ansiblelint/utils.py
```diff
@@ -573,3 +575,82 @@ def construct_mapping(node, deep=False):
     except (yaml.parser.ParserError, yaml.scanner.ScannerError) as e:
         raise SystemExit("Failed to parse YAML in %s: %s" % (filename, str(e)))
     return data
+
+
+def add_skipped_rules(orig_file_text, file_type):
+    yaml = ruamel.yaml.YAML()
+    data = yaml.load(orig_file_text)
```
webknjaz: load or load`s`?

lib/ansiblelint/utils.py
```diff
@@ -573,3 +575,82 @@ def construct_mapping(node, deep=False):
...
+            new_file_text = tf_read.read()
+
+    return new_file_text
+
+
+def _get_rule_skips_from_task(task):
+    def traverse_yaml(obj):
+        task_comment_obj_strs.append(str(obj.ca.items))
+        if isinstance(obj, dict):
+            for k in obj:
```
webknjaz:
```suggestion
            for k, v in obj.items():
```

lib/ansiblelint/utils.py
```diff
@@ -573,3 +575,82 @@ def construct_mapping(node, deep=False):
...
+                if isinstance(obj[k], (dict, list)):
+                    traverse_yaml(obj[k])
+        elif isinstance(obj, list):
+            for e in obj:
+                if isinstance(e, (dict, list)):
+                    traverse_yaml(e)
+        else:
+            return
```
webknjaz: This returns `None`. Shouldn't it return `obj`?

ssbarnea:
@webknjaz There are not linters supporting noqa for yaml files so far, so we are safe. Even if there would be a collision us very unlikely to cause problems, worst case would be to disable the same rule on two linters. 

I am more than happy to see the same syntax being used. The 2nd linter would have to play nice with ansible-lint not the other way around ;)

Most of the time noqa is used without parameters so no real reasons to worry. I am also worried that any longer requirements would cripple the UX because you may exceed the max line lengh. `  # noqa` adds 8 chars.

lib/ansiblelint/utils.py
willthames: Also, should we use safe_load?

setup.cfg
```diff
@@ -83,6 +83,7 @@ install_requires =
     ansible
     pyyaml
     six
+    ruamel.yaml
```
willthames: What is the reliance on ruamel for?

awcrosby: These 2 steps were added between existing ansible-lint functionality
1. (existing) `RulesCollection.run` reads yml file into text
2. (new) `ruamel.yaml` parses text while retaining access to comments through it’s python object
3. (new) `ansiblelint.utils.add_skipped_rules` looks at those comments for any that match a rule skip, if found it adds a `skipped_rules` key to the task with a list of rule ids. `ruamel.yaml` then dumps it’s serialized yaml back into text
4. (existing) Pass text into each rule’s match method such as `matchtasks` which parses using the `yaml` and `ansible` packages (this pr doesn’t try to change any of that parsing)

We are using `yaml = ruamel.yaml.YAML()` and their [docs](https://yaml.readthedocs.io/en/latest/dumpcls.html#dumping-python-classes) state:
> Only yaml = YAML(typ='unsafe') loads and dumps Python objects out-of-the-box. And since it loads any Python object, this can be unsafe.

lib/ansiblelint/utils.py
awcrosby: `ruamel.yaml` uses `load` instead of `loads`

We are using `yaml = ruamel.yaml.YAML()` and their [docs](https://yaml.readthedocs.io/en/latest/dumpcls.html#dumping-python-classes) state:
> Only yaml = YAML(typ='unsafe') loads and dumps Python objects out-of-the-box. And since it loads any Python object, this can be unsafe.

lib/ansiblelint/utils.py
awcrosby: The recursive method takes `obj`, but appends to the list defined outside the method, so returning `None` is expected.

test/TestSkipInsideYaml.py
```diff
@@ -0,0 +1,114 @@
+import unittest
```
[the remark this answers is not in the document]
awcrosby: A refactor of ansible-lint test to use pytest overall would probably be a good exercise

lib/ansiblelint/utils.py
awcrosby: `io.BytesIO()` worked in py2 and py3

lib/ansiblelint/__init__.py
awcrosby: Opened #469 to address this

awcrosby:
Rebased and fixed 2 issues found in this PR:
1. The previous implementation was not retaining line numbers since it was adding to yaml text, now it appends skipped_rules to the pyyaml objects
2. On playbook, fixed to look for skipped rules in all plays

Issue #373, opened 2018-10-02
Lint all yaml in tasks/ and handlers/ regardless of import or include


awcrosby:
# Issue Type
- Feature request

# Ansible and Ansible Lint details

```
(vgal) awc:code$ ansible --version
ansible 2.6.5
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/home/awc/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/awc/code/vgal/lib/python2.7/site-packages/ansible
  executable location = /home/awc/code/vgal/bin/ansible
  python version = 2.7.15 (default, Sep 21 2018, 23:26:48) [GCC 8.1.1 20180712 (Red Hat 8.1.1-5)]
(vgal) awc:code$ ansible-lint --version
ansible-lint 3.4.23
```

- ansible installation method: pip
- ansible-lint installation method: pip

# Desired Behaviour
Currently when linting a role, ansible-lint looks at tasks in `tasks/main.yml` and `handlers/main.yml`, and follows task files imported or included. 

Proposal: to lint all yaml files inside `tasks/` and `handlers/` regardless of whether they have been imported or included from `main.yml`

With an include in `main.yml`, `databases.yml` is linted:
```
(vgal) awc:code$ ls role_test1/tasks/
databases.yml  main.yml
(vgal) awc:code$ cat role_test1/tasks/main.yml 
---
- name: using git module
  action: git repo=blah

- include: databases.yml
(vgal) awc:code$ ansible-lint -p -r galaxy-lint-rules/rules/ role_test1
/home/awc/code/role_test1/tasks/databases.yml:2: [E105GAL] Deprecated with_X for loops
/home/awc/code/role_test1/tasks/main.yml:4: [E401] Git checkouts must contain explicit version
```

Currently without the include in `main.yml`, `databases.yml` is not linted:
```
(vgal) awc:code$ cat role_test1/tasks/main.yml 
---
- name: using git module
  action: git repo=blah
(vgal) awc:code$ ansible-lint -p -r galaxy-lint-rules/rules/ role_test1
/home/awc/code/role_test1/tasks/main.yml:4: [E401] Git checkouts must contain explicit version

```

lwm:
> Proposal: to lint all yaml files inside tasks/ and handlers/ regardless of whether they have been imported or included from main.yml

Sounds good to me :+1:

willthames:
[...]

PR #462, opened 2019-01-11
 Lint all yaml in tasks/ and handlers/ regardless of import or include

AlekseiSidorenko:
https://github.com/ansible/ansible-lint/issues/373


lwm:
After a cursory pass on this, LGTM <3

PR #465, opened 2019-01-16
Remove rule 405 checking for retry on package modules

### LINKED ISSUE #456: [405] Remote package tasks should have a retry - too nitpicky?
# Issue Type

- Feature request

```
$ ansible --version
ansible 2.7.5
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/Users/jgeerling/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python2.7/site-packages/ansible
  python version = 2.7.15 (default, Jul 17 2018, 19:41:20) [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)]

$ ansible-lint --version
ansible-lint 4.0.0
```

# Desired Behaviour

Using package modules (`apt`, `yum`, `dnf`, `package`, etc.) should not require the user to always add 3+ extra lines of configuration per use to resolve 405.

Rather, if it's truly important enough that we add retries that it should be a lint warning... maybe it should be the responsibility of ansible itself, and it could be a module parameter?

So I guess this is more of a 'question request'—since 405 hits pretty much every playbook, role, and tiny ansible script I've ever written, and would require me adding thousands of lines of the exact same thing to all those hundreds of instances:

```yaml
register: task_result
until: task_result is success
retries: 10
delay: 2
```

...might it not be better to make this the responsibility of Ansible itself? Otherwise, this is one of those warnings I think many might just drop. Seems extremely nitpicky to me (even when rated as a `LOW`).

The task:

```
- name: Update apt cache (Debian).
  apt: update_cache=true cache_valid_time=86400
  when: ansible_os_family == 'Debian'
```

(or any task with apt, yum, dnf, package, etc.)

results in:

```
[405] Remote package tasks should have a retry
/Users/jgeerling/Dropbox/VMs/roles/geerlingguy.git/tasks/main.yml:2
Task/Handler: Ensure git is installed (RedHat).
```

awcrosby:
Closes #456 


docs/docsite/rst/rules/default_rules.rst
```diff
@@ -8,51 +8,50 @@ Default Rules
...
-`E405 <https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/PackageHasRetryRule.py>`_                                                                                                                                  v4.0.0                                                                                                                                                                                                                                      Remote package tasks should have a retry                                                                                                                                                                                                    Package operations are unreliable as they require network communication and the availability of remote servers. To mitigate the potential problems, retries should be used via ``register: my_result`` and ``until: my_result | success``
```
aakarshg: Minor nit: seems that because the formatting was changed, github shows 11-57 lines as changed, when in reality only line 41 needs to be dropped. Perhaps this is normal and expected but it caught my eye so pointing it out.

awcrosby: Thanks for this catch - I think the rst -> html script had not be run since a formatting change in the script. In comparing the output, only 405 has changed so I think this is ok.

awcrosby: Ah yes the script changes cell width based on all data, so all lines are altered.

PR #478, opened 2019-02-06
Check ANSIBLE_ROLES_PATH before basedir

pilou-:
Check `ANSIBLE_ROLES_PATH` directories before `basedir`.

The two commands below should report the same output:
* `cd test && ansible-lint roles/test-role`
* `cd test/roles/test-role && ANSIBLE_ROLES_PATH=.. ansible-lint molecule/default/include-import-role.yml` (this is the way `molecule` calls `ansible-lint`)
but only the first checks tasks inside the `test-role` role.

This regression has been introduced by 88e9114eeaff9758dfe491f93aa5a647969fdf51 which have been included in `v4.0.0a0`.

Unit test included.

Fix ansible/molecule#1723.


pilou-:
I will push the 3rd commit (the bugfix) once the Travis build result is available (failure expected).

PR #483, opened 2019-02-18
Add an official GitHub Action to the project

webknjaz:
This will unlock GitHub Actions and workflows usage by playbook/role writers making it possible to easily run ansible-lint in their repos as a part of their workflows.

This is a very basic action which I hope to improve to use Checks API in future :)


willthames:
Nice, I love this!

.github/action/license
```diff
@@ -0,0 +1,21 @@
+The MIT License
+
+Copyright (c) Stefan Stölzle, http://stefan.stoelzle.me
```
webknjaz: @awcrosby I'm not sure how to correctly handle licensing here. Should we check this with Legal?

cc @stoe

webknjaz:
Great! I'm merging this and I hope that this can be extended to be better configurable and to post nice Checks API status updates to GitHub

.github/action/license
awcrosby: @webknjaz maybe so, maybe @gundalow could add thoughts

stoe: I’m happy to remove my name from that original license, but would appreciate some credits and maybe a similar license for it going forward.

stoe:
Did think about the ChecksAPI as well but then got swompt with other work. 
Let me know if I can help with the implementation.

webknjaz:
Actually, I have a CLI tool called `check-in` implementing Checks API interaction. Except it uses GitHub Apps integration. But yes, we'll need to figure out the best way to do it :)

I've also seen this implemented in JS: probot supports it out of the box...

webknjaz:
@willthames do you want to check it out on your repos?

.github/action/license
webknjaz: @stoe yeah, that's what I was thinking. I thought that maybe we don't need two licenses in the project because it's confusing. OTOH I've already put your name in the README in this dir. We'll make sure to keep the mention as we add info to docs or the main readme. I've even used git subtree split in order to properly extract your commits and add them here so that your history would be preserved :)

stoe:
Yeah. 

> Write permission for the Checks API is only available to GitHub Apps. OAuth Apps and authenticated users can view check runs and check suites, but they are not able to create them. 
via https://developer.github.com/v3/checks/runs/

So only GitHub Apps work.

webknjaz:
@stoe no, token exposed to Actions also works. It's just that my CLI tool uses a private key and there's no code for skipping that step and using a token right away.

.github/action/license
awcrosby: @webknjaz also with the root license being the same as this, I agree one license is less confusing, and we can maintain the mention in this dir's README.

webknjaz:
Hm... GitHub Developers docs recommend having separate action repos. I'm curious whether I can get it to Marketplace.

It's not in list https://github.com/marketplace/new for me and @jlaska doesn't see it there either. So they might be only supporting per-action repos...

stoe:
> It's not in list https://github.com/marketplace/new for me and @jlaska doesn't see it there either. So they might be only supporting per-action repos...

The issue might be that the action is in the "magic" `.github` folder which is not looked into when https://github.com/marketplace/new displays the list of possible Actions to be added.

webknjaz:
@stoe are you sure?

stoe:
> are you sure?

Not 💯, no.

stoe:
@webknjaz, I talked to our Marketplace product team, and here is what they said:

> **"One Repo = One Action" on Marketplace** 
> It allows to use git tags/releases for versioning.
> Follows a similar pattern as NPM.
> Plus  repository meta data (e.g. readme.md) can be used to populate Marketplace pages with info.

Now I'm sure, that it has to be it's own repo 😉

webknjaz:
@stoe thanks for clarifying with them. So I think we should move it to https://github.com/ansible/ansible-lint-action then...

webknjaz:
Hey @stoe, I've imported it to https://github.com/ansible/ansible-lint-action. Now I'm thinking about versioning. Do you have any thoughts wrt this?

Your original dockerfile has a version label, but also actions are versioned using git tags and it's not listed in the recommended labels list here either https://developer.github.com/actions/creating-github-actions/creating-a-docker-container/#label.
Should I remove it?

Also, another question is whether versioning should be aligned with the versioning of the ansible-lint itself. What do you think? Maybe I'd just create an "unversioned" or "latest" branch for folks who want that and tag + change dockerfile to install pinned versions in master? Another way might be vice versa: unversioned master + branches from it with a commit to change the pinned version in dockerfile + tag on it.

P.S. Do you want to co-maintain the action? If so, I think we could add you to that repo.

webknjaz:
https://github.com/marketplace/actions/ansible-lint

stoe:
> Your original dockerfile has a version label

Don’t remember why I did this 🤔

> Should I remove it?

Seems fine to me to remove, @webknjaz

> Do you want to co-maintain the action? If so, I think we could add you to that repo.

Honoured to.

webknjaz:
@stoe cool, I've invited you as a collaborator!

Do you happen to know why some people see error 404 on the marketplace page?

stoe:
> Do you happen to know why some people see error 404 on the marketplace page?

Currently only users who have signed up for the GitHub Actions Beta can see them on Marketplace. 

One can request to join the Beta over at https://github.com/features/actions/signup/

webknjaz:
Ah... So users in organizations where it's enabled don't count?

It's weird because they see other Actions on Marketplace, just not this one.

stoe:
> Ah... So users in organizations where it's enabled don't count?
> 
> It's weird because they see other Actions on Marketplace, just not this one.

I just tested with a temp user account that does not have the beta signup, and I can see no actions via https://github.com/marketplace?type=actions. It's only GitHub Apps that are listed.

image

vs my @stoe user

image

The behaviour is strange nonetheless.
I'll let our Marketplace team know.

stoe:
> It's weird because they see other Actions on Marketplace, just not this one.

@webknjaz, can you ask them to share a screenshot or a little gif, please?

webknjaz:
Sure, let's ask @jctanner and @mattclay, and maybe @rcarrillocruz

mattclay:
I've requested beta access for my account, but I'm not currently signed up, so I see the shorter list of only 61 actions which does not include ansible-lint.

webknjaz:
@mattclay can you show the screenshot? @stoe claims that you shouldn't be seeing any actions in the Marketplace at all in this case...

mattclay:
@webknjaz I see the same as in the first screenshot from @stoe showing 61 results.

webknjaz:
Ah, so that view is misleading. GitHub should show "no results" for Actions filter for such users but it shows results tricking them into thinking that those are Actions while they are not...

mscoutermarsh:
@webknjaz We shipped some updates based on your feedback.

https://github.com/marketplace/actions/ansible-lint now shows up for everyone.

And

https://github.com/marketplace?type=actions is only available for people in the Actions Beta.

Thanks 😄

webknjaz:
@mscoutermarsh thanks!

Issue #487, opened 2019-03-06
NoFormattingInWhenRule  shouldn't crash on empty roles:


bsibaud:
# Issue Type
- Bug report

# Ansible and Ansible Lint details
ansible 2.7.8
ansible-lint 4.1.0

- ansible installation method: pip
- ansible-lint installation method: pip

# Desired Behaviour

NoFormattingInWhenRule  shouldn't crash on the following playbook:

```
---
- hosts: localhost
  roles:
```

# Actual Behaviour (Bug report only)

ansible-lint previous_playbook.yml

```
Traceback (most recent call last):
(...)  File "/home/xxx/TEST/local/lib/python2.7/site-packages/ansiblelint/__init__.py", line 122, in matchyaml
    result = self.matchplay(file, play)
  File "/home/xxx/TEST/local/lib/python2.7/site-packages/ansiblelint/rules/NoFormattingInWhenRule.py", line 29, in matchplay
    for role in play['roles']:
TypeError: 'NoneType' object is not iterable
```

Fix:

```
--- orig/local/lib/python2.7/site-packages/ansiblelint/rules/NoFormattingInWhenRule.py   2019-03-06 15:38:39.815863000 +0000
+++ new/local/lib/python2.7/site-packages/ansiblelint/rules/NoFormattingInWhenRule.py   2019-03-06 15:29:11.755855540 +0000
@@ -24,6 +24,9 @@
         if isinstance(play, dict):
             if 'roles' not in play:
                 return errors
+            elif play['roles'] is None:
+                errors.append(('Play: empty roles:', self.shortdesc))
+                return errors
             for role in play['roles']:
                 if self.matchtask(file, role):
                     errors.append(({'when': role},
```

awcrosby:
Closed via https://github.com/ansible/ansible-lint/pull/495

PR #495, opened 2019-03-20
Don't crash on empty roles.

sargon:
Rule NoFormattingInWhen is crashing when rules is empty, which is
common in developing playbooks. 

Signed-off-by: Daniel Ehlers <sargon@toppoint.de>

This solves #487 but without the suggested warning.


ssbarnea:
@webknjaz @awcrosby ^ GTG. Thanks.

Issue #499, opened 2019-03-26
VariableHasSpace Rule matches {{- var }}


netzvieh:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version
ansible 2.7.9
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/sebastian/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/sebastian/.venvs/a2.7/lib/python3.6/site-packages/ansible
  executable location = /home/sebastian/.venvs/a2.7/bin/ansible
  python version = 3.6.5 (default, Mar 31 2018, 19:45:04) [GCC]

ansible-lint --version
ansible-lint 4.1.0
```

- ansible installation method: pip
- ansible-lint installation method: pip

# Desired Behaviour

ansible-lint should allow for Jinja whitespace control, e.g. `{{-` and `-}}`.

The reason for using yaml multiline string here is, that I need to build a string without newlines from multiple variables with defaults from other variables which leads to the line being longer than 160 chars.

# Actual Behaviour (Bug report only)

```
---
- hosts: localhost
  vars:
    fooo: asadsafdsafdsafdsafadsfdsa
    bar: 15.6
    baz: x86_64

  tasks:
    - debug:
        msg: |
          {{ fooo }}/
          {{- bar }}/
          {{- baz }}
```

```
ansible-lint /tmp/abc.yml
[206] Variables should have spaces before and after: {{ var_name }}
/tmp/abc.yml:12
          {{- bar }}/

[206] Variables should have spaces before and after: {{ var_name }}
/tmp/abc.yml:13
          {{- baz }}
```

ssbarnea:
I confirm that I enounter this bug in an alternative way: `{{ var_name -}}` as dash suffix is also part of the jinja2 specification and AFAIK there is no workaround.

awcrosby:
Closed by #500

PR #500, opened 2019-03-26
Allow Jinja2 whitespace control chars in vars

netzvieh:


netzvieh:
Fixes #499

ssbarnea:
@webknjaz Can you have a look at this?

lib/ansiblelint/rules/VariableHasSpacesRule.py
```diff
@@ -13,7 +13,7 @@ class VariableHasSpacesRule(AnsibleLintRule):
     tags = ['formatting']
     version_added = 'v4.0.0'
 
-    bracket_regex = re.compile("{{[^{' ]|[^ '}]}}")
+    bracket_regex = re.compile("{{[^{' -]|[^ '}-]}}")
```
webknjaz: please use raw-string literals for regexps
```suggestion
    bracket_regex = re.compile(r"{{[^{' -]|[^ '}-]}}")
```

test/TestVariableHasSpaces.py
```diff
@@ -11,11 +11,16 @@
...
   debug:
     msg: "{{ '{{' }}"
-- name: jijna escaping allowed
+- name: jinja escaping allowed
   shell: docker info --format '{{ '{{' }}json .Swarm.LocalNodeState{{ '}}' }}' | tr -d '"'
+- name: jinja whitespace control allowed
+  debug:
+    msg: |
+      {{ good_format }}/
+      {{- good_format }}
```
webknjaz: extend this test with @ssbarnea's one

lib/ansiblelint/rules/VariableHasSpacesRule.py
webknjaz: I'm not sure why this regexp does this check. Can anyone explain it to me? It seems wrong (even the original one)

test/TestVariableHasSpaces.py
ssbarnea: @netzvieh I hope you can make the small fix that was requested as I want to help merging this bugfix.

lib/ansiblelint/rules/VariableHasSpacesRule.py
netzvieh: As I understand we want to match everything that's starting with `{{` not followed by space or hyphen. Also we want to allow `{{ '{{' }}` so we need to allow single quotes after `{{` aswell. Same thing for `}}`. I'm not sure why we wouldn't want to match `{{{` or `}}}` though.

Maybe @awcrosby can tell us why it's needed?

Problem is, this actually allows doing `{{'asdas'}}` as `{{'` and `'}}` are not matched. Same for `{{-asassa-}}`. But the regex to capture that would be something like: `({{-(?! ))|({{'(?! ))|({{[^ '-])|([^' -]}})|((?<=[^ ])-}})|((?<=[^ ])'}})` and I'm not positive we want to do that.

I'm not sure if python does allow the regex to be split into multiple lines? Then we probably could introduce it with `(# comment)` to explain it. Basically like:

```
({{-(?! ))|(# match everything with {{- not followed by a space OR)
({{'(?! ))|(# match everything with {{' not followed by a space OR)
({{[^ '-])|(# match everything with {{ not followed by a space, single quotation mark or hyphen OR)
([^' -]}})|(# match everything with }} not preceded by a space, single quotation mark or hyphen OR)
((?<=[^ ])-}})|(# match everything with -}} not preceded by a space OR)
((?<=[^ ])'}})(# match everything with '}} not preceded by a space)
```

netzvieh: It does work with `\` at EOL, though I'm still not sure, we want to do that.

test/TestVariableHasSpaces.py
netzvieh: Added.

awcrosby:
The allowing of `{{{` and `}}}` was added to the original rule in ansible-review before the rule was added to ansible-lint, I am not sure the purpose.

Thanks for pointing out this rule allows `{{'asdas'}}`, this may be ok for now as we are catching the most common bad formats, and limiting the number of false positives.

The PR looks good, I will merge this in - thanks for the contribution!

PR #501, opened 2019-03-28
Syntax proposed by 104 should not fail 206

netzvieh:
Introduce additional spaces in the UsingBareVariablesIsDeprecatedRule
proposal output, so it's not failing VariableHasSpacesRule.

Signed-off-by: Sebastian Meyer <meyer@b1-systems.de>


ssbarnea:
Thanks! Any chance to see a release soon?

PR #502, opened 2019-03-29
📌 Set min supported Ansible version @ dist meta

webknjaz:
I just realized that there's literally no place specifying that we don't support EOL versions of Ansible.

This should be set in the package distribution metadata, at least.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,7 +80,7 @@ setup_requires =
 
 # These are required in actual runtime:
 install_requires =
-    ansible
+    ansible >= 2.5
     pyyaml
     six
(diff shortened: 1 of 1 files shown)


awcrosby:
I can add a line to the docs as well

webknjaz:
@awcrosby yeah, I wasn't sure whether it's fine with docs because it'd be more places to update once 2.8 is released... More places means more forgotten updates :)

awcrosby:
Agreed on maintaining versions numbers in the docs, might be best to state simply we do not support EOL versions of ansible

PR #504, opened 2019-04-02
Allow exceptions in CommandsInsteadOfModules

garnser:
As there's multiple subcommands of yum that doesn't have an equivialent function there's a need to have exceptions available to don't hit the rule when for example versionlock is used in a command or shell.

At this time it doesn't seam like versionlock will be added to the yum module.

https://github.com/ansible/ansible/issues/13086


webknjaz:
Closing since there's no tests. Feel free to send a new PR once you have the tests.
If somebody else is up to finishing this, feel free to do so but keep in mind that the tests are mandatory.

PR #510, opened 2019-05-07
Dockerfile was added

lexa-uw:
Dockerfile for building docker images for this cool tool.
Like https://hub.docker.com/r/lexauw/ansible-lint
Usage `docker run --rm -it -v $PWD:/app -w /app lexauw/ansible-lint:v4.1.0 --help`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,13 @@
+FROM python:3.7-alpine3.9
+
+ENV ANSIBLE_LINT_VERSION=4.1.0
+
+RUN set -xe \
+    && apk add --no-cache --progress --virtual .build-deps gcc musl-dev libffi-dev openssl-dev make \
+    \
+    && pip install ansible-lint==${ANSIBLE_LINT_VERSION} \
+    \
+    && echo "****** Remove unused system librabies ******" \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.7-alpine3.9
+
+ENV ANSIBLE_LINT_VERSION=4.1.0
```
Zempashi: ENV is overwritable at run time, so will not change the version of ansible-lint. 
something like

```
docker run -ti -e ANSIBLE_LINT_VERSION=4.0.0 ansible-lint
```
will have an environment variable to 4.0.0 but the version of ansible-lint of 4.1.0.
It looks like, you're looking for 'ARG' keyword, and build-args to change the image at build-time.

webknjaz:
Thanks for your contribution but I think it doesn't belong in this repo.

There's a Dockerfile created for GitHub Actions: https://github.com/ansible/ansible-lint-action. I believe that we don't need another one. And also I think having a separate docker container for a tool like this is an overkill anyway. It's like deploying a static website on K8s: you can do it but why would you.

PR #517, opened 2019-05-23
fix skipping when using import_playbook

kouk:
The `import_playbook` item doesn't have the `tasks` property, so an exception is raised in utils.py line 669 (in `append_skipped_rules`). Although the exception is caught and ignored the side effect is that the rest of the method doesn't finish it's work, so the line based `noqa XXX` statements are ignored.

For example, the following playbook should pass by default:

```yaml
- hosts: all

  tasks:
    - name: should be shell  # noqa 305 301
      shell: echo lol

- import_playbook: imported_playbook.yml
```

Instead, the following two errors are raised:

```
ansible/playbooks/test3.yml:4: [E301] Commands should not change things if nothing needs doing
ansible/playbooks/test3.yml:4: [E305] Use shell only when shell functionality is required
```

To fix this, I added the empty list as a default value in the calls to `.get('tasks')`.

Signed-off-by: Konstantinos Koukopoulos <koukopoulos@gmail.com>

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -613,8 +613,8 @@ def append_skipped_rules(pyyaml_data, file_text, file_type):
             ruamel_tasks = []
             pyyaml_tasks = []
             for ruamel_play, pyyaml_play in zip(ruamel_data, pyyaml_data):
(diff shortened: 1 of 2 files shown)


awcrosby:
@kouk thanks for this fix, and for adding a test as well!

PR #528, opened 2019-05-31
Support inline rule skipping inside a block

### LINKED ISSUE #484: Unable to skip rules inside a block
# Issue Type
- Bug report

```
ansible --version
ansible 2.5.5
  config file = /Users/atsvetkov/.ansible.cfg
  configured module search path = [u'/Users/atsvetkov/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /Library/Python/2.7/site-packages/ansible
  python version = 2.7.10 (default, Oct  6 2017, 22:29:07) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.31)]

ansible-lint --version
ansible-lint 4.1.0
```

# Desired Behaviour
```
  - name: some block name
    block:
      - name: block_1
        shell: yum install vim  # noqa 303
```
ansilbe lint should ignore `yum install ...` warning.

ansible lint produces the following output:
`[303] yum used in place of yum module`

But, ` # noqa 303` option will work if we put the same task outside of a block.

awcrosby:
Fixes #484

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -629,6 +629,10 @@ def append_skipped_rules(pyyaml_data, file_text, file_type):
     else:
         raise RuntimeError('Unexpected file type')
 
+    # get tasks from blocks of tasks
+    pyyaml_tasks = _get_tasks_from_blocks(pyyaml_task_blocks)
+    ruamel_tasks = _get_tasks_from_blocks(ruamel_task_blocks)
+
     if len(ruamel_tasks) != len(pyyaml_tasks):
         return pyyaml_data
 
@@ -654,6 +658,24 @@ def _get_task_blocks_from_playbook(playbook):
         for key in PLAYBOOK_TASK_KEYWORDS:
             task_blocks.extend(play.get(key, []))
     return task_blocks
+
+
+def _get_tasks_from_blocks(task_blocks):
+    """Get list of tasks from list made of tasks and nested tasks."""
+    NESTED_TASK_KEYS = [
+        'block',
+        'always',
+        'rescue',
+    ]
+
+    tasks = []
+    for item in task_blocks:
+        for key in NESTED_TASK_KEYS:
+            if item.get(key):
+                tasks.extend(item[key])
+                continue
+        tasks.append(item)
+    return tasks
 
 
 def _get_rule_skips_from_task(task):
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -21,6 +21,7 @@
 import glob
 import imp
 import os
+from itertools import product
 
 import six
 from ansible import constants
@@ -614,9 +615,9 @@ def append_skipped_rules(pyyaml_data, file_text, file_type):
 
     try:
         yaml_skip = _append_skipped_rules(pyyaml_data, file_text, file_type)
-    except Exception as exc:
+    except RuntimeError as exc:
         # Notify user of skip error, do not stop, do not change exit code
-        print('Error trying to append skipped rules: {}'.format(repr(exc)))
+        print('Error trying to append skipped rules: {!r}'.format(exc))
         return pyyaml_data
     return yaml_skip
 
@@ -626,11 +627,8 @@ def _append_skipped_rules(pyyaml_data, file_text, file_type):
     yaml = ruamel.yaml.YAML()
     ruamel_data = yaml.load(file_text)
 
-    # if file type is 'meta', add skips to metadata block and return
     if file_type == 'meta':
-        skipped_rules = _get_rule_skips_from_yaml(ruamel_data)
-        if skipped_rules:
-            pyyaml_data[0]['skipped_rules'] = skipped_rules
+        pyyaml_data[0]['skipped_rules'] = _get_rule_skips_from_yaml(ruamel_data)
         return pyyaml_data
 
     # create list of blocks of tasks or nested tasks
@@ -647,7 +645,7 @@ def _append_skipped_rules(pyyaml_data, file_text, file_type):
             # call stack, and can remove this except
             return pyyaml_data
     else:
-        raise RuntimeError('Unexpected file type')
+        raise RuntimeError('Unexpected file type: {}'.format(file_type))
 
     # get tasks from blocks of tasks
     pyyaml_tasks = _get_tasks_from_blocks(pyyaml_task_blocks)
@@ -665,7 +663,11 @@ def _append_skipped_rules(pyyaml_data, file_text, file_type):
 
 
 def _get_task_blocks_from_playbook(playbook):
-    """Return parts of playbook that contains tasks, and nested tasks."""
+    """Return parts of playbook that contains tasks, and nested tasks.
+
+    :param playbook: playbook yaml from yaml parser.
+    :returns: list of task dictionaries.
+    """
     PLAYBOOK_TASK_KEYWORDS = [
         'tasks',
         'pre_tasks',
@@ -674,9 +676,8 @@ def _get_task_blocks_from_playbook(playbook):
     ]
 
     task_blocks = []
-    for play in playbook:
-        for key in PLAYBOOK_TASK_KEYWORDS:
-            task_blocks.extend(play.get(key, []))
+    for play, key in product(playbook, PLAYBOOK_TASK_KEYWORDS):
+        task_blocks.extend(play.get(key, []))
     return task_blocks
 
 
@@ -688,17 +689,19 @@ def _get_tasks_from_blocks(task_blocks):
         'rescue',
     ]
 
+    def get_nested_tasks(task):
+        return [task[k] for k in NESTED_TASK_KEYS if task.get(k)]
+
     tasks = []
     for item in task_blocks:
-        for key in NESTED_TASK_KEYS:
-            if item.get(key):
-                tasks.extend(item[key])
-                continue
+        for sub_task in get_nested_tasks(item):
+            tasks.extend(sub_task)
         tasks.append(item)
+
     return tasks
 
 
-def _get_rule_skips_from_yaml(yaml):
+def _get_rule_skips_from_yaml(yaml_input):
     """Travese yaml for comments with rule skips and return list of rules."""
     def traverse_yaml(obj):
         yaml_comment_obj_strs.append(str(obj.ca.items))
@@ -714,7 +717,7 @@ def _get_rule_skips_from_yaml(yaml):
             return
 
     yaml_comment_obj_strs = []
-    traverse_yaml(yaml)
+    traverse_yaml(yaml_input)
 
     rule_id_list = []
     for comment_obj_str in yaml_comment_obj_strs:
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
 
-    PLAYBOOK_TASK_KEYWORDS = [
-        'tasks',
-        'pre_tasks',
-        'post_tasks',
-        'handlers',
-    ]
+    try:
+        yaml_skip = _append_skipped_rules(pyyaml_data, file_text, file_type)
+    except Exception as exc:
```
webknjaz: Could you please specify exact errors which may happen? Wholesale excepts is a bad practice.

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+        # Notify user of skip error, do not stop, do not change exit code
+        print('Error trying to append skipped rules: {}'.format(repr(exc)))
```
webknjaz: Can we use logging instead? Also, there's a format specifier for repr built-in:
```suggestion
        print('Error trying to append skipped rules: {!r}'.format(exc))
```

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+        return pyyaml_data
+    return yaml_skip
 
+
+def _append_skipped_rules(pyyaml_data, file_text, file_type):
+    # parse file text using 2nd parser library
     yaml = ruamel.yaml.YAML()
     ruamel_data = yaml.load(file_text)
 
+    # if file type is 'meta', add skips to metadata block and return
```
webknjaz: This comment explains the implementation which is right below. Better use comments to explain motivation and the high-level effect.

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+    for item in task_blocks:
+        for key in NESTED_TASK_KEYS:
+            if item.get(key):
+                tasks.extend(item[key])
+                continue
+        tasks.append(item)
+    return tasks
+
+
+def _get_rule_skips_from_yaml(yaml):
```
webknjaz: let's not use existing module names for vars to prevent accidental shadowing 
```suggestion
def _get_rule_skips_from_yaml(_yaml):
```

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
 
+
+def _append_skipped_rules(pyyaml_data, file_text, file_type):
+    # parse file text using 2nd parser library
     yaml = ruamel.yaml.YAML()
     ruamel_data = yaml.load(file_text)
 
+    # if file type is 'meta', add skips to metadata block and return
+    if file_type == 'meta':
+        skipped_rules = _get_rule_skips_from_yaml(ruamel_data)
```
webknjaz: I believe that the conditional is not needed:
```suggestion
        pyyaml_data[0]['skipped_rules'] = _get_rule_skips_from_yaml(ruamel_data)
```

It also makes it easy for the further processing code to iterate over this thing without having to do extra checks for the dict key existence.

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+            # call stack, and can remove this except
             return pyyaml_data
-    elif file_type == 'meta':
-        if not isinstance(pyyaml_data, list):
-            return pyyaml_data
-        ruamel_tasks = [ruamel_data]
-        pyyaml_tasks = pyyaml_data
     else:
-        return pyyaml_data
+        raise RuntimeError('Unexpected file type')
```
webknjaz: This error will probably benefit from the context added:
```suggestion
        raise RuntimeError('Unexpected file type: {}'.format(file_type))
```

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+        'tasks',
+        'pre_tasks',
+        'post_tasks',
+        'handlers',
+    ]
+
+    task_blocks = []
+    for play in playbook:
+        for key in PLAYBOOK_TASK_KEYWORDS:
+            task_blocks.extend(play.get(key, []))
```
webknjaz: I'll be simpler if you turn this into a generator
```suggestion
            yield play.get(key, [])
```
And than `list()` its call if needed.

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+def _get_task_blocks_from_playbook(playbook):
+    """Return parts of playbook that contains tasks, and nested tasks."""
+    PLAYBOOK_TASK_KEYWORDS = [
+        'tasks',
+        'pre_tasks',
+        'post_tasks',
+        'handlers',
+    ]
+
+    task_blocks = []
```
webknjaz: Alternatively, use a list comprehension
```suggestion
    from itertools import product
    return [play.get(key, []) for play, key in product(playbook, PLAYBOOK_TASK_KEYWORDS)]
```

lib/ansiblelint/utils.py
webknjaz: (use `itertools.product()` regardless of whether you choose to use a generator or a comprehension)

lib/ansiblelint/utils.py
webknjaz: Or a generator expression:
```suggestion
    from itertools import product
    return (play.get(key, []) for play, key in product(playbook, PLAYBOOK_TASK_KEYWORDS))
```

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+        'always',
+        'rescue',
+    ]
+
+    tasks = []
+    for item in task_blocks:
+        for key in NESTED_TASK_KEYS:
+            if item.get(key):
+                tasks.extend(item[key])
+                continue
```
webknjaz: Something is not right here: `continue` doesn't change the flow.
```suggestion
```

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+    """Get list of tasks from list made of tasks and nested tasks."""
+    NESTED_TASK_KEYS = [
+        'block',
+        'always',
+        'rescue',
+    ]
+
+    tasks = []
+    for item in task_blocks:
+        for key in NESTED_TASK_KEYS:
```
webknjaz: This can improve readability
```suggestion
        for key in (k for k in NESTED_TASK_KEYS if item.get(k)):
```

lib/ansiblelint/utils.py
```diff
@@ -599,55 +599,109 @@ def get_first_cmd_arg(task):
...
+
+def _get_tasks_from_blocks(task_blocks):
+    """Get list of tasks from list made of tasks and nested tasks."""
+    NESTED_TASK_KEYS = [
+        'block',
+        'always',
+        'rescue',
+    ]
+
+    tasks = []
```
webknjaz: Is this what you intended to do?
```suggestion
    def get_nested_tasks(item):
	    return [item[key] for k in NESTED_TASK_KEYS if item.get(k)]
	
	return [get_nested_tasks(i) + i for i in task_blocks]
```

lib/ansiblelint/utils.py
awcrosby: Agreed generic except should be avoided most times... however here I was thinking that regardless of any possible unexpected error within skipping portion, there is value in continuing program execution

lib/ansiblelint/utils.py
awcrosby: Other cli output uses print at this time, but a future move overall to logging would be good

lib/ansiblelint/utils.py
webknjaz: The danger of the wholesale exception swallowing is that you'll miss some important errors which may need to be processed in a better way. It also breaks the debugging experience.
+ Nobody is going to report such errors back.

awcrosby: Fair point, I made the error specific

lib/ansiblelint/utils.py
```diff
@@ -599,55 +600,111 @@ def get_first_cmd_arg(task):
...
-        ruamel_tasks = [ruamel_data]
-        pyyaml_tasks = pyyaml_data
     else:
-        return pyyaml_data
+        raise RuntimeError('Unexpected file type: {}'.format(file_type))
 
-    if len(ruamel_tasks) != len(pyyaml_tasks):
-        return pyyaml_data
+    # get tasks from blocks of tasks
+    pyyaml_tasks = _get_tasks_from_blocks(pyyaml_task_blocks)
```
webknjaz: I'd prefer that function to be a generator: there's no need to hold the whole list in memory and the next thing you do is feeding it to `zip()` and throwing the variable away. So why waste memory for this?

lib/ansiblelint/utils.py
```diff
@@ -599,55 +600,111 @@ def get_first_cmd_arg(task):
...
+def _get_tasks_from_blocks(task_blocks):
+    """Get list of tasks from list made of tasks and nested tasks."""
+    NESTED_TASK_KEYS = [
+        'block',
+        'always',
+        'rescue',
+    ]
+
+    def get_nested_tasks(task):
+        return [task[k] for k in NESTED_TASK_KEYS if task.get(k)]
```
webknjaz: Save memory and improve app responsiveness:
```suggestion
        return (task[k] for k in NESTED_TASK_KEYS if k in task)
```

lib/ansiblelint/utils.py
```diff
@@ -599,55 +600,111 @@ def get_first_cmd_arg(task):
...
+
+    tasks = []
+    for item in task_blocks:
+        for sub_task in get_nested_tasks(item):
+            tasks.extend(sub_task)
+        tasks.append(item)
```
webknjaz: With https://github.com/ansible/ansible-lint/pull/528/files#r290001518 you can reduce the number of explicit loops

lib/ansiblelint/utils.py
```diff
@@ -599,55 +600,111 @@ def get_first_cmd_arg(task):
...
+    pyyaml_tasks = _get_tasks_from_blocks(pyyaml_task_blocks)
+    ruamel_tasks = _get_tasks_from_blocks(ruamel_task_blocks)
 
+    # append skipped_rules for each task
     for ruamel_task, pyyaml_task in zip(ruamel_tasks, pyyaml_tasks):
-        skipped_rules = _get_rule_skips_from_task(ruamel_task)
+        if pyyaml_task.get('name') != ruamel_task.get('name'):
+            raise RuntimeError('Error in matching skip comment to a task')
+        skipped_rules = _get_rule_skips_from_yaml(ruamel_task)
         if skipped_rules:
```
webknjaz: Why do we need this check? I think, it's unnecessary

lib/ansiblelint/utils.py
awcrosby: Generator has been added here

PR #529, opened 2019-05-31
Break out skip_utils into separate module

awcrosby:
Break out `skip_utils` into separate module for readability

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -19,6 +19,7 @@
 # THE SOFTWARE.
 
 from __future__ import print_function
+import codecs
 from collections import defaultdict
 import os
 import re
@@ -27,7 +28,7 @@ import sys
 import six
 
 import ansiblelint.utils
-import codecs
+import ansiblelint.skip_utils
 
 default_rulesdir = os.path.join(os.path.dirname(ansiblelint.utils.__file__), 'rules')
 
@@ -58,7 +59,7 @@ class AnsibleLintRule(object):
             if line.lstrip().startswith('#'):
                 continue
 
-            rule_id_list = ansiblelint.utils.get_rule_skips_from_line(line)
+            rule_id_list = ansiblelint.skip_utils.get_rule_skips_from_line(line)
             if self.id in rule_id_list:
                 continue
 
@@ -84,7 +85,7 @@ class AnsibleLintRule(object):
         if not yaml:
             return matches
 
-        yaml = ansiblelint.utils.append_skipped_rules(yaml, text, file['type'])
+        yaml = ansiblelint.skip_utils.append_skipped_rules(yaml, text, file['type'])
 
         for task in ansiblelint.utils.get_normalized_tasks(yaml, file):
             if self.id in task.get('skipped_rules', ()):
@@ -116,7 +117,7 @@ class AnsibleLintRule(object):
         if isinstance(yaml, dict):
             yaml = [yaml]
 
-        yaml = ansiblelint.utils.append_skipped_rules(yaml, text, file['type'])
+        yaml = ansiblelint.skip_utils.append_skipped_rules(yaml, text, file['type'])
 
         for play in yaml:
             if self.id in play.get('skipped_rules', ()):
--- a/lib/ansiblelint/skip_utils.py
+++ b/lib/ansiblelint/skip_utils.py
@@ -24,12 +24,14 @@ from itertools import product
 
 import ruamel.yaml
 
+INLINE_SKIP_FLAG = '# noqa'
+
 
 def get_rule_skips_from_line(line):
     """Return list of rule ids skipped via comment on the line of yaml."""
     rule_id_list = []
-    if '# noqa' in line:
-        noqa_text = line.split('# noqa')[1]
+    if INLINE_SKIP_FLAG in line:
+        noqa_text = line.split(INLINE_SKIP_FLAG)[1]
         rule_id_list = noqa_text.split()
     return rule_id_list
 
--- a/lib/ansiblelint/skip_utils.py
+++ b/lib/ansiblelint/skip_utils.py
@@ -157,7 +157,7 @@ def _get_rule_skips_from_yaml(yaml_input):
 
     rule_id_list = []
     for comment_obj_str in yaml_comment_obj_strs:
-        for line in comment_obj_str.split('\\n'):
+        for line in comment_obj_str.split(r'\n'):
             rule_id_list.extend(get_rule_skips_from_line(line))
 
     return rule_id_list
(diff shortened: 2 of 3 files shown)


webknjaz:
plz rebase this

awcrosby:
@webknjaz - rebase complete

lib/ansiblelint/__init__.py
```diff
@@ -23,11 +23,12 @@
 import os
 import re
 import sys
+import codecs
```
webknjaz: Maybe keep this alphabetically ordered?

lib/ansiblelint/__init__.py
```diff
@@ -23,11 +23,12 @@
...
 
 import six
 
 import ansiblelint.utils
-import codecs
+from ansiblelint import skip_utils
```
webknjaz: Can we use relative imports?
(this may require `from __future__ import absolute_imports`)

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) 2019, Ansible Project
```
webknjaz: This probably needs some license lines.

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+from itertools import product
+
+import ruamel.yaml
+
+
+def get_rule_skips_from_line(line):
+    rule_id_list = []
+    if '# noqa' in line:
```
webknjaz: Let's put this string into a module-global constant.

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,140 @@
...
+                    traverse_yaml(e)
+        else:
+            return
+
+    yaml_comment_obj_strs = []
+    traverse_yaml(yaml_input)
+
+    rule_id_list = []
+    for comment_obj_str in yaml_comment_obj_strs:
+        for line in comment_obj_str.split('\\n'):
```
webknjaz: Are you sure this is what you want? It may be easier to understand with raw-literal in place:
```suggestion
        for line in comment_obj_str.split(r'\n'):
```

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,140 @@
...
+        yaml_comment_obj_strs.append(str(obj.ca.items))
+        if isinstance(obj, dict):
+            for key, val in obj.items():
+                if isinstance(val, (dict, list)):
+                    traverse_yaml(val)
+        elif isinstance(obj, list):
+            for e in obj:
+                if isinstance(e, (dict, list)):
+                    traverse_yaml(e)
+        else:
```
webknjaz: This else-block isn't really needed

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,140 @@
...
+            return
+
+    yaml_comment_obj_strs = []
```
webknjaz: I'd prefer to put the initialization of this accumulator inside of `traverse_yaml` and not use closure.

lib/ansiblelint/skip_utils.py
webknjaz: Also, can we have a module docstring here?

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) 2019, Ansible Project
+
+from itertools import product
+
+import ruamel.yaml
+
+
+def get_rule_skips_from_line(line):
```
webknjaz: Plz add a PEP257-compliant docstring here.

lib/ansiblelint/skip_utils.py
awcrosby: I'll look at updating the recursion to build the list within the function, I'll do that outside this PR, the focus here is to split existing functionality out.

webknjaz: okay

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,163 @@
+# (c) 2012-2019, Ansible by Red Hat
```
webknjaz: AFAIK the year should be of content creation/update. Which is 2019 for Ansible. Do we have anything written by the original author here? If yes, crediting them is also needed.
```suggestion
# (c) 2019, Ansible by Red Hat
```

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,163 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+"""This module contains utils related to inline skiping of rules."""
```
webknjaz: Let's simplify this. It's not important to state that it's a module, because it's quite obvious :)
```suggestion
"""Utils related to inline skiping of rules."""
```

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+from itertools import product
+
+import ruamel.yaml
+
+INLINE_SKIP_FLAG = '# noqa'
```
webknjaz: This probably needs an extra space:
```suggestion
INLINE_SKIP_FLAG = '# noqa '
```
what do you think?

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def get_rule_skips_from_line(line):
+    """Return list of rule ids skipped via comment on the line of yaml."""
+    rule_id_list = []
+    if INLINE_SKIP_FLAG in line:
+        noqa_text = line.split(INLINE_SKIP_FLAG)[1]
```
webknjaz: It'd be probably cleaner to replace the whole function body with just:
```suggestion
    _before_noqa, _noqa_marker, noqa_text = line.partition(INLINE_SKIP_FLAG)
    return noqa_text.split()
```
(though, it can wait till the next refactoring iteration)

webknjaz:
@slsh1o do you want to try completing this refactoring effort?

slsh1o:
@webknjaz why not? I'll do it.

Issue #534, opened 2019-06-06
How to run noqa on a YAML literal block


ianw:
# Issue Type
- Bug report
- Feature request(?)

# Desired Behaviour

Given something like

```yaml
exim_system_aliases_router: |
  driver = redirect
  allow_fail
  allow_defer
  data = ${lookup{$local_part}lsearch{/etc/aliases}}
  file_transport = address_file
  pipe_transport = address_pipe
```

This trips 

```
[206] Variables should have spaces before and after: {{ var_name }}
```

I'm at a loss how to actually skip this with noqa?  I don't want a ```#noqa XXX``` comment on the end of the line, because it will go into the string.  I tried

```yaml
exim_system_aliases_router: # noqa 206
 |
  driver = redirect
  allow_fail
  allow_defer
  data = ${lookup{$local_part}lsearch{/etc/aliases}}
  file_transport = address_file
  pipe_transport = address_pipe
```

which is ugly ... but also didn't work.

Happy to update documentation if I can figure it out ...

awcrosby:
The way ansible-lint analyses line-based rules (like 206) is by running over each line in the file text, not by parsing beforehand (like task-based rules). So unfortunately a `# noqa` that is not present on the specific line being analyzed will not work.

However I think we can cover this particular false positive by ensuring at least `{{` and `}}` both exist before applying further regex - raised PR #535

eigengrau:
Cross-reference #586

PR #535, opened 2019-06-06
Ensure variable syntax before matching in VariableHasSpacesRule

awcrosby:
Fixes false positive reported in #534


test/TestVariableHasSpaces.py
```diff
@@ -31,6 +31,10 @@
 - name: bad variable format
   debug:
     msg: "{{bad_format }}"
+- name: not a jinja variable
+  debug:
+    msg "test"
```
ianw: I guess this doesn't have to be vaild yaml but missing a ":"

awcrosby: thanks, i corrected

Issue #328, opened 2018-03-07
Is installation possible in windows?


jonmelia:
Not sure if it is possible to install in Windows, but would like confirmation if it isn't

# Issue Type
- Bug report
Unable to install via Python and Pip in Windows..

Error message:
c:\Python27amd64\Scripts>pip install ansible-lint
Collecting ansible-lint
  Using cached ansible-lint-3.4.20.tar.gz
Collecting ansible (from ansible-lint)
  Using cached ansible-2.4.3.0.tar.gz
Exception:
Traceback (most recent call last):
  File "c:\python27amd64\lib\site-packages\pip\basecommand.py", line 215, in main
    status = self.run(options, args)
  File "c:\python27amd64\lib\site-packages\pip\commands\install.py", line 324, in run
    requirement_set.prepare_files(finder)
  File "c:\python27amd64\lib\site-packages\pip\req\req_set.py", line 380, in prepare_files
    ignore_dependencies=self.ignore_dependencies))
  File "c:\python27amd64\lib\site-packages\pip\req\req_set.py", line 620, in _prepare_file
    session=self.session, hashes=hashes)
  File "c:\python27amd64\lib\site-packages\pip\download.py", line 821, in unpack_url
    hashes=hashes
  File "c:\python27amd64\lib\site-packages\pip\download.py", line 663, in unpack_http_url
    unpack_file(from_path, location, content_type, link)
  File "c:\python27amd64\lib\site-packages\pip\utils\__init__.py", line 605, in unpack_file
    untar_file(filename, location)
  File "c:\python27amd64\lib\site-packages\pip\utils\__init__.py", line 553, in untar_file
    ensure_dir(path)
  File "c:\python27amd64\lib\site-packages\pip\utils\__init__.py", line 83, in ensure_dir
    os.makedirs(path)
  File "c:\python27amd64\lib\os.py", line 157, in makedirs
    mkdir(name, mode)
WindowsError: [Error 206] The filename or extension is too long: 'c:\\users\\user\\appdata\\local\\temp\\pip-build-sz048m\\ansible\\test/integration/targets/copy/files/subdir/subdir1/circles/subdir1/circles/subdir1/circles/subdir1/circles/subdir1/circles/subdir1/circles/subdir1/circles/subdir1/circles/subdir1/circles/subdir1/'

# Ansible and Ansible Lint details

```
ansible --version
ansible-lint --version
```

- ansible installation method: one of source, pip, OS package
[...]

PR #549, opened 2019-07-16
Document that ansible-lint does not install on windows.

JorisVanEijden:
See also #328

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,8 @@ Installing
 
 .. installing-docs-inclusion-marker-do-not-remove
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -32,6 +32,8 @@ Installing
 
 .. installing-docs-inclusion-marker-do-not-remove
 
+Installing on windows is not supported because we use symlinks inside python packages.
```
gundalow:
```suggestion
Installing on Windows is not supported because we use symlinks inside Python packages.
```

PR #550, opened 2019-07-18
Add logging to configfile parsing

Davidsoff:
To ensure that it is clear why a config file is not being picked up, This change makes it so that an error message is logged if verbosity is greater than 1.
An option to skip the loading of any config file is also added and set to false by default

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -35,16 +35,18 @@ import yaml
 import os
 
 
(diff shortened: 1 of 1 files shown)


ssbarnea:
Can you please resolve conflicts on this?

ssbarnea:
@webknjaz I think we can close this since logging refactoring also addressed most of it and made this PR irrelevant.

PR #554, opened 2019-07-25
add with_filetree to bare variable exceptions

### LINKED ISSUE #494: ANSIBLE0015 False Positive with variable in "with_filetree"
# Issue Type
- Bug report

```
ansible 2.7.8
  python version = 3.6.8 (default, Jan 31 2019, 09:38:34) [GCC 8.2.1 20181215 (Red Hat 8.2.1-6)]

ansible-lint 3.4.23
```

# Desired Behaviour
[ANSIBLE0015] Found a bare variable '../templates/' used in a 'with_filetree' loop. You should use the full variable syntax ('{{../templates/}}')

But it's normal to pass non variable to the loop for with_filetree module.

```
- name: Generate configurations
  when: item.state == 'file'
  become: true
  template:
    src: "{{ item.src }}"
    dest: "{{ configuration_path }}/{{ item.path | replace('.j2', '') }}"
  with_filetree: "../templates/"
```
```
$ tree
├── playbooks
│      └── test_lint.yml
├── templates
        └── test.j2

$ ansible-lint playbooks/test_lint.yml 
[ANSIBLE0015] Found a bare variable '../templates/' used in a 'with_filetree' loop. You should use the full variable syntax ('{{../templates/}}')

$ cat playbooks/test_lint.yml 
- hosts: localhost
  vars:
    configuration_path: /tmp
  tasks:
    - name: Generate all configurations
      when: item.state == 'file'
      template:
        src: "{{ item.src }}"
        dest: "{{ configuration_path }}/{{ item.path | replace('.j2', '') }}"
      with_filetree: "../templates/"
$ cat templates/test.j2
{{ ansible_managed | comment }}
```

I see that for the `with_fileglob` there is `if` statement, so the simpliest way to resolve this is to update and add `with_filetree` module to the `if` statement:
```
def _matchvar(self, varstring, task, loop_type):
        if (isinstance(varstring, six.string_types) and
                not self._jinja.match(varstring)):
            if loop_type != 'with_fileglob' or not (self._jinja.search(varstring) or
                                                    self._glob.search(varstring)):
                message = "Found a bare variable '{0}' used in a '{1}' loop." + \
                          " You should use the full variable syntax ('{{{{{0}}}}}')"
                return message.format(task[loop_type], loop_type)
```

Like this issue: https://github.com/ansible/ansible-lint/issues/207

NotAFile:
This is an attempt to fix #494 

I'm not sure if this is the correct way to do this.

### DIFF
--- a/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
+++ b/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py
@@ -55,7 +55,7 @@ class UsingBareVariablesIsDeprecatedRule(AnsibleLintRule):
             elif loop_type == "with_subelements":
                 return self._matchvar(task[loop_type][0], task, loop_type)
             elif loop_type in ["with_sequence", "with_ini",
-                               "with_inventory_hostnames"]:
+                               "with_inventory_hostnames", "with_filetree"]:
                 pass
             else:
                 return self._matchvar(task[loop_type], task, loop_type)


NotAFile:
Any feedback? Just came across this false positive again, and it would be good if it could be fixed.

awcrosby:
This looks reasonable… the docs ask for full variable syntax for some loops like `with_items` but not for `with_filetree`, so excluding this seems ok.

CI seems stuck on queued, I will close and reopen.

awcrosby:
Closing as #582 covers this

PR #561, opened 2019-08-05
Ensure files are checked only once.

### LINKED ISSUE #560: Ansible lint does not effectively ensures a file has  already been checked
# Issue Type
- Bug report

```
$ ansible --version
ansible 2.8.3
  config file = ~/src/gitco/roles/user-make/ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/.virtualenvs/ansible-roling/lib/python3.7/site-packages/ansible
  executable location = ~/.virtualenvs/ansible-roling/bin/ansible
  python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.0]
$ ansible-lint --version
  ansible-lint 4.1.0
```

# Desired Behaviour

When files are include by others, they should not be recheck. E.g. when checking the role [cans/user-make](https://github.com/cans/user-make), I should get the following output:

```
$ ansible-lint tests/*.yml  # in there a playbook includes tests/test.yml

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:59
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:97
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:105
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:183
      local_action:
```

Instead file imported are reported several times.

```
[504] Do not use 'local_action', use 'delegate_to: localhost'
/home/ncaniart/src/gitco/roles/user-make/tests/test.yml:59
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
/home/ncaniart/src/gitco/roles/user-make/tests/test.yml:97
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
/home/ncaniart/src/gitco/roles/user-make/tests/test.yml:105
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
/home/ncaniart/src/gitco/roles/user-make/tests/test.yml:183
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:59
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:97
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:105
      local_action:

[504] Do not use 'local_action', use 'delegate_to: localhost'
tests/test.yml:183
      local_action:
```
Because the path are not properly normalized and despite a mechanism being in place in ansible-lint's implementation, the files are indeed checked and reported twice.

This issue has already been reported (cf. #335) but disregarded, though being fairly trivial to fix.

cans:
Ansible Lint has a mecanism to avoid checking
a file several times. Though file path not
being properly normalized, it is not effective.

This commit fixes that and adds a mean to still
keep the output of ansible short by remove the
files common path prefix (helps with cut pasting
path from several env., e.g. from your CI to
your local shell).

This should fix #560 .

There would be a few tests to perform implement though. In particular I have an exclude list, with relative path in it. They are indeed still ignored but I didn't check why yet). The config used is the following:
```yaml
exclude_paths:
  # Ignore third party roles
  - ./tests/roles/
```

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -221,9 +221,9 @@ class Runner(object):
         # assume role if directory
         if os.path.isdir(playbook):
             self.playbooks.add((os.path.join(playbook, ''), 'role'))
-            self.playbook_dir = playbook
+            self.playbook_dir = os.path.abspath(playbook)
         else:
-            self.playbooks.add((playbook, 'playbook'))
+            self.playbooks.add((os.path.abspath(playbook), 'playbook'))
             self.playbook_dir = os.path.dirname(playbook)
         self.tags = tags
         self.skip_list = skip_list
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -157,7 +157,7 @@ def find_children(playbook, playbook_dir):
         except AnsibleError as e:
             raise SystemExit(str(e))
     results = []
-    basedir = os.path.dirname(playbook[0])
+    basedir = os.path.abspath(os.path.dirname(playbook[0]))
     items = _playbook_items(playbook_ds)
     for item in items:
         for child in play_children(basedir, item, playbook[1], playbook_dir):


cans:
I have looked into why the 'exclude_path' configuration works. Path are indeed normalized to their [absolute value](https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/__init__.py#L246). Though I am not sure I totally agree with how it is done.

It seems that in most use case we have _ansible-lint "cwd" == ".ansible-lint" parent directory_. Though path are always normalized using ansible-lint "cwd". Which means if one uses the `-c` command-line option, path listed in `.ansible-lint` might not be resolved properly. Indeed, I would expect relative path listed in the configuration file to be relative to that file location, not to a completely independent and unpredictable run-time location like "cwd". That is only valid for command-line arguments passed with the `--exclude` option.

Any command like, `ansible-lint -c ../.ansible-lint` or `ansible-lint -c subdir/.ansible-lint` will likely not properly filter the files.

I guess that is yet another issue.

renskewierda:
Are there plans to merge this pull request? I also require this functionality to avoid multiple hits per rule.

ssbarnea:
@cans @webknjaz lets close this until is updated and ready to be reviewed.

webknjaz:
Marking this as a draft. I'll let @cans decide if he wants to move this forward or close.

ssbarnea:
Closing as abandoned. I also suspect is no longer a real issue.

PR #568, opened 2019-08-14
Remove PackageNotLatest rule

### LINKED ISSUE #479: Re-evaluate E0010 - Package installs should not use latest
# Issue Type
- Bug report

```
ansible-lint 4.0.1
ansible 2.7.6
```

# Summary
The error `#0010` says "Package installs should not use latest".  Further on rule's description one can read 'Package installs should use ``state=present`` with or without a version'.

The fact is that some times you really want to upgrade a package but you don't know exactly the version. Example, in an CI environment where you want to test the latest version of a package from an YUM repo that was built by other team.

I understand "should not use latest" is advisable for the 'package' module since its documentation already states "(...)Other states depend on the underlying package module, i.e latest.". But for packaging modules like 'yum' I still lack to understand the rationale. Mind to explain me?

evitalis:
Fixes #479 due to no response from original committer, and official Ansible documentation allows 'latest' as a valid state.


webknjaz:
@evitalis `ansible-lint` has no relation to what Ansible Core Team recommends. In general, we don't try to dictate what are the practices/styles that you have to follow — you are free to do what's best for *your* use case.
`ansible-lint` has a lot of rules that an average Core Team member would disagree with and, as such, is not a source of the "officially advised rules for writing playbooks".
So think of it as some *community-driven* recommendations that don't necessarily reflect what the Core Team thinks. Meaning that these rules should be changed if the community agrees, not just because Ansible can execute what you wrote.

evitalis:
@webknjaz That makes sense but there are a number of people commenting on the open issue that they also agree this rule should be disabled or removed as well. Those are also users of ansible and ansible-lint unless there is a different community you are referring to?

webknjaz:
@evitalis to me it looks like a rather exceptional use-case. Environments should be pinned to use well-defined versions of software in order to be reproducible. Using `latest` is possible but users should understand that it's not something that will create the same env every time they run this module. So I think it's reasonable to expect them to use a `# noqa` comment there in order to explicitly mark that they know what they are doing.

evitalis:
Would it be worth this conversation being moved to the referenced issue to get additional feedback and community comments please?

Issue #569, opened 2019-08-15
Include_tasks/import_tasks in subfolders result in wrong paths


wilmardo:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
# ansible --version
ansible 2.8.3
  config file = None
  configured module search path = ['/c/Users/wilmaro/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /c/Users/wilmaro/Development/Intermax/ansible-lint/venv/lib/python3.6/site-packages/ansible-2.8.3-py3.6.egg/ansible
  executable location = /c/Users/wilmaro/Development/Intermax/ansible-lint/venv/bin/ansible
  python version = 3.6.8 (default, Jan 14 2019, 11:02:34) [GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]

# ansible-lint --version
ansible-lint 4.1.0
```

- ansible installation method: pip
- ansible-lint installation method: pip

# Desired Behaviour

The paths correctly expanded and no warning shown.

# Actual Behaviour (Bug report only)

```
# ansible-lint .
WARNING: Couldn't open /c/Users/wilmaro/Development/Intermax/ansible-lint-issue-demo/tasks/tasks/includes/main.yml - No such file or directory
[502] All tasks should be named
/c/Users/wilmaro/Development/Intermax/ansible-lint-issue-demo/tasks/includes/main.yml:2
Task/Handler: ping
```

# Steps to reproduce

```
git clone https://github.com/wilmardo/ansible-lint-issue-demo.git
cd ansible-lint-issue-demo
ansible-lint .
```

PR #570, opened 2019-08-15
WIP: fix: Expand paths correctly from basepath

wilmardo:
Fixes: #569 

Ansible supports relative includes from the tasks folder. Check the demo here:
https://github.com/wilmardo/ansible-lint-issue-demo.git

Ansible lint assumed that the tasks dir is just one directory up but it could be more then one. My solutions split the basedir on the lasts tasks/ in the path and looks there for the include.
If that does not exists it falls back on the previous behavior (which in my opinion is better to remove but might break some existing scenarios).

If someone has a better solution please share! This solution assumes there is no nested tasks folder in the role (like `role/tasks/linux/tasks/main.yml`)
 
How Ansible handles relative imports/includes can be found here, which my be hard to reproduce since there is no reference to a previous include:

https://github.com/ansible/ansible/blob/7346b699eec99d7279da4175b99f07e08f0de283/lib/ansible/playbook/included_file.py#L117

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -240,7 +240,10 @@ def _include_children(basedir, k, v, parent_type):
 
     result = path_dwim(basedir, args[0])
     if not os.path.exists(result) and not basedir.endswith('tasks'):
-        result = path_dwim(os.path.join(basedir, '..', 'tasks'), v)
+        tasks_path = basedir.rsplit('tasks/')[0] + 'tasks/'
+        result = path_dwim(tasks_path, args[0])
+        if not os.path.exists(result):
+            result = path_dwim(os.path.join(basedir, '..', 'tasks'), v)
     return [{'path': result, 'type': parent_type}]
 
 
--- a/test/TestUtils.py
+++ b/test/TestUtils.py
@@ -157,6 +157,18 @@ def test_expand_paths_vars(monkeypatch):
     assert utils.expand_paths_vars(['~']) == [os.path.expanduser('~')]
     assert utils.expand_paths_vars(['$TEST_PATH']) == [test_path]
 
+@unittest.mock.patch('os.path.exists')
+def test_include_children_relative_from_root(self, mock_exists):
+    # Mock that the file exists relative from root
+    mock_exists.return_value = True
+
+    result = utils._include_children(
+        basedir='/role/tasks/subdir',
+        k='include_tasks',
+        v='includes/main.yml',
+        parent_type='tasks'
+    )
+    self.assertEqual(result, '/role/tasks/includes/main.yml')
 
 @pytest.mark.parametrize(
     ('reset_env_var', 'message_prefix'),


wilmardo:
@webknjaz @awcrosby Not to be rude but this bug is rather annoying and a month has passed ;)

webknjaz:
@wilmardo sorry, we may have different priorities right now. Especially, given that AnsibleFest is coming soon. Try pinging Andrew again in two weeks.

As for me, I'm not heavily involved in this project. I've taken a prompt look at the diff and it seems harmless but we'll need someone else to check this deeper.

Anyway, I'd advise you to always include/extend tests in your PRs so that it'd be visible that the change does what it's supposed to do. And also it's usually good for preventing regressions.

lib/ansiblelint/utils.py
```diff
@@ -231,7 +231,10 @@ def _include_children(basedir, k, v, parent_type):
 
     result = path_dwim(basedir, args[0])
     if not os.path.exists(result) and not basedir.endswith('tasks'):
-        result = path_dwim(os.path.join(basedir, '..', 'tasks'), v)
+        tasks_path = basedir.rsplit('tasks/')[0] + 'tasks/'
```
webknjaz: When working with paths, you should typically use `os.path.split()` instead of normal `str.split()`.

wilmardo: I looked at this today but I don't know if `os.path.split()` makes the code more robust:
```python
tasks_path = basedir
while not tasks_path[0].endswith('tasks'):
   tasks_path = os.path.split(tasks_path)
# Append slash since head of os.path.split has it stripped
tasks_path += '/'
```
(didn't test this approach)

Does it make the code better?

awcrosby: The rsplit approach seems more readable/maintainable. Although for joining, maybe `os.path.join()` could be used over `+`.
```
tasks_root = basedir.rsplit('tasks/')[0]
result = path_dwim(os.path.join(tasks_root, 'tasks'), args[0])
```

wilmardo:
@awcrosby Thanks for the review! Can't get tox to run nicely within WSL today, will pick this up later this week on my Ubuntu laptop :)

webknjaz:
@wilmardo not sure if this is still relevant (there's been some updates to the path expansion merged recently) but I see that this PR has a conflict with master so you may want to rebase it.

ssbarnea:
@webknjaz AFAIK, this wasalready fixed by another change we merged, the one about consistent paths. I think is safe to close this PR.

wilmardo:
Just tested this against the demo repo but the issue persists:

```
git clone https://github.com/wilmardo/ansible-lint-issue-demo
cd ansible-lint-issue-demo
pip install git+https://github.com/ansible/ansible-lint.git
```

```console
# ansible-lint .
WARNING: Couldn't open /home/wilmardo/Development/ansible-lint-issue-demo/tasks/tasks/includes/main.yml - No such file or directory
[502] All tasks should be named
tasks/includes/main.yml:2
Task/Handler: ping
```

Ansible will process the includes just fine :)

I will try to rebase this now.

wilmardo:
Did a quick rebase but the tests are failing now, I will try to correct this tomorrow morning :)

ssbarnea:
@wilmardo Any change to have another look at it?

ssbarnea:
@wilmardo Can you take a look at it? It would be great if it would make into 4.3.0.

ssbarnea:
Closing as abandoned.

PR #577, opened 2019-09-03
403: `state=latest` is fine with `version`

simon04:


ssbarnea:
@webknjaz @awcrosby do the honours. Thanks.

evitalis:
@ssbarnea How does this work with #479 and #568?

PR #582, opened 2019-10-01
with_filetree loop: reduce false positive

pilou-:
handle `with_filetree` loop:
- don't return an error when a string ending with a slash is used
- `with_filetree` loop accepts list too
- add tests


awcrosby:
LGTM, closes #494

Issue #590, opened 2019-10-16
ansible-lint fails to install from source: AttributeError: 'SpecifierSet' object has no attribute 'split'


ssbarnea:
Due to the complex workarounds collected in `setup.py` over the years ansible-lint may fail to install with something like:

```
pip install .                                                                                                                                                           [13:40:59]
Processing /Users/ssbarnea/.cache/pre-commit/repo5hdz7os2
ERROR: Command errored out with exit status 1:
command: /Users/ssbarnea/.cache/pre-commit/repo5hdz7os2/py3/bin/python3.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/pip-req-build-zpowp0qt/setup.py'"'"'; __file__='"'"'/private/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/pip-req-build-zpowp0qt/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'rn'"'"', '"'"'n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/pip-req-build-zpowp0qt/pip-egg-info
cwd: /private/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/pip-req-build-zpowp0qt/
Complete output (21 lines):
/private/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/pip-req-build-zpowp0qt/.eggs/setuptools_scm-3.3.3-py3.7.egg/setuptools_scm/git.py:68: UserWarning: "/private/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/pip-req-build-zpowp0qt" is shallow and may cause errors
warnings.warn('"%s" is shallow and may cause errors' % (wd.path,))
Traceback (most recent call last):
File "<string>", line 1, in <module>
File "/private/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/pip-req-build-zpowp0qt/setup.py", line 150, in <module>
__name__ == '__main__' and setuptools.setup(**setup_params)
[...]

PR #592, opened 2019-10-16
Refactor setup.py

ssbarnea:
Removes all workarounds from setup.py in order to avoid bugs like #590

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,7 @@ project_urls =
   Source Code = https://github.com/ansible/ansible-lint
 description = Checks playbooks for practices and behaviour that could potentially be improved
 long_description = file: README.rst
+long_description_content_type = text/x-rst
 author = Will Thames
 author_email = will@thames.id.au
 maintainer = Ansible by Red Hat
--- a/setup.py
+++ b/setup.py
@@ -1,147 +1,19 @@
 #! /usr/bin/env python
-"""Ansible-lint distribution package setuptools installer."""
-
 import setuptools
 
 
-try:
-    from setuptools.config import read_configuration, ConfigOptionsHandler
-    import setuptools.config
-    import setuptools.dist
-
-    # Set default value for 'use_scm_version'
-    setattr(setuptools.dist.Distribution, 'use_scm_version', False)
-
-    # Attach bool parser to 'use_scm_version' option
-    class ShimConfigOptionsHandler(ConfigOptionsHandler):
-        """Extension class for ConfigOptionsHandler."""
-
-        @property
-        def parsers(self):
-            """Return an option mapping with default data type parsers."""
-            _orig_parsers = super(ShimConfigOptionsHandler, self).parsers
-            return dict(use_scm_version=self._parse_bool, **_orig_parsers)
-
-    setuptools.config.ConfigOptionsHandler = ShimConfigOptionsHandler
-except ImportError:
-    """This is a shim for setuptools<30.3."""
-    import io
-    import json
-
-    try:
-        from configparser import ConfigParser, NoSectionError
-    except ImportError:
-        from ConfigParser import ConfigParser, NoSectionError
-        ConfigParser.read_file = ConfigParser.readfp
-
-    def maybe_read_files(d):
-        """Read files if the string starts with `file:` marker."""
-        d = d.strip()
-        if not d.startswith('file:'):
-            return d
-        descs = []
-        for fname in map(str.strip, str(d[5:]).split(',')):
-            with io.open(fname, encoding='utf-8') as f:
-                descs.append(f.read())
-        return ''.join(descs)
-
-    def cfg_val_to_list(v):
-        """Turn config val to list and filter out empty lines."""
-        return list(filter(bool, map(str.strip, str(v).strip().splitlines())))
-
-    def cfg_val_to_dict(v):
-        """Turn config val to dict and filter out empty lines."""
-        return dict(
-            map(lambda l: list(map(str.strip, l.split('=', 1))),
-                filter(bool, map(str.strip, str(v).strip().splitlines())))
-        )
-
-    def cfg_val_to_primitive(v):
-        """Parse primitive config val to appropriate data type."""
-        return json.loads(v.strip().lower())
-
-    def read_configuration(filepath):
-        """Read metadata and options from setup.cfg located at filepath."""
-        cfg = ConfigParser()
-        with io.open(filepath, encoding='utf-8') as f:
-            cfg.read_file(f)
-
-        md = dict(cfg.items('metadata'))
-        for list_key in 'classifiers', 'keywords':
-            try:
-                md[list_key] = cfg_val_to_list(md[list_key])
-            except KeyError:
-                pass
-        try:
-            md['long_description'] = maybe_read_files(md['long_description'])
-        except KeyError:
-            pass
-        opt = dict(cfg.items('options'))
-        for list_key in 'use_scm_version', 'zip_safe':
-            try:
-                opt[list_key] = cfg_val_to_primitive(opt[list_key])
-            except KeyError:
-                pass
-        for list_key in 'scripts', 'install_requires', 'setup_requires':
-            try:
-                opt[list_key] = cfg_val_to_list(opt[list_key])
-            except KeyError:
-                pass
-        try:
-            opt['package_dir'] = cfg_val_to_dict(opt['package_dir'])
-        except KeyError:
-            pass
-        try:
-            opt_package_data = dict(cfg.items('options.package_data'))
-            if not opt_package_data.get('', '').strip():
-                opt_package_data[''] = opt_package_data['*']
-                del opt_package_data['*']
-        except (KeyError, NoSectionError):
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -15,7 +15,11 @@ deps =
   pep8-naming
   wheel
 commands = nosetests []
-passenv = HOME
+passenv =
+    CURL_CA_BUNDLE
+    HOME
+    REQUESTS_CA_BUNDLE
+    SSL_CERT_FILE
```
ssbarnea: Before someone asks: these are needed if SSL proxies, othewise testing would fail with failure to validate certificates.

.travis.yml
```diff
@@ -150,3 +150,9 @@ install:
 
 script:
 - tox -v
+# test that install and uninstall from source works
```
webknjaz: This shouldn't be mangled into all jobs but should be its own job wrapped into a separate toxenv.

setup.cfg
```diff
@@ -20,6 +20,7 @@ project_urls =
     Source Code = https://github.com/ansible/ansible-lint
 description = Checks playbooks for practices and behaviour that could potentially be improved
 long_description = file: README.rst
+long_description_content_type = text/x-rst
```
webknjaz: Not really needed as it's the default but :+1:

tox.ini
webknjaz: Instead of a comment in PR, the best practice is to put a comment in the file itself.

tox.ini
```diff
@@ -31,10 +35,21 @@ usedevelop = True
...
+    twine>=1.13.0
+    wheel>=0.31.0
+
 usedevelop = False
 # Ref: https://twitter.com/di_codes/status/1044358639081975813
 commands =
-    python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
-    twine check .tox/dist/*
+    rm -rf {toxinidir}/dist
+    python -m pep517.build \
```
webknjaz: 1) this should be a separate testenv, you can then reuse parts of it in other places. also, you can specify deps between testenvs.
2) this is achievable by using `isolated_build = true` so not sure why you want to invoke it explicitly.

tox.ini
```diff
@@ -31,10 +35,21 @@ usedevelop = True
...
+    setuptools>=39.0
+    twine>=1.13.0
+    wheel>=0.31.0
+
 usedevelop = False
 # Ref: https://twitter.com/di_codes/status/1044358639081975813
 commands =
-    python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
-    twine check .tox/dist/*
+    rm -rf {toxinidir}/dist
```
webknjaz: This is not cross-platform. Better do
```suggestion
    python -c "import path; path.Path('{toxinidir}/dist').rmtree_p()"
```

tox.ini
```diff
@@ -31,10 +35,21 @@ usedevelop = True
 
 [testenv:metadata-validation]
 deps =
-    collective.checkdocs
-    twine
+    collective.checkdocs>=0.2
+    pep517 >= 0.5.0
```
webknjaz: plz, keep the same style of spacing around comparison operators.

setup.py
```diff
@@ -1,150 +1,5 @@
-#! /usr/bin/env python
-"""Ansible-lint distribution package setuptools installer."""
-
+#!/usr/bin/env python
```
webknjaz: Plz, don't remove the space after `#!`, it's there for readability.
```suggestion
#! /usr/bin/env python
```

setup.py
```diff
@@ -1,150 +1,5 @@
...
-    import setuptools_scm.version  # only present during setup time
-    IS_PYPI_UPLOAD = os.getenv('PYPI_UPLOAD') == 'true'
-    return (
-        '' if IS_PYPI_UPLOAD
-        else setuptools_scm.version.get_local_node_and_date(version)
-    )
-
-
-setup_params['use_scm_version'] = {
-    'local_scheme': cut_local_version_on_upload,
```
webknjaz: Removing this breaks uploads to Test PyPI.

tox.ini
ssbarnea: > This is not cross-platform. Better do

While 3rd party module are you referring to because this is clearly not part of standard library. I am not sure if for the sake of portability we need to introduce a new dependency.

Can you please elaborate? What is the library? Is it available for py27...py37?

.travis.yml
ssbarnea: We need to test this with all supported python versions but I also want to avoid exploding the number of environments we have (extra footprint). For the moment this should do the trick, we can improve it later. Lets not forget we are trying to fix an almost critical bug #590, not rewrite the entire product. I will be more than happy to add extra improvement latest if you can review them.

tox.ini
webknjaz: `path.py` and yes, it works on 2.7

setup.cfg
```diff
@@ -64,7 +65,7 @@ keywords =
 use_scm_version = True
 python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*
 package_dir =
-    = lib
+    =lib
```
webknjaz: this is not really needed

setup.py
```diff
@@ -1,150 +1,23 @@
...
 
-
+setup_params = {}
 setup_params['use_scm_version'] = {
     'local_scheme': cut_local_version_on_upload,
 }
 
 
-__name__ == '__main__' and setuptools.setup(**setup_params)
+__name__ == "__main__" and setuptools.setup(**setup_params)
```
webknjaz: Why did you change the style?

.travis.yml
webknjaz: FWIW the quality of this PR is unacceptable, it breaks compatibility with supported envs and mixes up several completely separate changes. So this at least should go into existing testenv in tox where it belongs, not into a selective CI service config.

setup.py
```diff
@@ -1,147 +1,20 @@
...
 def cut_local_version_on_upload(version):
+    """Generate a PEP440 local version if uploading to PyPI."""
     import os
     import setuptools_scm.version  # only present during setup time
-    IS_PYPI_UPLOAD = os.getenv('PYPI_UPLOAD') == 'true'
+
+    IS_PYPI_UPLOAD = os.getenv('PYPI_UPLOAD') == 'true'  # set in tox.ini
     return (
-        '' if IS_PYPI_UPLOAD
+        ""
```
webknjaz: Plz, don't restructure this, minimize the diff.

setup.py
```diff
@@ -1,147 +1,20 @@
...
+        if IS_PYPI_UPLOAD
         else setuptools_scm.version.get_local_node_and_date(version)
     )
 
-
+setup_params = {}
```
webknjaz:
```suggestion

setup_params = {}
```

tox.ini
```diff
@@ -35,10 +37,18 @@ usedevelop = True
 
 [testenv:metadata-validation]
 deps =
-    collective.checkdocs
-    twine
+  collective.checkdocs >= 0.2
```
webknjaz: Plz don't change the indentation.

setup.py
ssbarnea: Leftovers, fixing them. You can blame "black PTSD" for using double quotes. BTW, You should consider activating black on ansible-lint so we do not have to waste time discussing about formatting.

tox.ini
ssbarnea: I am afraid you cannot blame me for *FIXING* the identation. That makes the indentation on tox.ini deps consistent with previous deps line. Looks at the file, its a mix of 2 and 4 identations.

setup.py
webknjaz: No, black is still bad.

tox.ini
webknjaz: it's mostly 4 spaces. this PR is not about reformatting things. so yes, you're breaking it now.

webknjaz:
We cannot accept this atm but this example is good for reference for what we can do in the future. Closing, for now, we can reopen it later once somebody has time to work on fixing the shim or its removal is approved by a higher power.

setup.py
```diff
@@ -1,147 +1,19 @@
...
-    str(k): str(v)
-    for k, v in declarative_setup_params['options']['package_dir'].items()
-}
-
-setup_params = dict(setup_params, **declarative_setup_params['metadata'])
-setup_params = dict(setup_params, **declarative_setup_params['options'])
-
-
 def cut_local_version_on_upload(version):
+    """Generate a PEP440 local version if uploading to PyPI."""
```
webknjaz: This is not exactly true. It's more like "cut off a local version on upload"

setup.cfg
```diff
@@ -20,6 +20,7 @@ project_urls =
   Source Code = https://github.com/ansible/ansible-lint
 description = Checks playbooks for practices and behaviour that could potentially be improved
 long_description = file: README.rst
+long_description_content_type = text/x-rst
```
webknjaz: FTR this part of the PR will be accepted if submitted separately.

PR #598, opened 2019-10-23
Fixed deployment

ssbarnea:
Replaces older metada-validation targer with the more coprehensive
build-dist which builds packages using pep517 and also verifies
their metadata by usin twine check (replacement for the older method).

This also fixed deploy on travis which is wrongly calls setup.py when
found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,7 +116,9 @@ jobs:
     if: repo == "ansible/ansible-lint" AND tag IS present
     python: "3.7"
     env: &deploy-env
-      TOXENV: metadata-validation
+      TOXENV: build-dists,metadata-validation
+    before_deploy:
+    - echo > setup.py
     deploy: &deploy-step
       provider: pypi
       user: ansible-lint
--- a/tox.ini
+++ b/tox.ini
@@ -22,6 +22,24 @@ passenv =
     SSL_CERT_FILE  # https proxies
 # recreate = True
 
+[testenv:build-dists]
+basepython = python3
+isolated_build = true
+# `usedevelop = true` overrides `skip_install` instruction, it's unwanted
+usedevelop = false
+skip_install = true
+deps =
+  pep517 >= 0.5.0
+commands =
+  rm -rfv {toxinidir}/dist/
+  {envpython} -m pep517.build \
+    --source \
+    --binary \
+    --out-dir {toxinidir}/dist/ \
+    {toxinidir}
+whitelist_externals =
+  rm
+
 [testenv:flake8]
 deps = flake8
 commands = flake8
@@ -35,10 +53,14 @@ usedevelop = True
 
 [testenv:metadata-validation]
 deps =
+  {[testenv:build-dists]deps}
   collective.checkdocs
   twine
 usedevelop = False
+skip_install = {[testenv:build-dists]skip_install}
 # Ref: https://twitter.com/di_codes/status/1044358639081975813
 commands =
-  python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
-  twine check .tox/dist/*
+  {[testenv:build-dists]commands}
+  twine check {toxinidir}/dist/*
+whitelist_externals =
+  {[testenv:build-dists]whitelist_externals}


tox.ini
```diff
@@ -33,12 +56,3 @@ deps = sphinx
 commands = make -C docs/docsite htmldocs
 usedevelop = True
 
-[testenv:metadata-validation]
```
webknjaz: The validation env must remain separate. You may optionally specify an env dependency such as: `depends = build-dists`

.travis.yml
```diff
@@ -116,7 +116,9 @@ jobs:
     if: repo == "ansible/ansible-lint" AND tag IS present
     python: "3.7"
     env: &deploy-env
-      TOXENV: metadata-validation
+      TOXENV: build-dists
```
webknjaz: We should keep separate build and validation steps
```suggestion
      TOXENV: build-dists,metadata-validation
```

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.5.3
-envlist = flake8,py{37,36,35,27}-ansible{28,27,26,devel}
+envlist = flake8,build-dists,py{37,36,35,27}-ansible{28,27,26,devel}
```
webknjaz: I don't think it belongs here. We shouldn't run this env for normal testing.

.travis.yml
```diff
@@ -116,7 +116,9 @@ jobs:
...
+    before_deploy:
+      - echo > setup.py
```
webknjaz: This file doesn't use extra indentation for sequences.
```suggestion
    - echo > setup.py
```

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.5.3
-envlist = flake8,py{37,36,35,27}-ansible{28,27,26,devel}
+envlist = flake8,metadata-validation,py{37,36,35,27}-ansible{28,27,26,devel}
```
webknjaz: Let's not change this at all, please.

tox.ini
```diff
@@ -35,10 +55,14 @@ usedevelop = True
...
 usedevelop = False
+skip_install = true
 # Ref: https://twitter.com/di_codes/status/1044358639081975813
 commands =
-  python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
-  twine check .tox/dist/*
+  {[testenv:build-dists]commands}
+  twine check {toxinidir}/dist/*
+whitelist_externals =
+    rm
```
webknjaz: This is unnecessary

tox.ini
```diff
@@ -35,10 +55,14 @@ usedevelop = True
...
+  {[testenv:build-dists]deps}
   collective.checkdocs
   twine
 usedevelop = False
+skip_install = true
 # Ref: https://twitter.com/di_codes/status/1044358639081975813
 commands =
-  python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
-  twine check .tox/dist/*
+  {[testenv:build-dists]commands}
```
webknjaz: No, we don't need to embed building here. It should stay decoupled.

tox.ini
```diff
@@ -35,10 +55,14 @@ usedevelop = True
 
 [testenv:metadata-validation]
 deps =
```
webknjaz:
```suggestion
depends = build-dists
deps =
```

tox.ini
```diff
@@ -22,6 +22,26 @@ passenv =
...
+[testenv:build-dists]
+basepython = python3
+isolated_build = true
+# `usedevelop = true` overrides `skip_install` instruction, it's unwanted
+usedevelop = false
+skip_install = true
+deps =
+    pep517 >= 0.5.0
+setenv =
+    PYPI_UPLOAD = true
```
webknjaz: We shouldn't set this, only bypass the var from outside. This is necessary for proper validation. It must fail when wrongfully trying to upload to PyPI.

tox.ini
ssbarnea: @webknjaz This IS necessary because the inhered commands contain rm, and tox with throw a warning. If you remember few months back you were against setting a single whitelist_externals on molecule, so I suspect you didn't change your preference.

.travis.yml
```diff
@@ -116,7 +116,9 @@ jobs:
     if: repo == "ansible/ansible-lint" AND tag IS present
     python: "3.7"
     env: &deploy-env
-      TOXENV: metadata-validation
+      TOXENV: metadata-validation,build-dists
```
webknjaz: The order of execution matters. First, build, then validate.
```suggestion
      TOXENV: build-dists,metadata-validation
```
Alternatively, we could move `tox -e metadata-validation` to `before_deploy` step.

tox.ini
webknjaz: We shouldn't inherit the commands in the first place. This env should remain decoupled and we need to prefer composition over inheritance in this case.

tox.ini
ssbarnea: This is desired, we do not want to encourage developers to avoid testing the metadata when they do test code locally. Is listed the 2nd because is faster than running the unittests. The are no implications on CI.

webknjaz: Let's not discuss it here. It's unrelated to fixing deployment as the PR title states.

.travis.yml
webknjaz: Pro tip: it's important to accept suggested changes via GitHub UI since it adds a proper co-author commit trailers. You can then fetch such commit and do squashing once you know that metadata (it's not always public/easily accessible).

tox.ini
ssbarnea: @webknjaz Sorry but you are asking me to introduce a feature that added very recently to tox (3.7.0) and current minimal version is 3.5.3

Bumping minimal tox version would be outside the scope of this change.

tox.ini
```diff
@@ -35,10 +53,14 @@ usedevelop = True
...
 usedevelop = False
+skip_install = true
 # Ref: https://twitter.com/di_codes/status/1044358639081975813
 commands =
-  python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
-  twine check .tox/dist/*
+  {[testenv:build-dists]commands}
+  twine check {toxinidir}/dist/*
+whitelist_externals =
+  rm
```
webknjaz:
```suggestion
  {[testenv:build-dists]whitelist_externals}
```

tox.ini
ssbarnea: Not adding this here would be a serious bug, it would mean that we check a potentially empty folder, or an outdated one. Running tox on any environment should never made the assumption that the user already run successfully one of the other environments.

Tox environment executions are supposed to be predictable and easy to reproduce, not to depend on previous executions.

tox.ini
webknjaz: agreed

PR #601, opened 2019-10-24
Declare support for Python 3.8

webknjaz:


webknjaz:
This will need to wait until Ansible pins a newer version of ruamel.yaml

webknjaz:
Oh, wait, it's not Ansible's fault. We pin it.

PR #607, opened 2019-10-25
Fully depend on Pip having PEP 517 implementation

webknjaz:


pyproject.toml
```diff
@@ -4,5 +4,12 @@ requires = [
   "setuptools_scm >= 1.15.0",
   "setuptools_scm_git_archive >= 1.0",
   "wheel",
+
+  # needed because setuptools' attr import
+  # machinery (`importlib.import_module`)
+  # imports `__init__` first
+  "ansible",
```
ssbarnea: This could be a real problem as it would require installing ansible twice when installing from source. Why is ansible really needed for packaging ansible-lint?

lib/ansiblelint/versiontools.py
```diff
@@ -0,0 +1,37 @@
...
+        return get_version(
+            root=root,
+            relative_to=relative_to,
+            local_scheme=local_scheme,
+        )
+
+    return 'unknown'
+
+
+def cut_local_version_on_upload(version):
```
ssbarnea: I see this boilerplate added again and again in multiple places. What prevents adding it directly to setuptools_scm so we don't need it in each project?

pyproject.toml
webknjaz: Yes, because during the import of the version generation function it imports `__init__.py` which triggers `ImportError` otherwise.
Double install isn't a problem because of an isolated env.

lib/ansiblelint/versiontools.py
webknjaz: Feel free to open an issue there. I'm working with what I've got.

webknjaz:
@ssbarnea I've pushed `v4.1.1a2` so you can test it now.

Issue #611, opened 2019-10-31
when run as pre-commit hook ansible-lint silently fails to lint


ssbarnea:
I discovered that ansible-lint when run as a pre-commit hook ansible-lint gives false positives, mainly because it fails to identify the files types.

I will give a censored example:

```
# outside pre-commit
$ ansible-lint  -v .                                                                                                                                                 
Examining /Users/ssbarnea/rdo/ansible-role-collect-logs/tasks/publish.yml of type tasks
/Users/ssbarnea/rdo/ansible-role-collect-logs/tasks/ara_graphite.yml:2: [E301] Commands should not change things if nothing needs doing

# with pre-commit (same version)
$ pre-commit run -v -a ansible-lint                                                                                                                                       [ansible-lint] Ansible-lint..............................................Passed
hookid: ansible-lint
Examining /Users/ssbarnea/rdo/ansible-role-collect-logs/tasks/collect.yml of type playbook
```

Because pre-commit call ansible-lint for each file, this not only makes the process much slower but worse it prevents ansible lint from discovering the file types correctly. You can see that his guess was a playbook when in fact it was tasks inside roles.

One think for sure calling `ansible-lint /Users/ssbarnea/rdo/ansible-role-collect-logs/tasks/ara_graphite.yml` confuses it to think it is a playbook, but allowing it to find files seems to be working.

One quick fix for this would be to disable the file parsing support in

awcrosby:
@ssbarnea can you confirm this occurs with `v4.1.1a2` but not `v4.1.0`?

ssbarnea:
@awcrosby I am almost sure this bug was there from start, we just failed to see that when run as a hook it was linting *(differently)* than directly.

brucellino:
I landed here after I my CI was failing due to Ansible incorrectly linting my `.github/workflows` directory. Just wanted to say for the record that I implemented a workaround by passing `args` to `ansible-lint` instead of to `pre-commit`:

```yaml
---
repo:
  - repo: https://github.com/ansible-community/ansible-lint
    rev: v6.17.0
    hooks:
      - id: ansible-lint
        args: ["--exclude", ".github"]
```

ssbarnea:
@brucellino I advice you to avoid adding any custom arguments inside pre-commit config because it will make the linter behave different when run outside pre-commit hook, from command line or from other tools like vscode extension. Instead add needed settings inside linter config.

brucellino:
Excellent point, thank you!

PR #612, opened 2019-10-31
Fixed pre-commit hooks

ssbarnea:
Assures that when used as a pre-commit hook, ansible-lint is not
given each file as argument. This is needed because ansible-lint
needs to detect the code-base layout in order to figure-out which
YAML files are playbooks, tasks or vars files.

Previously the hook was passing all files that where matching its
filters and ansible-lint would assume they were playbooks. This made it
give false positives.

This change also adds the --force-color because without it ansible-lint
would fail to detect the TTY and avoid using colors.

Fixes: #611

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -6,6 +6,9 @@
 -   id: ansible-lint
     name: Ansible-lint
     description: This hook runs ansible-lint.
-    entry: ansible-lint
+    entry: ansible-lint --force-color .
     language: python
-    files: \.(yaml|yml)$
+    # do not pass files to ansible-lint, see:
+    # https://github.com/ansible/ansible-lint/issues/611
+    pass_filenames: false
+    always_run: true


ssbarnea:
@webknjaz I don't think that ansible-lint is the kind of linter that can be used on files without context. What if you try to install newer version? The newer version could find bugs in files that were not touched.

And yes, the idea about adding to CI is great!

webknjaz:
That command for CI doesn't work as I expected. It probably needs to be reported to pre-commit.

.pre-commit-hooks.yaml
```diff
@@ -6,6 +6,9 @@
 -   id: ansible-lint
     name: Ansible-lint
     description: This hook runs ansible-lint.
-    entry: ansible-lint
+    entry: ansible-lint --force-color .
```
asottile: `--force-color` shouldn't be necessary as of `pre-commit 1.19`

.pre-commit-hooks.yaml
```diff
@@ -6,6 +6,9 @@
...
     language: python
-    files: \.(yaml|yml)$
+    # do not pass files to ansible-lint, see:
+    # https://github.com/ansible/ansible-lint/issues/611
+    pass_filenames: false
+    always_run: true
```
asottile: this ~usually isn't the setting you want here -- this'll do a lot of unnecessary work on every commit (linting your entire codebase all the time) -- generally you want to optimize for quickness on commit and being thorough in CI (via `--all-files`, etc.)

.pre-commit-hooks.yaml
asottile: in fact, it's the _wrong thing to do_ if you're using a CI system, etc. which doesn't have color escapes

asottile:
> 
> 
> That command for CI doesn't work as I expected. It probably needs to be reported to pre-commit.

you'd want `pre-commit try-repo . --all-files` or something like that

Issue #613, opened 2019-11-01
ansible-lint needs a recursive discovery mode


ssbarnea:
# Issue Type
- Feature request

# Desired Behaviour

Currently ansible-lint cannot safely used like other linters which are able to look recursively inside the repository and lint whatever they find. Attempt to run `ansible-lint .` would result into a false-positive result, likely linting nothing if not playbook is present.

Once complex repository example can be seen at https://github.com/rdo-infra/ci-config/ as this exposes several important aspects:

- not all yaml files are Ansible related
- roles can be stored in multiple subdirectories: `ci-scripts/container-push/roles`,... and several others.
- playbooks can also be scattered around in multiple locations
- using exclude_paths can help but does not fix the inclusion part
- finding all .yaml files and passing them to ansible-lint (like pre-commit does by default) would confuse it because it may not be able to identify the nature of these files and may end-up trying to parse tasks or vars as playbooks.
- playbooks could be found inside `roledir/molecule/*.yml` with one notable exception: `molecule.yml` which is clearly not a playbook. I am not sure if we could just assume that the other files are playbooks or not.

While I do not mind making the assumption that `roles` are to found in folders with `roles` in their name, there is one signifiant exception: single-role repositories. 

Before doing any work on this, we may need some help from Ansible team regarding on what information can and should be used to determine what is an Ansible playbook or role.

# Desired outcome

If sorted correctly, **newly added playbooks and roles should be linted without having to reconfigure the ansible-linter**. 

The user should be able to run `ansible-lint` without any arguments to lint the entire repository, just like `flake8` and others are working.

Issue #564, opened 2019-08-08
4.1.1a0: Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)


ssbarnea:
```
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
```

# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible 2.8.1
ansible-lint 4.1.1a0
```

# Desired Behaviour

No errors or actionable ones.

# Actual Behaviour (Bug report only)

I will give only one example but this happens on multiple files. Content available at https://opendev.org/zuul/zuul-jobs/src/branch/master/roles/push-to-intermediate-registry/tasks/push-image.yaml --- which passed with older versions of Ansible lint.

```
Examining /Users/ssbarnea/rdo/zuul-jobs/roles/push-to-intermediate-registry/tasks/push-image.yaml of type block
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
Error trying to append skipped rules: RuntimeError('Unexpected file type: block',)
[...]

PR #615, opened 2019-11-01
Enable ansible-lint to auto-detect roles/playbooks

ssbarnea:
When called without any arguments, ansible-lint will now try to look
for all playbooks and roles inside current git repository.

Unrecognized YAML files will be displayed in verbose mode but will not
be considered errors. This will allow users to enable ansible-lint
on any repository without being forced to alter the way is called or
its configuration each time a new playbooks/role is added.

Fixes: #613
Fixes: #564

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -30,7 +30,7 @@ import ansiblelint.formatters as formatters
 import six
 from ansiblelint import default_rulesdir, RulesCollection, Runner
 from ansiblelint.version import __version__
-from ansiblelint.utils import normpath
+from ansiblelint.utils import get_playbooks_and_roles, normpath
 import yaml
 import os
 
@@ -52,7 +52,7 @@ def main():
 
     formatter = formatters.Formatter()
 
-    parser = optparse.OptionParser("%prog [options] playbook.yml [playbook2 ...]",
+    parser = optparse.OptionParser("%prog [options] [playbook.yml [playbook2 ...]]",
                                    version="%prog " + __version__)
 
     parser.add_option('-L', dest='listrules', default=False,
@@ -127,8 +127,8 @@ def main():
         if 'verbosity' in config:
             options.verbosity = options.verbosity + config['verbosity']
 
-        if 'exclude_paths' in config:
-            options.exclude_paths = options.exclude_paths + config['exclude_paths']
+        options.exclude_paths.extend(
+            config.get('exclude_paths', []))
 
         if 'rulesdir' in config:
             options.rulesdir = options.rulesdir + config['rulesdir']
@@ -148,9 +148,9 @@ def main():
     if options.parseable_severity:
         formatter = formatters.ParseableSeverityFormatter()
 
+    # no args triggers auto-detection mode
     if len(args) == 0 and not (options.listrules or options.listtags):
-        parser.print_help(file=sys.stderr)
-        return 1
+        args = get_playbooks_and_roles(options=options)
 
     if options.use_default_rules:
         rulesdirs = options.rulesdir + [default_rulesdir]
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -18,10 +18,17 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+from collections import OrderedDict
 import glob
 import imp
-import os
 from itertools import product
+import os
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+import subprocess
+
 
 import six
 from ansible import constants
@@ -53,6 +60,7 @@ except ImportError:
     from ansible.parsing.mod_args import ModuleArgsParser
     from ansible.parsing.yaml.constructor import AnsibleConstructor
     from ansible.parsing.yaml.loader import AnsibleLoader
+    from ansible.parsing.yaml.objects import AnsibleSequence
     from ansible.errors import AnsibleParserError
     ANSIBLE_VERSION = 2
 
@@ -742,3 +750,108 @@ def normpath(path):
     make this user configurable.
     """
     return os.path.relpath(path)
+
+
+def is_playbook(filename):
+    """
+    Check if the file is a playbook.
+
+    Given a filename, it should return true if it looks like a playbook. The
+    function is not supposed to raise exceptions.
+    """
+    # we assume is a playbook if we loaded a sequence of dictionaries where
+    # at least one of these keys is present:
+    playbooks_keys = {
+        "gather_facts",
+        "hosts",
+        "import_playbook",
+        "post_tasks",
+        "pre_tasks",
+        "roles"
+        "tasks",
+    }
+
+    # makes it work with Path objects by converting them to strings
+    if not isinstance(filename, six.string_types):
+        filename = str(filename)
+
+    try:
+        f = parse_yaml_from_file(filename)
+    except Exception as e:
+        print(
+            "Warning: Failed to load %s with %s, assuming is not a playbook."
+            % (filename, e))
+    else:
+        if (
+            isinstance(f, AnsibleSequence)
+            and playbooks_keys.intersection(next(iter(f), {}).keys())
+        ):
+            return True
+    return False
+
+
+def get_playbooks_and_roles(options=None):
+    """Find roles and playbooks."""
+    if options is None:
+        options = {}
+
+    # git is preferred as it also considers .gitignore
+    files = OrderedDict.fromkeys(sorted(subprocess.check_output(
+        ["git", "ls-files", "*.yaml", "*.yml"],
+        universal_newlines=True).split()))
+
+    playbooks = []
+    role_dirs = []
+    role_internals = {
+        'defaults',
+        'files',
+        'handlers',
+        'meta',
+        'tasks',
+        'templates',
+        'vars',
+    }
+
+    # detect role in repository root:
+    if 'tasks/main.yml' in files:
+        role_dirs.append('.')
+
+    for p in map(Path, files):
+
+        if any(str(p).startswith(file_path) for file_path in options.exclude_paths):
+            continue
+        elif (next((i for i in p.parts if i.endswith('playbooks')), None)
+                or 'playbook' in p.parts[-1]):
+            playbooks.append(normpath(p))
+            continue
+
+        # ignore if any folder ends with _vars
+        if next((i for i in p.parts if i.endswith('_vars')), None):
+            continue
+        elif 'roles' in p.parts or '.' in role_dirs:
+            if 'tasks' in p.parts and p.parts[-1] == 'main.yaml':
+                role_dirs.append(str(p.parents[1]))
+            elif role_internals.intersection(p.parts):
+                continue
+            elif 'tests' in p.parts:
+                playbooks.append(normpath(p))
+        if 'molecule' in p.parts:
+            if p.parts[-1] != 'molecule.yml':
+                playbooks.append(normpath(p))
+            continue
+        # hidden files are clearly not playbooks, likely config files.
+        if p.parts[-1].startswith('.'):
+            continue
+
+        if is_playbook(p):
+            playbooks.append(normpath(p))
+            continue
+
+        if options.verbosity:
+            print('Unknown file type: %s' % normpath(p))
+
+    if options.verbosity:
+        print('Found roles: ' + ' '.join(role_dirs))
+        print('Found playbooks: ' + ' '.join(playbooks))
+
+    return role_dirs + playbooks
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,7 @@ requires = [
   # machinery (`importlib.import_module`)
   # imports `__init__` first
   "ansible",
+  "pathlib2; python_version < '3.2'",
   "ruamel.yaml",
   "six",
 ]
--- a/setup.cfg
+++ b/setup.cfg
@@ -83,6 +83,7 @@ install_requires =
   ansible >= 2.7
   pyyaml
   six
+  pathlib2; python_version < "3.2"
   ruamel.yaml >= 0.15.34,<1; python_version < "3.7"
   ruamel.yaml >= 0.15.37,<1; python_version >= "3.7"
   # NOTE: per issue #509 0.15.34 included in debian backports


docs/docsite/rst/rules/default_rules.rst
```diff
@@ -20,7 +20,7 @@ ID
...
+`E202 <https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/OctalPermissionsRule.py>`_                                                                                                                                                                                                  historic                                                                                                                                                                                                                                                                                                     Octal file permissions must contain leading zero or be a string                                                                                                                                                                                                                                              Numeric file permissions without leading zero can behave in unexpected ways. See http://docs.ansible.com/ansible/file_module.html
```
webknjaz: Why does this show up in diff?

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,89 @@ def get_rule_skips_from_line(line):
...
+        'requirements.yml',
+    ]
+
+    # detect role in repository root:
+    if 'tasks/main.yml' in files:
+        role_dirs.append('.')
+        files.pop('tasks/main.yml')
+
+    for x in files.copy():
+        p = PurePath(x)
```
webknjaz: Why not just `Path`?

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,89 @@ def get_rule_skips_from_line(line):
         noqa_text = line.split('# noqa')[1]
         rule_id_list = noqa_text.split()
     return rule_id_list
+
+
+def is_playbook(filename):
+    try:
+        f = parse_yaml_from_file(filename)
+        if isinstance(f, AnsibleSequence) and hasattr(f[0], 'hosts'):
```
webknjaz: I think this would make more sense in the `else:` branch.

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,89 @@ def get_rule_skips_from_line(line):
...
+def get_playbooks_and_roles(verbose=False):
+
+    # git is preferred as it also considers .gitignore
+    files = OrderedDict.fromkeys(sorted(subprocess.check_output(
+        ["git", "ls-files", "*.yaml", "*.yml"],
+        universal_newlines=True).split()))
+
+    playbooks = []
+    role_dirs = []
+    role_internals = set(['defaults', 'vars', 'tasks', 'handlers', 'meta', 'files', 'templates'])
```
webknjaz: no need to create an intermediate in-memory list.
```suggestion
    role_internals = {'defaults', 'vars', 'tasks', 'handlers', 'meta', 'files', 'templates'}
```

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,89 @@ def get_rule_skips_from_line(line):
...
+    exclude_dirs = set([
```
webknjaz: Same here + I don't think these should be hardcoded right here. This needs a config with a default value.

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,89 @@ def get_rule_skips_from_line(line):
...
+        'inventory.yaml',
+        'requirements.yml',
+    ]
+
+    # detect role in repository root:
+    if 'tasks/main.yml' in files:
+        role_dirs.append('.')
+        files.pop('tasks/main.yml')
+
+    for x in files.copy():
```
webknjaz: This would be nicer:
```suggestion
    for p in map(PurePath, files.copy()):
```

tox.ini
```diff
@@ -16,6 +16,7 @@ deps =
   ansible29: ansible>=2.9,<2.10
   ansibledevel: git+https://github.com/ansible/ansible.git
   ruamel.yaml==0.16.5  # NOTE: 0.15.34 has issues with py37
+  pathlib2; python_version < '3.2'
```
webknjaz: This shouldn't be necessary as this dep should come from the dist.

ssbarnea: I agree with you but it does not! Please try it! My guess is that setup.cfg is not used at all due to removal of of setup.py — in fact we should not even see the ruamel here, all deps should be in a single place.

docs/docsite/rst/rules/default_rules.rst
ssbarnea: Running lint seems to reformat the file, it happened twice locally. I had no intention to touch it, still...

lib/ansiblelint/utils.py
ssbarnea: That's a conservative function, expected to return True only in idea conditions. I am not aware of a way to rewrite that logic in a more concise way (or at least not without affecting its readability). I am aware that I could remove
the last return and use the implicit None, but is kinda like it bit more explicit on that aspect.

Anyway, if you know a better way, propose a change here, I will gladly accept it.

lib/ansiblelint/utils.py

lib/ansiblelint/utils.py
ssbarnea: In fact we already have `exclude_paths` which are applied after this detection but I will reuse them. Since I introduced the is_playbook there is no danger in parsing all, so I could remove these from here. We share the sentiment about that hardcoded list.

lib/ansiblelint/utils.py
ssbarnea: Fixed

ssbarnea:
@webknjaz Please have another look on it, I hope I addressed all concerns.

If you want to see the output of calling it on multiple repositories, look at https://sbarnea.com/tmp/result.html -- i kinda scripted its execution in order to get an idea about what to expect in real-life once the code goes in.

tox.ini
webknjaz: Do you have evidence?

lib/ansiblelint/__init__.py
```diff
@@ -279,8 +279,11 @@ def run(self):
         # remove files that have already been checked
         files = [x for x in files if x['path'] not in self.checked_files]
         for file in files:
+            # normfile = ansiblelint.utils.normpath(file['path'])
```
webknjaz: What is this about?

lib/ansiblelint/__main__.py
```diff
@@ -127,7 +128,7 @@ def main():
             options.verbosity = options.verbosity + config['verbosity']
 
         if 'exclude_paths' in config:
-            options.exclude_paths = options.exclude_paths + config['exclude_paths']
+            options.exclude_paths.extend(config['exclude_paths'])
```
webknjaz: Since you are already refactoring it, let's just do
```suggestion
        options.exclude_paths.extend(
            config.get('exclude_paths', []),
        )
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -2,6 +2,7 @@
     from ansible import color
 except ImportError:
     from ansible.utils import color
+from ansiblelint.utils import normpath
```
webknjaz: Let's add an extra line to separate local imports from third party
```suggestion

from ansiblelint.utils import normpath
```

lib/ansiblelint/utils.py
```diff
@@ -22,6 +22,12 @@
 import imp
 import os
 from itertools import product
+import subprocess
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+from collections import OrderedDict
```
webknjaz: Please reorder this alphabetically

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,108 @@ def get_rule_skips_from_line(line):
...
+    except Exception as e:
+        print(
+            "Warning: Failed to load %s with %s, assuming is not a playbook."
+            % (filename, e))
+    return False
+
+
+def normpath(path):
+    """
+    Normalize a path in order to provide a more consistent output. Currently
```
webknjaz: Please separate a 72-char title sentence from the rest of the docstring by an empty line as required by PEP257

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,108 @@ def get_rule_skips_from_line(line):
...
+
+        if is_playbook(p):
+            playbooks.append(normpath(p))
+            continue
+
+        if options.verbosity:
+            print('Unknown file type: %s' % normpath(p))
+
+    if options.verbosity:
+        print("Found roles: " + " ".join(role_dirs))
```
webknjaz: Plz use single quotes to match the style of the rest of the module.

lib/ansiblelint/utils.py
```diff
@@ -53,6 +59,7 @@
     from ansible.parsing.mod_args import ModuleArgsParser
     from ansible.parsing.yaml.constructor import AnsibleConstructor
     from ansible.parsing.yaml.loader import AnsibleLoader
+    from ansible.parsing.yaml.objects import AnsibleSequence
     from ansible.errors import AnsibleParserError
     ANSIBLE_VERSION = 2
```
webknjaz: Wow, it looks like this whole try/except needs to be cleaned up because we don't support Ansible 1

webknjaz: But that's out of scope right now. My concern is that if someone is running it with Ansible 1 for whatever reason, they may hit `NameError`

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,108 @@ def get_rule_skips_from_line(line):
         noqa_text = line.split('# noqa')[1]
         rule_id_list = noqa_text.split()
     return rule_id_list
+
+
+def is_playbook(filename):
+    """
```
webknjaz:
```suggestion
    """Check if the file is a playbook.

```

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,108 @@ def get_rule_skips_from_line(line):
...
+        "tasks",
+    }
+
+    # makes it work with Path objects by converting them to strings
+    if not isinstance(filename, six.string_types):
+        filename = str(filename)
+
+    try:
+        f = parse_yaml_from_file(filename)
+        if isinstance(f, AnsibleSequence) and playbooks_keys.intersection(f[0].keys()):
```
webknjaz: This belongs to the else-block

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,108 @@ def get_rule_skips_from_line(line):
...
+
+    # detect role in repository root:
+    if 'tasks/main.yml' in files:
+        role_dirs.append('.')
+
+    for p in map(Path, files):
+
+        if any(str(p).startswith(file_path) for file_path in options.exclude_paths):
+            continue
+        elif next((i for i in p.parts if i.endswith('playbooks')), None) \
```
webknjaz: Please use braces instead of escaping EOL

lib/ansiblelint/utils.py
```diff
@@ -732,3 +739,108 @@ def get_rule_skips_from_line(line):
...
+    role_dirs = []
+    role_internals = {
+        'defaults',
+        'files',
+        'handlers',
+        'meta',
+        'tasks',
+        'templates',
+        'vars',
+        }
```
webknjaz:
```suggestion
    }
```

lib/ansiblelint/utils.py
ssbarnea: @webknjaz I had impression that code-formatting was sometime to be avoided in reviews. That is why we have linters and formatters to take care of these. Discussions about style are a waste of human time. I will do the change but lets find a way to avoid them in the first place.

lib/ansiblelint/utils.py
ssbarnea: @webknjaz Can you please suggest how you see it working? If I move it to a try-else block it will be more verbose and buggy because f[0] could throw a IndexError itself if the sequence is empty. Current code does run ok with it.

lib/ansiblelint/utils.py
ssbarnea: Including normalization in this review what not the best idea, I decided to create another PR at https://github.com/ansible/ansible-lint/pull/620 for implementing that, it would make review much easier. I will implement the changes on the new PR and rebase this once normalization lands.

ssbarnea: Re PEP257, lets address it the right way, and have CI check its compliance. See https://github.com/ansible/ansible-lint/pull/621 as a start.

lib/ansiblelint/__init__.py
ssbarnea: Leftover, fixed

tox.ini
ssbarnea: I will do a test, I will comment the line and see what CI tells us.

docs/docsite/rst/rules/default_rules.rst
webknjaz: Plz drop this change and let's deal with it separately. I want to understand what the problem is instead of blindly ignoring it.

lib/ansiblelint/utils.py
webknjaz: This the current implementation it's totally unobvious what exceptions each expression can raise. You could do a nested try/except inside. Or even suppress CM. Alternatively, do `next(iter(f), {}})`.

lib/ansiblelint/utils.py
```diff
@@ -732,3 +740,108 @@ def get_rule_skips_from_line(line):
...
+
+    try:
+        f = parse_yaml_from_file(filename)
+        if isinstance(f, AnsibleSequence) and playbooks_keys.intersection(f[0].keys()):
+            return True
+    except Exception as e:
+        print(
+            "Warning: Failed to load %s with %s, assuming is not a playbook."
+            % (filename, e))
+    return False
```
webknjaz: Option 1:
```suggestion
    else:
        with contextlib.suppress(IndexError):
            if isinstance(f, AnsibleSequence) and playbooks_keys.intersection(f[0].keys()):
                return True
    return False
```

webknjaz: Option 2:
```suggestion
    else:
        if (
            isinstance(f, AnsibleSequence)
            and playbooks_keys.intersection(next(iter(f), {}}).keys())
        ):
            return True
    return False
```

lib/ansiblelint/utils.py
```diff
@@ -742,3 +750,105 @@ def normpath(path):
...
+    else:
+        if (
+            isinstance(f, AnsibleSequence)
+            and playbooks_keys.intersection(next(iter(f), {}).keys())
+        ):
+            return True
+    return False
+
+
+def get_playbooks_and_roles(options={}):
```
webknjaz: Mutable default values are harmful
```suggestion
def get_playbooks_and_roles(options=None):
    if options is None:
        options = {}

```

ssbarnea:
Anything else preventing this to be merged?

lib/ansiblelint/utils.py
```diff
@@ -742,3 +750,108 @@ def normpath(path):
...
+    return False
+
+
+def get_playbooks_and_roles(options=None):
+    """Find roles and playbooks."""
+    if options is None:
+        options = {}
+
+    # git is preferred as it also considers .gitignore
+    files = OrderedDict.fromkeys(sorted(subprocess.check_output(
```
webknjaz: @ssbarnea why did you use `OrderedDict` and not a list here?

PR #621, opened 2019-11-03
Enable flake-docstrings to check for pep257

ssbarnea:
Enables checking of docstrings in order to assure compliance with
PEP257.

Because the codebase was never checked for this we are forced to disable
some of the rules and address them in follow-ups. This will ease code
reviews and avoid having too wide changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,3 +5,5 @@ repos:
   hooks:
   - id: flake8
     language_version: python3
+    additional_dependencies:
+    - flake8-docstrings
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -68,7 +68,7 @@ class AnsibleLintRule(object):
             message = None
             if isinstance(result, six.string_types):
                 message = result
-            matches.append(Match(prev_line_no+1, line,
+            matches.append(Match(prev_line_no + 1, line,
                            file['path'], self, message))
         return matches
 
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -119,7 +119,7 @@ def main():
 
         if 'parseable_severity' in config:
             options.parseable_severity = options.parseable_severity or \
-                                         config['parseable_severity']
+                config['parseable_severity']
 
         if 'use_default_rules' in config:
             options.use_default_rules = options.use_default_rules or config['use_default_rules']
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -1,0 +1,1 @@
+"""All internal ansible-lint rules."""
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,17 @@ universal = 1
 [flake8]
 max-line-length = 100
 exclude = .git,.hg,.svn,__pycache__,.tox,.eggs,env
+ignore =
+  # TODO(ssbarnea): gradually address exceptions below:
+  D100  # D100 Missing docstring in public module
+  D101  # D101 Missing docstring in public class
+  D102  # D102 Missing docstring in public method
+  D103  # D103 Missing docstring in public function
+  D104  # D104 Missing docstring in public package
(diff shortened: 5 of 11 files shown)


lib/ansiblelint/__init__.py
```diff
@@ -68,7 +68,7 @@ def matchlines(self, file, text):
             message = None
             if isinstance(result, six.string_types):
                 message = result
-            matches.append(Match(prev_line_no+1, line,
+            matches.append(Match(prev_line_no + 1, line,
```
webknjaz: This is not a docstring

tox.ini
```diff
@@ -54,7 +54,10 @@ commands =
     {toxinidir}
 
 [testenv:flake8]
-deps = flake8
+deps =
+  flake8
+  flake8-docstrings
```
webknjaz: I'd prefer pre-commit with pydocstyle, but this'll do for starters.

lib/ansiblelint/rules/__init__.py
```diff
@@ -0,0 +1 @@
+"""Hosts all internal ansible-lint rules."""
```
webknjaz: This is a package docstring, not a function docstring. For "objects" we describe what they are. For functions/methods we describe what they *do*. So use a verb only for those.
```suggestion
"""All internal ansible-lint rules."""
```

lib/ansiblelint/__init__.py
ssbarnea: I guess that is obvious **but** flake8-docstrings complained about it, so I fixed it.

lib/ansiblelint/__main__.py
```diff
@@ -118,7 +118,7 @@ def main():
 
         if 'parseable_severity' in config:
             options.parseable_severity = options.parseable_severity or \
-                                         config['parseable_severity']
+                config['parseable_severity']
```
webknjaz: This reads weirdly, I'd prefer braces. But again, this has nothing to do with docstrings so should be submitted separately.

lib/ansiblelint/__init__.py
webknjaz: really? it's hard to believe. are you sure it wasn't just flake8 itself? (w/o plugins)

tox.ini
ssbarnea: @webknjaz Glad to hear about pre-commit, I wanted to enable it myself but I was not sure what's your take on it.

.pre-commit-config.yaml
```diff
@@ -5,3 +5,5 @@ repos:
   hooks:
   - id: flake8
     language_version: python3
+    additional_dependencies:
+      - flake8-docstrings
```
webknjaz:
```suggestion
    - flake8-docstrings
```

ssbarnea:
@webknjaz My impression is that we should disable building docs with py38 as is currently broken, I raised https://github.com/sphinx-doc/sphinx/issues/6803 and I doubt is had anything to do with my change. I got the same error locally as on CI.

webknjaz:
@ssbarnea that is correct. Python 3.8 **under macOS** specifically. But I haven't had time to figure out an elegant way to do this.

setup.cfg
```diff
@@ -7,6 +7,17 @@ universal = 1
...
+ignore =
+  # TODO(ssbarnea): gradually address exceptions below:
+  D100  # D100 Missing docstring in public module
+  D101  # D101 Missing docstring in public class
+  D102  # D102 Missing docstring in public method
+  D103  # D103 Missing docstring in public function
+  D104  # D104 Missing docstring in public package
+  D105  # D105 Missing docstring in magic method
+  D107  # D107 Missing docstring in __init__
+  # W503 is incompatible with W504
```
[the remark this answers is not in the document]
webknjaz: Okay, please make a separate issue for this then.

.pre-commit-config.yaml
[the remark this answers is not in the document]
webknjaz: I'd argue that it's more illustrative as a separate check.

webknjaz:
Ideally, we need a way to apply `-j 0` just for that job.

PR #623, opened 2019-11-03
Adopt use of pre-commit for calling flake8

ssbarnea:
This enables us to easily add other linters and to better control
the linter versions, without risking of getting different results based
on their versions (flake8 version was not pinned before).

Removes flake8 excludes, so we cover our entire codebase.

Fixes few rules that were not covered yet by our codebase:
```
E201 whitespace after '['
E202 whitespace before ']'
E231 missing whitespace after ','
E251 unexpected spaces around keyword / parameter equals
E302 expected 2 blank lines, found 1
E501 line too long (108 > 100 characters)
F841 local variable 'result' is assigned to but never used
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+---
+repos:
+- repo: https://gitlab.com/pycqa/flake8.git
+  rev: 3.7.9
+  hooks:
+  - id: flake8
+    language_version: python3
--- a/test/TestUseHandlerRatherThanWhenChanged.py
+++ b/test/TestUseHandlerRatherThanWhenChanged.py
@@ -1,7 +1,8 @@
 import unittest
 
 from ansiblelint import RulesCollection
-from ansiblelint.rules.UseHandlerRatherThanWhenChangedRule import UseHandlerRatherThanWhenChangedRule
+from ansiblelint.rules.UseHandlerRatherThanWhenChangedRule import (
+    UseHandlerRatherThanWhenChangedRule)
 from test import RunFromText
 
 
--- a/test/TestUtils.py
+++ b/test/TestUtils.py
@@ -67,7 +67,9 @@ class TestUtils(unittest.TestCase):
     def test_normalize_simple_command(self):
         task1 = dict(name="hello", action="command chdir=abc echo hello world")
         task2 = dict(name="hello", command="chdir=abc echo hello world")
-        self.assertEqual(utils.normalize_task(task1, 'tasks.yml'), utils.normalize_task(task2, 'tasks.yml'))
+        self.assertEqual(
+            utils.normalize_task(task1, 'tasks.yml'),
+            utils.normalize_task(task2, 'tasks.yml'))
 
     def test_normalize_complex_command(self):
         task1 = dict(name="hello", action={'module': 'ec2',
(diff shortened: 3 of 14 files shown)


test/TestUtils.py
```diff
@@ -77,25 +79,37 @@ def test_normalize_complex_command(self):
...
+        self.assertEqual(
+            utils.normalize_task(task1, 'tasks.yml'),
+            utils.normalize_task(task4, 'tasks.yml'))
 
     def test_extract_from_list(self):
         block = dict(
-                block = [dict(tasks=[dict(name="hello",command="whoami")])],
-                test_none = None,
-                test_string = 'foo'
+                block=[dict(tasks=[dict(name="hello", command="whoami")])],
```
webknjaz: Let's maybe turn it into a proper dict literal then?

tox.ini
```diff
@@ -53,9 +53,10 @@ commands =
     --out-dir {toxinidir}/dist/ \
     {toxinidir}
 
-[testenv:flake8]
-deps = flake8
-commands = flake8
+[testenv:lint]
+deps = pre-commit>=1.20.0
+skip_install = true
+commands = python -m pre_commit run -a
```
webknjaz:
```suggestion
commands =
    python -m pre_commit run {posargs:--all-files}
```

tox.ini
```diff
@@ -53,9 +53,10 @@ commands =
     --out-dir {toxinidir}/dist/ \
     {toxinidir}
 
-[testenv:flake8]
-deps = flake8
-commands = flake8
+[testenv:lint]
```
webknjaz: I usually call it just
```suggestion
[testenv:pre-commit]
```

This also helps understand what's going on when you want to pass custom args via `tox -e pre-commit -- <custom_args_here>`.

webknjaz: Also needs to be changed @ GH Actions workflows

tox.ini
```diff
@@ -53,9 +53,10 @@ commands =
...
+deps = pre-commit>=1.20.0
```
webknjaz: Let's keep it on a separate line, it's better for diffs.
```suggestion
deps =
    pre-commit>=1.20.0
```

test/TestUseHandlerRatherThanWhenChanged.py
```diff
@@ -1,7 +1,8 @@
 import unittest
 
 from ansiblelint import RulesCollection
-from ansiblelint.rules.UseHandlerRatherThanWhenChangedRule import UseHandlerRatherThanWhenChangedRule
+from ansiblelint.rules.UseHandlerRatherThanWhenChangedRule \
```
webknjaz: Let's use braces plz

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+---
+repos:
+  - repo: https://gitlab.com/pycqa/flake8.git
+    rev: 3.7.9
+    hooks:
+      - id: flake8
```
webknjaz: Let's reduce sequence indents as per https://github.com/ansible/ansible-lint/blob/master/.yamllint

tox.ini
ssbarnea: Please lets not do this and I can explain why:
* tox -e lint is short and easy to type and future proof
* "pre-commit" is the current implementation of linting, maybe in the future we decide to sue something else, user/developer should not care about what to run to lint his change 

What we did on OpenStack which historically had lots of [testenv]s for each linter was to introduce `lint` or `linters` environment without removing the old ones and later (6mo-1y) removed them.  This was convenient for developers used to do just `tox -e flake9|pep8` as it have them time to adjust without swearing too often.

I am particularly opposed to use pre-commit because is more than double the number of chars to type, for something that any developer should run before creating a PR. Many forget to do it even when is short.

Does my reasoning makes sense?

test/TestUtils.py
ssbarnea: Sure

tox.ini
ssbarnea: sure!

tox.ini
```diff
@@ -53,9 +53,20 @@ commands =
...
 
+# deprecated: use more generic 'lint' instead
 [testenv:flake8]
-deps = flake8
-commands = flake8
+deps = {[testenv:lint]deps}
+envdir = {toxworkdir}/lint
+skip_install = true
+commands =
+  python -m pre_commit run -a flake8
```
webknjaz: Plz use the long version of the arg:
```suggestion
  python -m pre_commit run --all-files flake8
```

ssbarnea:
> Also, GH Actions and Azure Pipelines still need to use a new tox env specified

I am fully aware of this but it cannot be addressed with a PR due to reasons mentioned in commit message. You have to fix it manually using the web interface. GH forbits opening PRs that touch these files.

webknjaz:
Merged manually

PR #629, opened 2019-11-08
DNM: used to test CI status

ssbarnea:
Ignore this PR, we should used it to test if CI tests are working well on master.


webknjaz:
I believe this is no longer needed.

PR #630, opened 2019-11-11
Disable docs build on macos with py38

ssbarnea:
We avoid broken Sphinx on specific platform until upstream bug is fixed.

CI was broken by enablement of broken job via a merge that ignored CI
result.

### DIFF
--- a/.github/workflows/tox-linters.yml
+++ b/.github/workflows/tox-linters.yml
@@ -41,6 +41,11 @@ jobs:
         # - TOXENV: pre-commit-ci
         # - TOXENV: check-docs
         # - TOXENV: build-docs
+        exclude:
+        # Excluded until sphinx bug is fixed
+        # https://github.com/sphinx-doc/sphinx/issues/6803
+        - os: macOS-10.14
+          python-version: 3.8
 
     steps:
     - uses: actions/checkout@master


.github/workflows/tox-linters.yml
```diff
@@ -41,6 +41,11 @@ jobs:
         # - TOXENV: pre-commit-ci
         # - TOXENV: check-docs
         # - TOXENV: build-docs
+        exclude:
```
webknjaz: Oh, I didn't know there was an `exclude:`. Where's the docs for it?

.github/workflows/tox-linters.yml
```diff
@@ -41,6 +41,11 @@ jobs:
...
+        # Excluded until sphinx bug is fixed
+        # https://github.com/sphinx-doc/sphinx/issues/6803
+        - os: macOS-10.14
```
webknjaz: btw, this OS is being deprecated. this should be switched to `macOS-latest`

.github/workflows/tox-linters.yml
webknjaz: Found it: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix.

PR #631, opened 2019-11-12
Fix role detection to include tasks/main.yml

ssbarnea:
Previous role detection logic was looking only for tasks/main.yaml to
determine what looks to be a role. Now it also includes tasks/main.yml

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -786,7 +786,7 @@ def get_playbooks_and_roles(options=None):
     }
 
     # detect role in repository root:
-    if 'tasks/main.yml' in files:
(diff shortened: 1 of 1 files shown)


lib/ansiblelint/utils.py
```diff
@@ -786,7 +786,7 @@ def get_playbooks_and_roles(options=None):
     }
 
     # detect role in repository root:
-    if 'tasks/main.yml' in files:
+    if 'tasks/main.yml' in files or 'tasks/main.yml' in files:
```
webknjaz: This does nothing.
```suggestion
    if 'tasks/main.yml' in files:
```

ssbarnea: True, typo. Corrected.

PR #637, opened 2019-11-25
Re-enable GitHub actions workflows

webknjaz:


webknjaz:
@ssbarnea got Actions back ^

ssbarnea:
That part is good but also wish we would stop using emoji in code. Ok to use them in comments but I am not happy to see them in code, bots, emails from bots, too disruptive.

PR #642, opened 2019-12-05
[WIP] Support tasks using modules from collections

### LINKED ISSUE #372: Error for custom/local module
# Issue Type
- Bug report

```
$ ansible --version
ansible 2.6.4
  config file = /Users/bkaplan/si/ops/ansible/ansible.cfg
  configured module search path = [u'/Users/bkaplan/si/ops/ansible/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.6.4/libexec/lib/python2.7/site-packages/ansible
  python version = 2.7.15 (default, Sep 18 2018, 20:16:18) [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)]

$ ansible-lint --version
ansible-lint 3.4.23
```

- ansible installation method: one of source, pip, OS package
brew install ansible
- ansible-lint installation method: one of source, pip, OS package
brew install ansible-lint

# Desired Behaviour

I have a custom module, `modules/logrotate.py` which is used like
```
    - name: Configure logrotate
      logrotate:
        name: "{{item.name}}"
        path: "{{item.pattern}}"
        options: "{{logrotate_options}}"
      with_items: []
```
ansible.cfg contains
```
library = modules
```

When I run `ansible-lint ansible/playbooks/hosts/configure/agent-tunneller.yml ` I expect the module `logrotate` to not cause an error.

# Actual Behaviour

What I get instead is
```
❯ ansible-lint ansible/playbooks/hosts/configure/agent-tunneller.yml                                                 ops/git/master !
Couldn't parse task at ansible/playbooks/hosts/configure/agent-tunneller.yml:258 (no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '<unicode string>': line 258, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

(could not open file to display line))
{ u'logrotate': { '__file__': 'ansible/playbooks/hosts/configure/agent-tunneller.yml',
                  '__line__': 260,
                  u'name': u'{{item.name}}',
                  u'options': u'{{logrotate_options}}',
                  u'path': u'{{item.pattern}}'},
  u'name': u'Configure logrotate',
  u'with_items': []}
```

Where it seems that ansible-lint does not detect custom modules.

### LINKED ISSUE #538: Collection modules inside tasks cause exceptions
## Issue

## Background

## Proposal

awcrosby:
With this PR ansible-lint with ansible 2.9 uses updated ansible parser logic to determine the module action even though the collection is not installed (which it is not assumed to be during linting). So this allows linting of tasks using a collection module. This PR resolves #538, and resolves #372 (errors using local modules)

Notes:
- TestAlwaysRunRule.py can be removed since it is skipped for 2.7+ and ansible-lint requires 2.7+
- This PR skips testing against ansible devel since https://github.com/ansible/ansible/blob/devel/lib/ansible/release.py#L22 dev version is not semver compliant, this would work if it were `2.10.0-dev0` or `2.10.0+dev0`

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,16 +12,16 @@ strategy:
   matrix:
     Python27:
       python.version: '2.7'
-      TOXENV: ansible{29,28,27,devel}
+      TOXENV: ansible{29,28,27}
     Python35:
       python.version: '3.5'
-      TOXENV: ansible{29,28,27,devel}
+      TOXENV: ansible{29,28,27}
     Python36:
       python.version: '3.6'
-      TOXENV: ansible{29,28,27,devel}
+      TOXENV: ansible{29,28,27}
     Python37:
       python.version: '3.7'
-      TOXENV: ansible{29,28,27,devel}
+      TOXENV: ansible{29,28,27}
 
 steps:
 - task: UsePythonVersion@0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ requires = [
   "ansible",
   "pathlib2; python_version < '3.2'",
   "ruamel.yaml",
+  "semver",
   "six",
 ]
 build-backend = "setuptools.build_meta"
--- a/setup.cfg
+++ b/setup.cfg
@@ -93,6 +93,7 @@ setup_requires =
 install_requires =
   ansible >= 2.7
   pyyaml
+  semver
   six
   pathlib2; python_version < "3.2"
   ruamel.yaml >= 0.15.34,<1; python_version < "3.7"
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.5.3
-envlist = lint,py{38,37,36,35,27}-ansible{29,28,27,devel}
+envlist = lint,py{38,37,36,35,27}-ansible{29,28,27}
 isolated_build = true
 requires =
   setuptools >= 41.4.0
@@ -14,13 +14,13 @@ deps =
   ansible27: ansible>=2.7,<2.8
   ansible28: ansible>=2.8,<2.9
   ansible29: ansible>=2.9,<2.10
-  ansibledevel: git+https://github.com/ansible/ansible.git
   ruamel.yaml==0.16.5  # NOTE: 0.15.34 has issues with py37
   flake8
   pep8-naming
   pytest
   pytest-cov
   pytest-xdist
+  semver
   # Needed to avoid DeprecationWarning errors in pytest:
   setuptools >= 40.4.3
   wheel
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -29,6 +29,7 @@ except ImportError:
     from pathlib2 import Path
 import subprocess
 
+import semver
 import six
 import yaml
 from yaml.composer import Composer
@@ -45,6 +46,7 @@ from ansible.parsing.yaml.loader import AnsibleLoader
 from ansible.parsing.yaml.objects import AnsibleSequence
 from ansible.plugins.loader import module_loader
 from ansible.template import Templar
+from ansible import __version__ as ANSIBLE_VERSION
 
 
 # ansible-lint doesn't need/want to know about encrypted secrets, so we pass a
@@ -364,7 +366,11 @@ def normalize_task_v2(task):
     result = dict()
     mod_arg_parser = ModuleArgsParser(task)
     try:
-        action, arguments, result['delegate_to'] = mod_arg_parser.parse()
+        if semver.match(ANSIBLE_VERSION, '<2.9.0'):
+            action, arguments, result['delegate_to'] = mod_arg_parser.parse()
+        else:
+            action, arguments, result['delegate_to'] = mod_arg_parser.parse(
+                skip_action_validation=True)
     except AnsibleParserError as e:
         try:
             task_info = "%s:%s" % (task[FILENAME_KEY], task[LINE_NUMBER_KEY])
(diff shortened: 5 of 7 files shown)


ssbarnea:
Any update on this? It could prove very useful.

greg-hellings:
Stumbled here independent of @ssbarnea because I'm running into this issue linting roles inside of a collection. Would love to have this available.

ignatenkobrain:
@netzvieh ping?

ignatenkobrain:
@nitzmahone ping?

ignatenkobrain:
@webknjaz ping?

webknjaz:
> * dev version is not semver compliant, this would work if it were `2.10.0-dev0` or `2.10.0+dev0`

This is not entirely true. It's is a PEP440-based server. Python packaging ecosystem specifically requires following PEP440. `+` is reserved for "local" version part (e.g. when OS vendors patch things). And `-`/`_` are only allowed in some places but are normalized as `.` in a canonical form.

tox.ini
```diff
@@ -14,13 +14,13 @@ deps =
...
   ansible28: ansible>=2.8,<2.9
   ansible29: ansible>=2.9,<2.10
-  ansibledevel: git+https://github.com/ansible/ansible.git
   ruamel.yaml==0.16.5  # NOTE: 0.15.34 has issues with py37
   flake8
   pep8-naming
   pytest
   pytest-cov
   pytest-xdist
+  semver
```
webknjaz, line 23: Test deps don't require duplicate runtime entries.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.5.3
-envlist = lint,py{38,37,36,35,27}-ansible{29,28,27,devel}
+envlist = lint,py{38,37,36,35,27}-ansible{29,28,27}
```
webknjaz, line 3: When `packaging`'s used, this will be unnecessary.

setup.cfg
```diff
@@ -93,6 +93,7 @@ setup_requires =
 install_requires =
   ansible >= 2.7
   pyyaml
+  semver
```
webknjaz, line 96:
```suggestion
  importlib-metadata
  packaging
```

pyproject.toml
```diff
@@ -11,6 +11,7 @@ requires = [
   "ansible",
   "pathlib2; python_version < '3.2'",
   "ruamel.yaml",
+  "semver",
```
webknjaz, line 14: unnecessary
```suggestion
```

azure-pipelines.yml
```diff
@@ -12,16 +12,16 @@ strategy:
...
-      TOXENV: ansible{29,28,27,devel}
+      TOXENV: ansible{29,28,27}
     Python36:
       python.version: '3.6'
-      TOXENV: ansible{29,28,27,devel}
+      TOXENV: ansible{29,28,27}
     Python37:
       python.version: '3.7'
-      TOXENV: ansible{29,28,27,devel}
+      TOXENV: ansible{29,28,27}
```
webknjaz, line 24: these changes are unnecessary too

ssbarnea:
@awcrosby @webknjaz I think we can safely close this PR because current version of ansible-lint is able to recognize modules from collections without any problems. There is no change needed to fix that.

xenlo:
@ssbarnea, can you be more explicit on «_current version of ansible-lint is able to recognize modules from collections without any problems_»? 

Do you anything to do to enable this behaviour? Because I'm running ansible-lint v4.2.0, and I still get an error:
```
(389) xenlo@indigo-faraday:~/Projects/389$ ansible-lint roles/vmware-esxi/
Couldn't parse task at /home/xenlo/Projects/389/roles/vmware-esxi/tasks/san_preparations.yml:72 (couldn't resolve module/action 'vmware_host_iscsi'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in '<unicode string>': line 72, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

(could not open file to display line))
{ 'delegate_to': 'localhost',
  'name': 'Enable iSCSI Software adapter on cnodes',
  'register': 'cnode_iscsi_adapt',
  'skipped_rules': [],
  'vmware_host_iscsi': { '__file__': '/home/xenlo/Projects/389/roles/vmware-esxi/tasks/san_preparations.yml',
                         '__line__': 73,
                         'esxi_hostname': '{{ ansible_host }}',
                         'hostname': '{{ ansible_host }}',
                         'password': '{{ ansible_ssh_pass }}',
                         'state': 'enabled',
                         'username': '{{ ansible_user }}',
                         'validate_certs': False}}
```

Maybe, I'm doing something wrong… But my role is working correctly, able to found the `vmware_host_iscsi` module from vmware collection.
```
(389) xenlo@indigo-faraday:~/Projects/389$ grep collections etc/ansible.cfg
collections_paths      = ~/.ansible/collections:../collections

(389) xenlo@indigo-faraday:~/Projects/389$ ls -1 ~/.ansible/collections/ansible_collections/community/vmware/plugins/modules/vmware_host_iscsi.py 
/home/xenlo/.ansible/collections/ansible_collections/community/vmware/plugins/modules/vmware_host_iscsi.py
```

webknjaz:
@xenlo he meant master. We don't do backports so when saying "current", it's implied that no further changes need to be submitted as a PR, just wait for the release to happen. Or try out pre-releases (there's 4.3.0a4 on PyPI, for example).

barbosm:
I just tested ansible-lint 4.3.0a5.dev29+g8e6ba5d (installed from master) and I'm still not able to validate this feature (same error message).

ssbarnea:
I tested with ansible-devel (2.11) and it works exactly as expected.

```
- hosts: localhost
  # collections:
  #   - community.vmware
  tasks:
    - name: foo
      vmware_host_iscsi: {}
```

With the commented lines you get an error because ansible does not know where to find the collection, but once you uncomment the lines it will work as it will be able to compute the fully qualitied collection name. Alternative is to refer to the collection with its full name from start.

Issue #635, opened 2019-11-24
Feature Request: Replace deprecated libraries


mjtrangoni:
Hi,

I was trying `pylint-3` on the master branch, and pylint found one deprecated library and one deprecated method. See,

```bash
$ pylint-3 --disable=missing-docstring,invalid-name $(find . -type f -name "*.py") | grep depre
lib/ansiblelint/utils.py:560:31: W1505: Using deprecated method getargspec() (deprecated-method)
lib/ansiblelint/__main__.py:26:0: W0402: Uses of a deprecated module 'optparse' (deprecated-module)

```

webknjaz:
Feel free to send a PR.

evitalis:
I can take a shot at this if its still available. I didn't see a PR at this time, but doesn't mean someone isn't already working on it.

evitalis:
Sorry for the spam guys, I had to rebase to squash a few commits, guess it didn't reflect here.

#645 should fix the warning for getopt. I deprecated function cannot be fixed while py27 is supported as there is no replacement function until py3.0. Good news is the updated function is a drop in replacement and passes the tox testing for py3.x.

webknjaz:
Removal of `inspect.getargspec()` is tracked @ #603.

PR #645, opened 2019-12-10
Switch from getopt to argparse

evitalis:
I was able to run tox tests for linting and py{37,27}-ansible29 without issue.
This addresses the bulk of #635, the final warning cannot be resolved while Python 2.7 is supported.


lib/ansiblelint/__main__.py
```diff
@@ -177,7 +180,11 @@ def main():
         skip.update(str(s).split(','))
     options.skip_list = frozenset(skip)
 
-    playbooks = sorted(set(args))
+    if len(options.playbook) > 0:
```
webknjaz: Maybe just
```suggestion
playbooks = sorted(set(options.playbook or args))
```

lib/ansiblelint/__main__.py
```diff
@@ -149,7 +152,7 @@ def main():
         formatter = formatters.ParseableSeverityFormatter()
 
     # no args triggers auto-detection mode
-    if len(args) == 0 and not (options.listrules or options.listtags):
+    if len(options.playbook) == 0 and not (options.listrules or options.listtags):
```
webknjaz:
```suggestion
    if not options.playbook and not (options.listrules or options.listtags):
```

evitalis:
@webknjaz Both changes made and committed. I have no idea why this now has conflicts. I added your suggestions and added a commit with those minor changes. I also tried to squash the commits which didn't resolve the conflict.

Checked my files with `git diff` before I committed too and didn't see anything that was wrong. Thoughts on what happened?

evitalis:
Fixed the conflict, 8b7a4c4b87d5d3856d017e44ea5cede5e20dc1cb ended up conflicting due to some edits in that commit. Should be fine now.

evitalis:
ping @webknjaz

lib/ansiblelint/__main__.py
```diff
@@ -52,62 +52,64 @@ def main():
...
+                        help="Try force colored output (relying on ansible's code)")
+    parser.add_argument('--exclude', dest='exclude_paths', action='append',
+                        help='path to directories or files to skip. This option'
+                             ' is repeatable.',
+                        default=[])
+    parser.add_argument('-c', dest='config_file',
+                        help='Specify configuration file to use.  '
+                             'Defaults to ".ansible-lint"')
+    parser.add_argument('--version', action='version', version='%(prog)s ' + __version__)
+    parser.add_argument('playbook', nargs='*')
```
webknjaz: Why is this arg needed?

lib/ansiblelint/__main__.py
```diff
@@ -52,62 +52,64 @@ def main():
...
+    options = parser.parse_args()
```
webknjaz: This guide https://docs.python.org/3.9/library/argparse.html#upgrading-optparse-code says that it should be
```suggestion
    args = parser.parse_args()
```

So why did you keep options? OTOH it probably contains both args and options in a single var instead of two vars.

webknjaz: Also, why don't you keep `sys.argv[1:]`. Or even better — add an optional arg to `main()` to make custom in-script invocations possible.

evitalis:
@webknjaz I probably left both as they were there previously. That being said its been awhile so I no longer remember. Feel free to make what edits you feel are needed as I am not interested in contributing to this project further. As noted in #635 though the current code is deprecated and should be updated.

ssbarnea:
@evitalis Please rebase it, I cannot do it for you and this is likely the solution for closing #730

webknjaz:
@ssbarnea it'd be non-trivial to rebase this given that arg parsing stuff got moved to `cli` module by @cans recently.

ssbarnea:
@webknjaz feel free to ping me on irc if you want me to do a manual check of the change once done.

webknjaz:
I think this change is okay, I just need to adjust a small tests patch and will merge.

webknjaz:
This is tricky, postponing for now

PR #646, opened 2019-12-12
Added note regarding pip requirements

### LINKED ISSUE #643: Can't install from source: setup.py missing
# Issue Type
- Bug report

```
ansible: n.a. (doesn't matter for this bug)
current master tree (since release 4.2.0)
```

- ansible installation method: from source

# Desired Behaviour

The README states that `pip install git+https://github.com/ansible/ansible-lint.git` installs from source - this should work either with warning that a requirement isn't fulfilled or with successful installation

> 
>   Cloning https://github.com/ansible/ansible-lint.git to /tmp/pip-aS0W0q-build
>     Complete output from command python setup.py egg_info:
>     Traceback (most recent call last):
>       File "<string>", line 1, in <module>
>     IOError: [Errno 2] No such file or directory: '/tmp/pip-aS0W0q-build/setup.py'
>     
>     ----------------------------------------
> Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-aS0W0q-build/

Same method was working working all the time (at least till newer setuptools was blowing things up, and somehow a working setup.py was completely removed from the source tree).

Currently I'm unable to install it - neither from pypi nor from git

priv-kweihmann:
Since the project switched to PEP517 installation
only, it is required to have pip 19.0+ installed.
Earlier versions of pip don't support PEP517 style
installation.

Added note to README to inform about this

Relates to #643

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,6 +44,8 @@ Using Pip
 From Source
 -----------
 
+**Note**: pip 19.0+ is required for installation.
+
 .. code-block:: bash
 
     pip install git+https://github.com/ansible/ansible-lint.git
--- a/README.rst
+++ b/README.rst
@@ -44,11 +44,14 @@ Using Pip
 From Source
 -----------
 
-**Note**: pip 19.0+ is required for installation.
+**Note**: pip 19.0+ is required for installation. Please consult with the `PyPA User Guide`_
+to learn more about managing Pip versions.
 
 .. code-block:: bash
 
     pip install git+https://github.com/ansible/ansible-lint.git
+
+.. _PyPA User Guide: https://packaging.python.org/tutorials/installing-packages/#ensure-pip-setuptools-and-wheel-are-up-to-date
 
 .. installing-docs-inclusion-marker-end-do-not-remove
 


README.rst
```diff
@@ -44,6 +44,15 @@ Using Pip
 From Source
 -----------
 
+**Note**: pip 19.0+ is required for installation.
+To check run
+
+.. code-block:: bash
+
+    pip --version
```
webknjaz: I don't think that we should document how to use Pip CLI. Just point to some proper guide like https://packaging.python.org/tutorials/installing-packages/#ensure-pip-setuptools-and-wheel-are-up-to-date.

priv-kweihmann: Then the shown usage of pip for installing could be also removed, right? (just joking)
Don't get me wrong here, but installation requires a specific version (or higher) of pip to be installed, so it's good to show the user (which might not have that kind of experience with pip) how that can be ensured/checked.
Otherwise there could just be some unspecific note like `always use the latest pip/setuptools for installing`, which doesn't ensure that this really meets the required `pip 19.0+` version.

For me it comes down to the point/question, what kind of experience does the project require from a user - and then setup documentation accordingly.

webknjaz: I think it's unfair to push more maintenance burden into a project with next to 0 maintainers.

priv-kweihmann: Removed that note/block - Now everybody should be happy, right?
What's left is the note about the required pip version...

webknjaz:
@priv-kweihmann thanks!

PR #647, opened 2019-12-15
Update readme to include new usage options.

deekayen:
parseable-severity and extra verbosity are missing on the readme as usage options

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,7 +90,7 @@ The following is the output from ``ansible-lint --help``, providing an overview 
                             path to directories or files to skip. This option is
                             repeatable.
       -c /path/to/file      Specify configuration file to use.  Defaults to
-                            ".ansible-lint"
+                              ".ansible-lint"
 
 
 
--- a/README.rst
+++ b/README.rst
@@ -62,7 +62,7 @@ The following is the output from ``ansible-lint --help``, providing an overview 
 
 .. code-block:: bash
 
-    Usage: ansible-lint [options] [playbook.yml [playbook2 ...]]
+    Usage: ansible-lint [options] [playbook.yml [playbook2 ...]]|roledirectory
 
     Options:
       --version             show program's version number and exit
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -62,14 +62,15 @@ The following is the output from ``ansible-lint --help``, providing an overview
 
 .. code-block:: bash
 
-    Usage: ansible-lint playbook.yml|roledirectory ...
+    Usage: ansible-lint [options] [playbook.yml [playbook2 ...]]
```
awcrosby: we shouldn't remove `roledirectory` here

README.rst
```diff
@@ -80,15 +81,16 @@ The following is the output from ``ansible-lint --help``, providing an overview
...
+      --nocolor             disable colored output
+      --force-color         Try force colored output (relying on ansible's code)
       --exclude=EXCLUDE_PATHS
                             path to directories or files to skip. This option is
                             repeatable.
-      --force-color         Try force colored output (relying on ansible's code)
-      --nocolor             disable colored output
-      -c /path/to/file      Specify configuration file to use.  Defaults to
-                              ".ansible-lint"
+      -c C                  Specify configuration file to use.  Defaults to
```
awcrosby: maybe `CONFIGFILE` or existing `/path/to/file` is a bit more user-friendly

README.rst
deekayen: I didn't really remove it so much as use what was in the output of the actual utility. I put it back in, but I'd think it would be in the actual --help output too, not just appended in the docs.

README.rst
deekayen: Yeah, I don't know what C is. That was the literal output on my macbook.

README.rst
awcrosby: Ah, I didn't realize our `--help` was missing that, I will add it there. Thanks!

README.rst
awcrosby: I will also make this more readable in `--help`

PR #650, opened 2019-12-27
Fix: fallback to find when not a git repo

anryko:
__PROBLEM__
- `ansible-lint` is failing if run not in git repo.

__SOLUTION__
- Fallback to `find` in case if `git ls-files` fails.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -763,15 +763,48 @@ def is_playbook(filename):
     return False
 
 
+def get_yaml_files(options=None):
+    """Find all yaml files."""
+    if options is None:
+        options = {}
+
+    # git is preferred as it also considers .gitignore
+    git_command = ['git', 'ls-files', '*.yaml', '*.yml']
+    if options.verbosity:
+        print(
+            "Discovering files to lint: %s" % (' '.join(git_command))
+        )
+
+    try:
+        out = subprocess.check_output(
+            git_command,
+            stderr=subprocess.STDOUT,
+            universal_newlines=True
+        ).split()
+    except subprocess.CalledProcessError as exc:
+        if options.verbosity:
+            print(
+                "Warning: Failed to discover yaml files to lint using git: %s" % (
+                    exc.output.rstrip('\n')
+                )
+            )
+
+        out = [
+            os.path.join(root, name)
+            for root, dirs, files in os.walk('.')
+            for name in files
+            if name.endswith('.yaml') or name.endswith('.yml')
+        ]
+
+    return OrderedDict.fromkeys(sorted(out))
+
+
 def get_playbooks_and_roles(options=None):
     """Find roles and playbooks."""
     if options is None:
         options = {}
 
-    # git is preferred as it also considers .gitignore
-    files = OrderedDict.fromkeys(sorted(subprocess.check_output(
-        ["git", "ls-files", "*.yaml", "*.yml"],
-        universal_newlines=True).split()))
+    files = get_yaml_files(options)
 
     playbooks = []
     role_dirs = []
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -763,16 +763,15 @@ def is_playbook(filename):
     return False
 
 
-def get_yaml_files(options=None):
+def get_yaml_files(options):
     """Find all yaml files."""
-    if options is None:
-        options = {}
-
     # git is preferred as it also considers .gitignore
     git_command = ['git', 'ls-files', '*.yaml', '*.yml']
     if options.verbosity:
         print(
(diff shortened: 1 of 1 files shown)


lib/ansiblelint/utils.py
```diff
@@ -763,15 +763,55 @@ def is_playbook(filename):
...
+    )
+
+    out, err = proc.communicate()
+
+    if err and options.verbosity:
+        print(
+            "Warning: Failed to discover yaml files to lint: %s" % (err)
+        )
+
+    if proc.returncode:
```
webknjaz: Why don't you just use `subprocess.check_output()`? It seems cleaner.

lib/ansiblelint/utils.py
```diff
@@ -763,15 +763,55 @@ def is_playbook(filename):
...
         options = {}
 
-    # git is preferred as it also considers .gitignore
-    files = OrderedDict.fromkeys(sorted(subprocess.check_output(
-        ["git", "ls-files", "*.yaml", "*.yml"],
-        universal_newlines=True).split()))
+    try:
+        # git is preferred as it also considers .gitignore
+        files = get_yaml_files(options=options)
+    except subprocess.CalledProcessError:
```
webknjaz: Why don't you encapsulate the fallback logic in `get_yaml_files`?

lib/ansiblelint/utils.py
```diff
@@ -763,15 +763,55 @@ def is_playbook(filename):
...
+
+    out, err = proc.communicate()
+
+    if err and options.verbosity:
+        print(
+            "Warning: Failed to discover yaml files to lint: %s" % (err)
+        )
+
+    if proc.returncode:
+        raise subprocess.CalledProcessError(
```
webknjaz: LGTM linter is unhappy about this invocation.

anryko:
@webknjaz thanks! I have Addressed your comments.

ssbarnea:
If really needed we could improve the fallback mechanism to support non-git use-cases but we will need a detection mechanism and I am not sure who will have the time to write and maintain this code. Maybe it would make sense to propose a way to override the discovery command in the config file and we can avoid increasing the code size for uncommon use-cases.

anryko:
@ssbarnea done

lib/ansiblelint/utils.py
```diff
@@ -763,15 +763,48 @@ def is_playbook(filename):
...
+    try:
+        out = subprocess.check_output(
+            git_command,
+            stderr=subprocess.STDOUT,
+            universal_newlines=True
+        ).split()
+    except subprocess.CalledProcessError as exc:
+        if options.verbosity:
+            print(
+                "Warning: Failed to discover yaml files to lint using git: %s" % (
```
webknjaz: Let's use `str.format`
```suggestion
                'Warning: Failed to discover yaml files to lint using git: {git_out!s}'.
                format(git_out=exc.output.rstrip('\n'))
```

lib/ansiblelint/utils.py
```diff
@@ -763,15 +763,48 @@ def is_playbook(filename):
...
+    try:
+        out = subprocess.check_output(
+            git_command,
+            stderr=subprocess.STDOUT,
+            universal_newlines=True
+        ).split()
+    except subprocess.CalledProcessError as exc:
+        if options.verbosity:
+            print(
+                'Warning: Failed to discover yaml files to lint using git: {!s}'.format(
```
webknjaz: Please use an indexed str placeholder.

lib/ansiblelint/utils.py
```diff
@@ -763,15 +763,48 @@ def is_playbook(filename):
...
+def get_yaml_files(options=None):
+    """Find all yaml files."""
+    if options is None:
+        options = {}
+
+    # git is preferred as it also considers .gitignore
+    git_command = ['git', 'ls-files', '*.yaml', '*.yml']
+    if options.verbosity:
+        print(
+            "Discovering files to lint: %s" % (' '.join(git_command))
```
webknjaz: plz use str format with an indexed placeholder here too.

lib/ansiblelint/utils.py
```diff
@@ -763,15 +763,48 @@ def is_playbook(filename):
     return False
 
 
+def get_yaml_files(options=None):
```
webknjaz: I think this shouldn't be optional as it's only used in one way in only one place making this assumption premature.

anryko: Could you please elaborate?
Do you suggest to change this line to `def get_yaml_files(options):`?

webknjaz: Exactly. Remove `=None` and the following `is None` if-clause block.

anryko: done

webknjaz:
Great, thanks!

Issue #653, opened 2020-01-15
project is not installable in editable mode


ssbarnea:
We need to be sure that `pip install -e .` works on this project as that is the first action any new contributor will try for testing a PR or his own change.

```
ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /Users/ssbarnea/c/os/ansible-lint
(A "pyproject.toml" file was found, but editable mode currently requires a setup.py based build.)
```
That is huge problem because installing a package as editable and testing a change made to it is the first thing any contributor would want to do. 

Out of >50 popular python projects I have cloned locally ansible-lint is the only one where installing as editable is broken. I think that the decision to drop the `setup.py` may have being not the best one.

AFAIK there is no workaround to make editable mode work.

Mainly this bug is a big barrier for project development.

* https://github.com/pypa/pip/issues/6605 locked by pypa

ssbarnea:
^ that was quick and without explanations.

webknjaz:
It has all been weighted-in at the time of the refactoring. It's not needed.

webknjaz:
FTR it should be easy to do `pip install .` (in normal mode) and then set something like `export PYTHONPATH="lib/:$PYTHONPATH".

webknjaz:
resolved

PR #654, opened 2020-01-15
Make ansible-lint installable in editable mode

ssbarnea:
The change is also adding a regression protection in CI jobs by testing that editable installation is possible after it runs the normal tox environments.

Considering the CI determines what is supported or not I see no reason for not fixing this issue. 
When most (in not all) python development tools like tox, pytest, flake8, pre-commit, black are installable in editable mode, I am confident to challenge the reasons for preventing ansible-lint from doing the same.

Fixes: #653
Fixes: #716

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -158,3 +158,4 @@ install:
 
 script:
 - tox -v
+- pip install -e .  # Validates that project can install in editable mode
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,10 @@
+#! /usr/bin/env python
+"""ansible-lint distribution package setuptools installer."""
+
+import setuptools
+
+
+if __name__ == "__main__":
+    setuptools.setup(
+        use_scm_version=True, setup_requires=["setuptools_scm"],
+    )


webknjaz:
@sshnaidm the cost is more than a zero-maintainer team can afford. I'd like to be unbound from the old tech as much as possible.

I'm pretty sure that this is solvable w/o adding `setup.py`.

I haven't seen any compelling evidence that it's needed. "I want this command to work" is not a use-case.

@ssbarnea @sshnaidm please record your workflows using https://asciinema.org/ explaining what you're missing and I'll try to take this into account (I have a hunch that your cases are easy to work around). (Pro tip: asciinema records your terminal so make sure that no secrets appear in the recording + also it doesn't have sound but you can add explanations by typing in comments as in `# I'm demonstrating X and you're observing Y`)

setup.py
```diff
@@ -0,0 +1,22 @@
...
+        if IS_PYPI_UPLOAD
+        else setuptools_scm.version.get_local_node_and_date(version)
+    )
+
+
+setup_params = {}
+setup_params['use_scm_version'] = {'local_scheme': cut_local_version_on_upload}
+
+
+__name__ == '__main__' and setuptools.setup(**setup_params)
```
webknjaz: Even if you're recovering `setup.py`, you should call `setup()` with no args because version generation is already hooked up in `setup.cfg`. Duplicating things in multiple places is bad.

sshnaidm:
@webknjaz well, as I wrote before I found `pip -e` very useful for development, especially for ansible related stuff. 

>  "I want this command to work" is not a use-case.

But I really want it to work :smile: My use case is pretty simple, I develop python modules and most of them are from TripleO in Openstack or Ansible modules or plugins, running `pip -e` and linking source directory to my `sources` directory which has IDE coverage and allow me to develop and test immediately and not to deal with code, build, install, try, clean, ouch.. not cleaned enough, etc etc.
Maybe I do it wrong way, but I used to it :smile: If you have the better workflow when I can use module and develop its sources with IDE and try without install, I'd like to know of course, always happy to learn.
I mean there are a lot of developers and many workflows, `pip -e` has its own case and I personally and a lot of my colleagues use it, I can't say we want something bizarre and unusual. In our cases it allows to develop easier and it means more involvment in projects and more tested and qualified code. If this mode will allow more people to install ansible-lint, make change and test easily - would it be a good trade of for code maintenance?
Of course we can discuss details and how to do it separately if there is a general agreement about this feature.

webknjaz:
You still haven't answered what commands you run. Clicking "some buttons" in IDE is not going to be a supported workflow. We need something actionable.

webknjaz:
I was thinking about it and I think I'd settle for a compromise.

So modern setuptools injects a `setup.py` file (in-memory) containing exactly `import setuptools; setuptools.setup()`. That's it.

This got me thinking:
1. The main idea is not to have `setup.py` in the repo and it should stay that way.
2. To ensure this, we need to add it to `.gitignore`.
3. Once it's there, it's safe to create `setup.py` with whatever contents you want on the dev machine.
4. So we could do that and document something like `echo 'import setuptools; setuptools.setup()' > setup.py` in dev docs/readme.
5. If that's too complex, we could add a tox env that would do this.

ssbarnea:
That is not what I would call a compromise. A compromise would be to require a minimal version of pip or setuptools for the edit to work, which could be seen acceptable for development mode. Still, requiring that kind of hack does clearly not count as making easy to contribute.

ssbarnea:
Should I also mention that due to this marveleus change I takes >25s to run a single test with tox?
```
tox -e py38-ansible29 -- -k test_run_single_role_path_no_trailing_slash
```
This is because every time you run tox, it will repackage and install the module.

strider:
This change would be very helpful for projects development IMO.

Please find a quick and fair compromise and let's putting things forward!

cloudnull:
While I personally like to use `editable` all the time and find it very useful, I can completely understand @webknjaz point about reducing any and all maintenance burdens. Even though this change is tiny, and *shouldn't* incur any maintenance burdens, there is a risk that something could change in upstream setuptools which will force the noted "zero-maintainer team" to fix whatever is broken. I'm :+1:  on the content of this change, however, I'm hesitant to approve it until something can be worked out that satisfies both sides.

ssbarnea:
Please review https://github.com/ansible/ansible-lint/pull/697 which has the side effect of also addressing most of this ticket.

.travis.yml
```diff
@@ -158,3 +158,4 @@ install:
 
 script:
 - tox -v
+- pip install -e .  # Validates that project can install in editable mode
```
webknjaz, line 161: If you want this check, it should go to pre-commit as one of the linters there.

ssbarnea: I already considered this but I am afraid that pre-commit already knows to use PEP-518 and to avoid use of setup.py. Still, it could be possible to write a small in-repo custom hook that performs the packaging test. In fact it may prove to be good idea to have a generic hook for testing packaging. I will try to write and propose on to pre-commit-hooks.

Issue #652, opened 2020-01-15
4.2.0: TypeError: join() argument must be str or bytes, not 'NoneType'


ssbarnea:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version

ansible-lint --version
4.2.0
```

- ansible installation method: one of source, pip, OS package
- ansible-lint installation method: one of source, pip, OS package

# Desired Behaviour

Not to crash, passing or with meaningful error.

# Actual Behaviour (Bug report only)

```
  File "/Users/ssbarnea/c/opendev/zuul-jobs/.tox/linters/lib/python3.7/site-packages/ansiblelint/utils.py", line 299, in _rolepath
    path_dwim(basedir, os.path.join('roles', role)),
  File "/Users/ssbarnea/c/opendev/zuul-jobs/.tox/linters/lib/python3.7/posixpath.py", line 94, in join
    genericpath._check_arg_types('join', a, *p)
  File "/Users/ssbarnea/c/opendev/zuul-jobs/.tox/linters/lib/python3.7/genericpath.py", line 149, in _check_arg_types
    (funcname, s.__class__.__name__)) from None
TypeError: join() argument must be str or bytes, not 'NoneType'
```

## Reproducing

It seems to be caused by use of `incude_role`:
```
- hosts: all
  tasks:
    - name: foo
      include_role:
        role: run-buildset-registry
```

# Workaround

Rename `role` key to `name` and the linter will work fine, name is what official documentation mentions, role key is undocumented, probably there as an alias for compatibility.

geerlingguy:
I hit the same thing in my testing; just used:

```
- include_role: [role name here]
```

And couldn't figure out what was triggering the error until I googled this issue.

Workaround of using the following as @ssbarnea mentioned seems to pass linting:

```
    - name: Include a role
      include_role:
        name: [role name here]
```

freedemster:
I ran into the same error with a simple playbook with one `include_role` instruction.

PR #655, opened 2020-01-15
Avoid error when include_role uses role instead of name

ssbarnea:
While Ansible documents only `name` as accepted for `include_role` and `import_role` it does silently accept `role` alias.

We match Ansible behaviour instead of crashing during parsing.

Fixes: #652

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -237,13 +237,17 @@ def _taskshandlers_children(basedir, k, v, parent_type):
             append_children(th['import_playbook'], basedir, k, parent_type, results)
         elif 'import_tasks' in th:
             append_children(th['import_tasks'], basedir, k, parent_type, results)
-        elif 'import_role' in th:
+        elif 'include_role' in th or 'import_role' in th:
             th = normalize_task_v2(th)
-            results.extend(_roles_children(basedir, k, [th['action'].get('name')], parent_type,
-                                           main=th['action'].get('tasks_from', 'main')))
-        elif 'include_role' in th:
-            th = normalize_task_v2(th)
-            results.extend(_roles_children(basedir, k, [th['action'].get('name')],
+            module = th['action']['__ansible_module__']
+            if "name" not in th['action']:
+                raise RuntimeError(
+                    "Failed to find required 'name' key in %s" % module)
+            if not isinstance(th['action']["name"], six.string_types):
+                raise RuntimeError(
+                    "Value assigned to 'name' key on '%s' is not a string." %
+                    module)
+            results.extend(_roles_children(basedir, k, [th['action'].get("name")],
                                            parent_type,
                                            main=th['action'].get('tasks_from', 'main')))
         elif 'block' in th:
--- a/test/TestImportIncludeRole.py
+++ b/test/TestImportIncludeRole.py
@@ -1,4 +1,5 @@
 import os
+import pytest
 import unittest
 import tempfile
 import shutil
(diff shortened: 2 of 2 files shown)


lib/ansiblelint/utils.py
```diff
@@ -226,6 +226,13 @@ def _include_children(basedir, k, v, parent_type):
     return [{'path': result, 'type': parent_type}]
 
 
+def first(iterable, func=lambda L: L is not None, **kwargs):
```
webknjaz: If only `default` is used, there's no need to special-case it.

lib/ansiblelint/utils.py
```diff
@@ -226,6 +226,13 @@ def _include_children(basedir, k, v, parent_type):
...
+    it = (el for el in iterable if func(el))
+    if 'default' in kwargs:
+        return next(it, kwargs['default'])
+    return next(it)
```
webknjaz: just do
```suggestion
    args = () if 'default' in kwargs else (kwargs['default'], )
    return next(it, *args)
```

lib/ansiblelint/utils.py
```diff
@@ -237,13 +237,15 @@ def _taskshandlers_children(basedir, k, v, parent_type):
...
-        elif 'include_role' in th:
-            th = normalize_task_v2(th)
-            results.extend(_roles_children(basedir, k, [th['action'].get('name')],
+            # gets first property that is defined
+            role_name = next(
+                filter(bool, (th['action'].get(k, None) for k in ['name', 'role'])), None)
+            if not role_name:
+                raise RuntimeError(
+                    "Failed to find required 'name' key in %s" % th['action']['__ansible_module__'])
+            results.extend(_roles_children(basedir, k, [role_name],
```
webknjaz: it's hard to find where that runtime error ends. could you plz put align the closing bracket with `raise`?

lib/ansiblelint/utils.py
```diff
@@ -237,13 +237,15 @@ def _taskshandlers_children(basedir, k, v, parent_type):
...
+        elif 'include_role' in th or 'import_role' in th:
             th = normalize_task_v2(th)
-            results.extend(_roles_children(basedir, k, [th['action'].get('name')], parent_type,
-                                           main=th['action'].get('tasks_from', 'main')))
-        elif 'include_role' in th:
-            th = normalize_task_v2(th)
-            results.extend(_roles_children(basedir, k, [th['action'].get('name')],
+            # gets first property that is defined
+            role_name = next(
+                filter(bool, (th['action'].get(k, None) for k in ['name', 'role'])), None)
```
webknjaz: Usually combining functional style with generator expressions is discouraged.

ssbarnea: TBH, I am was very close to make the use of 'role' key raise the same exception as a missing one because is not documented. This would made the filter not needed.

lib/ansiblelint/utils.py
```diff
@@ -237,13 +237,13 @@ def _taskshandlers_children(basedir, k, v, parent_type):
...
             append_children(th['import_tasks'], basedir, k, parent_type, results)
-        elif 'import_role' in th:
+        elif 'include_role' in th or 'import_role' in th:
             th = normalize_task_v2(th)
-            results.extend(_roles_children(basedir, k, [th['action'].get('name')], parent_type,
-                                           main=th['action'].get('tasks_from', 'main')))
-        elif 'include_role' in th:
-            th = normalize_task_v2(th)
-            results.extend(_roles_children(basedir, k, [th['action'].get('name')],
+            role_name = th['action'].get("name", None)
```
webknjaz: Honestly, I'd prefer try/except KeyError, because checking for magic values is error-prone. Also, you'll get "Failed to find required key" even if it exists but is of a falsy type (`[]`, `{}`, `0`, `null`) which makes it confusing/misleading for users who try to debug it based on an error message produced.

PR #657, opened 2020-01-17
Enable Zuul CI Jobs

ssbarnea:
@gundalow Please activate https://github.com/apps/ansible-zuul for this repo so we can test it, activating it does not mean it will act as required.

## 2020-02-27 Update
As all molecule* projects where already succesfully migrated to Zuul CI, during the last months, we can say that now it should be the right time to enable it on ansible-lint too, but only as a 3rd party CI, without replacing the dual-CI config already configured by @webknjaz 

Recent bugs regarding failure to install as a pre-commit hook under centos-7, created more pressure for setting CI that runs on our platforms. Once we enable Zuul, we will be able to add extra coverage for bugs like that.

With @pabelanger help I can take care of configuring zuul project and the jobs running on it.

### DIFF
--- a/tools/test-setup.sh
+++ b/tools/test-setup.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euxo pipefail
+set -Eeuxo pipefail
 # Used by Zuul CI to perform extra bootstrapping
 
 # Bumping system tox because version from CentOS 7 is too old
--- a/tools/test-setup.sh
+++ b/tools/test-setup.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -euxo pipefail
+# Used by Zuul CI to perform extra bootstrapping
+
+# Bumping system tox because version from CentOS 7 is too old
+# We are not using pip --user due to few bugs in tox role which does not allow
+# us to override how is called. Once these are addressed we will switch back
+# non-sudo mode.
+PYTHON=$(command -v python3 python | head -n1)
+
+sudo "$PYTHON" -m pip install -U tox "zipp<0.6.0;python_version=='2.7'"
--- a/zuul.d/layout.yaml
+++ b/zuul.d/layout.yaml
@@ -1,0 +1,40 @@
+---
+# zuul.d/layout.yaml
+- job:
+    name: tox-py27-ansible28
+    parent: ansible-tox-py27
+    vars:
+      tox_envlist: py27-ansible28
+
+- job:
+    name: tox-py37-ansible28
+    parent: ansible-tox-py37
+    vars:
+      tox_envlist: py37-ansible28
+
+- project:
+    check:
+      jobs: &defaults
+        - tox-linters:
+            vars:
+              tox_envlist: lint
+        - tox-docs
+        - tox-py27:
+            vars:
+              tox_envlist: py27-ansible29
+        - tox-py27-ansible28
+        - tox-py35:
+            vars:
+              tox_envlist: py35-ansible29
+        - tox-py36:
+            vars:
+              tox_envlist: py36-ansible29
+        - tox-py37:
+            vars:
+              tox_envlist: py37-ansible29
+        - tox-py37-ansible28
+        - tox-py38:
+            vars:
+              tox_envlist: py38-ansible29
+    gate:
+      jobs: *defaults


webknjaz:
@ssbarnea have Zuul learned to report things via Checks API so far?

ssbarnea:
> @ssbarnea have Zuul learned to report things via Checks API so far?

Not the full check-api but current integration is more than usable. Take a look at https://github.com/ansible-community/pytest-molecule/pull/60 -- and remember that commit-messages you see can be disabled (they do annoy me too). The most important aspect is that now when you click the check it sends you to a page where you can easily see the results (that is a recent improvement).

For example since December we also starting to use zuul (rdo instance) to test podman packaging and it runs quite well for them, even if is not their primary CI. Example https://github.com/containers/libpod/pull/4898 (comments are disabled on that one).

I got good progress with Molecule but did not finish it because its CI has some special requirements (docker-ce) which are bit more tricky to get right. Still, for ansible-lint it should go quite smooth, similar to https://github.com/ansible-community/pytest-molecule/pull/60

gundalow:
I've requested the GitHub app be enabled

webknjaz:
> commit-messages you see can be disabled

I guess you mean comments on GitHub, not Git commit messages. Right?


> Not the full check-api

Ah, it's not Checks API, it's old Statuses API as far as I can see.


> The most important aspect is that now when you click the check it sends you to a page where you can easily see the results (that is a recent improvement).

Honestly, I still get confused on where to click next when I get to that page. Typical CIs would let me see the logs but I'm totally getting lost in that UI for some reason...

Spredzy:
Currently, many ansible upstream projects rely on Zuul for testing, namely:

  * `ansible/awx`
  * `ansible/ansible-runner`
  * `ansible/workshops` (as 3rd party)
  * `project-receptor/receptor`
  * `ansible-collections/vyos` (and all the network collections + security collections)

It would be great if we could get more project onboarding so we make sure we keep it DRY as much as possible and offer same testing experience across the board

webknjaz:
*/me pinged @mattclay about this*

tools/test-setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -euxo pipefail
```
webknjaz: I recently learned that adding `-E` will make it even safer:
```suggestion
set -Eeuxo pipefail
```

webknjaz:
recheck

webknjaz:
@ssbarnea @Spredzy the GitHub App has been installed.

webknjaz:
@ssbarnea I merged this but Zuul doesn't seem to pick it up. Feel free to submit follow-up config PRs.

pabelanger:
@webknjaz we need to do a few things for zuul.a.c to pick it up, so we might want to revert until that is done. If we are ready to test, I can propose the change to zuul.a.c to enable it for ansible-lint.

webknjaz:
I'm okay with this living in the repo, propose updates as you're ready.

pabelanger:
https://github.com/ansible/project-config/pull/296 is the PR, which I'll approve in a moment. Currently, I need to be an admin on the repo, so we can setup some things like merge state and branch protections. If I am following along, zuul will gate this repo right?

webknjaz:
No gating needed atm, just an extra CI. Will decide later on changing workflows.

pabelanger:
okay, this change needs to be reverted then. The jobs are setup to gate.  we need to remove the gate line in zuul.yaml

webknjaz:
So not revert, just remove one line?

pabelanger:
Actually, yes a partial revert. We need to remove the intree zuul.yaml file. This is because, we do not load zuul configuration files from project we don't gate. This is to help avoid the potential to wedge zuul.

We'd enable the job configuration in https://github.com/ansible/project-config for the moment. Then, if we move to the next stage, job stanza could move back intree.

webknjaz:
> Then, if we move to the next stage, job stanza could move back intree.

What is the next stage?

pabelanger:
jobs have been setup to run, you can see results on https://github.com/ansible/ansible-lint/pull/666

next steps, are to fix jobs to get them to properly work.

Issue #658, opened 2020-01-17
pipefail fail


utoddl:
# Issue Type
- Bug report

# Ansible and Ansible Lint details
ansible 2.9.2
ansible-lint 4.1.1a0

# Desired Behaviour

ansible-lint should not complain about bash's pipefail option if it's being set. There's only so much debugging of shell scripts a linter can do after all.

# Actual Behaviour (Bug report only)

After acknowledging 306 'Shells that use pipes should set the pipefail option', examining the shell script and determining that the default behavior is in fact the correct behavior in this case, and adding `set +o pipefile`, ansible-lint still throws a 306. This leads to the absurd solution:
```
set -o pipefail
set +o pipefail
```

Surely it would be better to give the programmer the benefit of the doubt if she says `set +o pipefail` that perhaps she knows what she's doing.

I'd recommend ammending the regex from this:
```
_pipefail_re = re.compile(r"^\s*set.*-[A-z]*o\s*pipefail")
```
to this:
```
_pipefail_re = re.compile(r"^\s*set.*[+-][A-z]*o\s*pipefail")
```

ssbarnea:
Not a bad idea, please make it a PR. Also keep in mind that there is the `# noqa 306` comment you can use to bypass the rule, which works fine in v4.2.0.

PR #660, opened 2020-01-19
accept explicit clearing of pipefail

utoddl:
Signed-off-by: Todd Lewis <todd_lewis@unc.edu>
See discussion with issue #658.
The is the RP requested by @ssbarnea from that issue.


ssbarnea:
@webknjaz GTG

PR #661, opened 2020-01-20
New Rule: Check if the role names are respecting the Ansible recommen…

strider:
…dations

Roles can no longer have a "-" in them according to the upstream ansible
documentation[0]. This change adds a new ansible-lint rule to check if
the role name is respecting the Ansible recommendations.

[0] https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#roles-directory

Signed-off-by: Gael Chamoulaud <gchamoul@redhat.com>


ssbarnea:
Very useful but I think it needs documentation update, release notes, lint fixes and a test.

lib/ansiblelint/rules/RoleNames.py
```diff
@@ -0,0 +1,55 @@
...
+        "with an alpha character in role names"
+    )
+    description = (
+        "Role names are now limited to contain only lowercase alphanumeric "
+        "characters, plus '_' and start with an alpha character"
+    )
+    severity = 'HIGH'
+    done = []  # already noticed roles list
+    tags = ['formatting']
+    version_added = '4.0.0'
```
ssbarnea: Put a 4.2.1 here, as 4.2.0 is current release.

lib/ansiblelint/rules/RoleNames.py
```diff
@@ -0,0 +1,55 @@
...
+
+class RoleNames(AnsibleLintRule):
+    id = '207'
+    shortdesc = (
+        "Use only lowercase alphanumeric characters, plus '_' and start "
+        "with an alpha character in role names"
+    )
+    description = (
+        "Role names are now limited to contain only lowercase alphanumeric "
+        "characters, plus '_' and start with an alpha character"
```
ssbarnea: If you can find a permalink to the doc page mentioning the rule, even better.

lib/ansiblelint/rules/RoleNames.py
```diff
@@ -0,0 +1,55 @@
...
+
+import re
+from ansiblelint import AnsibleLintRule
+
+
+class RoleNames(AnsibleLintRule):
+    id = '207'
+    shortdesc = (
+        "Use only lowercase alphanumeric characters, plus '_' and start "
+        "with an alpha character in role names"
```
ssbarnea: If is more than 50 chars is unlikely a shortdesc. Maybe `Role name breaks $regex collection naming`, which will also inform the user about what pattern he needs to use, and indicate that this is related to collections.

strider:
Duplicate of #666

PR #664, opened 2020-01-23
var-space-rule: Fix nested JSON obj false positive

simonkeyd:
When using compact form nested JSON object within a
Jinja2 context as shown in the following example:
  set_fact:"{{ {'test': {'subtest': variable}} }}"

'variable}}' will raise a false positive [206] error.

This commit adds an intermediate step within 206
(VariableHasSpacesRule.py) rule to exclude nested
JSON object before matching for an actual error.

Fixes: #665 

Signed-off-by: Simon Kheng <simon.kheng1337@gmail.com>


ssbarnea:
@simonkeyd Unless a test is added I doubt this change would end-up being merged, which would a be a loss for everyone.

PR #666, opened 2020-01-23
E106: Check if the role names match collection spec

ssbarnea:
Roles can no longer have a "-" in them according to the upstream ansible
documentation[0]. This change adds a new ansible-lint rule to check if
the role name is respecting the Ansible recommendations.

[0] https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#roles-directory

Signed-off-by: Gael Chamoulaud <gchamoul@redhat.com>


strider:
@ssbarnea Thanks for taking over, Sorin!

ssbarnea:
Can someone look at this, please?

pabelanger:
recheck

lib/ansiblelint/rules/RoleNames.py
```diff
@@ -0,0 +1,58 @@
...
+
+        path = file['path'].split("/")
+        if "roles" in path:
+            role_name = path[path.index("roles") + 1]
+            if role_name not in self.done:
+                self.done.append(role_name)
+                if not re.match(self.ROLE_NAME_REGEXP, role_name):
+                    return self.shortdesc.format(role_name)
+            else:
+                return False
```
webknjaz, line 57: Let's simplify it a bit:
```suggestion
            if role_name in self.done:
                return False
            self.done.append(role_name)
            if not re.match(self.ROLE_NAME_REGEXP, role_name):
                return self.shortdesc.format(role_name)
```

test/TestCliRolePaths.py
```diff
@@ -92,3 +92,10 @@ def test_run_playbook(self):
 
         result = self.run_ansible_lint(cwd=cwd, role_path=role_path, env=env)
         self.assertIn('Use shell only when shell functionality is required', str(result))
+
+    def test_run_role_name_invalid(self):
+        cwd = self.local_test_dir
+        role_path = 'roles/invalid-name'
+
+        result = self.run_ansible_lint(cwd=cwd, role_path=role_path)
+        self.assertIn('106 Role name invalid-name does not match', str(result))
```
webknjaz, line 101: As per #725:
```suggestion
        assert '106 Role name invalid-name does not match' in str(result)
```

PR #672, opened 2020-02-01
Expand environment vars and ~

### LINKED ISSUE #541: How do you exclude third party roles in requirements.yml
How do you exclude a third party roles, or any role for that matter. I have tried

I have tried
```exclude_paths:
  - ./roles/nexus
  - requirements.yml
  - compliance-playbook.yml
  - RHEL7-CIS
  - ./roles/RHEL7-CIS
  - ./.ansible/roles/RHEL7-CIS
  - ~/.ansible/roles/RHEL7-CIS
  - hardened-core.yml
  - site.yml
  - meta
  - ./roles/java```

I just want to exclude RHEL7-CIS and nothing seems to prevent it from running. I have even excluded files referencing it. I am running this as a pre-commit run -a with an exclude: .*RHEL.* even. I can not get RHEL7-CIS to stop being checked. And even some of the roles i have excluded like java continue to have files picked up.

### PR DESCRIPTION
**Pull request by evilhamsterman**
Signed-off-by: Dan Mills <evilhamsterman@gmail.com>

Simply adds the ability to expand environment vars and ~ in the exclude and rulesdirs config. This could close #541 by allowing you to easily exclude the traditional galaxy roles directory with the following 

```YAML
exclude_paths:
    - ~/.ansible/roles
```

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -196,7 +196,7 @@ class RulesCollection(object):
     @classmethod
     def create_from_directory(cls, rulesdir):
         result = cls()
-        result.rules = ansiblelint.utils.load_plugins(os.path.expanduser(rulesdir))
+        result.rules = ansiblelint.utils.load_plugins(rulesdir)
         return result
 
 
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -196,6 +196,8 @@ class RulesCollection(object):
     @classmethod
     def create_from_directory(cls, rulesdir):
         result = cls()
+        rulesdir = os.path.expanduser(rulesdir)
+        rulesdir = os.path.expanduser(rulesdir)
         result.rules = ansiblelint.utils.load_plugins(rulesdir)
         return result
 
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -140,10 +140,13 @@ class AnsibleLintRule(object):
 
 class RulesCollection(object):
 
-    def __init__(self, rulesdirs=[]):
-        self._update_rulesdirs(rulesdirs)
+    def __init__(self, rulesdirs=None):
+        if rulesdirs is None:
+            rulesdirs = []
+        self.rulesdirs = ansiblelint.utils.expand_paths_vars(rulesdirs)
         self.rules = []
-        self._update_rules(self.rulesdirs)
+        for rulesdir in self.rulesdirs:
+            self.extend(ansiblelint.utils.load_plugins(rulesdir))
 
     def register(self, obj):
         self.rules.append(obj)
@@ -153,14 +156,6 @@ class RulesCollection(object):
 
     def __len__(self):
         return len(self.rules)
-
-    def _update_rulesdirs(self, rulesdirs):
-        rulesdirs = ansiblelint.utils.expand_paths_vars(rulesdirs)
-        self.rulesdirs = rulesdirs
-
-    def _update_rules(self, rulesdirs):
-        for rulesdir in rulesdirs:
-            self.extend(ansiblelint.utils.load_plugins(rulesdir))
 
     def extend(self, more):
         self.rules.extend(more)
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -198,14 +198,6 @@ class RulesCollection(object):
             results.append("{0} {1}".format(tag, tags[tag]))
         return "\n".join(results)
 
-    @classmethod
-    def create_from_directory(cls, rulesdir):
-        result = cls()
-        rulesdir = os.path.expanduser(rulesdir)
-        rulesdir = os.path.expandvars(rulesdir)
-        result.rules = ansiblelint.utils.load_plugins(rulesdir)
-        return result
-
 
 class Match(object):
 
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -155,8 +155,7 @@ class RulesCollection(object):
         return len(self.rules)
 
     def _update_rulesdirs(self, rulesdirs):
-        rulesdirs = [os.path.expanduser(p) for p in rulesdirs]
-        rulesdirs = [os.path.expandvars(p) for p in rulesdirs]
+        rulesdirs = ansiblelint.utils.expand_paths_vars(rulesdirs)
         self.rulesdirs = rulesdirs
 
     def _update_rules(self, rulesdirs):
@@ -252,10 +251,7 @@ class Runner(object):
     def _update_exclude_paths(self, exclude_paths):
         if exclude_paths:
             # These will be (potentially) relative paths
-            paths = [s.strip() for s in exclude_paths]
-            # Expand ~ and other environment vars such as $HOME
-            paths = [os.path.expanduser(s) for s in paths]
-            paths = [os.path.expandvars(s) for s in paths]
+            paths = ansiblelint.utils.expand_paths_vars(exclude_paths)
             # Since ansiblelint.utils.find_children returns absolute paths,
             # and the list of files we create in `Runner.run` can contain both
             # relative and absolute paths, we need to cover both bases.
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -197,7 +197,7 @@ class RulesCollection(object):
     def create_from_directory(cls, rulesdir):
         result = cls()
         rulesdir = os.path.expanduser(rulesdir)
-        rulesdir = os.path.expanduser(rulesdir)
+        rulesdir = os.path.expandvars(rulesdir)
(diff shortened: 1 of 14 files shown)


lib/ansiblelint/__init__.py
```diff
@@ -196,7 +196,7 @@ def listtags(self):
     @classmethod
     def create_from_directory(cls, rulesdir):
         result = cls()
-        result.rules = ansiblelint.utils.load_plugins(os.path.expanduser(rulesdir))
+        result.rules = ansiblelint.utils.load_plugins(rulesdir)
```
webknjaz: What is the advantage of moving the expansion outside?

evilhamsterman: TBH I don't remember why I did that. Though now that it's been pointed out it would make more sense to have that in the class to ensure the rules are properly expanded.

evilhamsterman:
Looks like the 🚨 / lint/2.7@ubuntu-16.04 (pull_request) test is failing in general. It failed because tox was missing in the test not because of my code, and it looks like it failed on another recent PR.

webknjaz:
Yes, please ignore that failing jobs. I haven't had time to figure out what's causing that failure. Maybe it's because of recent releases of ecosystem-level dists.

lib/ansiblelint/__init__.py
```diff
@@ -196,7 +196,9 @@ def listtags(self):
     @classmethod
     def create_from_directory(cls, rulesdir):
         result = cls()
-        result.rules = ansiblelint.utils.load_plugins(os.path.expanduser(rulesdir))
+        rulesdir = os.path.expanduser(rulesdir)
+        rulesdir = os.path.expanduser(rulesdir)
```
webknjaz: Typo during copy-paste:
```suggestion
        rulesdir = os.path.expandvars(rulesdir)
```

evilhamsterman:
I did a bunch of refactoring. I broke out the expansion code to the utils module so it can be reused. I also change the RulesCollection class to work more like the Runner class and setup the rulesdir on class creation rather than having to use a class method after the creation. I couldn't see why it was done that way but this way seemed cleaner. I can clean up the tests to use the new methodology.

lib/ansiblelint/__init__.py
```diff
@@ -140,8 +140,10 @@ def matchyaml(self, file, text):
 
 class RulesCollection(object):
 
-    def __init__(self):
+    def __init__(self, rulesdirs=[]):
```
webknjaz: Please don't use mutable defaults. In Python, it's common to initialize such things in the init body:
```suggestion
    def __init__(self, rulesdirs=None):
        if rulesdirs is None:
            rulesdirs = []
```

lib/ansiblelint/__init__.py
```diff
@@ -140,8 +140,10 @@ def matchyaml(self, file, text):
...
+        self._update_rulesdirs(rulesdirs)
         self.rules = []
+        self._update_rules(self.rulesdirs)
```
webknjaz: `rulesdirs` is already accessible via `self` inside that method. You probably don't need to pass it as an arg. Also, if you look at the code in init, it's not obvious where this attribute appears from.

lib/ansiblelint/__init__.py
```diff
@@ -140,8 +140,10 @@ def matchyaml(self, file, text):
 
 class RulesCollection(object):
 
-    def __init__(self):
+    def __init__(self, rulesdirs=[]):
+        self._update_rulesdirs(rulesdirs)
```
webknjaz: I don't really understand why you need a separate method for this — initializing things belongs to init. Plus it's not clear what sort of magic the method does from its name.
I'd keep
```suggestion
        self.rulesdirs = ansiblelint.utils.expand_paths_vars(rulesdirs)
```
because that method is effectively a oneliner and there's no need to abstract one function call with another function call.

lib/ansiblelint/__init__.py
```diff
@@ -196,7 +206,9 @@ def listtags(self):
     @classmethod
     def create_from_directory(cls, rulesdir):
         result = cls()
```
webknjaz: After your change of passing rulesdir into the initializer, it's not really needed anymore:
```suggestion
        return cls(rulesdir)
```
which probably means that any invocations of `RulesCollection.create_from_directory(rulesdir)` should be replaced with `RulesCollection(rulesdir)` and this classmethod should be deleted. Or maybe you should implement a public method that populates `rules` and call it here as
```suggestion
        obj = cls()
        obj.populate_rule_dirs(rulesdirs)
        return obj
```
And then there'd be no need to rewrite the initializer. (But I think that the initializer rewrite seems to be a more favorable approach)

test/TestUtils.py
```diff
@@ -154,3 +155,15 @@ def test_normpath_with_string(self):
         self.assertEqual(
             utils.normpath("a/b/../"),
             "a")
+
+    def test_expand_path_vars(self):
```
webknjaz: Since this test suite doesn't share a state on the class level, please make it a normal pytest test function:
```suggestion
def test_expand_path_vars():
```

test/TestUtils.py
```diff
@@ -154,3 +155,15 @@ def test_normpath_with_string(self):
...
+        test_path = '/test/path'
+        os.environ['TEST_PATH'] = test_path
+        self.assertEqual(utils.expand_path_vars('~'), os.path.expanduser('~'))
```
webknjaz: And also use `assert` statements to do checks as we now use `pytest` as a test runner:
```suggestion
    assert utils.expand_path_vars('~') == os.path.expanduser('~')
```

test/TestUtils.py
```diff
@@ -154,3 +155,15 @@ def test_normpath_with_string(self):
         self.assertEqual(
             utils.normpath("a/b/../"),
             "a")
+
+    def test_expand_path_vars(self):
+        test_path = '/test/path'
+        os.environ['TEST_PATH'] = test_path
```
webknjaz: The method you used modifies the global env vars object for the whole process, affecting other tests.

Please use a built-in monkeypatch fixture to set the env var temporary just for this test. It'll automatically unset it on exit.
```suggestion
    monkeypatch.setenv('TEST_PATH', test_path)
```

test/TestRunner.py
```diff
@@ -70,6 +70,19 @@ def test_unicode_standard_color_formatting(self):
         formatter = ansiblelint.formatters.Formatter()
         formatter.format(matches[0], colored=True)
 
+    def test_runner_exclude_var_expansion(self):
+        filename = 'example/lots_of_warnings.yml'
+        os.environ['EXCLUDE_PATH'] = filename
```
webknjaz: This test also needs a monkeypatch fixture.

test/TestRulesCollection.py
```diff
@@ -74,7 +75,18 @@ def test_skip_non_existent_id(self):
         matches = self.rules.run(self.ematchtestfile, skip_list=['DOESNOTEXIST'])
         self.assertEqual(len(matches), 3)
 
+    def test_rulesdir_var_expansion(self):
```
webknjaz: plz change these tests as per recommendations I posted for other test modules

lib/ansiblelint/__init__.py
evilhamsterman: I agree the create_from_directory can probably go. I removed it from __main__.py and the only other places were a bunch of tests. I wasn't sure if I should go down that route in this PR. I will go ahead an do that now.

test/TestUtils.py
evilhamsterman: I was following the paradigm that was used in the file. I will change them

test/TestRunner.py
evilhamsterman: I'm not sure how you would like me to proceed here. I understand the need for the monkeypatch and that you are moving to pytest. But it uses the RulesCollection from the class to instantiate the Runner. To use the pytest monkeypatch I need to either pull this out of the class which would mean setting up Rules in the function, or setup pytest fixtures to decorate the class with, or I can use the unittest mocking features. How would you like me proceed.

evilhamsterman: Went ahead an moved them out and just instantiated the rules in each test.

evilhamsterman:
@webknjaz I believe I've made all the changes you requested

test/TestRunner.py
```diff
@@ -22,15 +22,15 @@
 import unittest
 
 import ansiblelint
-from ansiblelint import Runner, RulesCollection
+from ansiblelint import default_rulesdir, Runner, RulesCollection
```
webknjaz: I'm not sure if using this var makes sense in tests but let's keep it.

webknjaz:
Thanks for the PR @evilhamsterman! I'll merge it soon.

test/TestRunner.py
evilhamsterman: The default_rules is the same as what the the setup in the class was doing.

webknjaz: Yep, I just think that we could be testing a more generic case or many cases w/o depending on this.

PR #676, opened 2020-02-12
Update Readme based on actual output

akx:
Commit a5828ff235d83624bcb080af2a5b199a3f4b257f changed check IDs, but these were not reflected in the documentation.


webknjaz:
Thanks, @akx!

PR #686, opened 2020-02-25
Add a new rule for detecting nested jinja mustache syntax

europ:
A new rule for ansible lint was added that provides nested jinja bracket pattern detection. Nesting jinja patterns can lead to failures and other problems.

\cc @amarza-rh

### DIFF
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -32,7 +32,7 @@ class NestedJinjaRule(AnsibleLintRule):
     )
     severity = 'VERY_HIGH'
     tags = ['formatting']
-    version_added = 'v4.2.1'
+    version_added = 'v4.3.0'
 
     nested_jinja_pattern = re.compile(r"{{(?:[^{}]*)?{{")
 
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Author: Adrián Tóth <adtoth@redhat.com>
 #
 # Copyright (c) 2020, Red Hat, Inc.
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -41,12 +41,12 @@ class NestedJinjaRule(AnsibleLintRule):
 
     def matchtask(self, file, task):
 
-        command = "".join([
+        command = "".join(
             str(value)
             # task properties are stored in the 'action' key
             for key, value in task['action'].items()
             # exclude useless values of '__file__', '__ansible_module__', '__*__', etc.
             if not key.startswith('__') and not key.endswith('__')
-        ])
+        )
 
         return bool(self.pattern.search(command))
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -1,3 +1,5 @@
+# Author: Adrián Tóth <adtoth@redhat.com>
+#
 # Copyright (c) 2020, Red Hat, Inc.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -1,0 +1,45 @@
+# Copyright (c) 2020, Red Hat, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import re
+from ansiblelint import AnsibleLintRule
+
+
+class NestedJinjaRule(AnsibleLintRule):
+    id = '207'
+    shortdesc = 'Nested jinja pattern'
+    description = (
+        'There should not be any nested jinja pattern. '
+        'Example (bad): \'{{ list_one + {{ list_two | max }} }}\', '
+        'example (good): \'{{ list_one + max(list_two) }}\''
+    )
+    severity = 'VERY_HIGH'
+    tags = ['formatting']
+    version_added = 'v4.2.1'
+
+    nested_jinja_pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+        strings = (
+            val for key, val in task['action'].items()
+            if isinstance(val, str) and not key.startswith('__')
+        )
+
+        return any(self.nested_jinja_pattern.search(item) for item in strings)
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -30,19 +30,23 @@ class NestedJinjaRule(AnsibleLintRule):
     shortdesc = 'Nested jinja pattern'
     description = (
         "There should not be any nested jinja pattern. "
-        "Example (bad): '{{ list_one + {{ list_two | max }} }}', "
-        "example (good): '{{ list_one + max(list_two) }}'"
+        "Example (bad): ``{{ list_one + {{ list_two | max }} }}``, "
+        "example (good): ``{{ list_one + max(list_two) }}``"
     )
     severity = 'VERY_HIGH'
     tags = ['formatting']
     version_added = 'v4.3.0'
 
-    nested_jinja_pattern = re.compile(r"{{(?:[^{}]*)?{{")
+    pattern = re.compile(r"{{(?:[^{}]*)?{{")
 
     def matchtask(self, file, task):
-        strings = (
-            val for key, val in task['action'].items()
-            if isinstance(val, str) and not key.startswith('__')
-        )
 
-        return any(self.nested_jinja_pattern.search(item) for item in strings)
+        command = "".join([
+            str(value)
+            # task properties are stored in the 'action' key
+            for key, value in task['action'].items()
+            # exclude useless values of '__file__', '__ansible_module__', '__*__', etc.
+            if not key.startswith('__') and not key.endswith('__')
+        ])
+
+        return bool(self.pattern.search(command))
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -26,9 +26,9 @@ class NestedJinjaRule(AnsibleLintRule):
     id = '207'
     shortdesc = 'Nested jinja pattern'
     description = (
-        'There should not be any nested jinja pattern. '
-        'Example (bad): \'{{ list_one + {{ list_two | max }} }}\', '
-        'example (good): \'{{ list_one + max(list_two) }}\''
+        "There should not be any nested jinja pattern. "
+        "Example (bad): '{{ list_one + {{ list_two | max }} }}', "
+        "example (good): '{{ list_one + max(list_two) }}'"
     )
     severity = 'VERY_HIGH'
     tags = ['formatting']
--- a/lib/ansiblelint/rules/NestedJinjaRule.py
+++ b/lib/ansiblelint/rules/NestedJinjaRule.py
@@ -22,7 +22,7 @@
 # THE SOFTWARE.
 
 import re
-from ansiblelint import AnsibleLintRule
+from ansiblelint.rules import AnsibleLintRule
 
 
 class NestedJinjaRule(AnsibleLintRule):
--- a/test/TestNestedJinjaRule.py
+++ b/test/TestNestedJinjaRule.py
@@ -25,7 +25,6 @@ from collections import namedtuple
 
 import pytest
 
-from ansiblelint.rules.NestedJinjaRule import NestedJinjaRule
 from ansiblelint.runner import Runner
 
 
@@ -186,7 +185,7 @@ def _playbook_file(tmp_path, request):
 @pytest.mark.usefixtures('_playbook_file')
 def test_including_wrong_nested_jinja(runner):
     rule_violations = runner.run()
-    assert type(rule_violations[0].rule) is NestedJinjaRule  # avoid matching subclasses
+    assert rule_violations[0].rule.id == '207'
 
 
 @pytest.mark.parametrize(
--- a/test/TestNestedJinjaRule.py
+++ b/test/TestNestedJinjaRule.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Author: Adrián Tóth <adtoth@redhat.com>
 #
 # Copyright (c) 2020, Red Hat, Inc.
--- a/test/TestNestedJinjaRule.py
+++ b/test/TestNestedJinjaRule.py
@@ -21,128 +21,148 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-import os
-import pytest
 from collections import namedtuple
-from ansiblelint import Runner, RulesCollection
+
+import pytest
+
+from ansiblelint.runner import Runner
+
 
 PlayFile = namedtuple('PlayFile', ['name', 'content'])
 
-FAIL_TASK_1LN = PlayFile('playbook.yml', u'''
-- name: one-level nesting
-  set_fact:
-    var_one: "2*(1+2) is {{ 2 * {{ 1 + 2 }} }}"
+
+FAIL_TASK_1LN = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: one-level nesting
+      set_fact:
+        var_one: "2*(1+2) is {{ 2 * {{ 1 + 2 }} }}"
 ''')
 
-FAIL_TASK_1LN_M = PlayFile('playbook.yml', u'''
-- name: one-level multiline nesting
-  set_fact:
-    var_one_ml: >
-      2*(1+2) is {{ 2 *
-      {{ 1 + 2 }}
-      }}
+FAIL_TASK_1LN_M = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: one-level multiline nesting
+      set_fact:
+        var_one_ml: >
+          2*(1+2) is {{ 2 *
+          {{ 1 + 2 }}
+          }}
 ''')
 
-FAIL_TASK_2LN = PlayFile('playbook.yml', u'''
-- name: two-level nesting
-  set_fact:
-    var_two: "2*(1+(3-1)) is {{ 2 * {{ 1 + {{ 3 - 1 }} }} }}"
+FAIL_TASK_2LN = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: two-level nesting
+      set_fact:
+        var_two: "2*(1+(3-1)) is {{ 2 * {{ 1 + {{ 3 - 1 }} }} }}"
 ''')
 
-FAIL_TASK_2LN_M = PlayFile('playbook.yml', u'''
-- name: two-level multiline nesting
-  set_fact:
-    var_two_ml: >
-      2*(1+(3-1)) is {{ 2 *
-      {{ 1 +
-      {{ 3 - 1 }}
-      }} }}
+FAIL_TASK_2LN_M = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: two-level multiline nesting
+      set_fact:
+        var_two_ml: >
+          2*(1+(3-1)) is {{ 2 *
+          {{ 1 +
+          {{ 3 - 1 }}
+          }} }}
 ''')
 
-FAIL_TASK_W_5LN = PlayFile('playbook.yml', u'''
-- name: five-level wild nesting
-  set_fact:
-    var_three_wld: "{{ {{ {{ {{ {{ 234 }} }} }} }} }}"
+FAIL_TASK_W_5LN = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: five-level wild nesting
+      set_fact:
+        var_three_wld: "{{ {{ {{ {{ {{ 234 }} }} }} }} }}"
 ''')
 
-FAIL_TASK_W_5LN_M = PlayFile('playbook.yml', u'''
-- name: five-level wild multiline nesting
-  set_fact:
-    var_three_wld_ml: >
-      {{
-      {{
-      {{
-      {{
-      {{ 234 }}
-      }}
-      }}
-      }}
-      }}
+FAIL_TASK_W_5LN_M = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: five-level wild multiline nesting
+      set_fact:
+        var_three_wld_ml: >
+          {{
+          {{
+          {{
+          {{
+          {{ 234 }}
+          }}
+          }}
+          }}
+          }}
 ''')
 
-SUCCESS_TASK_P = PlayFile('playbook.yml', u'''
-- name: proper non-nested example
-  set_fact:
-    var_one: "number for 'one' is {{ 2 * 1 }}"
+SUCCESS_TASK_P = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: non-nested example
+      set_fact:
+        var_one: "number for 'one' is {{ 2 * 1 }}"
 ''')
 
-SUCCESS_TASK_P_M = PlayFile('playbook.yml', u'''
-- name: proper multiline non-nested example
-  set_fact:
-    var_one_ml: >
-      number for 'one' is {{
-      2 * 1 }}
+SUCCESS_TASK_P_M = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: multiline non-nested example
+      set_fact:
+        var_one_ml: >
+          number for 'one' is {{
+          2 * 1 }}
 ''')
 
-SUCCESS_TASK_2P = PlayFile('playbook.yml', u'''
-- name: proper nesting far from each other
-  set_fact:
-    var_two: "number for 'two' is {{ 2 * 1 }} and number for 'three' is {{ 4 - 1 }}"
+SUCCESS_TASK_2P = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: nesting far from each other
+      set_fact:
+        var_two: "number for 'two' is {{ 2 * 1 }} and number for 'three' is {{ 4 - 1 }}"
 ''')
 
-SUCCESS_TASK_2P_M = PlayFile('playbook.yml', u'''
-- name: proper multiline nesting far from each other
-  set_fact:
-    var_two_ml: >
-      number for 'two' is {{ 2 * 1
-      }} and number for 'three' is {{
-      4 - 1 }}
+SUCCESS_TASK_2P_M = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: multiline nesting far from each other
+      set_fact:
+        var_two_ml: >
+          number for 'two' is {{ 2 * 1
+          }} and number for 'three' is {{
+          4 - 1 }}
 ''')
 
-SUCCESS_TASK_C_2P = PlayFile('playbook.yml', u'''
-- name: proper nesting close to each other
-  set_fact:
-    var_three: "number for 'ten' is {{ 2 - 1 }}{{ 3 - 3 }}"
+SUCCESS_TASK_C_2P = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: nesting close to each other
+      set_fact:
+        var_three: "number for 'ten' is {{ 2 - 1 }}{{ 3 - 3 }}"
 ''')
 
-SUCCESS_TASK_C_2P_M = PlayFile('playbook.yml', u'''
-- name: proper multiline nesting close to each other
-  set_fact:
-    var_three_ml: >
-      number for 'ten' is {{
-      2 - 1
-      }}{{ 3 - 3 }}
+SUCCESS_TASK_C_2P_M = PlayFile('playbook.yml', '''
+- hosts: all
+  tasks:
+    - name: multiline nesting close to each other
+      set_fact:
+        var_three_ml: >
+          number for 'ten' is {{
+          2 - 1
+          }}{{ 3 - 3 }}
 ''')
+
 
 @pytest.fixture
-def play_file_path(tmp_path):
-    p = tmp_path / 'playbook.yml'
-    return str(p)
+def runner(tmp_path, default_rules_collection):
+    return Runner(
+        default_rules_collection,
+        str(tmp_path / 'playbook.yml'),
+        [], [], [],
+    )
 
 
 @pytest.fixture
-def rules():
-    rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
-    return RulesCollection([rulesdir])
-
-
-@pytest.fixture
-def runner(play_file_path, rules):
-    return Runner(rules, play_file_path, [], [], [])
-
-
-@pytest.fixture
-def play_files(tmp_path, request):
+def _playbook_file(tmp_path, request):
     if request.param is None:
         return
     for play_file in request.param:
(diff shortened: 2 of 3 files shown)


lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,45 @@
...
+# THE SOFTWARE.
+
+import re
+from ansiblelint import AnsibleLintRule
+
+
+class NestedJinjaRule(AnsibleLintRule):
+    id = '207'
+    shortdesc = 'Nested jinja pattern'
+    description = 'There should not be any nested jinja patterns.'
```
webknjaz: I think it'd be useful to show "bad" and "good" examples. I personally have trouble imagining what this check does exactly.

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,45 @@
...
+
+    nested_jinja_pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+        strings = [
+            val for key, val in task['action'].items()
+            if isinstance(val, str) and not key.startswith('__')
+        ]
+
+        for item in strings:
```
webknjaz: It'd probably read better semantically with `any()`
```suggestion
        return any(self.nested_jinja_pattern.search(item) for item in strings)
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,45 @@
...
+    tags = ['formatting']
+    version_added = 'v4.2.1'
+
+    nested_jinja_pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+        strings = [
+            val for key, val in task['action'].items()
+            if isinstance(val, str) and not key.startswith('__')
+        ]
```
webknjaz: Let's use a generator expression in order to avoid creating another in-memory list:
```suggestion
        strings = (
            val for key, val in task['action'].items()
            if isinstance(val, str) and not key.startswith('__')
        (
```

webknjaz:
recheck

lib/ansiblelint/rules/NestedJinjaRule.py
europ: May it be like this?
```py
description = (
    'There should not be any nested jinja pattern. '
    'Example (bad): \'{{ list_one + {{ list_two | max }} }}\', '
    'example (good): \'{{ list_one + max(list_two) }}\''
)
```

ssbarnea:
recheck

lib/ansiblelint/rules/NestedJinjaRule.py
webknjaz: Yeah, that looks good.

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,45 @@
...
+from ansiblelint import AnsibleLintRule
+
+
+class NestedJinjaRule(AnsibleLintRule):
+    id = '207'
+    shortdesc = 'Nested jinja pattern'
+    description = (
+        'There should not be any nested jinja pattern. '
+        'Example (bad): \'{{ list_one + {{ list_two | max }} }}\', '
+        'example (good): \'{{ list_one + max(list_two) }}\''
```
webknjaz: It's a good practice to invert quotes when it's possible to avoid escaping:
```suggestion
        "Example (bad): '{{ list_one + {{ list_two | max }} }}', "
        "example (good): '{{ list_one + max(list_two) }}'"
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    severity = 'VERY_HIGH'
+    tags = ['formatting']
+    version_added = 'v4.2.1'
```
webknjaz: I think it'll be
```suggestion
    version_added = 'v4.3.0'
```
because per semver the minor bit is incremented for features and the patch bit is incremented for bugfixes. And this is not a bugfix, it's a feature.

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,45 @@
...
+
+    nested_jinja_pattern = re.compile(r"{{(?:[^{}]*)?{{")
```
webknjaz: Does this really need to exist outside the `matchtask` method? If yes, I'd put it on the module level as `NESTED_JINJA_PATTERN = re.compile(r"{{(?:[^{}]*)?{{")` rather than in a class.

webknjaz:
@europ could you please also submit tests for this feature? It'd be hard to keep it working/healthy/supported without any CI checks in place.

europ:
Good evening @webknjaz,

Thank you for your added patches.

I am sorry that I did not add the tests yet. I am struggling with other current tasks. I will add it soon.

Best regards.

europ:
### TODO: add tests

lib/ansiblelint/rules/NestedJinjaRule.py
europ: I was inspired by other rules where it was done the same.

E.g.: https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/VariableHasSpacesRule.py#L16

It can be found in the following files too:
* `ComparisonToEmptyStringRule.py`
* `ComparisonToLiteralBoolRule.py`
* `MetaTagValidRule.py`
* `MetaVideoLinksRule.py`
* `MetaVideoLinksRule.py`
* `MetaVideoLinksRule.py`
* `NestedJinjaRule.py`
* `ShellWithoutPipefail.py`
* `ShellWithoutPipefail.py`
* `UsingBareVariablesIsDeprecatedRule.py`
* `UsingBareVariablesIsDeprecatedRule.py`
* `VariableHasSpacesRule.py`
* `VariableHasSpacesRule.py`

Do you want to move it there anyway?

webknjaz: Ah, nevermind then. A Foolish Consistency is the Hobgoblin of Little Minds.

europ:
### :heavy_check_mark: tests added

europ:
Hi @webknjaz, I have added the tests as soon as I could. I am sorry for the delay.

webknjaz:
@europ it's alright. No need to rush. Could you please check the failing CI jobs? Looks like the tests are failing.

europ:
There is some warning `RemovedInSphinx30Warning` due to that docs build fails.

```
 The HTML pages are in _build/html.
/home/runner/work/ansible-lint/ansible-lint/docs/docsite/_themes/sphinx_rtd_theme/search.html:21: RemovedInSphinx30Warning: To modify script_files in the theme is deprecated. Please insert a <script> tag directly in your theme instead.
  {% endblock %}
make[1]: Leaving directory '/home/runner/work/ansible-lint/ansible-lint/docs/docsite'
ERROR: Building docs changed tracked files, include them in commit to avoid this failure.
Makefile:38: recipe for target 'htmldocs' failed
make: *** [htmldocs] Error 1
make: Leaving directory '/home/runner/work/ansible-lint/ansible-lint/docs/docsite'
ERROR: InvocationError for command /usr/bin/make -C docs/docsite htmldocs (exited with code 2)
```

I tried some of the shared advice I found but I didn't succeed.

* https://stackoverflow.com/questions/57401254/fix-sphinx-removedinsphinx30warning
* https://github.com/readthedocs/sphinx_rtd_theme/issues/739
* https://github.com/gevent/gevent/issues/1404

I am not very familiar with jekyll / sphinx, do you have any idea how to fix this warning? It might be needed to rewrite some parts of the documentation to fix this.

webknjaz:
I think that sphinx errors are not related to your change. But there's failures in other jobs that might be.

webknjaz:
Just to be on the safe side, rebase your branch on top of the fresh master

webknjaz:
Oh, it looks like you need to regenerate `default_rules.rst`:
```
cd ../../lib/ansiblelint; python ./generate_docs.py
../../docs/docsite/rst/rules/default_rules.rst file written
...
ERROR: Building docs changed tracked files, include them in commit to avoid this failure.
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,47 @@
+# Author: Adrián Tóth <adtoth@redhat.com>
```
webknjaz:
```suggestion
# -*- coding: utf-8 -*-
# Author: Adrián Tóth <adtoth@redhat.com>
```

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,71 @@
+# Author: Adrián Tóth <adtoth@redhat.com>
```
webknjaz:
```suggestion
# -*- coding: utf-8 -*-
# Author: Adrián Tóth <adtoth@redhat.com>
```

lib/ansiblelint/rules/NestedJinjaRule.py
europ: Ou, yes, thanks for catching that.

europ:
BTW, should I squash commitsl into one?

webknjaz:
@europ I usually prefer squashing only related commits. Not all into one. Commits should be atomic: if you squash unrelated things they will be non-atomic.

webknjaz:
Now it's visible how your test is failing. It needs to be fixed before I can perform the merge:
- https://github.com/ansible/ansible-lint/pull/686/checks?check_run_id=528955398#step:8:188
- https://github.com/ansible/ansible-lint/pull/686/checks?check_run_id=528955393#step:8:188
- https://github.com/ansible/ansible-lint/pull/686/checks?check_run_id=528955439#step:8:188

amarza-rh:
I'd add at least a success and failed test case where the doubly-nested jinja is across multiple lines.

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import unittest
```
webknjaz: Plz avoid using unittest. We're trying to migrate towards pure pytest so just use it instead. inheriting from `unittest.TestCase` is not recommended in such setup and is harmful.

Here's an example of a test suite for reference that follows this rule:
https://github.com/ansible/ansible-lint/blob/master/test/TestIncludeMissingFileRule.py

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,120 @@
...
+from ansiblelint import RulesCollection
+from ansiblelint.rules.NestedJinjaRule import NestedJinjaRule
+from test import RunFromText
+
+FAIL_TASKS = '''
+- name: one-level nesting
+  set_fact:
+    var_one: "2*(1+2) is {{ 2 * {{ 1 + 2 }} }}"
+
+- name: one-level multiline nesting
```
webknjaz: Let's separate these cases into separate tests. Using `@pytest.mark.parametrize` is a way to go.

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,190 @@
...
+        [
+            SUCCESS_TASK_P,
+            SUCCESS_TASK_P_M,
+            SUCCESS_TASK_2P,
+            SUCCESS_TASK_2P_M,
+            SUCCESS_TASK_C_2P,
+            SUCCESS_TASK_C_2P_M
+        ],
+        id='file does not include nested jinja pattern'
+    )],
```
webknjaz, line 185: You created one param with a nested list of tasks. But you need multiple params with each of them having its own id so that it will generate a new test per param. Once done, each parametrized subtest will be visible in the pytest output.

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,190 @@
...
+
+@pytest.fixture
+def rules():
+    rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
+    return RulesCollection([rulesdir])
+
+
+@pytest.fixture
+def runner(play_file_path, rules):
+    return Runner(rules, play_file_path, [], [], [])
```
webknjaz, line 141: There's now `default_rules_collection` fixture in `conftest`. So you don't really need `play_file_path` and `rules` fixtures anymore:
```suggestion
def runner(tmp_path, default_rules_collection):
    return Runner(
        default_rules_collection,
        str(tmp_path / 'playbook.yml'),
        [], [], [],
    )
```
like @ https://github.com/ansible/ansible-lint/commit/6702769f4488529e16dc5f3dc320a363586364eb#diff-e21acb2ef339763de8d17d0a8be0f295R33-R34

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,190 @@
...
+@pytest.fixture
+def play_file_path(tmp_path):
+    p = tmp_path / 'playbook.yml'
+    return str(p)
+
+
+@pytest.fixture
+def rules():
+    rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
+    return RulesCollection([rulesdir])
```
webknjaz, line 136:
```suggestion
```

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,190 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import os
```
webknjaz, line 24:
```suggestion
```

webknjaz:
Oh, and besides the comments above, you can also check linters locally with `tox -e lint` if you need, they shouldn't fail the CI.

test/TestNestedJinjaRule.py
```diff
@@ -21,40 +21,51 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-import unittest
+import os
+import pytest
+from collections import namedtuple
+from ansiblelint import Runner, RulesCollection
```
webknjaz, line 27: collections is stdlib, pytest is thirdparty, ansiblelint is local:
```suggestion
from collections import namedtuple

import pytest

from ansiblelint import Runner
```

europ:
@webknjaz you are the fasted reviewer in the world :D

Thanks for the review and advices. Firstly, there are problems I need to fix beforehand:
* fix the regexp, we forget about the case when the user can escape the jinja patter
    * e.g. `{{ 'string {{ string' }}`
    * real example is here https://stackoverflow.com/a/32283447/11225802
* fix the rule, I am still wondering why it does not match the given 'bad' examples
* finally, fix the tests and other things such as linting, etc.

ssbarnea:
@europ Any chance to bring to green status? (you can ignore ansible/third-party-check)

europ:
changes:
* added change requests from the given review
* changed the logic of the pattern matching, here is an example:
    ```
    (Pdb) task
    {'delegate_to': <class 'ansible.utils.sentinel.Sentinel'>, 'name': 'task name', '__line__': 4, '__file__': 'adrian/test.yml', 'skipped_rules': [], 'action'
    : {'__ansible_module__': 'shell', '__ansible_arguments__': ['echo', '{{', '1', '+', '{{', '2+2', '}}}}']}, '__ansible_action_type__': 'task'}
    (Pdb) file
    {'path': 'adrian/test.yml', 'type': 'playbook'}
    (Pdb) command = "".join([val for key, val in task['action'].items()][1])
    (Pdb) command
    'echo{{1+{{2+2}}}}'
    (Pdb) True if self.pattern.search(command) else False
    True
    (Pdb) command = "{{a}}{{a}}"
    (Pdb) True if self.pattern.search(command) else False
    False
    (Pdb)
    ```

I have problem with the pytest, based on your review I did the changes. The pattern does match the given input correctly but the pytest unit tests do not work. I face some ValueError - duplicate 'play_files'. Can you give me a hit for that?

webknjaz:
@europ you shouldn't have multiple decorators for the same param. Read how the fixtures work here: https://docs.pytest.org/en/latest/fixture.html.

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,214 @@
...
+@pytest.mark.parametrize(
+    'play_files',
+    [pytest.param([FAIL_TASK_W_5LN], id='file includes nested jinja pattern')],
+    indirect=['play_files']
+)
+@pytest.mark.parametrize(
+    'play_files',
+    [pytest.param([FAIL_TASK_W_5LN_M], id='file includes nested jinja pattern')],
+    indirect=['play_files']
+)
```
webknjaz, line 176:
```suggestion
@pytest.mark.parametrize(
    'play_files',
    (
        pytest.param([FAIL_TASK_1LN], id='file includes nested jinja pattern'),  # FIXME: set a proper id
        pytest.param([FAIL_TASK_1LN_M], id='file includes nested jinja pattern'),  # FIXME: set a proper id
        pytest.param([FAIL_TASK_2LN], id='file includes nested jinja pattern'),  # FIXME: set a proper id
        pytest.param([FAIL_TASK_2LN_M], id='file includes nested jinja pattern'),  # FIXME: set a proper id
        pytest.param([FAIL_TASK_W_5LN], id='file includes nested jinja pattern'),  # FIXME: set a proper id
        pytest.param([FAIL_TASK_W_5LN_M], id='file includes nested jinja pattern'),  # FIXME: set a proper id
    ),
    indirect=('play_files', ),
)
```

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,214 @@
...
+def test_including_nested_jinja(runner, play_files):
```
webknjaz, line 177:
```suggestion
@pytest.mark.usefixtures('play_files')
def test_including_nested_jinja(runner):
```

test/TestNestedJinjaRule.py
europ: Yes, you are right https://github.com/ansible/ansible-lint/blob/master/test/conftest.py#L10

```py
@pytest.fixture
@pytest.mark.usefixtures('default_rules_collection')
def runner(tmp_path, default_rules_collection):
    return Runner(
        default_rules_collection,
        str(tmp_path / 'playbook.yml'),
        [], [], [],
    )
```
```
file /home/adrian/Documents/ansible-lint/test/TestNestedJinjaRule.py, line 129                                                                
  @pytest.fixture                                                                                                                             
  def runner(tmp_path, default_rules_collection):                                                                                             
E       fixture 'default_rules_collection' not found                                                                                          
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cov, doctest_namespace, monkeypatch, no_cover, play_files
, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, runner, testrun_uid, tmp_path, tmp_path_factory, tm
pdir, tmpdir_factory, worker_id
```

I tried to define a fixture using scope too as it should be, but I did not succeed.

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,46 @@
...
+from ansiblelint import AnsibleLintRule
+
+
+class NestedJinjaRule(AnsibleLintRule):
+    id = '207'
+    shortdesc = 'Nested jinja pattern'
+    description = (
+        "There should not be any nested jinja pattern. "
+        "Example (bad): '{{ list_one + {{ list_two | max }} }}', "
+        "example (good): '{{ list_one + max(list_two) }}'"
```
webknjaz, line 34: Let's mark these as inline code in RST:
```suggestion
        "Example (bad): ``{{ list_one + {{ list_two | max }} }}``, "
        "example (good): ``{{ list_one + max(list_two) }}``"
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,46 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import re
+from ansiblelint import AnsibleLintRule
```
webknjaz, line 25: I'd prefer this to be a relative import:
```suggestion
from . import AnsibleLintRule
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,46 @@
...
+    tags = ['formatting']
+    version_added = 'v4.3.0'
+
+    pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+
+        command = "".join([val for key, val in task['action'].items()][1])
+
+        return True if self.pattern.search(command) else False
```
webknjaz, line 46:
```suggestion
        return bool(self.pattern.search(command)
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    severity = 'VERY_HIGH'
+    tags = ['formatting']
+    version_added = 'v4.3.0'
+
+    pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+
+        command = "".join([val for key, val in task['action'].items()][1])
```
webknjaz, line 44:
```suggestion
        command = "".join(list(task['action'].values())[1])
```

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,184 @@
...
+
+from collections import namedtuple
+import pytest
+from ansiblelint import Runner
+
+
+PlayFile = namedtuple('PlayFile', ['name', 'content'])
+
+
+FAIL_TASK_1LN = PlayFile('playbook.yml', u'''
```
webknjaz, line 32: u-prefix isn't really necessary since we dropped support for Python 2:
```suggestion
FAIL_TASK_1LN = PlayFile('playbook.yml', '''
```

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,184 @@
...
+        pytest.param([FAIL_TASK_2LN_M], id='file includes two-level multiline nesting'),
+        pytest.param([FAIL_TASK_W_5LN], id='file includes five-level wild nesting'),
+        pytest.param([FAIL_TASK_W_5LN_M], id='file includes five-level wild multiline nesting'),
+    ),
+    indirect=['_playbook_file'],
+)
+@pytest.mark.usefixtures('_playbook_file')
+def test_including_wrong_nested_jinja(runner):
+    results = runner.run()
+    # FIXME: results == []
```
webknjaz, line 162: why?

test/TestNestedJinjaRule.py
webknjaz: I see it works now. I guess you forgot to rebase, right?

test/TestNestedJinjaRule.py
europ: **The runner does not return any value in the tests.**

It always returns only an empty list, whether the input is correct or incorrect. When I check if the mocked file includes the given rule, it does exist as it is supposed to be. The new rule is also included in that runner as it is expected.

The rule always returns a boolean data type but it cannot be cached in the test for now.

Whenever I launch ansible-lint as `ansible-lint -r <DIR> <FILE>` while the `<DIR>` includes that new rule, it works perfectly based on the correct/incorrect `<FILE>` input.

Should I change the method of runner launching?

test/TestNestedJinjaRule.py
europ: Yes. It was due to outdated changes. I have already rebased my branch against upstream/master.

lib/ansiblelint/rules/NestedJinjaRule.py
europ: hmm, `tox -e docs` fails on relative import like this

webknjaz: @europ okay, if it fails, revert this change.

test/TestNestedJinjaRule.py
webknjaz: Not sure what you mean. Are you saying that it doesn't work with the default rules dir?

europ: The created runner object includes all the rules (including the 'nested jinja rule') as well as the mocked file, which is properly created.
```
(Pdb) runner.checked_files
{'../../../../tmp/pytest-of-adrian/pytest-0/test_including_wrong_nested_ji0/playbook.yml'}
(Pdb) import os; os.system("cat ../../../../tmp/pytest-of-adrian/pytest-0/test_including_wrong_nested_ji0/playbook.yml")

- name: one-level nesting
  set_fact:
    var_one: "2*(1+2) is {{ 2 * {{ 1 + 2 }} }}"
0
(Pdb) runner.rules
101: ...
...
206: ...
  ...
207: Nested jinja pattern
  There should not be any nested jinja pattern. Example (bad): ``{{ list_one + {{ list_two | max }} }}``, example (good): ``{{ list_one + max(list_two) }}``
301: ...
  ...
...
901: ...
  ...
```

I use `runner.run()` accordingly to [`Runner` definition](https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/runner.py#L56)
```
class Runner(object):
    ...
    def run(self) -> List:
    ...
```

The runner should return a nonempty list in case of "bad" input. Do I launch the runner incorrectly in the tests that I always have an empty list?

webknjaz: Looks about right. Have you tried sticking debugger into your rule's `matchtask()`?

webknjaz: Have you tried converting tasks lists into actual playbooks?
(I mean add one more level of nesting and put tasks under `tasks:`)

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,186 @@
...
+from ansiblelint.runner import Runner
+
+
+PlayFile = namedtuple('PlayFile', ['name', 'content'])
+
+
+FAIL_TASK_1LN = PlayFile('playbook.yml', '''
+- name: one-level nesting
+  set_fact:
+    var_one: "2*(1+2) is {{ 2 * {{ 1 + 2 }} }}"
```
webknjaz, line 37:
```suggestion
- hosts: all
  tasks:
  - name: one-level nesting
    set_fact:
      var_one: "2*(1+2) is {{ 2 * {{ 1 + 2 }} }}"
```

webknjaz: try this

europ: huge thaaaaaaaanks for this!

test/TestNestedJinjaRule.py
europ: The `matchtask` behavior has changed, I had to fix it again. Now, the runner in the tests returns a list of errors as expected. This thread has been resolved.

europ:
@webknjaz it should be done, please do a final review, let me know if there is anything left to enhance

* tests are working
* the rule does the purpose
* the branch is rebased and up to date

Screenshot from 2020-06-16 23-10-18

webknjaz:
@europ looks like `test/TestRunner.py::test_runner[test/become.yml-exclude4-0]` is failing. Also, I think it'd be great to use `ids=` arg in `@pytest.mark.parametrize()` because autogenerated test param names (like `test/become.yml-exclude4-0`) are not very informative in the test run report, in logs. Would you improve this, please?

webknjaz:
Oh, wait... You didn't touch `TestRunner.py`. So you don't have to fix its IDs. But it looks like your code broke that test meaning you introduced a bug that should be fixed before proceeding: https://github.com/ansible/ansible-lint/pull/686/checks?check_run_id=778145463#step:9:299.

Also, please mark your PR as "Ready for review" once the CI checks are green.

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,209 @@
...
+    '_playbook_file',
+    (
+        pytest.param([SUCCESS_TASK_P], id='file includes non-nested example'),
+        pytest.param([SUCCESS_TASK_P_M], id='file includes multiline non-nested example'),
+        pytest.param([SUCCESS_TASK_2P], id='file includes nesting far from each other'),
+        pytest.param([SUCCESS_TASK_2P_M], id='file includes multiline nesting far from each other'),
+        pytest.param([SUCCESS_TASK_C_2P], id='file includes nesting close to each other'),
+        pytest.param(
+            [SUCCESS_TASK_C_2P_M],
+            id='file includes multiline nesting close to each other'
```
webknjaz, line 201: Nitpick: let's have a trailing comma here too, for consistency.
```suggestion
            id='file includes multiline nesting close to each other',
```

europ:
Hmm... `TypeError: sequence item 2: expected str instance, bool found`. Going to fix it.

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,209 @@
...
+        pytest.param([FAIL_TASK_2LN_M], id='file includes two-level multiline nesting'),
+        pytest.param([FAIL_TASK_W_5LN], id='file includes five-level wild nesting'),
+        pytest.param([FAIL_TASK_W_5LN_M], id='file includes five-level wild multiline nesting'),
+    ),
+    indirect=['_playbook_file'],
+)
+@pytest.mark.usefixtures('_playbook_file')
+def test_including_wrong_nested_jinja(runner):
+    results = runner.run()
+    assert results[0].rule.shortdesc == 'Nested jinja pattern'
```
webknjaz, line 188: I think it's better to check the id because the description may change over time and it'd be annoying to also update tests just because of that:
```suggestion
    assert results[0].rule.id == '207'
```

europ: Should I add `assert results` before? Or it is good as it is?

```py
def test_including_wrong_nested_jinja(runner):
    results = runner.run()
    assert results
    assert results[0].rule.id == '207'
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,50 @@
...
+    severity = 'VERY_HIGH'
+    tags = ['formatting']
+    version_added = 'v4.3.0'
+
+    pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+
+        command = "".join([
+            value
```
webknjaz, line 45: This could be
```suggestion
            str(value)
```
but also for some types it may be misleading. Like `b'something'` may be turned into `"b'something'"`. So you may want to be careful here — bytes should be converted differently.

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,50 @@
...
+            for key, value in task['action'].items()
```
webknjaz, line 46: why action exactly? I think this chunk of code would benefit from having a code comment committed next to it and explaining the motivation of such logic.

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,50 @@
...
+            if ((not key.startswith('__')) and (not key.endswith('__')))
```
webknjaz, line 47: This would also need an explanation. I can read what it does but it's hard to guess *why* just by looking at this code. Also, the braces are redundant:
```suggestion
            if not key.startswith('__') and not key.endswith('__')
```

webknjaz: So you're filtering out things that don't both start and with `__`? Maybe it'd read better as
```suggestion
            if not (key.startswith('__') and key.endswith('__'))
```

or even could be moved to a separate function like `is_dunder_name()` that'd explain its meaning.

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,209 @@
...
+            [SUCCESS_TASK_C_2P_M],
+            id='file includes multiline nesting close to each other'
+        ),
+    ),
+    indirect=['_playbook_file'],
+)
+@pytest.mark.usefixtures('_playbook_file')
+def test_including_proper_nested_jinja(runner):
+    results = runner.run()
+    assert not results
```
webknjaz, line 209: How about we add a bit more meaning here?
```suggestion
    rule_violations = runner.run()
    assert not rule_violations
```

lib/ansiblelint/rules/NestedJinjaRule.py
europ: added explanatory comment

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,209 @@
...
+        pytest.param([FAIL_TASK_2LN_M], id='file includes two-level multiline nesting'),
+        pytest.param([FAIL_TASK_W_5LN], id='file includes five-level wild nesting'),
+        pytest.param([FAIL_TASK_W_5LN_M], id='file includes five-level wild multiline nesting'),
+    ),
+    indirect=['_playbook_file'],
+)
+@pytest.mark.usefixtures('_playbook_file')
+def test_including_wrong_nested_jinja(runner):
+    rule_violations = runner.run()
+    assert rule_violations[0].rule.id == '207'
```
webknjaz, line 188: On the second thought, it's even better to just check if the rule is an instance of our target.
```suggestion
    assert type(rule_violations[0].rule) is NestedJinjaRule  # avoid matching subclasses
```

test/TestNestedJinjaRule.py
```diff
@@ -0,0 +1,209 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from collections import namedtuple
+
+import pytest
+
+from ansiblelint.runner import Runner
```
webknjaz, line 28:
```suggestion
from ansiblelint.rules.NestedJinjaRule import NestedJinjaRule
from ansiblelint.runner import Runner
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,52 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import re
+from ansiblelint import AnsibleLintRule
```
webknjaz, line 25: The top-level import is now deprecated so we need to use a nested one.
```suggestion
from ansiblelint.rules import AnsibleLintRule
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,52 @@
...
+
+    pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+
+        command = "".join([
+            str(value)
+            # task properties are stored in the 'action' key
+            for key, value in task['action'].items()
+            # exclude useless values of '__file__', '__ansible_module__', '__*__', etc.
```
webknjaz, line 48: This shouldn't explain *how* the code works. Reading it is pretty obvious. Code comments must be used to explain *why* this code is here.

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    severity = 'VERY_HIGH'
+    tags = ['formatting']
+    version_added = 'v4.3.0'
+
+    pattern = re.compile(r"{{(?:[^{}]*)?{{")
+
+    def matchtask(self, file, task):
+
+        command = "".join([
```
webknjaz, line 44: No need to create a "throw-away" in-memory list. It only consumes memory. Better use a generator expression:
```suggestion
        command = "".join(
```

lib/ansiblelint/rules/NestedJinjaRule.py
```diff
@@ -0,0 +1,52 @@
...
+            str(value)
+            # task properties are stored in the 'action' key
+            for key, value in task['action'].items()
+            # exclude useless values of '__file__', '__ansible_module__', '__*__', etc.
+            if not key.startswith('__') and not key.endswith('__')
+        ])
```
webknjaz, line 50: Same
```suggestion
        )
```

lib/ansiblelint/rules/NestedJinjaRule.py
webknjaz: I guess this is not very important atm so I won't wait for the rephrase.

webknjaz:
I didn't take into account that we have a custom dynamic rule importer and the class won't match. Reverting that.

europ:
@webknjaz thanks for the help & review!

PR #691, opened 2020-03-13
Add rules for verifying the existence of imported and included files

jlusiardi:
This new rule checks if file referenced in import and include are
actually existing and issues an error if not. This can be switched of
with a noqa comment (e.g. on files that depend on a variable).

The change requires also a little adjustment to `Runner.run` to get 
an absolute path of the listed file (important for relative references
to files)


lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2016, Will Thames and contributors
```
webknjaz: These copyright lines are invalid

lib/ansiblelint/__init__.py
```diff
@@ -259,7 +259,11 @@ def run(self):
                 continue
             if playbook[1] == 'role':
                 continue
-            files.append({'path': ansiblelint.utils.normpath(playbook[0]), 'type': playbook[1]})
+            files.append({'path': ansiblelint.utils.normpath(playbook[0]),
+                          'type': playbook[1],
+                          # add an absolute path here, so rules are able to validate if
+                          # referenced files exist
+                          'abs_path': os.path.dirname(playbook[0])})
```
webknjaz: this name is weird: it looks like it's a dir path but the implication seems to be that it's a file path.

lib/ansiblelint/rules/IncludeMissingFileRule.py
jlusiardi: Should it be `# Copyright (c) 2020, Will Thames and contributors` then?

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,44 @@
...
+        results = []
+        for task in data.get('tasks', []):
+
+            # check if the id of the current rule is not in list of skipped rules for this play
+            if self.id in task['skipped_rules']:
+                continue
+
+            # collect information which file was referenced for include / import
+            referenced_file = None
+            for key in task.keys():
```
webknjaz: Use items instead of doing this manually
```suggestion
            for key, val in task.items():
```

lib/ansiblelint/rules/IncludeMissingFileRule.py
webknjaz: No, it's a new module, all code is yours. You can write yourself and Ansible Project.

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,44 @@
...
+    def _get_play_file(self, playbook_text):
+        with open(os.path.join(self.play_root, 'playbook.yml'), 'w') as f_play:
+            f_play.write(playbook_text)
+        return f_play
+
+    def test_include(self):
+        fh = self._get_play_file(PLAY_INCLUDE)
+        runner = Runner(self.rules, fh.name, [], [], [])
+        results = runner.run()
+        print(results)
```
webknjaz: this is unnecessary in tests

lib/ansiblelint/rules/IncludeMissingFileRule.py
jlusiardi: Is this ok then?
```
# Copyright (c) 2020, Joachim Lusiardi
# Copyright (c) 2020, Ansible Project
```

webknjaz: Yes. Usually, the copyright states when the content was created/updated.

test/TestIncludeMissingFileRule.py
jlusiardi: agreed, will remove the `print` in the next commit

lib/ansiblelint/__init__.py
jlusiardi: would `absolute_directory` be a better name?

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import print_function
+from ansiblelint import AnsibleLintRule
+import os.path
+import ansible.parsing.yaml.objects
+
+
+class IncludeMissingFileRule(AnsibleLintRule):
+    id = '4711'
+    shortdesc = 'an include task references a non existent file'
+    description = 'Tabs can cause unexpected display issues, use spaces'
```
webknjaz: This doesn't seem right

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+    severity = 'HIGH'
+    tags = ['formatting']
+    version_added = 'v4.2.0'
```
webknjaz: It's already released so it's probably `v4.3.0` if this gets in before we cut the release.

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2020, Joachim Lusiardi
+# Copyright (c) 2020, Ansible Project
+
+from __future__ import print_function
+from ansiblelint import AnsibleLintRule
```
webknjaz: Local imports should be the last ones.

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2020, Joachim Lusiardi
+# Copyright (c) 2020, Ansible Project
+
+from __future__ import print_function
```
webknjaz: future imports should be separated

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+from ansiblelint import AnsibleLintRule
+import os.path
```
webknjaz: stdlib imports should go right after the future imports

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+import ansible.parsing.yaml.objects
```
webknjaz: third-party imports should go after stdlib and before locals

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+            for key, val in task.items():
+                if key.startswith('include_') or key.startswith('import_') or key == 'include':
+                    if isinstance(val, ansible.parsing.yaml.objects.AnsibleMapping):
+                        referenced_file = val.get('file', None)
+                    else:
+                        referenced_file = val
+
+            if referenced_file:
+                # make sure we have a absolute path here and check if it is a file
+                referenced_file = os.path.join(absolute_directory, referenced_file)
```
webknjaz: What if it's `None`?

lib/ansiblelint/__init__.py
webknjaz: probably. but I'm still not sure if it's the right design decision

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+            # collect information which file was referenced for include / import
+            referenced_file = None
+            for key, val in task.items():
+                if key.startswith('include_') or key.startswith('import_') or key == 'include':
+                    if isinstance(val, ansible.parsing.yaml.objects.AnsibleMapping):
+                        referenced_file = val.get('file', None)
+                    else:
+                        referenced_file = val
+
+            if referenced_file:
```
webknjaz: Invert the check to interrupt early and dedent the block of code
```suggestion
            if not referenced_file:
                break
```

webknjaz: Also, it seems like you actually should be checking for `None`:
```suggestion
            if referenced_file is None:
```

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+                continue
+
+            # collect information which file was referenced for include / import
+            referenced_file = None
+            for key, val in task.items():
+                if key.startswith('include_') or key.startswith('import_') or key == 'include':
+                    if isinstance(val, ansible.parsing.yaml.objects.AnsibleMapping):
+                        referenced_file = val.get('file', None)
+                    else:
+                        referenced_file = val
```
webknjaz: Is there a break missing here?
```suggestion
                        referenced_file = val
                    break
```

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+        for task in data.get('tasks', []):
+
+            # check if the id of the current rule is not in list of skipped rules for this play
+            if self.id in task['skipped_rules']:
+                continue
+
+            # collect information which file was referenced for include / import
+            referenced_file = None
+            for key, val in task.items():
+                if key.startswith('include_') or key.startswith('import_') or key == 'include':
```
webknjaz: Invert this check and `continue` early.

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,43 @@
...
+                    if isinstance(val, ansible.parsing.yaml.objects.AnsibleMapping):
+                        referenced_file = val.get('file', None)
+                    else:
+                        referenced_file = val
+
+            if referenced_file:
+                # make sure we have a absolute path here and check if it is a file
+                referenced_file = os.path.join(absolute_directory, referenced_file)
+                if not os.path.isfile(referenced_file):
```
webknjaz: make it flat
```suggestion
                if os.path.isfile(referenced_file):
                    continue
```

lib/ansiblelint/rules/IncludeMissingFileRule.py
jlusiardi: ok, i'll change that

lib/ansiblelint/rules/IncludeMissingFileRule.py
jlusiardi: will be fixed

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,37 @@
...
+            f_play.write(playbook_text)
+        return f_play
+
+    def test_include(self):
+        fh = self._get_play_file(PLAY_INCLUDE)
+        runner = Runner(self.rules, fh.name, [], [], [])
+        results = runner.run()
+        assert 'referenced missing file in' in str(results)
+        assert 'playbook.yml:2' in str(results)
+        assert 'some_file.yml' in str(results)
```
webknjaz: Here's a pytest based implementation:
```suggestion
def play_file_path(tmp_path):
    d = tmp_path / 'include-missing-file'
    d.mkdir()
    p = d / 'playbook.yml'
    p.write_text(PLAY_INCLUDE)
    return str(p)


def rules(play_file_path):
    rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
    return = RulesCollection([rulesdir])


def runner(play_file_path, rules):
    return Runner(rules, play_file_path, [], [], [])


def test_include(runner):
    results = str(runner.run())

    assert 'referenced missing file in' in results
    assert 'playbook.yml:2' in results
    assert 'some_file.yml' in results
```

You can accept it from the GitHub UI and then check if my in-browser coding doesn't have mistakes.

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,37 @@
+import os
+import unittest
+import tempfile
+import shutil
```
webknjaz: These would be unnecessary w/ pytest.

lib/ansiblelint/__init__.py
jlusiardi: fair enough. i did not know where this information was available for rules so i put it there.

lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -0,0 +1,63 @@
...
+            referenced_file = None
+            for key, val in task.items():
+                if not (key.startswith('include_') or
+                        key.startswith('import_') or
+                        key == 'include'):
+                    continue
+                if isinstance(val, ansible.parsing.yaml.objects.AnsibleMapping):
+                    referenced_file = val.get('file', None)
+                else:
+                    referenced_file = val
```
webknjaz: Still missing `break`

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
+from __future__ import print_function
```
webknjaz: This is unnecessary here.

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
...
+
+def test_include_missing_file(linter_rules, including_playbook):
+    runner = Runner(linter_rules, including_playbook.name, [], [], [])
+    results = runner.run()
+    assert 'referenced missing file in' in str(results)
+    assert 'playbook.yml:2' in str(results)
+    assert 'some_file.yml' in str(results)
+
+
+def test_include_found_file(linter_rules, including_playbook, included_playbook):
```
webknjaz: Why do you need two fixtures here? `including_playbook`/`including_playbook`

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
...
+@pytest.fixture
+def including_playbook_jinja2(tmp_path):
+    with open(os.path.join(str(tmp_path), 'playbook.yml'), 'w') as f_play:
+        f_play.write(PLAY_INCLUDING_JINJA2)
+    return f_play
+
+
+@pytest.fixture
+def included_playbook(tmp_path):
+    with open(os.path.join(str(tmp_path), 'some_file.yml'), 'w') as f_play:
```
webknjaz: Just use pytest-provided interfaces instead of this. Like in the snippet that I posted earlier.

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
...
+        f_play.write(PLAY_INCLUDED)
+    return f_play
```
webknjaz: No need for fh here if you only use the path.

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
...
+'''
+
+PLAY_INCLUDED = '''
+- debug:
+    msg: 'was found & included'
+'''
+
+
+@pytest.fixture
+def linter_rules(scope="module"):
```
webknjaz: It doesn't work like this. `scope` is an argument of the fixture decorator, not of the decorated function.
```suggestion
@pytest.fixture(scope='module')
def linter_rules():
```

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
...
+
+@pytest.fixture
+def included_playbook(tmp_path):
+    with open(os.path.join(str(tmp_path), 'some_file.yml'), 'w') as f_play:
+        f_play.write(PLAY_INCLUDED)
+    return f_play
+
+
+def test_include_missing_file(linter_rules, including_playbook):
+    runner = Runner(linter_rules, including_playbook.name, [], [], [])
```
webknjaz: the runner should be provided by a fixture. you can parameterize fixtures using `request` built-in fixture

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
...
+    results = runner.run()
+    assert 'referenced missing file in' in str(results)
+    assert 'playbook.yml:2' in str(results)
+    assert 'some_file.yml' in str(results)
```
webknjaz: you don't need to call `str()` multiple times. this should be optimized by calling in once.
```suggestion
    results = str(runner.run())
    assert 'referenced missing file in' in results
    assert 'playbook.yml:2' in results
    assert 'some_file.yml' in results
```

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,90 @@
+from __future__ import print_function
+
+import os
```
webknjaz:
```suggestion
import os
```

test/TestIncludeMissingFileRule.py
jlusiardi: :+1: will be fixed

test/TestIncludeMissingFileRule.py
jlusiardi: ok, sorry the naming might be suboptimal:
includ**ing**_playbook vs  includ**ed**_playbook

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,88 @@
...
+def test_include_noqa_working(linter_rules, including_playbook_noqa):
+    runner = Runner(linter_rules, including_playbook_noqa.name, [], [], [])
+    results = runner.run()
+    assert 'referenced missing file in' not in str(results)
+
+
+def test_include_omit_templated_references(linter_rules, including_playbook_jinja2):
+    runner = Runner(linter_rules, including_playbook_jinja2.name, [], [], [])
+    results = runner.run()
+    assert 'referenced missing file in' not in str(results)
```
webknjaz: These look like the same test mostly copy-pasted. This could be fixed by using `@pytest.mark.parametrize`

test/TestIncludeMissingFileRule.py
jlusiardi: sorry, I don't understand?

webknjaz: In every test you do `f_play.name` but you should `return f_play.name` here instead.

webknjaz: (really you should move the runner construction into its own fixture too)

docs/docsite/rst/rules/default_rules.rst
```diff
@@ -45,6 +45,7 @@ ID
...
+`E505 <https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/IncludeMissingFileRule.py>`_                                                                                                                                                                                                v4.3.0                                                                                                                                                                                                                                                                                                       referenced files must exist                                                                                                                                                                                                                                                                                  All files referenced by by include / import tasksmust exist. The check excludes files with jinja2 templates in the filename.
```
ssbarnea: `taskmust` missed a space. that file is generated but you need to sure is still ok.

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,85 @@
+import os
+import pytest
+
+from ansiblelint import Runner, RulesCollection
+
+
+class PlayFile:
+    def __init__(self, name, content):
+        self.name = name
+        self.content = content
```
webknjaz: you could just use tuples, having a class is not really necessary

test/TestIncludeMissingFileRule.py
```diff
@@ -0,0 +1,85 @@
+import os
+import pytest
```
webknjaz: pytest is a third-party lib
```suggestion

import pytest
```

test/TestIncludeMissingFileRule.py
webknjaz: or you could just use a namedtuple instead of creating a class manually
```suggestion
from collections import namedtuple
import os

import pytest

from ansiblelint import Runner, RulesCollection


PlayFile = namedtuple('PlayFile', ['name', 'content'])
```

PR #573, opened 2019-08-20
Normalize paths from config file deterministically

cans:
Ensure relative path listed in the configuration file are expanded relatively to the directory containing that file. Not relatively to the current working directory.

Closes #572

As I find very fast feedback loops during development to be really important, I allowed myself to do a bit of refactoring. Tests run time is substantially reduced (from >30s to ~12s on my machine). There are still similarly low hanging fruits to gain even more (e.g. in `test/TestCliRolePaths.py`). This is achieved by avoiding to use sub-processes  whenever possible. Another benefit is that it makes tests way easier to debug.

I am happy to discuss any possible improvements, splitting the PR in smaller chunks, _etc._

PR #692, opened 2020-03-14
Consistent relative path display

cans:
See discussion in PR #573 (this PR is built on top of the latter, only have a look at the last commit).

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -21,6 +21,7 @@
 # THE SOFTWARE.
 
 import errno
+import pathlib
 import sys
 
 import ansiblelint.formatters as formatters
@@ -29,18 +30,21 @@ from ansiblelint.utils import normpath
 
 
 def main():
+    cwd = pathlib.Path.cwd()
 
-    formatter = formatters.Formatter()
     options, args = cli.get_config(sys.argv[1:])
 
+    formatter_factory = formatters.Formatter
     if options.quiet:
-        formatter = formatters.QuietFormatter()
+        formatter_factory = formatters.QuietFormatter
 
     if options.parseable:
-        formatter = formatters.ParseableFormatter()
+        formatter_factory = formatters.ParseableFormatter
 
     if options.parseable_severity:
-        formatter = formatters.ParseableSeverityFormatter()
+        formatter_factory = formatters.ParseableSeverityFormatter
+
+    formatter = formatter_factory(cwd, options.display_relative_path)
 
     # no args triggers auto-detection mode
     if len(args) == 0 and not (options.listrules or options.listtags):
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -98,6 +98,9 @@ def get_cli_parser():
                            "rules directories specified with -r. There is "
                            "no need to specify this if no -r flags are used"
                            % ansiblelint.default_rulesdir)
+    parser.add_option('--show-relpath', dest='display_relative_path', action='store_false',
+                      default=True,
+                      help="Display path relative to CWD")
     parser.add_option('-t', dest='tags',
                       action='append',
                       default=[],
@@ -142,6 +145,10 @@ def merge_config(file_config, cli_config):
     if 'parseable_severity' in file_config:
         cli_config.parseable_severity = (cli_config.parseable_severity or
                                          file_config['parseable_severity'])
+
+    if 'display_relative_path' in file_config:
+        cli_config.display_relative_path = (cli_config.display_relative_path or
+                                            file_config['display_relative_path'])
 
     if 'use_default_rules' in file_config:
         cli_config.use_default_rules = (cli_config.use_default_rules or
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -1,12 +1,47 @@
 """Output formatters."""
+import os
+from pathlib import Path
+
 try:
     from ansible import color
 except ImportError:
     from ansible.utils import color
-from ansiblelint.utils import normpath
 
 
-class Formatter(object):
+class BaseFormatter:
+    """Formatter of ansible-lint output.
+
+    Base class for output formatters.
+
+    Args:
+        base_dir (str|Path): reference directory against which display relative path.
+        display_relative_path (bool): whether to show path as relatvie or absolute
+    """
+
+    def __init__(self, base_dir, display_relative_path):
+        if isinstance(base_dir, str):
+            base_dir = Path(base_dir)
+        if base_dir:  # can be None
+            base_dir = base_dir.absolute()
+
+        # Required 'cause os.path.relpath() does not accept Path before 3.6
+        if isinstance(base_dir, Path):
+            base_dir = str(base_dir)  # Drop when Python 3.5 is no longer supported
+
+        self._base_dir = base_dir if display_relative_path else None
+
+    def _format_path(self, path):
+        # Required 'cause os.path.relpath() does not accept Path before 3.6
+        if isinstance(path, Path):
+            path = str(path)  # Drop when Python 3.5 is no longer supported
+
+        if not self._base_dir:
+            return path
+        # Use os.path.relpath 'cause Path.relative_to() misbehaves
+        return os.path.relpath(path, start=self._base_dir)
+
+
+class Formatter(BaseFormatter):
 
     def format(self, match, colored=False):
         formatstr = u"{0} {1}\n{2}:{3}\n{4}\n"
@@ -14,7 +49,7 @@ class Formatter(object):
             color.ANSIBLE_COLOR = True
             return formatstr.format(color.stringc(u"[{0}]".format(match.rule.id), 'bright red'),
                                     color.stringc(match.message, 'red'),
-                                    color.stringc(normpath(match.filename), 'blue'),
+                                    color.stringc(self._format_path(match.filename), 'blue'),
                                     color.stringc(str(match.linenumber), 'cyan'),
                                     color.stringc(u"{0}".format(match.line), 'purple'))
         else:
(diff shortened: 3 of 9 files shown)


ssbarnea:
recheck

webknjaz:
recheck

cans:
@ssbarnea Branch rebased. There are still 1 that fails. An issue with tox not finding a the configparser module... I don't think it has anything to do with this.

BE CAREFUL: as mentioned above this branch is base on top of #573.

webknjaz:
#573's been merged.

webknjaz:
@cans this needs rebase

lib/ansiblelint/__main__.py
```diff
@@ -31,18 +32,19 @@
...
-        formatter = formatters.QuietFormatter()
+        formatter = formatters.QuietFormatter(cwd, options.relative_path)
 
     if options.parseable:
-        formatter = formatters.ParseableFormatter()
+        formatter = formatters.ParseableFormatter(cwd, options.relative_path)
 
     if options.parseable_severity:
-        formatter = formatters.ParseableSeverityFormatter()
+        formatter = formatters.ParseableSeverityFormatter(cwd, options.relative_path)
```
webknjaz, line 47: Maybe call it in one place?
```suggestion
    formatter_factory = formatters.Formatter

    if options.quiet:
        formatter_factory = formatters.QuietFormatter

    if options.parseable:
        formatter_factory = formatters.ParseableFormatter

    if options.parseable_severity:
        formatter_factory = formatters.ParseableSeverityFormatter

    formatter = formatter_factory(cwd, options.relative_path)
```

lib/ansiblelint/cli.py
```diff
@@ -98,6 +98,9 @@ def get_cli_parser():
                            "rules directories specified with -r. There is "
                            "no need to specify this if no -r flags are used"
                            % ansiblelint.default_rulesdir)
+    parser.add_option('-s', dest='relative_path', action='store_false',
```
webknjaz, line 101: Why `-s`? How does it relate to paths?

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,40 @@
...
+
 try:
     from ansible import color
 except ImportError:
     from ansible.utils import color
-from ansiblelint.utils import normpath
 
 
-class Formatter(object):
+class BaseFormatter(object):
```
webknjaz, line 10: Now that we dropped Python 2, no need to explicitly make classes new-style.
```suggestion
class BaseFormatter:
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,40 @@
...
+    """Formats ansible lint output.
```
webknjaz, line 11: Class docstring shouldn't start with a verb. Classes don't do actions, unlike methods.
```suggestion
    """Formatter of ansible-lint output.
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,40 @@
...
+        relative_path (bool): whether to show path as relatvie or absolute
+    """
+
+    def __init__(self, base_dir, relative_path):
+        self._base_dir = base_dir if relative_path else None
+
+    def normpath(self, path):
+        if self._base_dir:
+            return os.path.relpath(path, start=self._base_dir)
+        return path
```
webknjaz, line 26: Maybe we could have shorter lines
```suggestion
        if not self._base_dir:
            return path
        return os.path.relpath(path, start=self._base_dir)
```

lib/ansiblelint/cli.py
cans: No clue... did I cut paste something ? :thinking: 
Honestly, I don't really have an opinion on this so if you want to pick on or only use a long option name... please be my guest, choose something.

webknjaz: @ssbarnea do you think this should be configurable?

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,40 @@
...
 
 
-class Formatter(object):
+class BaseFormatter(object):
+    """Formats ansible lint output.
+
+    Base class for output formatters.
+
+    Args:
+        base_dir: reference directory against which display relative path.
```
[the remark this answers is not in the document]
cans: Do you use `pathlib` ? Should this support `Path` objects too ?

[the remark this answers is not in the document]
webknjaz: I guess, but it's not specified in this line.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,40 @@
...
+        relative_path (bool): whether to show path as relatvie or absolute
+    """
+
+    def __init__(self, base_dir, relative_path):
```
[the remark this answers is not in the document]
cans: I am gonna go for `display_relative_path` (same verb as used in the docstring)

lib/ansiblelint/cli.py
cans: Most options changing the output have long names: `--nocolor`, `--force-color`. We could go from something like `--show-relpath` (changed, but still open to suggestion).

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,46 @@
...
+        display_relative_path (bool): whether to show path as relatvie or absolute
+    """
+
+    def __init__(self, base_dir, display_relative_path):
+        if isinstance(base_dir, Path):
+            base_dir = base_dir.absolute()
+
+        self._base_dir = base_dir if display_relative_path else None
+
+    def _format_path(self, path):
```
cans, line 27: Renamed 'cause `normpath` (in `os.path`) as a defined meaning and does almost opposite to what we do here. And I find that confusing [like this is](https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/utils.py#L720)

test/TestCommandLineInvocationSameAsConfig.py
```diff
@@ -22,6 +22,8 @@ def base_arguments():
                           "test/fixtures/skip-tags.yml"),
                          (["--exclude", "test/"],
                           "test/fixtures/exclude-paths.yml"),
+                         (["--show-relpath"],
+                          "test/fixtures/show-relpath.yml"),
```
cans, line 28: Seen your [comment](https://github.com/ansible/ansible-lint/pull/573#discussion_r402256100) but I am not sure how to not make a mess of this, while placing the files content in the code.

webknjaz: Well, this https://github.com/ansible/ansible-lint/blob/master/test/TestIncludeMissingFileRule.py#L9-L16 looks like a good solution to me. What kind of mess are you afraid of?

lib/ansiblelint/formatters/__init__.py
webknjaz: good point

lib/ansiblelint/cli.py
webknjaz: This is probably fine

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,46 @@
...
+
+    def __init__(self, base_dir, display_relative_path):
+        if isinstance(base_dir, Path):
+            base_dir = str(base_dir.absolute())
+
+        self._base_dir = base_dir if display_relative_path else None
+
+    def _format_path(self, path):
+        if isinstance(path, Path):
+            path = str(path)
```
webknjaz, line 29: I don't think we need to do the conversion. It should be done when producing the output and will probably happen automatically anyway.

webknjaz: I think we should do the opposite and use pathlib representation wherever possible instead.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,20 +1,46 @@
...
+        if not self._base_dir:
+            return path
+        return os.path.relpath(path, start=self._base_dir)
```
webknjaz, line 32: I'd rather stick to pathlib which also allows to avoid needing `os.path`:
```suggestion
        return path.relative_to(self._base_dir)
```

test/TestBaseFormatter.py
```diff
@@ -0,0 +1,47 @@
...
+    # Given
+    base_formatter = BaseFormatter(base_dir, True)
+
+    # When
+    output_path = base_formatter._format_path(path)
+
+    # Then
+    assert isinstance(output_path, str)
+    assert isinstance(base_formatter._base_dir, str)
+    assert output_path == os.path.basename(str(path))
```
webknjaz, line 44: This way we can drop `import os`:
```suggestion
    assert output_path == Path(path).name
```

test/TestFormatter.py
```diff
@@ -29,7 +30,7 @@ class TestFormatter(unittest.TestCase):
     def setUp(self):
         self.rule = AnsibleLintRule()
         self.rule.id = "TCF0001"
-        self.formatter = Formatter()
+        self.formatter = Formatter(os.getcwd(), True)
```
webknjaz, line 33: How about we get rid of `os`?
```suggestion
        self.formatter = Formatter(pathlib.Path.cwd(), True)
```

test/TestFormatter.py
```diff
@@ -17,6 +17,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
+import os
```
webknjaz, line 20:
```suggestion
import pathlib
```

lib/ansiblelint/__main__.py
```diff
@@ -31,18 +32,21 @@
 
 
 def main():
+    cwd = os.getcwd()
```
webknjaz, line 35: same suggestion here
```suggestion
    cwd = pathlib.Path.cwd()
```

lib/ansiblelint/formatters/__init__.py
webknjaz: Agreed.

> The reason for the cast from path to string is that os.path.relpath() does not accepts Path in all supported version of Python.

I assume you're talking about Python 3.5, right?

PR #694, opened 2020-03-15
Improve linting with generic checks

ssbarnea:
- Enables several generic checks and fixes.
- Removed BOM from 3 files as is not really needed
- Included fixes made by pre-commit itself regarding whitespace
- Assures that git tracked generated doc file pass linters and prevent regression


ssbarnea:
recheck

.pre-commit-config.yaml
```diff
@@ -1,5 +1,17 @@
 ---
 repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks.git
+  rev: v2.4.0
+  hooks:
+  - id: end-of-file-fixer
+  - id: trailing-whitespace
+    exclude: (test/with-skip-tag-id.yml|examples/example.yml|test/unicode.yml)
```
webknjaz: Let's multiline it for better maintainability:
```suggestion
    exclude: |
      ^x
      (
          test/(with-skip-tag-id|unicode).yml|
          examples/example.yml
      )
```

docs/docsite/Makefile
```diff
@@ -37,6 +37,7 @@ generate_rst: staticmin
 htmldocs: generate_rst
 	cd ../../lib/ansiblelint; python ./generate_docs.py
 	CPUS=$(CPUS) $(MAKE) -f Makefile.sphinx html
+	@sh -c "git diff --quiet || { echo 'ERROR: Building docs changed tracked files, include them in commit to avoid this failure.'; exit 1; }"
```
webknjaz: Honestly, I'd rather get rid of the makefile and run things directly via tox.

lib/ansiblelint/generate_docs.py
```diff
@@ -95,6 +95,8 @@ def make_table(grid):
             block = block + num_cols * ((cell_width) * '=' + ' ') + '\n'
         header = False
     block = block + num_cols * ((cell_width) * '=' + ' ') + '\n'
+    # remove trailing whitelines from block
+    block = '\n'.join([line.rstrip() for line in block.splitlines()]) + '\n'
```
webknjaz: no need for an extra in-memory list:
```suggestion
    block = '\n'.join(line.rstrip() for line in block.splitlines()) + '\n'
```

docs/docsite/Makefile
ssbarnea: That worth considering but not part of this chance. Lets try to improve it one thing at a time.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,21 @@
 ---
 repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks.git
+  rev: v2.4.0
+  hooks:
+  - id: end-of-file-fixer
+  - id: trailing-whitespace
+    exclude: |
+      (?x)(
+        ^test/(with-skip-tag-id|unicode).yml|
```
webknjaz: @ssbarnea is it possible to move this caret outside of the braces and not repeat it on every line? It's really not necessary here.

webknjaz: In fact, the outer braces aren't necessary w/o it

webknjaz: Here's a better example: https://pre-commit.com/#regular-expressions

.pre-commit-config.yaml
```diff
@@ -1,5 +1,21 @@
...
+        ^examples/example.yml
+      )
```
webknjaz:
```suggestion
    exclude: >
      (?x)^(
        test/(with-skip-tag-id|unicode).yml|
        examples/example.yml
      )$
```

ssbarnea: I am really not confident that his would work correctly based on how ">" yaml joining is working. I know for sure that the proposed one works well. Why to waste another 2-3h on a variation?

PR #695, opened 2020-03-16
POC: verify tox upgrading on zuul

ssbarnea:
Depends-On: https://github.com/ansible/project-config/pull/400

### DIFF
--- a/tools/test-setup.sh
+++ b/tools/test-setup.sh
@@ -6,6 +6,15 @@ set -Eeuxo pipefail
 # We are not using pip --user due to few bugs in tox role which does not allow
 # us to override how is called. Once these are addressed we will switch back
 # non-sudo mode.
+sudo yum remove -y python-tox || true
+
 PYTHON=$(command -v python3 python | head -n1)
 
-sudo "$PYTHON" -m pip install -U tox "zipp<0.6.0;python_version=='2.7'"
+$PYTHON -m pip install --user "pip>=20.0.2"
+$PYTHON -m pip install --user "virtualenv>=20.0.10"
+$PYTHON -m pip install --user "tox>=3.8.0" "tox-venv" "zipp<0.6.0;python_version=='2.7'"
+
+# testing
+$PYTHON -m virtualenv --version
+$PYTHON -m tox --version
+tox --version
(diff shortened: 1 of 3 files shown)


tools/test-setup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -Eeuxo pipefail
+set -euxo pipefail
```
webknjaz: I'd rather not remove `-E`

tools/test-setup.sh
```diff
@@ -8,4 +8,8 @@ set -Eeuxo pipefail
 # non-sudo mode.
 PYTHON=$(command -v python3 python | head -n1)
 
-sudo "$PYTHON" -m pip install -U tox "zipp<0.6.0;python_version=='2.7'"
+# $PYTHON -m pip install --user "tox>=3.8.0" "zipp<0.6.0;python_version=='2.7'"
```
webknjaz: Upgrade virtualenv before doing anything else. Use a separate `pip install` command.

tox.ini
```diff
@@ -5,6 +5,7 @@ isolated_build = true
 requires =
   setuptools >= 41.4.0
   pip >= 19.3.0
+  six
```
webknjaz: I don't think it's needed.

tools/test-setup.sh
```diff
@@ -8,4 +8,10 @@ set -Eeuxo pipefail
 # non-sudo mode.
 PYTHON=$(command -v python3 python | head -n1)
 
-sudo "$PYTHON" -m pip install -U tox "zipp<0.6.0;python_version=='2.7'"
+$PYTHON -m pip install --user "virtualenv>=20.0.10"
+$PYTHON -m pip install --user "tox>=3.8.0" "zipp<0.6.0;python_version=='2.7'"
```
webknjaz: why is zipp needed?

ssbarnea: There is a bug in older versions of pip where it installs a version incompatible with py27.

BTW, this PR is not ready for review yet, I am using it to test various things. I will remove draft status when I consider it ready.

ssbarnea:
recheck

Issue #683, opened 2020-02-24
ImportError: No module named six


ssbarnea:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
4.2.0
```

# Desired Behaviour

Ansible-lint must assure it installs six all the time.

# Actual Behaviour (Bug report only)
Full log https://logserver.rdoproject.org/46/25246/13/check/tox-linters/2a3ccec/job-output.txt

```
2020-02-24 09:13:28.954259 | rdo-centos-7 |       Installing collected packages: setuptools, setuptools-scm, setuptools-scm-git-archive, wheel, MarkupSafe, jinja2, PyYAML, six, ipaddress, pycparser, cffi, enum34, cryptography, ansible, scandir, pathlib2, ruamel.ordereddict, ruamel.yaml.clib, ruamel.yaml
2020-02-24 09:13:28.954327 | rdo-centos-7 |       ERROR: Exception:
2020-02-24 09:13:28.954395 | rdo-centos-7 |       Traceback (most recent call last):
2020-02-24 09:13:28.954608 | rdo-centos-7 |         File "/home/zuul/.cache/pre-commit/reporf1mbW/py_env-python2.7/lib/python2.7/site-packages/pip/_internal/cli/base_command.py", line 186, in _main
2020-02-24 09:13:28.954701 | rdo-centos-7 |           status = self.run(options, args)
2020-02-24 09:13:28.954867 | rdo-centos-7 |         File "/home/zuul/.cache/pre-commit/reporf1mbW/py_env-python2.7/lib/python2.7/site-packages/pip/_internal/commands/install.py", line 404, in run
2020-02-24 09:13:28.954931 | rdo-centos-7 |           use_user_site=options.use_user_site,
2020-02-24 09:13:28.955078 | rdo-centos-7 |         File "/home/zuul/.cache/pre-commit/reporf1mbW/py_env-python2.7/lib/python2.7/site-packages/pip/_internal/req/__init__.py", line 71, in install_given_reqs
2020-02-24 09:13:28.955125 | rdo-centos-7 |           **kwargs
2020-02-24 09:13:28.955279 | rdo-centos-7 |         File "/home/zuul/.cache/pre-commit/reporf1mbW/py_env-python2.7/lib/python2.7/site-packages/pip/_internal/req/req_install.py", line 787, in install
2020-02-24 09:13:28.955330 | rdo-centos-7 |           prefix=prefix,
[...]

PR #697, opened 2020-03-21
Fix package installation

ssbarnea:
Afters many experiments, I was able to narrow down what caused the multiple installation failures on Zuul CI nodepool images, it was ansible-lint "unique" packaging. This change implements the smallest change needed in order to address this issue.

The root-cause of this is the use of `version` option inside `setup.cfg`, which was also a mitigating factor in introduction of unreasonable list of dependencies inside `build-system` section of `pyproject.toml`, where even ansible was listed. We should all agree that building the distribution should have never asked for something like this. 

By relying on setuptools-scm versioning, which is  extremely popular, we remove all these problems, hugely improve installation speed and, as bonus, we also reinstate the ability to install the package as editable.


ssbarnea:
@webknjaz @awcrosby Please review this as this is a long-standing issue that also bloked adoption of Zuul CI jobs.

ssbarnea:
By looking at build history I find very hard to magine, that multiple nodepool test images (centos-7 and centos-8 at least) are affected by such claimed miconfiguration that happens to break only ansible-lint project, all the other ones not happening to be affected.
[zuul-screenshot](https://dashboard.zuul.ansible.com/t/ansible/builds?project=ansible%2Fansible-lint#)

The fact that the entire ansible package happens to be required during the package building and that is installed **twice** each time is not an issue on how the packaging was configured with this project but instead is caused by external factors>

I would put it in different view: one of primary goal of maintaining a package is to assure it can successfully install on as many scenarios (platforms, configuration) as users may encounter in production.  This means that a package should be installable using a various range of virtualenv/pip versions and especially tested with versions shipped with supported operating systems and also preferably with the latest releases, so we harden the installation.

webknjaz:
If there's no reproducer, there's nothing to fix. It's https://github.com/ansible/ansible-lint/issues/683#issuecomment-594764105 unless proven otherwise. Band-aids only make things worse in the long run.

Besides, the main supported case is installing from wheels. Other cases are non-essential.

pyproject.toml
```diff
@@ -1,16 +1,9 @@
 [build-system]
 requires = [
+  "pip >= 19.3.1",
```
webknjaz: It looks like you're trying to rely on some weird side-effect, which is wrong.

ssbarnea:
I saw lots of statements referring to what is supposed to be supported or not which do not have any backing, like links to documentation, mailing-list or irc logs. Sorry to say it, but they look more like personal-preferences being enforced to others.

Ansible project itself has a `setup.py` file, same with galaxy project, and they are installable as editable. `setup.py` file was removed in a [change](https://github.com/ansible/ansible-lint/commit/7fa1bde06ea21e451d61bbdf8c6a9eded0edc066) that was merged without any peer-review. While we can all understand that there are cases where we are unable to perform reviews, that was not really the case with ansible-lint, which is a popular project, especially with high-risk  changes.

IMHO most of the problems started with the forced introduction of the experimental PEP-518 and when its issues started to show a deliberate decision to drop support for everything that looked old was made. The reality is that python packaging is hard and requires time to make it work, but that should not be used as an excuse for testing experimental feature at users expense. I am not aware of any other major or minor python package that dropped `setup.py` completely.

In fact a previous attempt to address that was made on https://github.com/ansible/ansible-lint/pull/654 and was backed by not less than 3 others with valid comments. Then people came with additional arguments and all we got was a two months silence. 

Now, when we have extra reasons to undo the past damage as we do have CI jobs as proofs of the failures, we get a very quick dismissal of a PR (2h).

webknjaz:
> I saw lots of statements referring to what is supposed to be supported or not which do not have any backing, like links to documentation, mailing-list or irc logs

Supporting only wheels-based install was discussed with @awcrosby and also agreed @ #ansible-galaxy on Freenode at some point, you were okay with this too back then but this wasn't in a recorded meeting.


> Ansible project itself has a `setup.py` file

And it also supports Python 2.6. The reason for `setup.py` there is also to have some voodoo magic with symlinks. But many agree if that wasn't the case and we'd be able to ship wheels, it'd better to not execute arbitrary code during install time which is bad.

> and they are installable as editable.

This is not a reason to support editable installs.



> I am not aware of any other major or minor python package that dropped `setup.py` completely.

Let's stay reasonable. Other packages are not the metric. In fact, a lot of popular projects make horrible decisions.

> Then people came with additional arguments

In particular, agreeing that the maintenance burden is not worth it.



> extra reasons to undo the past damage as we do have CI jobs as proofs of the failures

I have proof that all other CIs work if configured properly. Those failing jobs clearly are misconfigured and trying to enforce an unsupported case is not a reason to shadow real problems and make the future debugging harder.

PR #518, opened 2019-05-23
Support to_nice_yaml filter

kouk:
`ansiblelint.utils.template` can't handle the `to_nice_yaml` filter, so we should add the `RepresenterError` to the ignored exception list.

Example task:
```yaml
- name: output some nice yaml
  debug: msg="{{ some_var | to_nice_yaml }}"
```
Causes this error:

```python-traceback
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansiblelint/utils.py", line 224, in play_children
    fail_on_undefined=False)
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansiblelint/utils.py", line 191, in template
    **dict(kwargs, fail_on_undefined=fail_on_undefined))
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansiblelint/utils.py", line 78, in ansible_template
    return templar.template(varname, **kwargs)
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansible/template/__init__.py", line 572, in template
    ) for v in variable]
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansible/template/__init__.py", line 572, in <listcomp>
    ) for v in variable]
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansible/template/__init__.py", line 584, in template
    disable_lookups=disable_lookups,
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansible/template/__init__.py", line 584, in template
    disable_lookups=disable_lookups,
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansible/template/__init__.py", line 539, in template
    disable_lookups=disable_lookups,
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansible/template/__init__.py", line 804, in do_template
    res = j2_concat(rf)
  File "<template>", line 12, in root
  File "/root/.cache/pre-commit/repogevfadgi/py_env-python3.6/lib/python3.6/site-packages/ansible/plugins/filter/core.py", line 71, in to_nice_yaml
[...]

PR #701, opened 2020-03-23
Revert "Support to_nice_yaml filter"

webknjaz:
Reverts ansible/ansible-lint#518 _(this is to verify if that PR causes the current macOS failures or they are related to some env upgrades)_


webknjaz:
(no, it's not what's broken the CI)

Issue #705, opened 2020-03-27
become_user inside block does not inherit become applied to block, triggers 501


m-a-leclercq:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version : ansible 2.9.6
ansible-lint --version : 4.2.0
```

- ansible installation method: pip
- ansible-lint installation method: pip

# Desired Behaviour
No error.

# Actual Behaviour (Bug report only)
Rule 501 triggered

Please give some details of what is actually happening.
Include a [minimum complete verifiable example](http://stackoverflow.com/help/mcve)
with:

```
- name: create user with group, schedule cron and do a one time run of said cron
  become: true
  block:
    - name: create user and group
      import_tasks: system.yml

    - name: create .sh file and schedule cron
      import_tasks: config.yml

    - name: run .sh once
      become_user: my_newly_created_user
      command: "{{ my_shell_path }}/myshell.sh"
```

webknjaz:
@m-a-leclercq please submit a PR with a test showing the issue.

m-a-leclercq:
@webknjaz I'm behind a corporate proxy which makes it difficult to clone locally from github. I was however able to push to github a working example : https://github.com/m-a-leclercq/linter-test

webknjaz:
We need to have this exactly in our tests so that it'd be run in our CI and prevent regressions.

PR #706, opened 2020-03-27
become_user inside block does not inherit become applied to block, triggers 501

m-a-leclercq:
Shows #705

### DIFF
--- a/test/role-block/tasks/main.yml
+++ b/test/role-block/tasks/main.yml
@@ -31,4 +31,4 @@
 
     - name: run .sh once
       become_user: my_newly_created_user
-      command: "/etc/myshell.sh"
+      command: "/etc/test.sh"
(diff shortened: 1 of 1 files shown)


test/role-block/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
+---
```
webknjaz: This needs to be integrated into tests. Meaning there should be a test that runs this and then checks that there's no errors of a specific type.

m-a-leclercq: I'm a bit lost as to what needs to be done, I've read `contributing.md` but that doesn't really tell me much as to how I'm supposed to create my test and push my code. Any documentation I migh have missed that would help me? 
Thanks

webknjaz: Sorry, I missed your comment. @ssbarnea has created a reproducer test @ #793. Also, I've added a few references on adding pytest based tests @ https://github.com/ansible/ansible-lint/issues/725.

ssbarnea: @webknjaz You can close this one now.

PR #709, opened 2020-04-01
Avoid exception from 505 rule

ssbarnea:
This should fix exception when absolute_directory is None, like I seen on current master branch:
```
  File "/Users/ssbarnea/.cache/pre-commit/repo71i9evtb/py_env-python3/bin/ansible-lint", line 8, in <module>
    sys.exit(main())
  File "/Users/ssbarnea/.cache/pre-commit/repo71i9evtb/py_env-python3/lib/python3.7/site-packages/ansiblelint/__main__.py", line 185, in main
    matches.extend(runner.run())
  File "/Users/ssbarnea/.cache/pre-commit/repo71i9evtb/py_env-python3/lib/python3.7/site-packages/ansiblelint/__init__.py", line 290, in run
    skip_list=self.skip_list))
  File "/Users/ssbarnea/.cache/pre-commit/repo71i9evtb/py_env-python3/lib/python3.7/site-packages/ansiblelint/__init__.py", line 183, in run
    matches.extend(rule.matchyaml(playbookfile, text))
  File "/Users/ssbarnea/.cache/pre-commit/repo71i9evtb/py_env-python3/lib/python3.7/site-packages/ansiblelint/__init__.py", line 125, in matchyaml
    result = self.matchplay(file, play)
  File "/Users/ssbarnea/.cache/pre-commit/repo71i9evtb/py_env-python3/lib/python3.7/site-packages/ansiblelint/rules/IncludeMissingFileRule.py", line 53, in matchplay
    referenced_file = os.path.join(absolute_directory, referenced_file)
  File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/posixpath.py", line 80, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

### DIFF
--- a/lib/ansiblelint/rules/IncludeMissingFileRule.py
+++ b/lib/ansiblelint/rules/IncludeMissingFileRule.py
@@ -46,7 +46,7 @@ class IncludeMissingFileRule(AnsibleLintRule):
                 if referenced_file:
                     break
 
-            if referenced_file is None:
+            if referenced_file is None or absolute_directory is None:
                 continue
 
             # make sure we have a absolute path here and check if it is a file


lib/ansiblelint/rules/IncludeMissingFileRule.py
```diff
@@ -46,7 +46,7 @@ def matchplay(self, file, data):
                 if referenced_file:
                     break
 
-            if referenced_file is None:
+            if referenced_file is None or absolute_directory is None:
```
webknjaz, line 49: This looks like a band-aid. Maybe we should use a relative dir in this case instead?

webknjaz: @jlusiardi ^

webknjaz: I've merged this PR but I'm pretty sure that this change https://github.com/ansible/ansible-lint/pull/691/files#diff-3a0d2814ee14f24f804a7c3bd1c7863aR266 needs to be applied to other places with `files.append()` in that method.

ssbarnea: @webknjaz I can rephrase it if you want: I am ok to revert this rule until it gets better testing. I do not think is fair to ask me to fix it, especially as doing any development on ansible-lint is a PITA due to the lack of editable install.

ssbarnea:
Thanks. I will work on adding some extra testing, so we lower regression risks.

webknjaz:
@ssbarnea also this will need to be reverted later with the fix that I suggested in the comments above.

PR #710, opened 2020-04-01
Remove unused zuul config files

ssbarnea:
Currently, ansible-lint is not using Zuul gating so the configuration
is not loaded from within. These files where not used by the current
third-party-check implementation removing them is better in order to
avoid confusing users.


webknjaz:
> proved to be reliable

Reliable how?

ssbarnea:
recheck

ssbarnea:
@pabelanger I guess that using depends-on https://github.com/ansible/project-config/pull/400 does not work due unsecure-repo status and we will have to merge 400 and its dependency first, before being able to test this.

ssbarnea:
@pabelanger Can you confirm that this is the right action as long we are using `third-party-check` integration?

Issue #699, opened 2020-03-21
proposal: Adopt and document smart CI-testing matrix


ssbarnea:
At this moment ansible-lint is testing using 3 different CI/CD systems with a huge number of jobs (over 100) while most of them are very-unlikely to cover real-world use cases, being only the results of an exploding all permutations found so far. 

This not only produces a considerable resource loading and delays on getting results, but also imposed additional maintenance costs with managing these jobs.

This proposal is about cherry-picking only these configurations that are most likely to be used by the users. With a more careful picking of combinations we can even provide a better practical coverage with far smaller number of jobs. For example, it can prove far more useful to test ansiblelint with default version of python shipped with multiple operating systems instead of trying to test all python versions on the same operating system, especially as some of these may not be even supported by the operating system vendor.

Variables in testing matrix:
* platforms: `centos-8`, `fedora-stable`, `ubuntu-bionic`, `macos-latest`
* python: `py36`, `py37`, maybe even `py38` as experimental
* ansible versions: `2.9`, `2.8`, `devel`
* jobs: simple like `linters`, `docs` or complex like unit/functional with the main difference being that the simple ones are used for development, and they do not need to have the same requirements.

Proposed matrix for simple jobs (2x2=4):
* linters + docs tested only with centos-8 and py36
* linters + docs tested only with and macOS and py36

Proposed matrix for complex jobs (5x3=15)
* centos-8-py36-ansible2x
* ubuntu-xenial-py36-ansible2x
* fedora-stable-py37-ansible-2x
* macos-latest-py36-ansible-2x

This list is missing py38 and we may want to make few adjustments to it but the entire idea is that we should try to test only combinations that are high-likely to happen in practice, or those that are going to be popular. The list is not set in stone, and we would be able to adjust it based on needs, even temporary enabling new jobs if we discover that this is needed.

There are also few other aspects related to this which are currently not covered and maybe we should cover for them: testing the installation of the package: to system, to virtualenv, to user site packages, installation as editable, installation from git, sdist or wheel, installation with various versions of python packaging tools lower/upper constraints. The reality is that these are known to caused lots of problems, but we do not really test them. Obviously that it would make no sense to create full jobs for each of these variations, but we could easily find ways to perform basic sanity checks of these within the other jobs that run on the same platforms. 

@awcrosby @gundalow @webknjaz I am looking forward to receive feedback on this proposal

[...]

PR #714, opened 2020-04-02
Remove six and pathlib2 dependencies

ssbarnea:
As we removed support for py27, we can now drop two dependencies that where used in order to provide support for it.

Partial-Fix: #699

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -24,8 +24,6 @@ import os
 import re
 import sys
 
-import six
-
 import ansiblelint.utils
 import codecs
 
@@ -66,7 +64,7 @@ class AnsibleLintRule(object):
             if not result:
                 continue
             message = None
-            if isinstance(result, six.string_types):
+            if isinstance(result, str):
                 message = result
             matches.append(Match(prev_line_no + 1, line,
                            file['path'], self, message))
@@ -97,7 +95,7 @@ class AnsibleLintRule(object):
                 continue
 
             message = None
-            if isinstance(result, six.string_types):
+            if isinstance(result, str):
                 message = result
             task_msg = "Task/Handler: " + ansiblelint.utils.task_to_str(task)
             matches.append(Match(task[ansiblelint.utils.LINE_NUMBER_KEY], task_msg,
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -23,13 +23,10 @@ import glob
 import imp
 from itertools import product
 import os
-try:
-    from pathlib import Path
-except ImportError:
-    from pathlib2 import Path
+from pathlib import Path
+
 import subprocess
 
-import six
 import yaml
 from yaml.composer import Composer
 from yaml.representer import RepresenterError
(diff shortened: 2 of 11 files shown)


docs/docsite/rst/conf.py
```diff
@@ -228,7 +228,7 @@
 
 autoclass_content = 'both'
 
-intersphinx_mapping = {'python': ('https://docs.python.org/2/', (None, '../python2-2.7.13.inv')),
+intersphinx_mapping = {'python': ('https://docs.python.org/3/', (None, '../python3-3.6.2.inv')),
```
webknjaz, line 231: I'm not sure if we should do this now. The docs may as well still reference things from py2. Let's omit this.

lib/ansiblelint/__init__.py
```diff
@@ -66,7 +64,7 @@ def matchlines(self, file, text):
             if not result:
                 continue
             message = None
-            if isinstance(result, six.string_types):
+            if isinstance(result, str):
```
webknjaz, line 67: Let's revert all the modifications to `six.string_types` for now as what you're doing effectively modifies logic. Before all checks were "bytes or unicode" not it's just "unicode" which may lead to unforeseen behavior changes. So I want to see less of dangerous changes in this PR.

lib/ansiblelint/utils.py
```diff
@@ -23,13 +23,10 @@
 import imp
 from itertools import product
 import os
-try:
-    from pathlib import Path
-except ImportError:
-    from pathlib2 import Path
+from pathlib import Path
```
webknjaz, line 26: LGTM

zuul.d/layout.yaml
```diff
@@ -1,40 +0,0 @@
----
```
webknjaz, line 1: What does this have to do with Python 2 removal?

lib/ansiblelint/__init__.py
ssbarnea: I really doubt there is any change in behavior, take a look at https://six.readthedocs.io/#six.string_types -- I did this switch on other projects.

As long we run on python3 there is no "unicode" anymore. The only case where it would be a change in behavior is if we endup running on py2, which we do not allow or support.

zuul.d/layout.yaml
ssbarnea: In fact the entire zuul.d/ needs to removed as is not used by zuul.

lib/ansiblelint/__init__.py
webknjaz: Okay, keep it then.

zuul.d/layout.yaml
webknjaz: Remove it separately then

ssbarnea:
I will stage it, first step is at https://github.com/ansible/ansible-lint/pull/715 which removes the jobs. I will also do the other changes but separated by subject, making easier to review.

webknjaz:
Great, thanks! This usually makes time to merge smaller for non-controversial changes in general and fixing one thing doesn't affect merging other things.

Issue #698, opened 2020-03-21
Reduce support matrix to last two versions of ansible


ssbarnea:
At this moment ansible-lint is tested using 4 versions of ansible: N-2, N-1, N and N+1 (devel) and this puts a serious strain on CI and maintenance.

Few months back molecule **dropped support for N-2 version of ansible** after we did a survey, explaining users why we are doing it. So far we had no problem with this approach, not even a single complaint.

This is why I propose making the same change and document it, so users would not be surprised. 
The goal is to follow Ansible releases as soon we can after a release. 

Testing devel branch has a special nature as these jobs do not perform an exhaustive testing and their failure should not prevent a merge. This allows us to prepare for future changes without being forced to fix bugs after the release, also allows us to give feedback to ansible team about a possible breaking changes before the release is made.

I raised this ticked based on a discussion I had on #ansible-galaxy channel but I hope to get feedback (thumbs up/down) about it before creating a PR to implement the change.

ssbarnea:
@awcrosby @webknjaz ^

webknjaz:
It's okay to do so

PR #721, opened 2020-04-06
Switch support to last two stable version of Ansible

ssbarnea:
Drops support for ansible 2.7 and documents the support matrix.

Fixes: #698


ssbarnea:
@webknjaz Doe this looks ok? zuul failure is unrelated, I will investigate with @pabelanger as it appears we now have an older version of tox installed on ubuntu, one that is not able to upgrade itself when needed.

PR #724, opened 2020-04-07
POC: Enable packaging test using pre-commit

ssbarnea:
Adds pre-commit hooks that validates python packaging.

That is more of a POC/WIP because I am not yet sure that pre-commit is the best place to test packaging, mainly because it adds ~40s, which is a lot for linting.

I am open to suggestions regarding when to run and what/how to test. I already suspect that we may need to perform two rounds of testing, one with minimal version of packaging tools supported and one with latest-versions, so we catch regressions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,6 +18,26 @@ repos:
     language: system
     files: >-
       ^setup\.py$
+  - id: check-python-packaging
+    name: Check python packaging
+    description: |
+        Performs validations of python packaging using both PEP-517 and old
+        setup.py while also validating metadata. Also verifies that the package
+        can be insalled in editable mode.
+    entry: >
+      bash -c "
+      rm -rf $VIRTUAL_ENV/dist.old $VIRTUAL_ENV/dist.new &&
+      python setup.py sdist bdist_wheel --dist-dir $VIRTUAL_ENV/dist.old &&
+      python -m twine check $VIRTUAL_ENV/dist.old/* &&
+      python -m pep517.build --source --binary --out-dir $VIRTUAL_ENV/dist.new . &&
+      python -m twine check $VIRTUAL_ENV/dist.new/* &&
+      python -m pip install -e ."
+    language: python
+    pass_filenames: false
+    always_run: true
+    additional_dependencies:
+      - pep517>=0.8.2
+      - twine>=3.1.1
 - repo: https://github.com/pre-commit/pre-commit-hooks.git
   rev: v2.4.0
   hooks:


.pre-commit-config.yaml
```diff
@@ -18,6 +18,26 @@ repos:
...
       ^setup\.py$
+  - id: check-python-packaging
+    name: Check python packaging
+    description: |
+        Performs validations of python packaging using both PEP-517 and old
+        setup.py while also validating metadata. Also verifies that the package
+        can be insalled in editable mode.
+    entry: >
+      bash -c "
+      rm -rf $VIRTUAL_ENV/dist.old $VIRTUAL_ENV/dist.new &&
```
webknjaz, line 29: What's this `$VIRTUAL_ENV` env var? Also, maybe you could use `mktemp -d` as a temporary location for the test env?

.pre-commit-config.yaml
```diff
@@ -18,6 +18,26 @@ repos:
...
+      python setup.py sdist bdist_wheel --dist-dir $VIRTUAL_ENV/dist.old &&
```
webknjaz, line 30: This isn't supposed to be used directly. Use it through PEP517. Also, it's unnecessary given that it's already being tested in the CI anyway.

.pre-commit-config.yaml
```diff
@@ -18,6 +18,26 @@ repos:
...
+      python -m twine check $VIRTUAL_ENV/dist.old/* &&
+      python -m pep517.build --source --binary --out-dir $VIRTUAL_ENV/dist.new . &&
+      python -m twine check $VIRTUAL_ENV/dist.new/* &&
+      python -m pip install -e ."
+    language: python
+    pass_filenames: false
+    always_run: true
```
webknjaz, line 37: I don't think it's necessary. It should only run on changes to known packaging-related files.

.pre-commit-config.yaml
```diff
@@ -18,6 +18,26 @@ repos:
...
+    additional_dependencies:
+      - pep517>=0.8.2
+      - twine>=3.1.1
```
webknjaz, line 40:
```suggestion
    - pep517>=0.8.2
    - twine>=3.1.1
```
(https://github.com/ansible/ansible-lint/blob/master/.yamllint)

.pre-commit-config.yaml
webknjaz: (when we dropped `setup.py` shims we made an assumption that end-users will get wheels that we publish and won't build from sdist most of the time and if they do it's documented that that have to use a good version of pip)

PR #726, opened 2020-04-07
Make config loading failures visible

ssbarnea:
Corrects implementation of config file loading in order to return an error code in case of failure, with a meaningful error message.

### DIFF
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -10,6 +10,7 @@ from ansiblelint.version import __version__
 
 
 _PATH_VARS = ['exclude_paths', 'rulesdir', ]
+INVALID_CONFIG_RC = 2
 
 
 def abspath(path, base_dir):
@@ -46,17 +47,33 @@ def expand_to_normalized_paths(config, base_dir=None):
 
 
 def load_config(config_file):
-    config_path = config_file if config_file else ".ansible-lint"
-    config_path = os.path.abspath(config_path)
+    config_path = os.path.abspath(config_file or '.ansible-lint')
 
-    if not os.path.exists(config_path):
+    if config_file:
+        if not os.path.exists(config_path):
+            print(
+                "Config file not found '{cfg!s}'.".format(cfg=config_path),
+                file=sys.stderr,
+            )
+            sys.exit(INVALID_CONFIG_RC)
+    elif not os.path.exists(config_path):
+        # a missing default config file should not trigger an error
         return
 
     try:
         with open(config_path, "r") as stream:
             config = yaml.safe_load(stream)
-    except yaml.YAMLError:
-        return None
+    except yaml.YAMLError as e:
+        print(e, file=sys.stderr)
+        sys.exit(INVALID_CONFIG_RC)
+    # TODO(ssbarnea): implement schema validation for config file
+    if isinstance(config, list):
+        print(
+            "Invalid configuration '{cfg!s}', expected YAML mapping in the config file.".
+            format(cfg=config_path),
+            file=sys.stderr,
+        )
+        sys.exit(INVALID_CONFIG_RC)
 
     config_dir = os.path.dirname(config_path)
     expand_to_normalized_paths(config, config_dir)
--- a/test/TestCommandLineInvocationSameAsConfig.py
+++ b/test/TestCommandLineInvocationSameAsConfig.py
@@ -74,3 +74,17 @@ def test_path_from_cli_depend_on_cwd(base_arguments, monkeypatch):  # Issue 572
 
     assert 'test/test' not in options1.exclude_paths[0]
     assert 'test/test' in options2.exclude_paths[0]
+
+
+@pytest.mark.parametrize(
+    "config_file",
+    [
+        pytest.param("test/fixtures/ansible-config-invalid.yml", id="invalid"),
+        pytest.param("/dev/null/ansible-config-missing.yml", id="missing")
+    ]
+)
+def test_config_failure(base_arguments, config_file):
+    """Ensures specific config files produce error code 2."""
+    with pytest.raises(SystemExit, match="^2$"):
+        cli.get_config(base_arguments +
+                       ["-c", config_file])
--- a/test/fixtures/ansible-config-invalid.yml
+++ b/test/fixtures/ansible-config-invalid.yml
@@ -1,0 +1,3 @@
+# invalid .ansible-lint config file
+- foo
+- bar


lib/ansiblelint/cli.py
```diff
@@ -55,8 +55,13 @@ def load_config(config_file):
...
         with open(config_path, "r") as stream:
             config = yaml.safe_load(stream)
-    except yaml.YAMLError:
-        return None
+    except yaml.YAMLError as e:
+        print(e, file=sys.stderr)
+        sys.exit(2)
+    # TODO(ssbarnea): implement schema validation for config file
+    if isinstance(config, list):
+        print("Invalid configuration '%s', expected dictionary." % config_path, file=sys.stderr)
```
webknjaz, line 63: It's called a mapping in terms of YAML. Also, it's better to use `str.format()` for interpolation.
```suggestion
        print(
        	"Invalid configuration '{cfg!s}', expected YAML mapping in the config file.".
        	format(cfg=config_path),
        	file=sys.stderr,
        )
```

webknjaz:
Please add regression tests for this

test/TestCommandLineInvocationSameAsConfig.py
```diff
@@ -74,3 +74,19 @@ def test_path_from_cli_depend_on_cwd(base_arguments, monkeypatch):  # Issue 572
...
+        cli.get_config(base_arguments +
+                       ["-c", "test/fixtures/ansible-config-invalid.yml"])
+    assert excinfo.value.code == 2
+
+
+def test_missing_config_file(base_arguments):
+    """Ensures missing config_file produces error code 2."""
+    with pytest.raises(SystemExit) as excinfo:
+        cli.get_config(base_arguments +
+                       ["-c", "test/fixtures/ansible-config-missing.yml"])
```
webknjaz, line 91: What if a file with this name appears there?
```suggestion
                       ["-c", "/dev/null/ansible-config-missing.yml"])
```

test/fixtures/ansible-config-invalid.yml
```diff
@@ -0,0 +1,3 @@
+# invalid .ansible-lint config file
```
webknjaz, line 1: Since there's no fast way to track the fixture file to the test where it's used, it's better to bundle it into the test module.

test/TestCommandLineInvocationSameAsConfig.py
```diff
@@ -74,3 +74,19 @@ def test_path_from_cli_depend_on_cwd(base_arguments, monkeypatch):  # Issue 572
 
     assert 'test/test' not in options1.exclude_paths[0]
     assert 'test/test' in options2.exclude_paths[0]
+
+
+def test_invalid_config_file(base_arguments):
+    """Ensures invalid config_file produces error code 2."""
+    with pytest.raises(SystemExit) as excinfo:
```
webknjaz, line 81: Have you tried 
```suggestion
    with pytest.raises(SystemExit, match='^2$'):
```
(not sure if it'll work, but it'd be cleaner if yes)

lib/ansiblelint/cli.py
```diff
@@ -46,17 +46,31 @@ def expand_to_normalized_paths(config, base_dir=None):
...
+        if not os.path.exists(config_path):
+            print(
+                "Config file not found '{cfg!s}'.".format(cfg=config_path),
+                file=sys.stderr,
+            )
+            sys.exit(2)
+    elif not os.path.exists(config_path):
+        # a missing default config file should trigger not error
         return
-
```
webknjaz, line 54: I'd keep this separator for readability.

lib/ansiblelint/cli.py
```diff
@@ -46,17 +46,31 @@ def expand_to_normalized_paths(config, base_dir=None):
...
     try:
```
webknjaz, line 60:
```suggestion

    try:
```

test/TestCommandLineInvocationSameAsConfig.py
```diff
@@ -74,3 +74,17 @@ def test_path_from_cli_depend_on_cwd(base_arguments, monkeypatch):  # Issue 572
...
+    with pytest.raises(SystemExit, match="^2$"):
+        cli.get_config(base_arguments +
+                       ["-c", "test/fixtures/ansible-config-invalid.yml"])
+
+
+def test_missing_config_file(base_arguments):
+    """Ensures missing config_file produces error code 2."""
+    with pytest.raises(SystemExit, match="^2$"):
+        cli.get_config(base_arguments +
+                       ["-c", "/dev/null/ansible-config-missing.yml"])
```
webknjaz, line 90: We should parametrize this, you're copy-pasting too many lines. Also, I'd check for the message in stderr using the `capsys` fixture.

test/fixtures/ansible-config-invalid.yml
webknjaz: No, not a joke. Doing it without cloning the repo is slower. Also, you misunderstood the direction of the tracking. I'm talking about tracking the test function name when seeing the fixture file.

test/TestCommandLineInvocationSameAsConfig.py
ssbarnea: For 3 tests maybe it would have made sense, but for only two is going to add complexity. Still, I will do it.

Not testing stderr/stdout output was on purpose because the only contract I want to offer is the result code. The content of the error message and where it is sent may change in time, especially as we are likely to do some refactoring and adopt python logging.

lib/ansiblelint/cli.py
```diff
@@ -46,17 +46,33 @@ def expand_to_normalized_paths(config, base_dir=None):
...
-    if not os.path.exists(config_path):
+    config_path = os.path.abspath(config_file or '.ansible-lint')
+
+    if config_file:
+        if not os.path.exists(config_path):
+            print(
+                "Config file not found '{cfg!s}'.".format(cfg=config_path),
+                file=sys.stderr,
+            )
+            sys.exit(2)
```
webknjaz, line 57: Could you pls move `2` to a global constant like `INVALID_CONFIG_RC = 2`?
```suggestion
            sys.exit(INVALID_CONFIG_RC)
```

I think it'd communicate the intent of the magic literal better. And would improve the consistency too.

test/TestCommandLineInvocationSameAsConfig.py
```diff
@@ -74,3 +74,17 @@ def test_path_from_cli_depend_on_cwd(base_arguments, monkeypatch):  # Issue 572
 
     assert 'test/test' not in options1.exclude_paths[0]
     assert 'test/test' in options2.exclude_paths[0]
+
+
+@pytest.mark.parametrize(
+    "config_file",
+    [
+        pytest.param("test/fixtures/ansible-config-invalid.yml", id="invalid"),
+        pytest.param("/dev/null/ansible-config-missing.yml", id="missing")
```
webknjaz, line 83: :+1:

test/TestCommandLineInvocationSameAsConfig.py
```diff
@@ -74,3 +74,17 @@ def test_path_from_cli_depend_on_cwd(base_arguments, monkeypatch):  # Issue 572
...
+    ]
+)
+def test_config_failure(base_arguments, config_file):
+    """Ensures specific config files produce error code 2."""
+    with pytest.raises(SystemExit, match="^2$"):
+        cli.get_config(base_arguments +
+                       ["-c", config_file])
```
webknjaz, line 90: Nitpick: this doesn't need to be multilined anymore. But won't block the PR if you decide to keep it.

Issue #707, opened 2020-04-01
ansible-lint raises AttributeError for malformed import_tasks


mdaniel:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

This is as of 6efa16e7 installed via `pip install .`

```console
$ ansible --version
ansible 2.9.6
  config file = None
  configured module search path = ['/usr/share/ansible/plugins/modules']
  ansible python module location = /proj/ansible/ansible-lint/venv/lib/python3.7/site-packages/ansible
  executable location = /proj/ansible/ansible-lint/venv/bin/ansible
  python version = 3.7.7 (default, Mar 11 2020, 11:44:20) [GCC 9.2.1 20191008]

$ ansible-lint --version
ansible-lint 4.2.1.dev63+g6efa16e
```

- ansible installation method: it came from `pip install .`
- ansible-lint installation method: `pip install .`

# Desired Behaviour

```console
$ ansible-lint -vvvv pb.yml 
  File "/proj/ansible/ansible-lint/venv/lib/python3.7/site-packages/ansiblelint/utils.py", line 151, in _playbook_items
    raise RuntimeError("What kind of nonsense is %r" % pb_data)
RuntimeError: What kind of nonsense is ['oops']
```

# Actual Behaviour (Bug report only)

Be aware this is the ansible-lint side of ansible/ansible#68605

`bug.yml`

```yaml
---
- oops
```
`pb.yml`

```yaml
- hosts: all
  gather_facts: no
  tasks:
  - import_tasks: bug.yml
```

```console
$ ansible-lint -vvvv pb.yml 
## snip
    return [item for play in pb_data for item in play.items()]
AttributeError: 'AnsibleUnicode' object has no attribute 'items'
```

webknjaz:
> This is as of [6efa16e](https://github.com/ansible/ansible-lint/commit/6efa16e70f330d5398c7b0240713a700606b0e02)

I believe you're mistaken here. That commit is unrelated to your error in any way.

Also, your error looks like it's a part of a traceback. Please post the complete trace w/o cutting it.

Could you please submit a PR with test cases (for our pytest runner) that demonstrate this issue?

mdaniel:
> I believe you're mistaken here. That commit is unrelated to your error in any way.

My apologies, I only meant that I had tested the most up-to-date code as of that sha; I didn't mean to blame it on that commit. Sorry for the confusion

[...]

PR #728, opened 2020-04-08
Rewrite #720 with pytest

slsh1o:
Relates: #707 
Close: #720

### DIFF
--- a/test/TestImportWithMalformed.py
+++ b/test/TestImportWithMalformed.py
@@ -1,0 +1,42 @@
+import os
+import unittest
+import tempfile
+import shutil
+
+from ansiblelint import Runner, RulesCollection
+
+IMPORT_TASKS_MAIN = '''
+- oops this is invalid
+'''
+
+PLAY_IMPORT_TASKS = '''
+- hosts: all
+
+  tasks:
+    - import_tasks: import-tasks-main.yml
+'''
+
+
+class TestImportIncludeRole(unittest.TestCase):
+    def setUp(self):
+        rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
+        self.rules = RulesCollection([rulesdir])
+
+        # make dir and write role tasks to import or include
+        self.play_root = tempfile.mkdtemp()
+        with open(os.path.join(self.play_root, 'import-tasks-main.yml'), 'w') as f_main:
+            f_main.write(IMPORT_TASKS_MAIN)
+
+    def tearDown(self):
+        shutil.rmtree(self.play_root)
+
+    def _get_play_file(self, playbook_text):
+        with open(os.path.join(self.play_root, 'playbook.yml'), 'w') as f_play:
+            f_play.write(playbook_text)
+        return f_play
+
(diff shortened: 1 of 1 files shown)


test/TestImportWithMalformed.py
```diff
@@ -0,0 +1,59 @@
...
+    rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
+    return RulesCollection([rulesdir])
+
+
+@pytest.fixture
+def runner(play_file_path, rules):
+    return Runner(rules, play_file_path, [], [], [])
+
+
+
```
webknjaz, line 40:
```suggestion
```

test/TestImportWithMalformed.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+
+from ansiblelint import Runner, RulesCollection
+
+
+PlayFile = namedtuple('PlayFile', ['name', 'content'])
+
+
+IMPORT_TASKS_MAIN = PlayFile('import-tasks-main.yml', '''
+- oops this is invalid
```
webknjaz, line 13: I think I'd add another playbook but with `- ping:` instead and add that to parametrize

test/TestImportWithMalformed.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.fixture
+def play_file_path(tmp_path):
+    p = tmp_path / 'playbook.yml'
+    return str(p)
+
+
+@pytest.fixture
+def rules():
+    rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
```
webknjaz, line 31: Let's use pathlib instead
```suggestion
    rulesdir = str(Path('lib') / 'ansiblelint' / 'rules')
```

test/TestImportWithMalformed.py
```diff
@@ -0,0 +1,59 @@
+from collections import namedtuple
+import os
```
webknjaz, line 2:
```suggestion
from pathlib import Path
```

slsh1o:
I believe I made it wrong with `-ping:` or we need something else because this _ping_ doesn't fit assert msg.
As we fix it I'll be able to add xfail to the parametrize.

webknjaz:
@slsh1o what's left now is to add an xfail marker to the parameter that causes `AttributeError`.

test/TestImportWithMalformed.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@pytest.mark.parametrize(
+    'play_files',
+    [
+        pytest.param([IMPORT_SHELL_PIP, PLAY_IMPORT_TASKS], id='Import shell w/ pip'),
+        pytest.param(
+            [IMPORT_TASKS_MAIN, PLAY_IMPORT_TASKS],
+            id='import_tasks w/ malformed import',
+            marks=pytest.mark.xfail(raises=AttributeError)
```
webknjaz, line 60: Nitpick: normally it's a good practice to also specify a `reason=` argument pointing the issue.

PR #729, opened 2020-04-08
Adopt black formatting

ssbarnea:
This should externalize code-styling to black and avoid code-styling subjects during code reviews.


test/rules/EMatcherRule.py
```diff
@@ -3,8 +3,9 @@
 
 class EMatcherRule(AnsibleLintRule):
     id = 'TEST0001'
-    description = 'This is a test rule that looks for lines ' + \
-                  'containing the letter e'
+    description = (
+        'This is a test rule that looks for lines ' + 'containing the letter e'
```
webknjaz, line 7: This is a good example of why autoformatters are terrible.

ssbarnea: Congrats! You found a bug in hundreads of lines, one very easy to fix. 

Note. I do not expect this PR to get merged but I hope to use it to learn about things we could imporve and make specific changes.

If it look overall I see a huge number of good changes, some debatable ones and one of two where I do not like the outcome.

ssbarnea:
I read that github now allows to make PR back to DRAFT but I do not see the option. I wonder if is  a beta feature that needs to be enabled by a repo-admin?

ssbarnea:
Closing for the moment as we have more important changes to do, and less controversial.

PR #732, opened 2020-04-09
Use python logging

ssbarnea:
Instead of using print use logging module for messages that
are not direct result of the tool (linting error).

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -19,13 +19,14 @@
 # THE SOFTWARE.
 
 from collections import defaultdict
+import logging
 import os
 import re
-import sys
 
 import ansiblelint.utils
 
 default_rulesdir = os.path.join(os.path.dirname(ansiblelint.utils.__file__), 'rules')
+_logger = logging.getLogger(__name__)
 
 
 class AnsibleLintRule(object):
@@ -164,9 +165,10 @@ class RulesCollection(object):
             with open(playbookfile['path'], mode='r', encoding='utf-8') as f:
                 text = f.read()
         except IOError as e:
-            print("WARNING: Couldn't open %s - %s" %
-                  (playbookfile['path'], e.strerror),
-                  file=sys.stderr)
+            _logger.warning(
+                "Couldn't open %s - %s",
+                playbookfile['path'],
+                e.strerror)
             return matches
 
         for rule in self.rules:
@@ -278,10 +280,10 @@ class Runner(object):
         # remove files that have already been checked
         files = [x for x in files if x['path'] not in self.checked_files]
         for file in files:
-            if self.verbosity > 0:
-                print("Examining %s of type %s" % (
-                    ansiblelint.utils.normpath(file['path']),
-                    file['type']))
+            _logger.debug(
+                "Examining %s of type %s",
+                ansiblelint.utils.normpath(file['path']),
+                file['type'])
             matches.extend(self.rules.run(file, tags=set(self.tags),
                            skip_list=self.skip_list))
         # update list of checked files
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import argparse
+import logging
 import os
 from pathlib import Path
 import sys
@@ -10,8 +11,8 @@ import ansiblelint
 from ansiblelint.version import __version__
 
 
+_logger = logging.getLogger(__name__)
 _PATH_VARS = ['exclude_paths', 'rulesdir', ]
-INVALID_CONFIG_RC = 2
 
 
 def abspath(path, base_dir):
@@ -52,11 +53,8 @@ def load_config(config_file):
 
     if config_file:
         if not os.path.exists(config_path):
-            print(
-                "Config file not found '{cfg!s}'.".format(cfg=config_path),
-                file=sys.stderr,
-            )
-            sys.exit(INVALID_CONFIG_RC)
+            _logger.error("Config file not found '%s'", config_path)
+            sys.exit(ansiblelint.utils.INVALID_CONFIG_RC)
     elif not os.path.exists(config_path):
         # a missing default config file should not trigger an error
         return
@@ -65,16 +63,14 @@ def load_config(config_file):
         with open(config_path, "r") as stream:
             config = yaml.safe_load(stream)
     except yaml.YAMLError as e:
-        print(e, file=sys.stderr)
-        sys.exit(INVALID_CONFIG_RC)
+        _logger.error(e)
+        sys.exit(ansiblelint.utils.INVALID_CONFIG_RC)
     # TODO(ssbarnea): implement schema validation for config file
     if isinstance(config, list):
-        print(
-            "Invalid configuration '{cfg!s}', expected YAML mapping in the config file.".
-            format(cfg=config_path),
-            file=sys.stderr,
-        )
-        sys.exit(INVALID_CONFIG_RC)
+        _logger.error(
+            "Invalid configuration '%s', expected YAML mapping in the config file.",
+            config_path)
+        sys.exit(ansiblelint.utils.INVALID_CONFIG_RC)
 
     config_dir = os.path.dirname(config_path)
     expand_to_normalized_paths(config, config_dir)
--- a/lib/ansiblelint/generate_docs.py
+++ b/lib/ansiblelint/generate_docs.py
@@ -3,6 +3,7 @@
 import os
 import importlib
 from inspect import getmembers, ismodule, isclass
+import logging
 import rules
 from ansiblelint import AnsibleLintRule
 from functools import reduce
@@ -18,6 +19,8 @@ Default Rules
 The table below shows the the default rules used by Ansible Lint to evaluate playbooks and roles:
 
 """
+
+_logger = logging.getLogger(__name__)
 
 
 def main():
@@ -45,7 +48,7 @@ def main():
     filename = '../../docs/docsite/rst/rules/default_rules.rst'
     with open(filename, 'w') as file:
         file.write(make_table(grid))
-        print('{} file written'.format(filename))
+        _logger.info('%s file written', filename)
 
 
 def import_all_rules():
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -22,6 +22,7 @@ from collections import OrderedDict
 import glob
 import imp
 from itertools import product
+import logging
 import os
 from pathlib import Path
 import pprint
@@ -51,6 +52,27 @@ from ansible.template import Templar
 DEFAULT_VAULT_PASSWORD = 'x'
 
 PLAYBOOK_DIR = os.environ.get('ANSIBLE_PLAYBOOK_DIR', None)
+
+INVALID_CONFIG_RC = 2
+ANSIBLE_FAILURE_RC = 3
+
+_logger = logging.getLogger(__name__)
+
+
+def initialize_logger(level=0):
+    """Set up the global logging level based on the verbosity number."""
+    VERBOSITY_MAP = {
+        0: logging.NOTSET,
+        1: logging.INFO,
+        2: logging.DEBUG
+    }
+
+    handler = logging.StreamHandler()
+    formatter = logging.Formatter('%(levelname)-8s %(message)s')
+    handler.setFormatter(formatter)
+    _logger.addHandler(handler)
+    # unknown logging level is treated as DEBUG
+    _logger.setLevel(VERBOSITY_MAP.get(level, logging.DEBUG))
 
 
 def parse_yaml_from_file(filepath):
@@ -379,7 +401,8 @@ def normalize_task_v2(task):
         pp = pprint.PrettyPrinter(indent=2)
         task_pprint = pp.pformat(task)
 
-        raise SystemExit("Couldn't parse task at %s (%s)\n%s" % (task_info, e.message, task_pprint))
+        _logger.critical("Couldn't parse task at %s (%s)\n%s", task_info, e.message, task_pprint)
+        raise SystemExit(ANSIBLE_FAILURE_RC)
 
     # denormalize shell -> command conversion
     if '_uses_shell' in arguments:
@@ -597,7 +620,7 @@ def append_skipped_rules(pyyaml_data, file_text, file_type):
         yaml_skip = _append_skipped_rules(pyyaml_data, file_text, file_type)
     except RuntimeError as exc:
         # Notify user of skip error, do not stop, do not change exit code
-        print('Error trying to append skipped rules: {!r}'.format(exc))
+        _logger.error('Error trying to append skipped rules: %s', exc)
         return pyyaml_data
     return yaml_skip
 
@@ -751,9 +774,9 @@ def is_playbook(filename):
     try:
         f = parse_yaml_from_file(filename)
     except Exception as e:
-        print(
-            "Warning: Failed to load %s with %s, assuming is not a playbook."
-            % (filename, e))
+        _logger.warning(
+            "Failed to load %s with %s, assuming is not a playbook.",
+            filename, e)
     else:
         if (
             isinstance(f, AnsibleSequence) and
@@ -767,12 +790,7 @@ def get_yaml_files(options):
     """Find all yaml files."""
     # git is preferred as it also considers .gitignore
     git_command = ['git', 'ls-files', '*.yaml', '*.yml']
-    if options.verbosity:
-        print(
-            "Discovering files to lint: {command}".format(
-                command=' '.join(git_command)
-            )
-        )
+    _logger.info("Discovering files to lint: %s", ' '.join(git_command))
 
     out = None
 
@@ -783,19 +801,14 @@ def get_yaml_files(options):
             universal_newlines=True
         ).split()
     except subprocess.CalledProcessError as exc:
-        if options.verbosity:
-            print(
-                "Warning: Failed to discover yaml files to lint using git:"
-                " {error_msg}".format(
-                    error_msg=exc.output.rstrip('\n')
-                )
-            )
+        _logger.warning(
+            "Failed to discover yaml files to lint using git: %s",
+            exc.output.rstrip('\n')
+        )
     except FileNotFoundError as exc:
         if options.verbosity:
-            print(
-                "Warning: Failed to locate command: {error_msg!s}".format(
-                    error_msg=exc
-                )
+            _logger.warning(
+                "Failed to locate command: %s", exc
             )
 
     if out is None:
@@ -863,12 +876,10 @@ def get_playbooks_and_roles(options=None):
             playbooks.append(normpath(p))
             continue
 
-        if options.verbosity:
-            print('Unknown file type: %s' % normpath(p))
+        _logger.info('Unknown file type: %s', normpath(p))
 
-    if options.verbosity:
-        print('Found roles: ' + ' '.join(role_dirs))
-        print('Found playbooks: ' + ' '.join(playbooks))
+    _logger.info('Found roles: %s', ' '.join(role_dirs))
+    _logger.info('Found playbooks: %s', ' '.join(playbooks))
 
     return role_dirs + playbooks
 
--- a/test/TestCliRolePaths.py
+++ b/test/TestCliRolePaths.py
@@ -20,8 +20,6 @@ class TestCliRolePaths(unittest.TestCase):
             env=env,
         ).communicate()
 
-        self.assertFalse(err, 'Expected no error but was ' + str(err))
-
         return result
 
     def test_run_single_role_path_no_trailing_slash(self):
--- a/test/TestUtils.py
+++ b/test/TestUtils.py
@@ -196,6 +196,8 @@ def test_get_yaml_files_git_verbose(
 def test_get_yaml_files_silent(is_in_git, monkeypatch, capsys):
     """Verify that no stderr output is displayed while discovering yaml files.
 
+    (when the verbosity is off, regardless of the Git or Git-repo presence)
+
     Also checks expected number of files are detected.
     """
     options = cli.get_config([])
--- a/test/TestUtils.py
+++ b/test/TestUtils.py
@@ -20,6 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+import logging
 import os
 from pathlib import Path
 
@@ -160,8 +161,10 @@ def test_expand_paths_vars(monkeypatch):
 @pytest.mark.parametrize(
     ('reset_env_var', 'message_prefix'),
     (
-        ('PATH', 'Warning: Failed to locate command:'),
-        ('GIT_DIR', 'Warning: Failed to discover yaml files to lint using git:')
+        ('PATH',
+            "Failed to locate command: "),
+        ('GIT_DIR',
+            "Failed to discover yaml files to lint using git: ")
     ),
     ids=('no Git installed', 'outside Git repository'),
 )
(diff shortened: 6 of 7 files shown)


lib/ansiblelint/utils.py
```diff
@@ -52,6 +53,26 @@
...
+
+logger = logging.getLogger()
+handler = logging.StreamHandler()
+formatter = logging.Formatter('%(levelname)-8s %(message)s')
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+logger.setLevel(logging.DEBUG)
+
+
+def logger_verbosity(level):
```
webknjaz, line 73: It should be a verb
```suggestion
def set_logger_verbosity(level):
```

webknjaz: Also plz add a docstring

lib/ansiblelint/utils.py
```diff
@@ -382,7 +403,8 @@ def normalize_task_v2(task):
             task_pprint = pp.pformat(task)
         except ImportError:
             task_pprint = task
-        raise SystemExit("Couldn't parse task at %s (%s)\n%s" % (task_info, e.message, task_pprint))
+        logger.fatal("Couldn't parse task at %s (%s)\n%s" % (task_info, e.message, task_pprint))
+        raise SystemExit(ANSIBLE_FAILURE_RC)
```
webknjaz, line 407: Changing the return code conventions is out of the scope of this PR.

test/TestCliRolePaths.py
```diff
@@ -20,16 +20,14 @@ def run_ansible_lint(self, cwd, bin, role_path, env=None):
             env=env,
         ).communicate()
 
-        self.assertFalse(err, 'Expected no error but was ' + str(err))
-
-        return result
+        return result, stderr
```
webknjaz, line 23: I don't see any point in returning stderr, you're not using it
```suggestion
        return result
```

test/TestCliRolePaths.py
```diff
@@ -10,7 +10,7 @@ def setUp(self):
     def run_ansible_lint(self, cwd, bin, role_path, env=None):
         command = '{} {}'.format(bin, role_path)
 
-        result, err = subprocess.Popen(
+        result, stderr = subprocess.Popen(
```
webknjaz, line 13: this looks unrelated

lib/ansiblelint/utils.py
```diff
@@ -52,6 +53,26 @@
...
+INVALID_CONFIG_RC = 2
+ANSIBLE_FAILURE_RC = 3
+
+VERBOSITY_MAP = {
+    0: logging.NOTSET,
+    1: logging.INFO,
+    2: logging.DEBUG
+}
+
+logger = logging.getLogger()
```
webknjaz, line 65: Not sure if I want only one global logger to be used. Why didn't you go in the direction of per-module loggers?

ssbarnea: Yep, better even if ansiblelint is a cli tool.

lib/ansiblelint/__init__.py
```diff
@@ -19,12 +19,12 @@
 # THE SOFTWARE.
 
 from collections import defaultdict
+import codecs
```
webknjaz, line 22: (future hint) it's probably unnecessary now that we don't have to support Python 2. Out of scope of this PR, of course, but would be a nice improvement separately.

lib/ansiblelint/cli.py
```diff
@@ -51,11 +52,10 @@ def load_config(config_file):
 
     if config_file:
         if not os.path.exists(config_path):
-            print(
-                "Config file not found '{cfg!s}'.".format(cfg=config_path),
-                file=sys.stderr,
+            logger.info(
+                "Config file not found '{cfg!s}'.".format(cfg=config_path)
```
webknjaz, line 56: Linters (pylint?) usually yell at manual interpolation in logger calls, it's preferable to pass arguments separately. Would something like this work?
```suggestion
                "Config file not found '%(cfg)s'.", cfg=config_path)
```

lib/ansiblelint/cli.py
```diff
@@ -51,11 +52,10 @@ def load_config(config_file):
 
     if config_file:
         if not os.path.exists(config_path):
-            print(
-                "Config file not found '{cfg!s}'.".format(cfg=config_path),
-                file=sys.stderr,
+            logger.info(
```
webknjaz, line 55: Why is this one info but the one below is error?

lib/ansiblelint/generate_docs.py
```diff
@@ -45,7 +46,7 @@ def main():
     filename = '../../docs/docsite/rst/rules/default_rules.rst'
     with open(filename, 'w') as file:
         file.write(make_table(grid))
-        print('{} file written'.format(filename))
+        logger.info('{} file written'.format(filename))
```
webknjaz, line 49:
```suggestion
        logger.info('%s file written', filename)
```

lib/ansiblelint/utils.py
webknjaz: Could you please elaborate on this? Not sure I can parse this.

lib/ansiblelint/utils.py
webknjaz: @ssbarnea ^ functions perform actions, so their names should hint that some action is done by using a verb. Otherwise, it reads as a property/variable.

test/TestCliRolePaths.py
```diff
@@ -20,8 +20,6 @@ def run_ansible_lint(self, cwd, bin, role_path, env=None):
             env=env,
         ).communicate()
 
-        self.assertFalse(err, 'Expected no error but was ' + str(err))
```
webknjaz, line 23: Why? Can this be changed to use a test using a `caplog` fixture?

lib/ansiblelint/utils.py
```diff
@@ -857,12 +877,10 @@ def get_playbooks_and_roles(options=None):
             playbooks.append(normpath(p))
             continue
 
-        if options.verbosity:
-            print('Unknown file type: %s' % normpath(p))
+        logger.info('Unknown file type: %s' % normpath(p))
```
webknjaz, line 880:
```suggestion
        logger.info('Unknown file type: %s', normpath(p))
```

lib/ansiblelint/utils.py
```diff
@@ -857,12 +877,10 @@ def get_playbooks_and_roles(options=None):
...
 
-    if options.verbosity:
-        print('Found roles: ' + ' '.join(role_dirs))
-        print('Found playbooks: ' + ' '.join(playbooks))
+    logger.info('Found roles: ' + ' '.join(role_dirs))
+    logger.info('Found playbooks: ' + ' '.join(playbooks))
```
webknjaz, line 883:
```suggestion
    logger.info('Found roles: %s', ' '.join(role_dirs))
    logger.info('Found playbooks: %s', ' '.join(playbooks))
```

lib/ansiblelint/utils.py
```diff
@@ -785,13 +806,12 @@ def get_yaml_files(options):
...
-                "Warning: Failed to discover yaml files to lint using git:"
-                " {error_msg}".format(
-                    error_msg=exc.output.rstrip('\n')
-                )
+        logger.warning(
+            "Failed to discover yaml files to lint using git:"
+            " {error_msg}".format(
+                error_msg=exc.output.rstrip('\n')
             )
+        )
```
webknjaz, line 814:
```suggestion
        logger.warning(
            "Failed to discover yaml files to lint using git:"
            " %(error_msg)s",
            error_msg=exc.output.rstrip('\n'),
        )
```

lib/ansiblelint/utils.py
```diff
@@ -771,12 +793,11 @@ def get_yaml_files(options):
...
-    if options.verbosity:
-        print(
-            "Discovering files to lint: {command}".format(
-                command=' '.join(git_command)
-            )
+    logger.info(
+        "Discovering files to lint: {command}".format(
+            command=' '.join(git_command)
         )
+    )
```
webknjaz, line 800:
```suggestion
    logger.info(
        "Discovering files to lint: %(command)s",
        command=' '.join(git_command),
    )
```

lib/ansiblelint/utils.py
```diff
@@ -600,7 +622,7 @@ def append_skipped_rules(pyyaml_data, file_text, file_type):
         yaml_skip = _append_skipped_rules(pyyaml_data, file_text, file_type)
     except RuntimeError as exc:
         # Notify user of skip error, do not stop, do not change exit code
-        print('Error trying to append skipped rules: {!r}'.format(exc))
+        logger.error('Error trying to append skipped rules: {!r}'.format(exc))
```
webknjaz, line 625:
```suggestion
        logger.error('Error trying to append skipped rules: %r', exc)
```

webknjaz: Even better
```suggestion
        logger.exception('Error trying to append skipped rules')
```

webknjaz, line 625: `exc` is not really needed
```suggestion
    except RuntimeError:
        # Notify user of skip error, do not stop, do not change exit code
        logger.exception('Error trying to append skipped rules')
```

lib/ansiblelint/utils.py
```diff
@@ -382,7 +403,8 @@ def normalize_task_v2(task):
             task_pprint = pp.pformat(task)
         except ImportError:
             task_pprint = task
-        raise SystemExit("Couldn't parse task at %s (%s)\n%s" % (task_info, e.message, task_pprint))
+        logger.fatal("Couldn't parse task at %s (%s)\n%s" % (task_info, e.message, task_pprint))
```
webknjaz, line 406: There's no `fatal` here: https://docs.python.org/3/library/logging.html. Did you mean
```suggestion
        logger.critical("Couldn't parse task at %s (%s)\n%s", task_info, e.message, task_pprint)
```

lib/ansiblelint/cli.py
ssbarnea: Clearly it would work. I did not touch it because use of new formatter was your suggestion last week. Another alternative is to use one of the hackish solutions that add formatting support to logging (a custom formatter). Still, i am bit afraid that we may endup introducing more problems than solving. Better to go to older but fully supported syntax.

test/TestCliRolePaths.py
ssbarnea: The reason for removing this line is that now we get some valid output on stderr. Presence of output on stderr no longer means an error happened.

At this moment I do not know what kind of test we can reliably do about stderr, it can be empty, or have various content even on perfect valid executions.

ssbarnea:
Current failure seems to be CI issue, which can be avoided if we merge https://github.com/ansible/ansible-lint/pull/737/files first.

lib/ansiblelint/cli.py
webknjaz: Oh, let me make it clear then. Formatting of the usual strings should be done with the use of `str.format()` (and f-strings in the future) but it's a bit different for logging: when you pass a separate pre-rendered string it doesn't know about the vars but when you pass them separately it does. I think the idea of separating the args comes from structlog.

Also, pylint has [logging-not-lazy (W1201) rule](http://pylint.pycqa.org/en/latest/technical_reference/features.html#logging-checker-messages) that implies that lazy interpolation is preferable (meaning that it won't happen if not really necessary per logging config).

test/TestCliRolePaths.py
webknjaz: I noticed that you changed the return code too. So maybe you could assert that? Or how about using `subprocess.check_output()` instead?

webknjaz:
@ssbarnea I just realized that I was commenting on the outdated chain (https://github.com/ansible/ansible-lint/pull/732#discussion_r410730458) so let me just copy this here for discoverability:

Formatting of the usual strings should be done with the use of `str.format()` (and f-strings in the future) but it's a bit different for logging: when you pass a separate pre-rendered string it doesn't know about the vars but when you pass them separately it does. I think the idea of separating the args comes from structlog.

Besides, tools that aggregate logs (like Sentry) can group them properly: https://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html#comment-2063573645.

Also, pylint has [logging-not-lazy (W1201) rule](http://pylint.pycqa.org/en/latest/technical_reference/features.html#logging-checker-messages) that implies that lazy interpolation is preferable (meaning that it won't happen if not really necessary per logging config).

The [official logging doc](https://docs.python.org/3/howto/logging.html#optimization) also recommends lazy interpolation: https://stackoverflow.com/a/34634301/595220.

lib/ansiblelint/utils.py
```diff
@@ -52,6 +53,27 @@
...
+    1: logging.INFO,
+    2: logging.DEBUG
+}
+
+logger = logging.getLogger("ansiblelint")
+handler = logging.StreamHandler()
+formatter = logging.Formatter('%(levelname)-8s %(message)s')
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+logger.setLevel(logging.DEBUG)
```
webknjaz, line 70: I think it's not a good idea to have this on the module level globally. Import-time evaluation is usually not a good thing. I'd prefer a setup function called when the app starts.

webknjaz:
@ssbarnea it looks like you forgot to answer the question https://github.com/ansible/ansible-lint/pull/732#discussion_r407121559:
Why do you prefer one global logger vs. multiple loggers (one per module)?

lib/ansiblelint/utils.py
ssbarnea: @webknjaz Can you please give an existing logging implementation as an reference example so I can  follow that? 

I understand what you mean and I agree it may not be the best implementation but I also want to avoid rewriting this PR forever. For example, I would have avoided implementing set_logging_verbosity_level and create an init_logging() function instead if that remark was added earlier.

Another approach is to merge it in its current form and implement the other changes as a follow-up, as the logging calls would not change.

test/TestCliRolePaths.py
ssbarnea: @webknjaz This method has no knowledge of what is the outcome of the execution and is called 7 times in tests. Doing any change on how we validate the tests should count as outside the scope of implementing the use of python logging.

ssbarnea:
> @ssbarnea it looks like you forgot to answer the question [#732 (comment)](https://github.com/ansible/ansible-lint/pull/732#discussion_r407121559):
> Why do you prefer one global logger vs. multiple loggers (one per module)?

Latest implementation uses a single logger named `ansiblelint` for the entire package. There is not need to use different loggers as it would make controlling of logging harder.

lib/ansiblelint/utils.py
webknjaz: Basically, I want some `initialize_logger()` implemented here and `main()` calling it at the very beginning, before anything else. I understand that you want to merge incrementally so I won't block this PR if you postpone this specific change.

test/TestCliRolePaths.py
webknjaz: Okay

webknjaz:
> Latest implementation uses a single logger named `ansiblelint` for the entire package. There is not need to use different loggers as it would make controlling of logging harder.

I understand this is you opinion and my question is: why do you think so? This looks like a result of some chain of thoughts, it'd be great if you could write them down here.

I have a slightly different opinion:
1) Using a global logger that's always imported from utils creates more inter-module dependencies, can lead to potential circular imports and it's unpredictable when the logger gets initialized.
2) One common practice is to use something like
   ```python
   _logger = logging.getLogger()
   ```
   This method
   1. Makes it easier to trace messages back to the modules that produced them
   2. Allows us to not depend on utils and just initialize loggers where they are needed w/o having to maintain a singleton.
3) Python's stdlib `logging` is done in a few stages:
   ```
   (1) add a log message -> (2) filter by levels -> (3) output if needed
   ```
   We should configure (2) and (3) globally and not care what logger the log messages are coming from. It'd allow modules to have logger instances that are not tightly bound to global output and filtering configuration.

ssbarnea:
As I said, I do not care very much about initial implementation, but I care not to spend endless hours reworking it and failing to switch print with logging calls. As far we get rid of print() I am more than happy.

In fact many people are just calling the module functions directly without an instance, which avoids the singleton issue.

So before I rewrite it, **are you ok if I just use the logging module directly?**
```
import logging
logging.warning(...)
```

webknjaz:
I'd rather not do that. Logger instances should be created per module, inputs should be distinctive, while the output may be aggregated.
Different loggers will have different names and it's easier to figure out where messages came from when you have the emitter identifiers.

lib/ansiblelint/utils.py
```diff
@@ -781,13 +799,10 @@ def get_yaml_files(options):
...
-            print(
-                "Warning: Failed to discover yaml files to lint using git:"
-                " {error_msg}".format(
-                    error_msg=exc.output.rstrip('\n')
-                )
-            )
+        logger.warning(
+            "Failed to discover yaml files to lint using git: %s",
+            exc.output.rstrip('\n')
+        )
```
webknjaz, line 805: Note: #738 adds a new print statement below this place, be sure to replace it if it's merged first.

webknjaz: @ssbarnea it's merged. And besides adding an extra except block, it also has tests that check stdout so you'll need to adjust them as well.

test/TestUtils.py
```diff
@@ -160,23 +161,34 @@ def test_expand_paths_vars(monkeypatch):
...
+        'Discovering files to lint: git ls-files *.yaml *.yml')
+
+    assert expected_info in caplog.record_tuples
+    for msg in caplog.messages:
+        if msg.startswith(message_prefix):
+            return
+    pytest.fail('%s not found in logged messages: %s', msg, caplog.messages)
 
 
 @pytest.mark.parametrize(
```
webknjaz, line 194: @ssbarnea you missed `capsys` here too:

test/TestUtils.py
```diff
@@ -160,23 +161,34 @@ def test_expand_paths_vars(monkeypatch):
...
+    expected_info = (
+        "ansiblelint",
+        logging.INFO,
+        'Discovering files to lint: git ls-files *.yaml *.yml')
+
+    assert expected_info in caplog.record_tuples
+    for msg in caplog.messages:
+        if msg.startswith(message_prefix):
+            return
+    pytest.fail('%s not found in logged messages: %s', msg, caplog.messages)
```
webknjaz, line 191: It's better to avoid manual `pytest.fail()` and also it reads better semantically if you do something like "`assert` there's `any` message in captured `messages` that `startswith` `message_prefix`":
```suggestion
    assert any(m.startswith(message_prefix) for m in caplog.messages)
```

test/TestUtils.py
```diff
@@ -160,23 +161,34 @@ def test_expand_paths_vars(monkeypatch):
...
 )
 def test_get_yaml_files_git_verbose(
     reset_env_var,
     message_prefix,
     monkeypatch,
-    capsys
+    caplog
 ):
     options = cli.get_config(['-v'])
+    utils.initialize_logger(options.verbosity)
```
webknjaz, line 178: I think this should be put into some sort of a generator-based fixture that would reset the logging level on teardown.

lib/ansiblelint/utils.py
```diff
@@ -863,12 +878,10 @@ def get_playbooks_and_roles(options=None):
             playbooks.append(normpath(p))
             continue
 
-        if options.verbosity:
-            print('Unknown file type: %s' % normpath(p))
+        _logger.info('Unknown file type: %s', normpath(p))
```
webknjaz, line 881: It looks like you're changing the logic here. It's probably a warning

lib/ansiblelint/utils.py
```diff
@@ -863,12 +878,10 @@ def get_playbooks_and_roles(options=None):
...
 
-    if options.verbosity:
-        print('Found roles: ' + ' '.join(role_dirs))
-        print('Found playbooks: ' + ' '.join(playbooks))
+    _logger.info('Found roles: %s', ' '.join(role_dirs))
+    _logger.info('Found playbooks: %s', ' '.join(playbooks))
```
webknjaz, line 884: should this be debug?

lib/ansiblelint/utils.py
```diff
@@ -52,6 +53,29 @@
 
 PLAYBOOK_DIR = os.environ.get('ANSIBLE_PLAYBOOK_DIR', None)
 
+INVALID_CONFIG_RC = 2
+ANSIBLE_FAILURE_RC = 3
+
+_logger = logging.getLogger("ansiblelint")
```
webknjaz, line 59:
```suggestion
_logger = logging.getLogger(__name__)
```

lib/ansiblelint/utils.py
ssbarnea: I do not think so because the logging will happen or not based on verbosity level, is just that now we rely on logging to make the decision.

lib/ansiblelint/utils.py
```diff
@@ -52,6 +53,29 @@
...
+    }
+
+    handler = logging.StreamHandler()
+    formatter = logging.Formatter('%(levelname)-8s %(message)s')
+    handler.setFormatter(formatter)
+    _logger.addHandler(handler)
+    _logger.setLevel(logging.DEBUG)
+
+    if level is not None:
+        _logger.setLevel(VERBOSITY_MAP.get(level, logging.DEBUG))
```
webknjaz, line 77: this is essentially
```suggestion
    _logger.setLevel(VERBOSITY_MAP.get(level, logging.DEBUG))
```

lib/ansiblelint/__init__.py
```diff
@@ -278,10 +280,10 @@ def run(self):
         # remove files that have already been checked
         files = [x for x in files if x['path'] not in self.checked_files]
         for file in files:
-            if self.verbosity > 0:
-                print("Examining %s of type %s" % (
-                    ansiblelint.utils.normpath(file['path']),
-                    file['type']))
+            _logger.info(
```
webknjaz, line 283: maybe
```suggestion
            _logger.debug(
```

lib/ansiblelint/utils.py
ssbarnea: Debug will not be displayed unless extra verbosity is asked and this commands are far too important to not be enabled by default (as we had several bugs around collection failures, like false-positives or false-negatives). I think is desirable to keep them like this, at least for the moment. In the long run, we may downgrade them to debug level. I would wait for a user to raise it as a request first, and later make a mini-survey.

lib/ansiblelint/utils.py

lib/ansiblelint/utils.py

test/TestUtils.py
ssbarnea: done

test/TestUtils.py
ssbarnea: In fact I did not miss it, i removed it as it was not used anymore.

webknjaz: No, you removed it in the test above, it's a different test that still has it. (`test_get_yaml_files_silent`)

test/TestUtils.py
ssbarnea: Read https://docs.pytest.org/en/latest/logging.html#caplog-fixture  which states that caplog fixture takes care of restoring logging setup after each test run.

test/TestUtils.py
ssbarnea: In that case there was a bug, wrongly named variable. We still want to check that no output is produced on stderr. The code was functional but it used misleading variable name. I fixed it and also documented the purpose of the test.

test/TestUtils.py
```diff
@@ -185,6 +194,10 @@ def test_get_yaml_files_git_verbose(
     ids=('in Git', 'outside Git'),
 )
 def test_get_yaml_files_silent(is_in_git, monkeypatch, capsys):
+    """Verify that no stderr output is displayed while discovering yaml files.
```
webknjaz, line 197:
```suggestion
    """Verify that no stderr output is displayed while discovering yaml files.

    (when the verbosity is off, regardless of the Git or Git-repo presence)
```

test/TestUtils.py
```diff
@@ -185,6 +194,12 @@ def test_get_yaml_files_git_verbose(
     ids=('in Git', 'outside Git'),
 )
 def test_get_yaml_files_silent(is_in_git, monkeypatch, capsys):
```
webknjaz, line 196: You should also check it with `caplog` because originally this test was added to check that there's no warnings when falling back to pure-python lookup.

ssbarnea: I am inclined to believe that using capsys is more comprehensive than caplog, mainly because what caplog captures is a subset of what capsys does, usually they are the same but there is no guarantee that someone does not slip a `print()` by mistake.

Anyway, I do not have any strong feelings about the test so I would likely accept any edit made on it.

PR #733, opened 2020-04-12
ci: tox upgrade workaround for zuul

ssbarnea:
Upgrades tox on Zuul CI in order to avoid build failures due to a version that is unable to upgrade itself.


ssbarnea:
@webknjaz ^ if you got tired to see the 3rd party ci check red.

Issue #602, opened 2019-10-24
[HELP WANTED] [SEND PR] Get rid of the deprecated `imp` module usage


webknjaz:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

N/A

- ansible installation method: tox tests
- ansible-lint installation method: tox tests

# Desired Behaviour

`pytest` should not fail in the strict warning mode (`filterwarnings = error`)

# Actual Behaviour (Bug report only)

`DeprecationWarning` is raised and pytest errors out. To see and fix this, remove the corresponding ignore lines from `pytest.ini`

webknjaz:
@slsh1o do you want to try this one out?

slsh1o:
Yep. I take it.

PR #734, opened 2020-04-14
[WIP] Get rid of the imp module usage

slsh1o:
The imp library deprecated since python 3.3 so it was changed to importlib

Close: #602

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -102,6 +102,7 @@ def load_plugins(directory):
     result = []
 
     for pluginfile in glob.glob(os.path.join(directory, '[A-Za-z]*.py')):
+
         pluginname = os.path.basename(pluginfile.replace('.py', ''))
         spec = importlib.util.spec_from_file_location(pluginname, pluginfile)
         module = importlib.util.module_from_spec(spec)
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -20,7 +20,7 @@
 
 from collections import OrderedDict
 import glob
-import imp
+import importlib
 from itertools import product
 import os
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/utils.py
```diff
@@ -100,19 +100,14 @@ def ansible_template(basedir, varname, templatevars, **kwargs):
 
 def load_plugins(directory):
     result = []
-    fh = None
 
     for pluginfile in glob.glob(os.path.join(directory, '[A-Za-z]*.py')):
-
```
webknjaz, line 106: Please avoid doing unrelated formatting changes.

pytest.ini
```diff
@@ -24,11 +24,6 @@ doctest_optionflags = ALLOW_UNICODE ELLIPSIS
 filterwarnings =
     error
```
webknjaz, line 26: Since the deprecation warning is happening in the third-party lib that we still support, we should ignore `DeprecationWarning` but only in that lib, not our own code base:
```suggestion
    # TODO: delete the following ignores once Ansible that we support gets rid of `imp`	
    # Ref: https://github.com/ansible/ansible-lint/pull/734
    ignore:the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses:DeprecationWarning:ansible.plugins.loader
```

webknjaz: (alternatively, it could be possible to not use that API from the third party that causes this, but only if it provides an alternate implementation)

lib/ansiblelint/utils.py
```diff
@@ -100,19 +100,14 @@ def ansible_template(basedir, varname, templatevars, **kwargs):
 
 def load_plugins(directory):
     result = []
-    fh = None
 
     for pluginfile in glob.glob(os.path.join(directory, '[A-Za-z]*.py')):
```
webknjaz, line 104:
```suggestion
    for pluginfile in glob.glob(os.path.join(directory, '[A-Za-z]*.py')):

```

ssbarnea:
@webknjaz This got merged while still having a "WIP" title...

webknjaz:
@ssbarnea no, `pytest.ini` needs to ignore things that happen in Ansible, not in ansible-lint.

ssbarnea:
@webknjaz Corrected title, good. I was more worried about the generating release notes with it. Can we try to enable Release Drafter?

webknjaz:
Honestly, I prefer towncrier. Also, this change does not look like it's worth including in the changelog because there's no immediate impact for the end-user.

PR #736, opened 2020-04-16
Cleanup py2 leftovers

ssbarnea:
Removes some code that was needed for py27.

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -24,7 +24,6 @@ import re
 import sys
 
 import ansiblelint.utils
-import codecs
 
 default_rulesdir = os.path.join(os.path.dirname(ansiblelint.utils.__file__), 'rules')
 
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/__init__.py
```diff
@@ -162,7 +161,7 @@ def run(self, playbookfile, tags=set(), skip_list=frozenset()):
         matches = list()
 
         try:
-            with codecs.open(playbookfile['path'], mode='rb', encoding='utf-8') as f:
+            with open(playbookfile['path'], mode='r', encoding='utf-8') as f:
```
webknjaz, line 164: Maybe it's better to keep binary mode?

ssbarnea: That is not an option, removal of "b" is mandatory as some supported versions of Python raise exception. In fact it makes no sense to use an encoding with a binary, these being exclusive. You can say that allowing it was a bug in codecs.

webknjaz: Ah, okay. I was concerned by us changing the type of the result variable and how this can affect the rest of the code.

lib/ansiblelint/utils.py
```diff
@@ -724,8 +721,7 @@ def normpath(path):
     Currently it generates a relative path but in the future we may want to
     make this user configurable.
     """
-    # convertion to string in order to allow receiving non string objects as
-    # arguments which would have failed under python2
+    # convertion to string in order to allow receiving non string objects
     return os.path.relpath(str(path))
```
webknjaz, line 725: _(note)_ In the future we may want to migrate the internal paths representation to pathlib.

PR #737, opened 2020-04-16
Run linters job groups on a single platform

ssbarnea:
Lower the CI footprint and chance of being affected by random failures. Non-core cannot ever retrigger CI without closing and reopening a PR.

Partial-Fix: #699

### DIFF
--- a/.github/workflows/tox-linters.yml
+++ b/.github/workflows/tox-linters.yml
@@ -19,13 +19,8 @@ jobs:
       matrix:
         python-version:
         - 3.8
-        - 3.7
-        - 3.6
-        - 3.5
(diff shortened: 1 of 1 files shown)


.github/workflows/tox-linters.yml
```diff
@@ -18,13 +18,13 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-        - 3.8
-        - 3.7
+        # - 3.8
+        # - 3.7
         - 3.6
```
webknjaz, line 23: I prefer having 3.8 instead and just remove the lines, no need to comment them out anymore.

PR #738, opened 2020-04-16
Gracefully process a missing git binary when falling-back to pure-python discovery

### LINKED ISSUE #735: [HELP WANTED] [SEND PR] Missing Git breaks the auto-discovery of YAML files
FTR here's the traceback from that log:
```python-traceback
  File "/usr/local/bin/ansible-lint", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/site-packages/ansiblelint/__main__.py", line 153, in main
    args = get_playbooks_and_roles(options=options)
  File "/usr/local/lib/python3.8/site-packages/ansiblelint/utils.py", line 772, in get_playbooks_and_roles
    files = OrderedDict.fromkeys(sorted(subprocess.check_output(
  File "/usr/local/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/local/lib/python3.8/subprocess.py", line 489, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/local/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'git'
```

_Originally posted by @webknjaz in https://github.com/ansible/ansible-lint-action/pull/26#issuecomment-614621653_

anryko:
This change fixes #735 by catching an extra corner case
exception `FileNotFoundError`.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -778,6 +778,8 @@ def get_yaml_files(options):
             )
         )
 
+    out = None
+
     try:
         out = subprocess.check_output(
             git_command,
@@ -792,7 +794,15 @@ def get_yaml_files(options):
                     error_msg=exc.output.rstrip('\n')
                 )
             )
+    except FileNotFoundError as exc:
+        if options.verbosity:
+            print(
+                "Warning: Failed to locate command: {error_msg!s}".format(
+                    error_msg=exc
+                )
+            )
 
+    if out is None:
         out = [
             os.path.join(root, name)
             for root, dirs, files in os.walk('.')
--- a/test/TestUtils.py
+++ b/test/TestUtils.py
@@ -25,6 +25,7 @@ import os
 from pathlib import Path
 
 import ansiblelint.utils as utils
+from ansiblelint import cli
 
 from importlib_metadata import version as get_dist_version
 from packaging.version import Version
@@ -185,3 +186,52 @@ def test_expand_paths_vars(monkeypatch):
     monkeypatch.setenv('TEST_PATH', test_path)
     assert utils.expand_paths_vars(['~']) == [os.path.expanduser('~')]
     assert utils.expand_paths_vars(['$TEST_PATH']) == [test_path]
+
+
+@pytest.mark.parametrize(
+    ('reset_env_var', 'message_prefix'),
+    (
+        ('PATH', 'Warning: Failed to locate command:'),
+        ('GIT_DIR', 'Warning: Failed to discover yaml files to lint using git:')
+    ),
+    ids=('no Git installed', 'outside Git repository'),
+)
+def test_get_yaml_files_git_verbose(
+    reset_env_var,
+    message_prefix,
+    monkeypatch,
+    capsys
+):
+    options, _ = cli.get_config(['-v'])
+    monkeypatch.setenv(reset_env_var, '')
+    utils.get_yaml_files(options)
+    msg1, msg2 = capsys.readouterr().out.splitlines()[:2]
+    assert msg1 == 'Discovering files to lint: git ls-files *.yaml *.yml'
+    assert msg2.startswith(message_prefix)
+
+
+@pytest.mark.parametrize(
+    'is_in_git',
+    (True, False),
+    ids=('in Git', 'outside Git'),
+)
+def test_get_yaml_files_silent(is_in_git, monkeypatch, capsys):
+    options, _ = cli.get_config([])
+    test_dir = Path(__file__).resolve().parent
+    lint_path = test_dir / 'roles' / 'test-role'
+    if not is_in_git:
+        monkeypatch.setenv('GIT_DIR', '')
+
+    yaml_count = (
+        len(list(lint_path.glob('**/*.yml'))) + len(list(lint_path.glob('**/*.yaml')))
+    )
+
+    monkeypatch.chdir(str(lint_path))
+    files = utils.get_yaml_files(options)
+    stdout = capsys.readouterr().out
+    assert not stdout, 'No output is expected when the verbosity is off'
+    assert len(files) == yaml_count, (
+        "Expected to find {yaml_count} yaml files in {lint_path}".format_map(
+            locals(),
+        )
+    )


lib/ansiblelint/utils.py
```diff
@@ -784,7 +784,7 @@ def get_yaml_files(options):
             stderr=subprocess.STDOUT,
             universal_newlines=True
         ).split()
-    except subprocess.CalledProcessError as exc:
+    except (subprocess.CalledProcessError, FileNotFoundError) as exc:
```
webknjaz, line 787: I think you need to also customize access to `exc.output` vs. `str(exc)` because `FileNotFoundError` probably doesn't have `.output` attr.

anryko: Sorry. Didn't check verbosity option. Fixed.

anryko:
@webknjaz could you suggest how the test should work?

webknjaz:
@anryko `monkeypatch.setenv('PATH', '/dev/null')` and then use `capsys` to check stdout for the verbose message (both are built-in pytest fixtures).

test/TestUtils.py
```diff
@@ -185,3 +186,36 @@ def test_expand_paths_vars(monkeypatch):
...
+    assert stdout[1].startswith('Warning: Failed to discover yaml files to lint using git:')
+    monkeypatch.chdir(sys_dir)
+
+    # Silent and returned dict is not empty
+    monkeypatch.chdir(role_path)
+    files = utils.get_yaml_files(options_silent)
+    stdout = capsys.readouterr().out
+    assert not stdout
+    assert len(files)
+    monkeypatch.chdir(sys_dir)
```
webknjaz, line 221: You don't need to restore anything. `monkeypatch` will revert the change upon exiting the test automatically.

test/TestUtils.py
```diff
@@ -185,3 +186,36 @@ def test_expand_paths_vars(monkeypatch):
     monkeypatch.setenv('TEST_PATH', test_path)
     assert utils.expand_paths_vars(['~']) == [os.path.expanduser('~')]
     assert utils.expand_paths_vars(['$TEST_PATH']) == [test_path]
+
+
+def test_get_yaml_files(monkeypatch, capsys):
+    options_silent, _ = cli.get_config([])
+    options_verbose, _ = cli.get_config(['-v'])
+    sys_path = os.getenv('PATH')
```
webknjaz, line 194: this is unnecessary

test/TestUtils.py
```diff
@@ -185,3 +186,36 @@ def test_expand_paths_vars(monkeypatch):
     monkeypatch.setenv('TEST_PATH', test_path)
     assert utils.expand_paths_vars(['~']) == [os.path.expanduser('~')]
     assert utils.expand_paths_vars(['$TEST_PATH']) == [test_path]
+
+
+def test_get_yaml_files(monkeypatch, capsys):
```
webknjaz, line 191: This should be either converted to use `@pytest.mark.parametrize` or separated into different tests. Currently, it contains too many tests in one. One test should assert one behavior.

anryko: Couldn't figure out a nice way to utilize `@pytest.mark.parametrize` for this case. Splitted.

webknjaz: Right, this'd be tricky to do but probably doable. OTOH we should always try to assess whether it's worth the complications because readability matters.

test/TestUtils.py
```diff
@@ -185,3 +186,32 @@ def test_expand_paths_vars(monkeypatch):
...
     assert utils.expand_paths_vars(['$TEST_PATH']) == [test_path]
+
+
+def test_get_yaml_files_no_git_verbose(monkeypatch, capsys):
+    options, _ = cli.get_config(['-v'])
+    monkeypatch.setenv('PATH', '')
+    utils.get_yaml_files(options)
+    stdout = capsys.readouterr().out.split('\n')
+    assert stdout[0] == 'Discovering files to lint: git ls-files *.yaml *.yml'
+    assert stdout[1].startswith('Warning: Failed to locate command:')
```
webknjaz, line 197: Can this be `==`?

webknjaz, line 197: [`str.splitlines()`](https://docs.python.org/3/library/stdtypes.html#str.splitlines) seems like a way to go here. Also, it'd be more readable if you named the lines:
```suggestion
    msg1, msg2 = capsys.readouterr().out.splitlines()
    assert msg1 == 'Discovering files to lint: git ls-files *.yaml *.yml'
    assert msg2.startswith('Warning: Failed to locate command:')
```

test/TestUtils.py
```diff
@@ -185,3 +186,32 @@ def test_expand_paths_vars(monkeypatch):
...
+
+
+def test_get_yaml_files_not_git_verbose(monkeypatch, capsys):
+    options, _ = cli.get_config(['-v'])
+    monkeypatch.chdir('/tmp')
+    utils.get_yaml_files(options)
+    stdout = capsys.readouterr().out.split('\n')
+    assert stdout[0] == 'Discovering files to lint: git ls-files *.yaml *.yml'
+    assert stdout[1].startswith('Warning: Failed to discover yaml files to lint using git:')
```
webknjaz, line 206: I think this is missing the check about the amount of files.

webknjaz: Also, we probably need another test with rules but no git repo.
You can use the [`tmp_path` fixture](https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture) for getting a tmp dir, copy rules there and `monkeypatch.chdir()` to that location.

test/TestUtils.py
```diff
@@ -185,3 +186,32 @@ def test_expand_paths_vars(monkeypatch):
...
+
+def test_get_yaml_files_silent(monkeypatch, capsys):
+    options, _ = cli.get_config([])
+    test_dir = os.path.dirname(os.path.realpath(__file__))
+    role_path = os.path.join(test_dir, 'roles/test-role')
+    monkeypatch.chdir(role_path)
+    files = utils.get_yaml_files(options)
+    stdout = capsys.readouterr().out
+    assert not stdout
+    assert len(files)
```
webknjaz, line 217: The semantical meaning of these assertions is not quite obvious:
```suggestion
    assert not stdout, 'No output is expected when the verbosity is off'
    assert len(files), 'The fallback is expected to find some files'
```
(but was is fallback or git? this could be a great case for `@pytest.mark.parametrize` with the same asserts but with git binary present&git repo, no binary or no repo)

test/TestUtils.py
anryko: After the warning, original exception message is printed, which theoretically might differ depending on python platform or version.

anryko: Exception message might be multiline (end with '\n'). But yep, this is nice, I'll do a variation of that.

webknjaz: Maybe match until the exception name then?

test/TestUtils.py
```diff
@@ -185,3 +187,54 @@ def test_expand_paths_vars(monkeypatch):
...
+    assert msg2.startswith('Warning: Failed to discover yaml files to lint using git:')
+
+
+@pytest.mark.parametrize(
+    'is_in_git',
+    [
+        (True),
+        (False)
+    ]
+)
```
webknjaz, line 216: Just `(value)` is that same as `value` so braces are unnecessary. But if you do `(value, )` it's a tuple. Still, it's not needed there because you have just one param.

This linter error https://github.com/ansible/ansible-lint/pull/738/checks?check_run_id=611810102#step:10:26 is about having that outer list converted into a tuple. Not the params.

Also, the params list is short enough so it could fit in one line (unless you want to multiline it intentionally):
```suggestion
@pytest.mark.parametrize('is_in_git', (True, False))
```

test/TestUtils.py
```diff
@@ -185,3 +187,48 @@ def test_expand_paths_vars(monkeypatch):
...
+    options, _ = cli.get_config([])
+    test_dir = Path(os.path.dirname(os.path.realpath(__file__)))
+    role_sub_dir = Path('roles/test-role')
+    role_dir = test_dir / 'roles/test-role'
+
+    if is_in_git:
+        lint_path = role_dir
+    else:
+        lint_path = tmp_path / role_sub_dir
+        shutil.copytree(role_dir, lint_path)
```
webknjaz, line 221: What if we tried patching the env var instead?
https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_repository_locations
```suggestion
    lint_path = role_dir
    if not is_in_git:
        monkeypatch.setenv('GIT_DIR', '/dev/null')
```

test/TestUtils.py
```diff
@@ -185,3 +187,48 @@ def test_expand_paths_vars(monkeypatch):
...
+    assert msg1 == 'Discovering files to lint: git ls-files *.yaml *.yml'
+    assert msg2.startswith('Warning: Failed to discover yaml files to lint using git:')
+
+
+@pytest.mark.parametrize('is_in_git', (True, False))
+def test_get_yaml_files_silent(is_in_git, monkeypatch, capsys, tmp_path):
+    options, _ = cli.get_config([])
+    test_dir = Path(os.path.dirname(os.path.realpath(__file__)))
+    role_sub_dir = Path('roles/test-role')
+    role_dir = test_dir / 'roles/test-role'
```
webknjaz, line 215: typo?
```suggestion
    role_sub_dir = Path('roles/test-role')
    role_dir = test_dir / role_sub_dir
```

test/TestUtils.py
```diff
@@ -185,3 +187,48 @@ def test_expand_paths_vars(monkeypatch):
...
+    monkeypatch.setenv('PATH', '')
+    utils.get_yaml_files(options)
+    msg1, msg2 = capsys.readouterr().out.splitlines()[:2]
+    assert msg1 == 'Discovering files to lint: git ls-files *.yaml *.yml'
+    assert msg2.startswith('Warning: Failed to locate command:')
+
+
+def test_get_yaml_files_not_git_verbose(monkeypatch, capsys):
+    options, _ = cli.get_config(['-v'])
+    monkeypatch.chdir('/tmp')
```
webknjaz, line 203: Use a tmp path from the fixture. Otherwise, pointing to a global top may result in a slow and unstable test (you never know what's on somebody else's machine in `/tmp` or how bit it is)
```suggestion
    monkeypatch.chdir(tmp_path)
```

test/TestUtils.py
```diff
@@ -185,3 +187,48 @@ def test_expand_paths_vars(monkeypatch):
...
+    role_dir = test_dir / 'roles/test-role'
+
+    if is_in_git:
+        lint_path = role_dir
+    else:
+        lint_path = tmp_path / role_sub_dir
+        shutil.copytree(role_dir, lint_path)
+
+    yaml_count = (
+        len(list(lint_path.glob('**/*.yml'))) + len(list(lint_path.glob('**/*.yaml')))
```
webknjaz, line 224: maybe smth like
```suggestion
        sum(1 for _ in itertools.chain.from_iterable(lint_path.glob('**/*.yml'), lint_path.glob('**/*.yaml')))
```

test/TestUtils.py
```diff
@@ -185,3 +187,48 @@ def test_expand_paths_vars(monkeypatch):
...
+    )
+
+    monkeypatch.chdir(lint_path)
+    files = utils.get_yaml_files(options)
+    stdout = capsys.readouterr().out
+    assert not stdout, 'No output is expected when the verbosity is off'
+    assert len(files) == yaml_count, "Expected to find {count} yaml files in {path}".format(
+        count=yaml_count,
+        path=lint_path
+    )
```
webknjaz, line 234: There's a nice trick for f-string-like interpolation:
```suggestion
    assert len(files) == yaml_count, "Expected to find {count} yaml files in {path}".format_map(
        locals(),
    )
```

test/TestUtils.py
```diff
@@ -185,3 +187,48 @@ def test_expand_paths_vars(monkeypatch):
...
+def test_get_yaml_files_not_git_verbose(monkeypatch, capsys):
+    options, _ = cli.get_config(['-v'])
+    monkeypatch.chdir('/tmp')
+    utils.get_yaml_files(options)
+    msg1, msg2 = capsys.readouterr().out.splitlines()[:2]
+    assert msg1 == 'Discovering files to lint: git ls-files *.yaml *.yml'
+    assert msg2.startswith('Warning: Failed to discover yaml files to lint using git:')
+
+
+@pytest.mark.parametrize('is_in_git', (True, False))
```
webknjaz, line 210: On the second thought, it's probably worth labeling params:
```suggestion
@pytest.mark.parametrize(
    'is_in_git',
    (True, False),
    ids=('in Git', 'outside Git'),
)
```

test/TestUtils.py
anryko: The full message in my case is `Warning: Failed to locate command: [Errno 2] No such file or directory: 'git': 'git'`.

test/TestUtils.py
anryko: Yep, this is much better!

test/TestUtils.py
anryko: The `itertools.chain.from_iterable` takes only one iterable.

webknjaz: Hm.. I'm pretty sure I made it work with many in the past. Nevermind then.

test/TestUtils.py
```diff
@@ -185,3 +186,53 @@ def test_expand_paths_vars(monkeypatch):
     monkeypatch.setenv('TEST_PATH', test_path)
     assert utils.expand_paths_vars(['~']) == [os.path.expanduser('~')]
     assert utils.expand_paths_vars(['$TEST_PATH']) == [test_path]
+
+
+@pytest.mark.parametrize(
+    'reset_env_var, message_prefix',
```
webknjaz, line 192: this will fix the linter violation
```suggestion
    ('reset_env_var', 'message_prefix'),
```

test/TestUtils.py
```diff
@@ -185,3 +186,53 @@ def test_expand_paths_vars(monkeypatch):
...
+@pytest.mark.parametrize(
+    'is_in_git',
+    (True, False),
+    ids=('in Git', 'outside Git'),
+)
+def test_get_yaml_files_silent(is_in_git, monkeypatch, capsys):
+    options, _ = cli.get_config([])
+    test_dir = Path(os.path.dirname(os.path.realpath(__file__)))
+    role_sub_dir = Path('roles/test-role')
+    lint_path = test_dir / role_sub_dir
```
webknjaz, line 222: It's only used once:
```suggestion
    lint_path = test_dir / 'roles' / 'test-role'
```

test/TestUtils.py
```diff
@@ -185,3 +186,52 @@ def test_expand_paths_vars(monkeypatch):
...
+
+
+@pytest.mark.parametrize(
+    'is_in_git',
+    (True, False),
+    ids=('in Git', 'outside Git'),
+)
+def test_get_yaml_files_silent(is_in_git, monkeypatch, capsys):
+    options, _ = cli.get_config([])
+    test_dir = Path(os.path.dirname(os.path.realpath(__file__)))
```
webknjaz, line 220: `__file__` is pre-populated by the interpreter, it's probably already resolved. But anyway, `os.path` is unnecessary here:
```suggestion
    test_dir = Path(__file__).resolve().parent
```

test/TestUtils.py
```diff
@@ -185,3 +186,52 @@ def test_expand_paths_vars(monkeypatch):
...
+    test_dir = Path(__file__).resolve().parent
+    lint_path = test_dir / 'roles' / 'test-role'
+    if not is_in_git:
+        monkeypatch.setenv('GIT_DIR', '')
+
+    yaml_count = (
+        len(list(lint_path.glob('**/*.yml'))) + len(list(lint_path.glob('**/*.yaml')))
+    )
+
+    monkeypatch.chdir(lint_path)
```
webknjaz, line 229: py 3.5 compat
```suggestion
    monkeypatch.chdir(str(lint_path))
```

anryko:
@webknjaz thanks! Great review.

Issue #739, opened 2020-04-19
Tests load python modules from hazardous locations (~, ...)


cans:
# Issue Type

- Bug report

# Ansible and Ansible Lint details

```
ansible --version
ansible 2.9.7
  config file = ~/.ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/.virtualenvs/ansible-lint/lib/python3.7/site-packages/ansible
  executable location = ~/.virtualenvs/ansible-lint/bin/ansible
  python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.0]
ansible-lint --version
ansible-lint 4.3.0a1.dev18+g4171317
```

- ansible installation method: pip
- ansible-lint installation method: source

# Desired Behaviour

Ansible-lint test suite should not load Python modules from arbitrary locations.

# Actual Behaviour (Bug report only)

When running the tests, on my machine, the following test failed.
```
test/TestRulesCollection.py::test_rulesdir_user_expansion FAILED
```
 The reason is that I have a .py file in my home directory that imports a module that is not available in my ansible-lint development virtual env. This is a security hazard: I could have had a python module that did far worst than fail to load and fail the test.

In previous PR (#572) we laid some ground work to make sure passed reading the configuration, and merging it with command-line options, ansible-lint only works with absolute path. Additional work was made in #561 (not yet merged, may need additional work) to ensure ansible-lint does not work with relative path internally. 

There is one missing bit which is variable expansion, for `rulesdir` values. Hence I believe the `ansiblelint.cli.get_config()` function should be either modified to use `ansiblelint.utils.expand_paths_vars()` or current `ansibleline.cli.abspath()` function be adapted to allow variable expansion.

Then the call to the former within the `RulesCollection.__init__()` should no longer be necessary. The test above should become useless, thus removed along with its hazardous behavior.

evilhamsterman:
I added the path expansion because I and many others like having a set of rules and excluded paths (galaxy installed roles) that are shared among multiple projects. If there's a cleaner way to accomplish that goal I'm fine.

I do think the path expansion is necessary for the excluded paths because that may need to point to an environment variable such as one that points to the Galaxy Roles install path. That shouldn't be an issue because its not loading code just paths to specifically not load.

[...]

Issue #725, opened 2020-04-07
[HELP WANTED] [SEND PRs] Make use of `pytest` features when running tests


webknjaz:
We've switched over the test runner to `pytest` and there are a few tests
that already use it natively too. Still, the transition isn't over yet, it's an ongoing effort and we ask for the help of the community to get this over with faster.

Help wanted
===========

Here's how you can help:
1. When adding new test modules (or just test functions), never import `unittest`. Follow, `pytest` conventions, use fixtures. Use test functions, classes are unnecessary here.
Use `@pytest.mark.parametrize` to have the same test body run against multiple inputs: each one of them will show up in the report separately. Use `assert` statement, don't inherit from `unittest.TestCase`. Use [built-in `tmp_path` fixture](https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture) (*not `tmpdir` — it's deprecated*) to work with temporary data on disk and make sure that multiple tests don't hit the resources at the same time.
2. When reporting a bug, write an integration test as a new `pytest` test function. It should be failing. Cover as many edge and negative cases as you can.
Send a PR demonstrating the red CI. After that, add one separate commit on top of that in the PR branch adding `@pytest.mark.xfail` as per https://pganssle-talks.github.io/xfail-lightning. Use `raises=` arg if it's not the assertion that's failing. Follow the rules from (1).
3. Send us separate PRs migrating legacy `unittest`-based modules to `pytest`. Follow the rules from (1).

`pytest`-based test module example
==================================

https://github.com/ansible/ansible-lint/blob/065faa5/test/TestIncludeMissingFileRule.py

webknjaz:
UPD: https://github.com/ansible/ansible-lint/pull/945 is going to help with enforcing the use of pytest over unittest. If one were to send PRs updating the ignored modules, they'll need to drop the ignore-comments from the top.

PR #740, opened 2020-04-19
Migrate some test to pytest

cans:
Contributes to #725 but does not pretend to be a complete migration.

Those changes are made to factor in codes that will be impacted by fixing issue #739.

### DIFF
--- a/test/TestDependenciesInMeta.py
+++ b/test/TestDependenciesInMeta.py
@@ -1,35 +1,22 @@
-import os
-import unittest
+import pytest
 
-from ansiblelint import Runner, RulesCollection
+from ansiblelint import Runner
 
 
-class TestDependenciesInMeta(unittest.TestCase):
-
-    def setUp(self):
-        rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
-        self.rules = RulesCollection([rulesdir])
-
-    def test_bitbucket_in_meta_dependency_is_ok(self):
-        filename = 'test/dependency-in-meta/bitbucket.yml'
-        self.assertEqual([], Runner(self.rules, filename, [], [], []).run())
-
-    def test_galaxy_dependency_is_ok(self):
-        filename = 'test/dependency-in-meta/galaxy.yml'
-        good_runner = Runner(self.rules, filename, [], [], [])
-        self.assertEqual([], good_runner.run())
-
-    def test_github_dependency_is_ok(self):
-        filename = 'test/dependency-in-meta/github.yml'
-        good_runner = Runner(self.rules, filename, [], [], [])
-        self.assertEqual([], good_runner.run())
-
-    def test_webserver_dependency_is_ok(self):
-        filename = 'test/dependency-in-meta/webserver.yml'
-        good_runner = Runner(self.rules, filename, [], [], [])
-        self.assertEqual([], good_runner.run())
-
-    def test_gitlab_dependency_is_ok(self):
-        filename = 'test/dependency-in-meta/gitlab.yml'
-        good_runner = Runner(self.rules, filename, [], [], [])
-        self.assertEqual([], good_runner.run())
+@pytest.mark.parametrize(
+    'filename',
+    (
+        'bitbucket',
+        'galaxy',
+        'github',
+        'webserver',
+        'gitlab',
+    ),
+)
+def test_external_dependency_is_ok(default_rules_collection, filename):
+    playbook_path = (
+        'test/dependency-in-meta/{filename}.yml'.
+        format_map(locals())
+    )
+    good_runner = Runner(default_rules_collection, playbook_path, [], [], [])
+    assert [] == good_runner.run()
--- a/test/TestImportIncludeRole.py
+++ b/test/TestImportIncludeRole.py
@@ -1,10 +1,7 @@
-import os
 import pytest
-import unittest
-import tempfile
-import shutil
 
-from ansiblelint import Runner, RulesCollection
+from ansiblelint import Runner
+
 
 ROLE_TASKS_MAIN = '''
 - name: shell instead of command
@@ -55,56 +52,52 @@ PLAY_INCLUDE_ROLE_INLINE = '''
 '''
 
 
-class TestImportIncludeRole(unittest.TestCase):
-    def setUp(self):
-        rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
-        self.rules = RulesCollection([rulesdir])
+@pytest.fixture
+def playbook_path(request, tmp_path):
+    playbook_text = request.param
+    role_tasks_dir = tmp_path / 'test-role' / 'tasks'
+    role_tasks_dir.mkdir(parents=True)
+    (role_tasks_dir / 'main.yml').write_text(ROLE_TASKS_MAIN)
+    (role_tasks_dir / 'world.yml').write_text(ROLE_TASKS_WORLD)
+    play_path = tmp_path / 'playbook.yml'
+    play_path.write_text(playbook_text)
+    return str(play_path)
 
-        # make dir and write role tasks to import or include
-        self.play_root = tempfile.mkdtemp()
-        role_path = os.path.join(self.play_root, 'test-role', 'tasks')
-        os.makedirs(role_path)
-        with open(os.path.join(role_path, 'main.yml'), 'w') as f_main:
-            f_main.write(ROLE_TASKS_MAIN)
-        with open(os.path.join(role_path, 'world.yml'), 'w') as f_world:
-            f_world.write(ROLE_TASKS_WORLD)
 
-    def tearDown(self):
-        shutil.rmtree(self.play_root)
+@pytest.mark.parametrize(('playbook_path', 'messages'), (
+    pytest.param(PLAY_IMPORT_ROLE,
+                 ['only when shell functionality is required'],
+                 id='IMPORT_ROLE',
+                 ),
+    pytest.param(PLAY_IMPORT_ROLE_INLINE,
+                 ['only when shell functionality is require'],
+                 id='IMPORT_ROLE_INLINE',
+                 ),
+    pytest.param(PLAY_INCLUDE_ROLE,
+                 ['only when shell functionality is require',
+                  'All tasks should be named'],
+                 id='INCLUDE_ROLE',
+                 ),
+    pytest.param(PLAY_INCLUDE_ROLE_INLINE,
+                 ['only when shell functionality is require',
+                  'All tasks should be named'],
+                 id='INCLUDE_ROLE_INLINE',
+                 ),
+), indirect=('playbook_path', ))
+def test_import_role2(default_rules_collection, playbook_path, messages):
+    runner = Runner(default_rules_collection, playbook_path, [], [], [])
+    results = runner.run()
+    for message in messages:
+        assert message in str(results)
 
-    def _get_play_file(self, playbook_text):
-        with open(os.path.join(self.play_root, 'playbook.yml'), 'w') as f_play:
-            f_play.write(playbook_text)
-        return f_play
 
-    def test_import_role(self):
-        fh = self._get_play_file(PLAY_IMPORT_ROLE)
-        runner = Runner(self.rules, fh.name, [], [], [])
-        results = runner.run()
-        assert 'only when shell functionality is required' in str(results)
-
-    def test_import_role_inline_args(self):
-        fh = self._get_play_file(PLAY_IMPORT_ROLE_INLINE)
-        runner = Runner(self.rules, fh.name, [], [], [])
-        results = runner.run()
-        assert 'only when shell functionality is required' in str(results)
-
-    def test_include_role(self):
-        fh = self._get_play_file(PLAY_INCLUDE_ROLE)
-        runner = Runner(self.rules, fh.name, [], [], [])
-        results = runner.run()
-        assert 'only when shell functionality is required' in str(results)
-        assert 'All tasks should be named' in str(results)
-
-    def test_include_role_inline_args(self):
-        fh = self._get_play_file(PLAY_INCLUDE_ROLE_INLINE)
-        runner = Runner(self.rules, fh.name, [], [], [])
-        results = runner.run()
-        assert 'only when shell functionality is required' in str(results)
-        assert 'All tasks should be named' in str(results)
-
-    def test_invalid_import_role(self):
-        fh = self._get_play_file(PLAY_IMPORT_ROLE_INCOMPLETE)
-        runner = Runner(self.rules, fh.name, [], [], [])
-        with pytest.raises(RuntimeError, match="Failed to find required 'name' key in import_role"):
-            runner.run()
+@pytest.mark.parametrize(('playbook_path', 'error'), (
+    pytest.param(PLAY_IMPORT_ROLE_INCOMPLETE,
+                 "Failed to find required 'name' key in import_role",
+                 id='IMPORT_ROLE_INCOMPLETE',
+                 ),
+), indirect=('playbook_path', ))
+def test_invalid_import_role(default_rules_collection, playbook_path, error):
+    runner = Runner(default_rules_collection, playbook_path, [], [], [])
+    with pytest.raises(RuntimeError, match=error):
+        runner.run()
--- a/test/TestPretaskIncludePlaybook.py
+++ b/test/TestPretaskIncludePlaybook.py
@@ -1,35 +1,20 @@
-import os
-import unittest
+import pytest
 
-from ansiblelint import Runner, RulesCollection
+from ansiblelint import Runner
 
 
-class TestTaskIncludes(unittest.TestCase):
+@pytest.mark.parametrize('stage', ('pre', 'post'))
+def test_task_hook_include_playbook(default_rules_collection, stage):
+    playbook_path = (
+        'test/playbook-include/playbook_{stage}.yml'.
+        format_map(locals())
+    )
+    runner = Runner(default_rules_collection, playbook_path, [], [], [])
+    results = runner.run()
 
-    def setUp(self):
-        rulesdir = os.path.join('lib', 'ansiblelint', 'rules')
-        self.rules = RulesCollection([rulesdir])
-
-    def test_pre_task_include_playbook(self):
-        filename = 'test/playbook-include/playbook_pre.yml'
-        runner = Runner(self.rules, filename, [], [], [])
-        results = runner.run()
-
-        self.assertEqual(len(runner.playbooks), 2)
-        self.assertEqual(len(results), 3)
-        self.assertIn('Commands should not change things', str(results))
-
-        self.assertNotIn('502', str(results))
-        self.assertNotIn('All tasks should be named', str(results))
-
-    def test_post_task_include_playbook(self):
-        filename = 'test/playbook-include/playbook_post.yml'
-        runner = Runner(self.rules, filename, [], [], [])
-        results = runner.run()
-
-        self.assertEqual(len(runner.playbooks), 2)
-        self.assertEqual(len(results), 3)
-        self.assertIn('Commands should not change things', str(results))
-
-        self.assertNotIn('502', str(results))
-        self.assertNotIn('All tasks should be named', str(results))
+    results_text = str(results)
+    assert len(runner.playbooks) == 2
+    assert len(results) == 3
+    assert 'Commands should not change things' in results_text
+    assert '502' not in results_text
+    assert 'All tasks should be named' not in results_text
--- a/test/TestRulesCollection.py
+++ b/test/TestRulesCollection.py
@@ -19,76 +19,76 @@
 # THE SOFTWARE.
 
 import collections
-import unittest
 import os
+
+import pytest
 
 from ansiblelint import RulesCollection
 
 
-class TestRulesCollection(unittest.TestCase):
-
-    rules = None
-    ematchtestfile = dict(path='test/ematchtest.yml', type='playbook')
-    bracketsmatchtestfile = dict(path='test/bracketsmatchtest.yml', type='playbook')
-
-    def setUp(self):
-        self.rules = RulesCollection(['./test/rules'])
-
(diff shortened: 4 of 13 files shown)


test/TestImportIncludeRole.py
```diff
@@ -55,56 +57,68 @@
...
+    os.makedirs(role_path)
+    with open(os.path.join(role_path, 'main.yml'), 'w') as f_main:
+        f_main.write(ROLE_TASKS_MAIN)
+    with open(os.path.join(role_path, 'world.yml'), 'w') as f_world:
+        f_world.write(ROLE_TASKS_WORLD)
+
+    play_path = os.path.join(tmpdir, 'playbook.yml')
+    with open(play_path, 'w') as f_play:
+        f_play.write(playbook_text)
+    return play_path
```
webknjaz, line 72: FYI tmpdir is a pathlib.Path-like object. So instead of converting things to old-fashioned strings, you should use its interface:

```suggestion
    role_tasks_dir = tmpdir / 'test-role' / 'tasks'
    role_tasks_dir.mkdir(parents=True)

    (role_tasks_dir / 'main.yml').write_text(ROLE_TASKS_MAIN)
    (role_tasks_dir / 'world.yml').write_text(ROLE_TASKS_WORLD)

    play_path = tmpdir / 'playbook.yml'
    play_path.write_text(playbook_text)
    return play_path
```

Ref: https://docs.pytest.org/en/latest/tmpdir.html

test/TestImportIncludeRole.py
```diff
@@ -55,56 +57,68 @@
...
-        results = runner.run()
-        assert 'only when shell functionality is required' in str(results)
-        assert 'All tasks should be named' in str(results)
-
-    def test_invalid_import_role(self):
-        fh = self._get_play_file(PLAY_IMPORT_ROLE_INCOMPLETE)
-        runner = Runner(self.rules, fh.name, [], [], [])
-        with pytest.raises(RuntimeError, match="Failed to find required 'name' key in import_role"):
-            runner.run()
+def _setup_play(tmpdir, playbook_text):
```
webknjaz, line 60: This should be a fixture

test/conftest.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8; -*-
```
webknjaz, line 1: This is unnecessary, we don't care about Python 2 anymore.

test/TestTaskIncludes.py
```diff
@@ -1,53 +1,50 @@
...
+    runner = Runner(default_rules_collection, filename, [], [], [])
+    runner.run()
+    assert len(runner.playbooks) == 3
+
+
+def test_include_tasks_in_role(default_rules_collection):
+    filename = 'test/include-import-tasks-in-role.yml'
+    runner = Runner(default_rules_collection, filename, [], [], [])
+    runner.run()
+    assert len(runner.playbooks) == 4
```
webknjaz, line 50: This is essentially a single test with different params:

```suggestion
import pytest


@pytest.mark.parametrize(
    ('filename', 'playbooks_count'),
    (
        pytest.param('blockincludes', 4, id='block included tasks'),
        pytest.param(
            'blockincludes2', 4,
            id='block included tasks with rescue and always',
        ),
        pytest.param('taskincludes', 4, id='included tasks'),
        pytest.param(
            'taskincludes_2_4_style', 4,
            id='include tasks 2.4 style',
        ),
        pytest.param('taskimports', 4, id='import tasks 2 4 style'),
        pytest.param(
            'include-in-block', 3,
            id='include tasks with block include',
        ),
        pytest.param(
            'include-import-tasks-in-role', 4,
            id='include tasks in role',
        ),
    ),
)
def test_block_included_tasks(
        default_rules_collection,
        filename, playbooks_count,
):
    playbook_path = 'test/{filename}.yml'.format(locals())
    runner = Runner(default_rules_collection, playbook_path, [], [], [])

    runner.run()
    assert len(runner.playbooks) == playbooks_count
```

test/TestRulesCollection.py
```diff
@@ -19,76 +19,76 @@
...
+
+def test_skip_non_existent_id(test_rules_collection, ematchtestfile):
+    matches = test_rules_collection.run(ematchtestfile, skip_list=['DOESNOTEXIST'])
+    assert len(matches) == 3
+
+
+def test_no_duplicate_rule_ids(test_rules_collection):
+    real_rules = RulesCollection([os.path.abspath('./lib/ansiblelint/rules')])
+    rule_ids = [rule.id for rule in real_rules]
+    assert [x for x, y in collections.Counter(rule_ids).items() if y > 1] == []
```
webknjaz, line 94: I think this could be simplified:
```suggestion
    assert not any(y > 1 for x in collections.Counter(rule_ids).values())
```

webknjaz:
@cans I've added some pytest linters so this needs rebasing

test/TestImportIncludeRole.py
cans: Yes and No. Changelog says yes, since pytest 3.9, then around 4.x it seems to revert back to something else (referred to as `py.path.local`). On my computer with PyTest 5.2.2 / Py3.7 this is not a `pathlib.Path`, it is a `py._path.local.LocalPath` (displayed as `local('<path>')`) and that does not behave like a `Path` (_e.g._ its `mkdir()` method does not support the `parents` keyword argument). I guess there is some portability stuff still around in PyTest's code. Maybe the situation will become clearer in the future.

In the meantime, If I switch to `tmp_path`, which is a `pathlib.Path`, then later on the playbook fails to load. It happens somewhere in Ansible's API which, for portability's sake we have to avoid. So I went for the swtich to `tmp_path` + `return str(play_path)` on the last statement. Not as elegant as what you suggested, but hopefully close enough.

Traceback relative to the ansible issue:
```
test/TestImportIncludeRole.py:124: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
lib/ansiblelint/__init__.py:266: in run
    for child in ansiblelint.utils.find_children(arg, self.playbook_dir):
lib/ansiblelint/utils.py:156: in find_children
    playbook_ds = parse_yaml_from_file(playbook[0])
lib/ansiblelint/utils.py:60: in parse_yaml_from_file
    return dl.load_from_file(filepath)
../../../../.virtualenvs/ansible-lint/lib/python3.7/site-packages/ansible/parsing/dataloader.py:85: in load_from_file
    file_name = self.path_dwim(file_name)
../../../../.virtualenvs/ansible-lint/lib/python3.7/site-packages/ansible/parsing/dataloader.py:186: in path_dwim
    given = unquote(given)
../../../../.virtualenvs/ansible-lint/lib/python3.7/site-packages/ansible/parsing/quoting.py:29: in unquote
    if is_quoted(data):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

data = PosixPath('/tmp/pytest-of-user/pytest-47/test_invalid_import_role_IMPOR0/playbook.yml')

    def is_quoted(data):
>       return len(data) > 1 and data[0] == data[-1] and data[0] in ('"', "'") and data[-2] != '\\'
E       TypeError: object of type 'PosixPath' has no len()

data       = PosixPath('/tmp/pytest-of-user/pytest-47/test_invalid_import_role_IMPOR0/playbook.yml')

../../../../.virtualenvs/ansible-lint/lib/python3.7/site-packages/ansible/parsing/quoting.py:24: TypeError
```

test/TestImportIncludeRole.py
cans: Not sure, I use it as a helper to define fixtures (see a few lines below) but I needs to keep its arguments free. Hand you cannot call a fixture function directly, so making this a fixture would defeat the purpose. Though to avoid confusion I renamed the `tmpdir` parameter to `base_dir`.

test/TestTaskIncludes.py
cans: Indeed, fixed.

test/TestRulesCollection.py
cans: Fixed (modulo a typo `any(**x** > 1...)`).

test/conftest.py
cans: This is not just about Python 2... though I would not fight too hard on this (I am old, I still work with Emacs)

cans:
@webknjaz I don't know how your CI is setup, but I never understood how to ask pre-commit to install flake8 plugins (I'd be happy to know if you find a way), and they indeed did not get installed on my machine. Hence the checks you run on your CI are stricter than those pre-commit will run locally, resulting in failing tests (e.g. https://github.com/ansible/ansible-lint/pull/740/checks?check_run_id=603029144)

This is a bit annoying.

test/conftest.py
webknjaz: *works in Vim*

test/TestTaskIncludes.py
webknjaz: FYI it's a good tone to accept suggestions on GitHub UI because it preserves authorship. If you want to squash, you can then do this by preserving `Co-Authored-By` commit trailers

test/TestRulesCollection.py
webknjaz: ah, right. coding in browser is dangerous :)

test/TestImportIncludeRole.py
webknjaz: Well, you can call fixture functions by using smth like `request.getfixture('fixturename')` but I guess you cannot pass params explicitly.

test/TestImportIncludeRole.py
webknjaz: Well, in this case, I'd do `str(path)` when passing the value to the library calls but keep the pathlib object where you can control it. But I guess we could skip this for now.

webknjaz:
> I don't know how your CI is setup, but I never understood how to ask pre-commit to install flake8 plugins (I'd be happy to know if you have/find a way), and they indeed did not get installed on my machine.

pre-commit installs extra pip deps automatically, like this:
https://github.com/ansible/ansible-lint/blob/cd14978/.pre-commit-config.yaml#L42-L45. If it doesn't for you, it's a bug or you have an outdated version of it.

Are you running it via `tox`? `tox -e lint` should do the trick (that's what the CIs execute and is tested by many people under many OSs).
If it doesn't work, try `rm -rf ~/.pre-commit ~/.cache/pre-commit .tox && pip install --upgrade tox virtualenv && tox -e lint` and see if that fixes it.



> Tests failed 'cause of a network error:

Right, GH seems to be having outages today and I saw some unicorns yesterday too. Restarted.

webknjaz:
recheck

test/TestDependenciesInMeta.py
```diff
@@ -1,35 +1,31 @@
...
 
-    def test_gitlab_dependency_is_ok(self):
-        filename = 'test/dependency-in-meta/gitlab.yml'
-        good_runner = Runner(self.rules, filename, [], [], [])
-        self.assertEqual([], good_runner.run())
+
+def test_gitlab_dependency_is_ok(default_rules_collection):
+    filename = 'test/dependency-in-meta/gitlab.yml'
+    good_runner = Runner(default_rules_collection, filename, [], [], [])
+    assert [] == good_runner.run()
```
webknjaz, line 31: This is also a single test:
```suggestion
import pytest


@pytest.mark.parametrize(
    'filename',
    (
        'bitbucket',
        'galaxy',
        'github',
        'webserver',
        'gitlab',
    ),
)
def test_external_dependency_is_ok(default_rules_collection, filename):
    playbook_path = (
        'test/dependency-in-meta/{filename}.yml'.
        format_map(locals())
    )
    good_runner = Runner(default_rules_collection, playbook_path, [], [], [])

    assert [] == good_runner.run()
```

test/TestImportIncludeRole.py
```diff
@@ -55,56 +56,67 @@
...
+    playbook, messages = play_and_messages
+    runner = Runner(default_rules_collection, playbook, [], [], [])
+
+    results = runner.run()
+
+    for message in messages:
+        assert message in str(results)
+
+
+def test_invalid_import_role(default_rules_collection, play_and_error):
```
webknjaz, line 118: It'd be easier to follow if parametrize was applied here and not hidden in the fixture implementation or maybe if there were fixtures decoupling playbook&error back into two args

test/TestPretaskIncludePlaybook.py
```diff
@@ -1,35 +1,27 @@
...
-        results = runner.run()
-
-        self.assertEqual(len(runner.playbooks), 2)
-        self.assertEqual(len(results), 3)
-        self.assertIn('Commands should not change things', str(results))
-
-        self.assertNotIn('502', str(results))
-        self.assertNotIn('All tasks should be named', str(results))
+    assert '502' not in str(results)
+    assert 'All tasks should be named' not in str(results)
```
webknjaz, line 27: This is a single test:
```suggestion
import pytest


@pytest.mark.parametrize('stage', ('pre', 'post'))
def test_post_task_include_playbook(default_rules_collection, stage):
    playbook_path = (
        'test/playbook-include/playbook_{stage}.yml'.
        format_map(locals())
    )
    runner = Runner(default_rules_collection, playbook_path, [], [], [])
    results = runner.run()
    results_text = str(results)

    assert len(runner.playbooks) == 2
    assert len(results) == 3

    assert 'Commands should not change things' in results_text
    assert '502' not in results_text
    assert 'All tasks should be named' not in results_text
```

test/TestRunner.py
```diff
@@ -17,121 +17,109 @@
...
+
+    filename = os.path.abspath('test/common-include-1.yml')
+    runner = Runner(default_rules_collection, filename, [], [], [], 0, checked_files)
+    run1 = runner.run()
+
+    filename = os.path.abspath('test/common-include-2.yml')
+    runner = Runner(default_rules_collection, filename, [], [], [], 0, checked_files)
+    run2 = runner.run()
+
+    assert ((len(run1) + len(run2)) == 1)
```
webknjaz, line 125: note: test in this module are also parametrizable but this can be postponed.

webknjaz:
@cans I've figured out where the confusion with the `tmpdir` fixture `pathlib.Path` vs `py.path.local` is coming from: `tmpdir` is deprecated and is supposed to be replaced by `tmp_path`. As per [the pytest fixtures list] the suggests that there's better alternatives for `tmpdir` and `tmpdir_factory` that return old-style shims:
> `tmp_path`
>    Provide a `pathlib.Path` object to a temporary directory which is unique to each test function.
> `tmp_path_factory`
>    Make session-scoped temporary directories and return `pathlib.Path` objects.
> `tmpdir`
>    Provide a `py.path.local` object to a temporary directory which is unique to each test function; **replaced by `tmp_path`**.
> `tmpdir_factory`
>    Make session-scoped temporary directories and return `py.path.local` objects; **replaced by `tmp_path_factory`**.

I saw this in a few other places in the docs but it's not emphasized enough. I often don't remember which one to use and add `tmpdir` expecting `pathlib(2).Path` objects there when I should use `tmp_path` instead to achieve this. Every time I have to go to docs to find out which one is old.


*/me filed a bug about it @ https://github.com/m-burst/flake8-pytest-style/issues/23*


[The pytest fixtures list]: https://docs.pytest.org/en/latest/fixture.html

test/TestSkipImportPlaybook.py
```diff
@@ -24,26 +23,18 @@
...
-
-        # make dir and write role tasks to import or include
-        self.play_root = tempfile.mkdtemp()
-        with open(os.path.join(self.play_root, 'imported_playbook.yml'), 'w') as f_main:
-            f_main.write(IMPORTED_PLAYBOOK)
-
-    def tearDown(self):
-        shutil.rmtree(self.play_root)
+@pytest.fixture
+def playbook(tmpdir):
```
webknjaz, line 27: [`tmp_path` fixture] should be used instead since `tmpdir` is supposed to go away as per https://github.com/ansible/ansible-lint/pull/740#issuecomment-617896072.

[`tmp_path` fixture]:
https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture

webknjaz: (and I see you already use the proper `tmp_path` fixture in another place in this PR)

test/TestImportIncludeRole.py
webknjaz: https://github.com/ansible/ansible-lint/pull/740#issuecomment-617896072

test/TestRunner.py
cans: This is now three parametrized tests (but the last that stands on its own).

test/TestImportIncludeRole.py
cans: I think I found a better way.

webknjaz:
@cans I noticed you're trying out the co-authors feature. You used a github handle but that is an incorrect format. Compare https://github.com/ansible/ansible-lint/pull/740/commits/fb88e6139df9d74b814ea8299ac35e8cd15cc2d5 (one comitter, one avatar) and https://github.com/ansible/ansible-lint/commit/41d597d0583967fc21c0f1b5a1281b4205b14d37 (two committers with avatars linked to both of their GH profiles).
<details>
<summary>Screenshots</summary>
Screen Shot 2020-04-23 at 2 24 01 PM
Screen Shot 2020-04-23 at 2 24 14 PM
</details>

The proper form would be
```console
Co-Authored-By: Sviatoslav Sydorenko <webknjaz@redhat.com>
```

You can also use users' no-reply emails if they don't want to use public ones.

Ref: https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

test/TestImportIncludeRole.py
```diff
@@ -55,56 +52,54 @@
...
+
+@pytest.mark.parametrize(('playbook_filename', 'error'), (
+    pytest.param(PLAY_IMPORT_ROLE_INCOMPLETE,
+                 "Failed to find required 'name' key in import_role",
+                 id='IMPORT_ROLE_INCOMPLETE',
+                 ),
+))
+def test_invalid_import_role(tmp_path, default_rules_collection, playbook_filename, error):
+    playbook = _setup_play(tmp_path, playbook_filename)
+    runner = Runner(default_rules_collection, playbook, [], [], [])
```
webknjaz, line 103: How about a more classical approach?
```suggestion
@pytest.fixture
def playbook_path(request, tmp_path):
    playbook_text = request.param
    role_tasks_dir = tmp_path / 'test-role' / 'tasks'
    role_tasks_dir.mkdir(parents=True)
    (role_tasks_dir / 'main.yml').write_text(ROLE_TASKS_MAIN)
    (role_tasks_dir / 'world.yml').write_text(ROLE_TASKS_WORLD)
    play_path = tmp_path / 'playbook.yml'
    play_path.write_text(playbook_text)
    return str(play_path)


@pytest.mark.parametrize(('playbook_path', 'messages'), (
    pytest.param(PLAY_IMPORT_ROLE,
                 ['only when shell functionality is required'],
                 id='IMPORT_ROLE',
                 ),
    pytest.param(PLAY_IMPORT_ROLE_INLINE,
                 ['only when shell functionality is require'],
                 id='IMPORT_ROLE_INLINE',
                 ),
    pytest.param(PLAY_INCLUDE_ROLE,
                 ['only when shell functionality is require',
                  'All tasks should be named'],
                 id='INCLUDE_ROLE',
                 ),
    pytest.param(PLAY_INCLUDE_ROLE_INLINE,
                 ['only when shell functionality is require',
                  'All tasks should be named'],
                 id='INCLUDE_ROLE_INLINE',
                 ),
), indirect=('playbook_path', ))
def test_import_role2(default_rules_collection, playbook_path, messages):
    runner = Runner(default_rules_collection, playbook_path, [], [], [])

    results = runner.run()

    for message in messages:
        assert message in str(results)


@pytest.mark.parametrize(('playbook_path', 'error'), (
    pytest.param(PLAY_IMPORT_ROLE_INCOMPLETE,
                 "Failed to find required 'name' key in import_role",
                 id='IMPORT_ROLE_INCOMPLETE',
                 ),
), indirect=('playbook_path', ))
def test_invalid_import_role(default_rules_collection, playbook_path, error):
    runner = Runner(default_rules_collection, playbook_path, [], [], [])
```

cans: Didn't know about  `indirect`, neat. Though I am not sure that at first glance, it makes things much easier to follow. It's a lot of pytest _"shazam !"_, but when you know the trick, there is no more magic...

webknjaz: > but when you know the trick, there is no more magic...

Right: “Any sufficiently advanced technology is indistinguishable from magic.” --Arthur C. Clarke

I think this is true for anything new to us. I knew about this feature for years before I actually stopped by to read the docs on how do I apply it to my fixtures.

In my opinion, it's important to document such things as a usage guide. Mostly, it'd be useful to have such fixtures in `conftest.py`, reused everywhere along with some guidelines. It'd confuse newbies, no doubt. And I am often having a hard time finding that golden mean between "too complicated" and "too much copy-paste".

webknjaz:
Thanks, @cans!

PR #749, opened 2020-04-26
Fix playbook auto-detection

### LINKED ISSUE #731: [HELP WANTED] [SEND PR] Recover calling `get_playbooks_and_roles`
# Issue Type
- Bug report

# What's up?

While updating #645, I noticed that [the call to `get_playbooks_and_roles()` got removed](https://github.com/ansible/ansible-lint/pull/573/files#diff-e5d3af2dcf25a4d665f0ef298f146a85L154) and it seems like now it's not called from anywhere. It needs to be recovered. I guess it should go to `get_config()` in the `cli` module.

cans:
Closes #731

### DIFF
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -9,6 +9,7 @@ import yaml
 
 import ansiblelint
 from ansiblelint.version import __version__
+from ansiblelint.utils import get_playbooks_and_roles
 
 
 _logger = logging.getLogger(__name__)
@@ -57,7 +58,7 @@ def load_config(config_file):
             sys.exit(ansiblelint.utils.INVALID_CONFIG_RC)
     elif not os.path.exists(config_path):
         # a missing default config file should not trigger an error
-        return
+        return {}
 
     try:
         with open(config_path, "r") as stream:
@@ -156,9 +157,6 @@ def get_cli_parser():
 
 
 def merge_config(file_config, cli_config):
-    if not file_config:
-        return cli_config
-
     if 'quiet' in file_config:
         cli_config.quiet = cli_config.quiet or file_config['quiet']
 
@@ -185,6 +183,8 @@ def merge_config(file_config, cli_config):
     cli_config.exclude_paths.extend(file_config.get('exclude_paths', []))
 
     cli_config.rulesdir.extend(file_config.get('rulesdir', []))
+    if cli_config.use_default_rules or not cli_config.rulesdir:
+        cli_config.rulesdir.extend([Path(ansiblelint.default_rulesdir)])
 
     if 'skip_list' in file_config:
         cli_config.skip_list = cli_config.skip_list + file_config['skip_list']
@@ -200,12 +200,29 @@ def get_config(arguments):
     options = parser.parse_args(arguments)
 
     config = load_config(options.config_file)
+    config = merge_config(config, options)
 
-    return merge_config(config, options)
+    if config.listrules or config.listtags:
+        return config
 
+    # Exclude playbooks according to exclude path.
+    for playbook in list(config.playbook):
+        for excluded in config.exclude_paths:
+            try:
+                playbook.relative_to(excluded)
+            except ValueError:
+                pass
+            else:
+                config.playbook.remove(playbook)
 
-def print_help(file=sys.stdout):
-    get_cli_parser().print_help(file=file)
+    # no args triggers auto-detection mode
+    if not config.playbook:
+        config.playbook = get_playbooks_and_roles(options=config)
+        if not config.playbook:
+            parser.print_help(file=sys.stderr)
+            sys.exit(ansiblelint.utils.ANSIBLE_FAILURE_RC)
+
+    return config
 
 
 # vim: et:sw=4:syntax=python:ts=4:
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -19,7 +19,6 @@
 # THE SOFTWARE.
 
 from collections import OrderedDict
-import glob
 import importlib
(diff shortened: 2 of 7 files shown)


test/TestCliRolePaths.py
```diff
@@ -1,92 +1,139 @@
...
+    # Call ansible-lint the way molecule does.
+    pytest.param(os.path.normpath(os.path.join(local_test_dir, 'roles/test-role')),
+                 os.path.normpath(os.path.join(local_test_dir, '../bin/ansible-lint')),
+                 'molecule/default/include-import-role.yml',
+                 {'ANSIBLE_ROLES_PATH': os.path.join(local_test_dir, 'roles')},
+                 marks=pytest.mark.xfail,  # Cannot figure out why yet...
+                 id='playbook'),
+))
+def test_main(cwd, capsys, mocker, monkeypatch, binary, role_path, env):
+    mocker.patch.object(sys, 'argv', [binary, role_path])
```
webknjaz, line 69: `pytest-mock` is not installed which breaks CI. Also, you can do the same with built-in `monkeypatch`:
```suggestion
def test_main(cwd, capsys, monkeypatch, binary, role_path, env):
    monkeypatch.setattr(sys, 'argv', [binary, role_path])
```

lib/ansiblelint/cli.py
```diff
@@ -160,15 +161,12 @@ def get_cli_parser():
...
-    if not file_config:
-        return cli_config
-
     if 'quiet' in file_config:
         cli_config.quiet = cli_config.quiet or file_config['quiet']
 
     if 'parseable' in file_config:
         cli_config.parseable = (cli_config.parseable or
-                                file_config['parseable'])
+                                file_config.get['parseable'])
```
webknjaz, line 169: typo
```suggestion
                                file_config.get('parseable'))
```

lib/ansiblelint/utils.py
```diff
@@ -834,11 +838,9 @@ def get_playbooks_and_roles(options=None):
 
     for p in map(Path, files):
 
-        if any(str(p).startswith(file_path) for file_path in options.exclude_paths):
-            continue
-        elif (next((i for i in p.parts if i.endswith('playbooks')), None) or
+        if (next((i for i in p.parts if i.endswith('playbooks')), None) or
```
webknjaz, line 841: It's easier to read it like this:
```suggestion
        if any(i.endswith('playbooks') for i in p.parts) or
```

lib/ansiblelint/cli.py
```diff
@@ -204,8 +204,29 @@ def get_config(arguments):
...
+                pass
+            else:
+                config.playbook.remove(playbook)
+
+    # no args triggers auto-detection mode
+    if not config.playbook:
+        config.playbook = get_playbooks_and_roles(options=config)
+        if not config.playbook:
+            print_help(file=sys.stderr)
+            return 1
```
webknjaz, line 227: It's weird to return int where every other place returns dict. Did you mean to raise an exception maybe?

webknjaz:
@cans bad conflict resolution: https://github.com/ansible/ansible-lint/pull/749/files#diff-8a54e9ace4c09bec09fd356284e1853eR22

ssbarnea:
@cans if you can rebase this it would be great.

lib/ansiblelint/cli.py
cans: I just moved code from the `main()` function to here, and forgot to change that. But yes, I guess raising `SystemExit()` would be more explicit.

webknjaz: Right, but also it'd be great to consolidate places where we terminate the program. So that they are not distributed across multiple modules.

ssbarnea:
Any exits should use exit code defined in a single place: https://github.com/ansible/ansible-lint/blob/8673e5b27a55b88e0752338403045e9ac1cb2ebd/lib/ansiblelint/utils.py#L56-L57

That is a recent addition, but should avoid the case where we create overlapping exit codes.

lib/ansiblelint/cli.py
cans: The only other places where the program exits I could find are:
- in the `__main__` module;
- in the `utils` module;

Those in in the `__main__` module, I believe could be removed (and those don't use the constants you defined for return codes).

In the `utils` module, I find the use of `SystemExit()` debatable. If a file cannot be found, cannot be parsed, _etc._ that should be reported as an error like any other, no reason to stop the linting of other files. I generally prefer linting tools to spit out as much issues as they can so I can fix them in batch, without the need to run the command again and again. Maybe that's just me. 

That would put all termination points in the `cli` module but for when the main function returns.

If we want to change that, can we do it in another PR ? Let's stay focus on the issue at hand which, I believe, prevents the next release of the tool. The above are only internal changes that don't have any impact on the users.

ssbarnea:
recheck

ssbarnea:
The code is not ready, I tried to run it in a random repository and I got:
```
File "/Users/ssbarnea/.pyenv/versions/3.7.6/bin/ansible-lint", line 11, in <module>
load_entry_point('ansible-lint', 'console_scripts', 'ansible-lint')()
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/__main__.py", line 35, in main
options = cli.get_config(sys.argv[1:])
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/cli.py", line 212, in get_config
playbook.relative_to(excluded)
AttributeError: 'str' object has no attribute 'relative_to'
```

webknjaz:
Fixed by @ssbarnea via https://github.com/ansible/ansible-lint/pull/767. Feel free to re-submit the related changes separately.

PR #750, opened 2020-04-27
Avoid printing coverage report on test failures

ssbarnea:
Improves development experience by avoiding printing a coverage
report when tests are failing. This avoids forcing the user to
do additional scrolling in order to see the error.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -20,6 +20,7 @@ addopts =
     --cov=ansiblelint
     --cov-report term-missing:skip-covered
     --cov-report xml:.test-results/pytest/cov.xml
+    --no-cov-on-fail
 doctest_optionflags = ALLOW_UNICODE ELLIPSIS
 filterwarnings =
     error


pytest.ini
```diff
@@ -20,6 +20,7 @@ addopts =
     --cov=ansiblelint
     --cov-report term-missing:skip-covered
     --cov-report xml:.test-results/pytest/cov.xml
+    --no-cov-on-fail
```
webknjaz, line 23: I still want it to be printed out in the CIs. So please re-enable it there if you remove it from the default config.

ssbarnea: Did you happen to ever look at coverage result when there were test failures? The same remark is useful for CI: the developer may have to scroll extra on CI to see which test failed. 

TBH, I never found a case where I also found coverage useful when test failed.

In fact, even if we manage to identify one case where coverage report is useful with test failures we can ask ourselves another question: how often is this needed compared with the case where you just want to see which test failed? I bet is less than 1%, which would be a good reason for avoiding cluttering the UX for the other 99% cases.

The reality is that no change is to be merged without passing all tests, meaning that we would have at least one run with full coverage reports anyway. In fact the review process does not even start before CI reports green, so I do not see reasons to worry about lack of information.

webknjaz: ok

PR #751, opened 2020-04-29
Relocate skip utility functions into a separate `skip_utils` module

slsh1o:
This change rebases and fixes #529.

### DIFF
--- a/lib/ansiblelint/skip_utils.py
+++ b/lib/ansiblelint/skip_utils.py
@@ -31,11 +31,8 @@ _logger = logging.getLogger(__name__)
 
 def get_rule_skips_from_line(line):
     """Return list of rule ids skipped via comment on the line of yaml."""
-    rule_id_list = []
-    if INLINE_SKIP_FLAG in line:
-        noqa_text = line.split(INLINE_SKIP_FLAG)[1]
-        rule_id_list = noqa_text.split()
-    return rule_id_list
+    _before_noqa, _noqa_marker, noqa_text = line.partition(INLINE_SKIP_FLAG)
+    return noqa_text.split()
 
 
 def append_skipped_rules(pyyaml_data, file_text, file_type):
--- a/lib/ansiblelint/skip_utils.py
+++ b/lib/ansiblelint/skip_utils.py
@@ -51,7 +51,6 @@ def append_skipped_rules(pyyaml_data, file_text, file_type):
     :returns: original pyyaml_data altered with a 'skipped_rules' list added
     to individual tasks, or added to the single metadata block.
     """
-
     try:
         yaml_skip = _append_skipped_rules(pyyaml_data, file_text, file_type)
     except RuntimeError as exc:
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,164 @@
...
+INLINE_SKIP_FLAG = '# noqa '
+
+_logger = logging.getLogger(__name__)
+
+
+def get_rule_skips_from_line(line):
+    """Return list of rule ids skipped via comment on the line of yaml."""
+    rule_id_list = []
+    if INLINE_SKIP_FLAG in line:
+        noqa_text = line.split(INLINE_SKIP_FLAG)[1]
```
webknjaz, line 36: Plz apply https://github.com/ansible/ansible-lint/pull/529#discussion_r291989753 here

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,164 @@
...
+    :param file_text: raw file text.
+    :param file_type: type of file: tasks, handlers or meta.
+    :returns: original pyyaml_data altered with a 'skipped_rules' list added
+    to individual tasks, or added to the single metadata block.
+    """
+    try:
+        yaml_skip = _append_skipped_rules(pyyaml_data, file_text, file_type)
+    except RuntimeError as exc:
+        # Notify user of skip error, do not stop, do not change exit code
+        _logger.error('Error trying to append skipped rules: %s', exc)
```
webknjaz, line 58: No need to render exc manually:
```suggestion
    except RuntimeError:
        # Notify user of skip error, do not stop, do not change exit code
        _logger.error('Error trying to append skipped rules', exc_info=True)
```

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,164 @@
...
+
+INLINE_SKIP_FLAG = '# noqa '
+
+_logger = logging.getLogger(__name__)
+
+
+def get_rule_skips_from_line(line):
+    """Return list of rule ids skipped via comment on the line of yaml."""
+    rule_id_list = []
+    if INLINE_SKIP_FLAG in line:
```
webknjaz, line 35: Also could reverse this check into a guard style:
```suggestion
    if INLINE_SKIP_FLAG not in line:
        return []
```

lib/ansiblelint/skip_utils.py
```diff
@@ -0,0 +1,164 @@
+# (c) 2019, Ansible by Red Hat
```
webknjaz, line 1: Since we're applying extra changes, let's extend the range:
```suggestion
# (c) 2019–2020, Ansible by Red Hat
```

lib/ansiblelint/skip_utils.py
slsh1o: I would like to apply changes from [comment here](https://github.com/ansible/ansible-lint/pull/529#discussion_r291989753) as you suggested above and preserve func current behavior.

PR #754, opened 2020-05-01
Update cli output on README

ssbarnea:


ssbarnea:
recheck

docs/build
```diff
@@ -0,0 +1 @@
+../docs/docsite/_build
```
webknjaz, line 1: this is weird. why is this not `./docsite/_build` then? also, why can't this be fixed on Zuul side? I'd rather not have symlinks in this repo

ssbarnea: I can change the location using a parameter for jobs defined inside the repo. For the moment I am only trying to see if new location is enough. 

If I can avoid the symlink, I will do it.

README.rst
```diff
@@ -86,36 +86,43 @@ The following is the output from ``ansible-lint --help``, providing an overview
...
       -q                    quieter, although not silent output
       -p                    parseable output in the format of pep8
       --parseable-severity  parseable output including severity of rule
       -r RULESDIR           specify one or more rules directories using one or
                             more -r arguments. Any -r flags override the default
-                            rules in ['/path/to/ansible-
-                            lint/lib/ansiblelint/rules'], unless -R is also used.
-      -R                    Use default rules ['/path/to/ansible-
-                            lint/lib/ansiblelint/rules'] in addition to any extra
+                            rules in /Users/ssbarnea/c/os/ansible-
```
danielmellado, line 106: I'd leave this as it was before ;)

README.rst
```diff
@@ -86,36 +86,43 @@ The following is the output from ``ansible-lint --help``, providing an overview
...
+                            lint/lib/ansiblelint/rules, unless -R is also used.
+      -R                    Use default rules in /Users/ssbarnea/c/os/ansible-
```
danielmellado, line 108: Same here, please keep the default ['path/to/ansible-lint...

webknjaz:
> Creates a relative symlink to the standard location of sphinx build output in order to allow Zuul CI to find the built docs.

@ssbarnea you copy-paste foo has failed you :)

PR #756, opened 2020-05-03
Allow returning line number in matchplay

albinvass:
Currently it's not possible to set the line number ansible-lint should show when using matchplay.
 This can be a bit confusing since the line number could be very off in case of long blocks where the line number would point to the beginning of the block instead of the task that is faulty.

Update matchyaml to check if linenumber is also returned and use that if it is.


lib/ansiblelint/__init__.py
```diff
@@ -101,6 +101,14 @@ def matchtasks(self, file, text):
                            file['path'], self, message))
         return matches
 
+    def _unpack_result(self, play, result):
```
webknjaz, line 104: Since this method doesn't use `self`, it should be a plain function instead.

lib/ansiblelint/__init__.py
```diff
@@ -101,6 +101,14 @@ def matchtasks(self, file, text):
...
+        linenumber = play[ansiblelint.utils.LINE_NUMBER_KEY]
```
webknjaz, line 105: Why do you extract the line number here if you don't need this in one of the branches?

lib/ansiblelint/__init__.py
```diff
@@ -129,8 +137,9 @@ def matchyaml(self, file, text):
             if not isinstance(result, list):
                 raise TypeError("{} is not a list".format(result))
 
-            for section, message in result:
-                matches.append(Match(play[ansiblelint.utils.LINE_NUMBER_KEY],
+            for match in result:
+                section, linenumber, message = self._unpack_result(play, match)
```
webknjaz, line 141: I don't think that we should have a helper that returns 3 values.

lib/ansiblelint/__init__.py
```diff
@@ -129,8 +137,9 @@ def matchyaml(self, file, text):
             if not isinstance(result, list):
                 raise TypeError("{} is not a list".format(result))
 
-            for section, message in result:
-                matches.append(Match(play[ansiblelint.utils.LINE_NUMBER_KEY],
+            for match in result:
```
webknjaz, line 140: It's probably easier and cleaner to just do it in place:
```suggestion
            for section, *optional_line_num, message in result:
                try:
                    line_number, = optional_line_num
                except ValueError:
                    line_number = play[ansiblelint.utils.LINE_NUMBER_KEY]
```

lib/ansiblelint/__init__.py
albinvass: I'm trying to lint for loop variables to enforce the loop var policy in zuul-jobs but ended up in a bit of a rabbit hole trying to solve that. matchtask didn't seem to match includes or blocks so I ended up using matchplay instead to solve that, but that makes it so it gives the linenumber of the block containing the task instead of the task the linter fails at.

see:
- https://review.opendev.org/#/c/724910/
- https://review.opendev.org/#/c/724967/

lib/ansiblelint/__init__.py
albinvass: I wanted to do a simple if/else first but flake8 complained about code complexity so I broke it out to a separate function.

albinvass: To me it seems to be better to just check number of returned values instead of try/catching this.

To be honest, I'd much rather return a key:value store so it's easily extendable in the future.

albinvass:
recheck

lib/ansiblelint/__init__.py
albinvass: can do

albinvass:
I'll see if I can create a test for this

lib/ansiblelint/__init__.py
albinvass: Oh, this policy by the way: https://zuul-ci.org/docs/zuul-jobs/policy.html#ansible-loops-in-roles

lib/ansiblelint/__init__.py
webknjaz: EAFP is considered more pythonic: https://stackoverflow.com/a/11360880/595220.

As for flake8, it may be right, I enabled the McCabe complexity check recently because I want to make the project easier to follow and maintain in general.

If it needs to be a separate function, I still think that there's no reason for it to return more than one value: we don't need one-ring-to-rule-them-all type of complexity either.

webknjaz: Please try applying this via GitHub UI and see if it'll be okay with the linter.

albinvass:
Something like this then?

I'll push another commit with a test for it.

albinvass:
I'm confused to why this error doesn't appear when testing locally.

lib/ansiblelint/__init__.py
```diff
@@ -137,8 +137,8 @@ def matchyaml(self, file, text):
             if not isinstance(result, list):
                 raise TypeError("{} is not a list".format(result))
 
-            for match in result:
-                section, linenumber, message = self._unpack_result(play, match)
+            for section, message, *optional_linenumber in result:
```
webknjaz, line 140: Wait, I thought the line number was in the middle, not in the end.

lib/ansiblelint/__init__.py
```diff
@@ -129,8 +137,9 @@ def matchyaml(self, file, text):
             if not isinstance(result, list):
                 raise TypeError("{} is not a list".format(result))
 
-            for section, message in result:
-                matches.append(Match(play[ansiblelint.utils.LINE_NUMBER_KEY],
+            for section, message, *optional_linenumber in result:
```
webknjaz, line 140: I think the fallback extraction doesn't belong in the helper and should be done outside of the loop to avoid multiple access operations:
```suggestion
            fallback_line_number = play[ansiblelint.utils.LINE_NUMBER_KEY]
            for section, message, *optional_linenumber in result:
```

lib/ansiblelint/__init__.py
```diff
@@ -129,8 +137,9 @@ def matchyaml(self, file, text):
...
+                linenumber = self._matchplay_linenumber(play, optional_linenumber)
```
webknjaz, line 141: atm I don't see the need to have an external helper here

test/TestLineNumber.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2013-2014 Albin Vass <albin.vass@gmail.com>
```
webknjaz, line 1: Did you really create this test class 7 years ago?

albinvass: Oops ;)

webknjaz:
@albinvass you need an interactive rebase to get rid of the commits you copied

webknjaz:
@ssbarnea WDYT?

albinvass:
Sure!

test/TestLineNumber.py
```diff
@@ -0,0 +1,52 @@
...
+    def matchplay(self, file, play):
+        return [('Linenumber returned', self.shortdesc, MAGIC_NUMBER)]
+
+
+def test_rule_linenumber(monkeypatch):
+
+    def mock_response(*args, **kwargs):
+        return [{'skipped_rules': []}]
+
+    monkeypatch.setattr(ansiblelint.skip_utils,
```
webknjaz, line 45: What happens if you don't patch it?

test/TestLineNumber.py
```diff
@@ -0,0 +1,52 @@
...
+
+import ansiblelint.skip_utils
+
+from ansiblelint import AnsibleLintRule
+
+
+MAGIC_NUMBER = 7007
+
+
+class LinenumberRule(AnsibleLintRule):
```
webknjaz, line 29: Now that you've updated a real rule why not use it instead of a stub?

albinvass: sure, yeah I can delete this

albinvass:
Why not just return a Match object?

ssbarnea:
@albinvass I am perfectly fine to return a Match object, maybe we can even enforce that in the future with mypy.

lib/ansiblelint/rules/SudoRule.py
```diff
@@ -15,10 +15,10 @@ def _check_value(self, play_frag):
         if isinstance(play_frag, dict):
             if 'sudo' in play_frag:
                 results.append(({'sudo': play_frag['sudo']},
-                                'Deprecated sudo feature'))
+                                'Deprecated sudo feature', play_frag['__line__']))
```
webknjaz, line 18: This doesn't seem related to this PR.

webknjaz: Oh, wait. I see. You modified it for testing. But does the test for this exist? I guess you should unremove the test and use this rule there.

albinvass: hmm, yes I think you're right.

albinvass: I think I just misunderstood what you meant, fixing

test/TestLineNumber.py
```diff
@@ -0,0 +1,37 @@
...
+- debug:
+    msg: test
+
+- command: echo test
+  sudo: true
+"""
+
+
+def test_rule_linenumber(monkeypatch):
+
```
webknjaz, line 34:
```suggestion
    """Check that SudoRule offense contains a line number."""
```

PR #757, opened 2020-05-04
Fix logging configuration

ssbarnea:
Assure logging is initialized in such way that all submodules use
the module settings.

Previously initialize_setup() did configure logging only for
`ansiblelint.utils`, which means that any other ansiblelint.* would
not use the same settings.

Adds a debug level logging of loaded options.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -68,9 +68,13 @@ def initialize_logger(level=0):
     handler = logging.StreamHandler()
     formatter = logging.Formatter('%(levelname)-8s %(message)s')
     handler.setFormatter(formatter)
-    _logger.addHandler(handler)
-    # unknown logging level is treated as DEBUG
-    _logger.setLevel(VERBOSITY_MAP.get(level, logging.DEBUG))
+    logger = logging.getLogger(__package__)
+    logger.addHandler(handler)
+    # Unknown logging level is treated as DEBUG
+    logging_level = VERBOSITY_MAP.get(level, logging.DEBUG)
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/utils.py
```diff
@@ -68,9 +68,12 @@ def initialize_logger(level=0):
...
     formatter = logging.Formatter('%(levelname)-8s %(message)s')
     handler.setFormatter(formatter)
-    _logger.addHandler(handler)
+    logger = logging.getLogger("ansiblelint")
+    logger.addHandler(handler)
     # unknown logging level is treated as DEBUG
-    _logger.setLevel(VERBOSITY_MAP.get(level, logging.DEBUG))
+    logging_level = VERBOSITY_MAP.get(level, logging.DEBUG)
+    logger.setLevel(logging_level)
+    logger.debug("Logging initialized to level %s", logging_level)
```
webknjaz, line 76: Let's use `_logger` here in order to test that the config above works.
```suggestion
    _logger.debug("Logging initialized to level %s", logging_level)
```

lib/ansiblelint/utils.py
```diff
@@ -68,9 +68,12 @@ def initialize_logger(level=0):
     handler = logging.StreamHandler()
     formatter = logging.Formatter('%(levelname)-8s %(message)s')
     handler.setFormatter(formatter)
-    _logger.addHandler(handler)
+    logger = logging.getLogger("ansiblelint")
```
webknjaz, line 71: How about we don't hardcode the name?
```suggestion
    logger = logging.getLogger(__package__)
```

PR #766, opened 2020-05-06
Remove azure-pipelines integration

ssbarnea:
The project does not need 4 CI/CD systems running to perform
redundant jobs as this not only is wasting resources but also creates
more confusing status reports and more chances to get PRs reported
as failed due to CI bugs.


ssbarnea:
@awcrosby @gundalow @webknjaz ^ please review. I can provide additional reasons about why azure-pipelines should be removed in private.

Issue #763, opened 2020-05-06
ansible-lint fails if not running in a directory that is a .git repo


geerlingguy:
##### Summary

It would seem that in https://github.com/ansible/ansible-lint/pull/650, functionality was added that allowed `ansible-lint` to discover YAML files to lint if a git repository was not available.

However, a test playbook I created today didn't work, when I was in a directory that was not part of a git repo.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible 2.9.7
ansible-lint 4.2.0
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

macOS Catalina

##### STEPS TO REPRODUCE

Create a new directory outside of a git repo:

    mkdir playbook && cd playbook

Create a playbook `main.yml` inside with the contents:

```yaml
- hosts: localhost
  gather_facts: false
  connection: local

  tasks:
    - name: Register the output of the 'uptime' command.
      shell: test
      register: system_uptime
```

Run `ansible-lint` in the same directory as the playbook.

##### Desired Behaviour

ansible-lint detects yaml files without a .git repo present in the directory heirarchy.

##### Actual Behaviour

```
$ ansible-lint   
fatal: not a git repository (or any of the parent directories): .git
Traceback (most recent call last):
  File "/usr/local/bin/ansible-lint", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/__main__.py", line 153, in main
    args = get_playbooks_and_roles(options=options)
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/utils.py", line 774, in get_playbooks_and_roles
    universal_newlines=True).split()))
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 411, in check_output
    **kwargs).stdout
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['git', 'ls-files', '*.yaml', '*.yml']' returned non-zero exit status 128.
```

##### Workarounds

For now, the best workaround is to init a git repo in the same directory, and add any playbooks you want `ansible-lint` to discover to it:

    git init .
    git add -A
[...]

PR #767, opened 2020-05-06
Restore playbook auto-detection

ssbarnea:
Fixes: #763

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,0 +1,2 @@
+exclude_paths:
+- .github/
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -27,6 +27,7 @@ import sys
 
 import ansiblelint.formatters as formatters
 from ansiblelint import cli, default_rulesdir, RulesCollection, Runner
+from ansiblelint.utils import get_playbooks_and_roles
 from ansiblelint.utils import normpath, initialize_logger
 
 
@@ -53,11 +54,6 @@ def main():
 
     formatter = formatter_factory(cwd, options.display_relative_path)
 
-    # no args triggers auto-detection mode
-    if not options.playbook and not (options.listrules or options.listtags):
-        cli.print_help(file=sys.stderr)
-        return 1
-
     if options.use_default_rules:
         rulesdirs = options.rulesdir + [default_rulesdir]
     else:
@@ -80,7 +76,12 @@ def main():
         skip.update(str(s).split(','))
     options.skip_list = frozenset(skip)
 
-    playbooks = sorted(set(options.playbook))
+    if not options.playbook:
+        # no args triggers auto-detection mode
+        playbooks = get_playbooks_and_roles(options=options)
+    else:
+        playbooks = sorted(set(options.playbook))
+
     matches = list()
     checked_files = set()
     for playbook in playbooks:
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -150,7 +150,9 @@ def get_cli_parser():
     parser.add_argument('--version', action='version',
                         version='%(prog)s {ver!s}'.format(ver=__version__),
                         )
-    parser.add_argument('playbook', nargs='*')
+    parser.add_argument(dest='playbook', nargs='*',
+                        help="One or more files or paths. When missing it will "
+                        " enable auto-detection mode.")
 
     return parser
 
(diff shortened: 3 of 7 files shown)


ssbarnea:
@geerlingguy @webknjaz please review, bringing back auto-detection is a top priority.

webknjaz:
@ssbarnea zuul is stuck here too

.ansible-lint
```diff
@@ -0,0 +1,2 @@
+exclude_paths:
+  - .github/
```
webknjaz, line 2:
```suggestion
- .github/
```
https://github.com/ansible/ansible-lint/blob/master/.yamllint#L2

geerlingguy:
```
indent-sequences: false
```

Aww :(

lib/ansiblelint/cli.py
```diff
@@ -150,7 +150,9 @@ def get_cli_parser():
     parser.add_argument('--version', action='version',
                         version='%(prog)s {ver!s}'.format(ver=__version__),
                         )
-    parser.add_argument('playbook', nargs='*')
+    parser.add_argument(dest='playbook', nargs='*',
+                        help="One or more files or paths. When missing it will "
```
webknjaz, line 154: I guess this should go to README too

test/TestCliRolePaths.py
```diff
@@ -8,10 +8,13 @@ class TestCliRolePaths(unittest.TestCase):
...
 
-    def run_ansible_lint(self, cwd, role_path, bin=None, env=None):
+    def run_ansible_lint(self, cwd, bin=None, role_path=None, env=None):
         if not bin:
             bin = sys.executable + " -m ansiblelint"
-        command = '{} {}'.format(bin, role_path)
+        if role_path:
+            command = '{} {}'.format(bin, role_path)
+        else:
+            command = '{} -v'.format(bin)
```
webknjaz, line 17: Would something like this work?
```suggestion
        command = '{bin!s} {role_path or "-v"!s}'.format_map(locals())
```

.ansible-lint
webknjaz: It's more readable. B/c the indentation is for mappings.

test/TestCliRolePaths.py
ssbarnea: That one would not work and even if it would work it would not add any real benefits to the code.

Ditching an easy to read if for a complex one-liner does not make code easier to understand, especially as it would still hide the conditional. I will try another approach.

.ansible-lint
ssbarnea: @webknjaz While I dislike the non-indented YAML, I can understand that is the current standard used in this repo and I will fix it. That debate is not very far from tabs vs spaces one :)

Still, I would prefer to avoid human-comments related to formatting for which there are existing linters available (yamllint).

test/TestCliRolePaths.py
```diff
@@ -8,10 +8,12 @@ class TestCliRolePaths(unittest.TestCase):
     def setUp(self):
         self.local_test_dir = os.path.dirname(os.path.realpath(__file__))
 
-    def run_ansible_lint(self, cwd, role_path, bin=None, env=None):
+    def run_ansible_lint(self, cwd, role_path=None, bin=None, env=None):
         if not bin:
             bin = sys.executable + " -m ansiblelint"
-        command = '{} {}'.format(bin, role_path)
+        command = '{} -v {}'.format(
+            bin or (sys.executable + " -m ansiblelint"),
```
webknjaz, line 15: This part doesn't make sense with the if-clause above. Either remove that block or this fallback.

test/TestUtils.py
```diff
@@ -233,3 +233,24 @@ def test_logger_debug(caplog):
     )
 
     assert expected_info in caplog.record_tuples
+
+
+def test_cli_auto_detect(capfd):
+    """Test that run without arguments it will detect and lint the entire repository."""
+    result = os.system("ansible-lint -v -p --nocolor")
```
webknjaz, line 240: Let's use a bit more high-level subprocess API. Also I prefer to be explicit about the interpreter in order to not hit some other installation of ansible-lint accidentally:
```suggestion
    cmd = sys.executable, "-m", "ansible-lint", "-v", "-p", "--nocolor"
    result = subprocess.run(cmd).returncode
```

lib/ansiblelint/__main__.py
```diff
@@ -80,7 +76,12 @@ def main():
         skip.update(str(s).split(','))
     options.skip_list = frozenset(skip)
 
-    playbooks = sorted(set(options.playbook))
+    # no args triggers auto-detection mode
+    if not options.playbook:
```
webknjaz, line 80: I'd move the comment inside the block to explicitly indicate which branch corresponds to the auto-detection:
```suggestion
    if not options.playbook:
        # no args triggers auto-detection mode
```

PR #769, opened 2020-05-06
Enable yamllint

ssbarnea:
- fixed config bug which prevented custom rule from being loaded
- adds yamllint to pre-commit so we run it

While this will fail to lint it will uncover how inconsistent is our current yaml formatting.

### DIFF
--- a/.azure-pipelines/tox.yml
+++ b/.azure-pipelines/tox.yml
@@ -26,11 +26,11 @@ jobs:
     #   with_toxenv: 'coverage' # generate .tox/.coverage, .tox/coverage.xml after test run
     #   for_envs: [py38, py37, py36, pypy3, pypy]
     before:
-      - task: UsePythonVersion@0
-        condition: and(succeeded(), in(variables['TOXENV'], 'pypy'))
-        displayName: provision pypy 3
-        inputs:
-          versionSpec: 'pypy3'
+    - task: UsePythonVersion@0
+      condition: and(succeeded(), in(variables['TOXENV'], 'pypy'))
+      displayName: provision pypy 3
+      inputs:
+        versionSpec: 'pypy3'
 # - ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/') }}:
 #   - template: publish-pypi.yml@tox
 #     parameters:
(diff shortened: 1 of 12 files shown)


.pre-commit-config.yaml
```diff
@@ -34,6 +34,13 @@ repos:
   - id: check-merge-conflict
   - id: debug-statements
     language_version: python3
+- repo: https://github.com/adrienverge/yamllint.git
```
webknjaz, line 37: Oh, I didn't know that we don't run it...

.pre-commit-config.yaml
```diff
@@ -34,6 +34,13 @@ repos:
...
+  rev: v1.23.0
+  hooks:
+  - id: yamllint
+    files: \.(yaml|yml)$
+    types: [file, yaml]
```
webknjaz, line 42: I'm pretty sure this should be
```suggestion
    types: [yaml]
```

.azure-pipelines/tox.yml
```diff
@@ -4,36 +4,36 @@ parameters:
...
 jobs:
-- template: run-tox-env.yml@tox
-  parameters:
-    jobs:
-      docs: null
-      lint:
-        before:
-        - task: CacheBeta@0
-          displayName: cache pre-commit
+  - template: run-tox-env.yml@tox
```
webknjaz, line 7: Sequences shouldn't be indented.

.pre-commit-config.yaml
ssbarnea: See https://github.com/adrienverge/yamllint/blob/master/.pre-commit-hooks.yaml -- that is the recommended setup. Keep in mind that these work with AND and not OR.

.azure-pipelines/tox.yml
ssbarnea: For Ansible code in particular that is a serious issue, especially with blocks where it makes extremely hard to find block/rescue/always sections. 

Also non identing makes is much easier to misplace a when condition between block level or last task within the block.

For these reasons all the Ansible code used by OpenStack does enforce sequence-indentation, something this change does **not**. This change is more relaxed, only request to be consistent inside the file.

So basically this touched only the files that reported inconsistencies, not others.

PR #770, opened 2020-05-07
Update ComparisonToEmptyStringRule.py

vbotka:
The conditionals with bare variables will be deprecated. It's causing a lint warning atm. See CONDITIONAL_BARE_VARS https://docs.ansible.com/ansible/latest/reference_appendices/config.html#conditional-bare-vars


webknjaz:
While the change is ok, I cannot accept it like this because you should also update the docs so that the CI doesn't fail: https://github.com/ansible/ansible-lint/pull/770/checks?check_run_id=652398319#step:10:53.

vbotka:
> ... update the docs so that the CI doesn't fail: https://github.com/ansible/ansible-lint/pull/770/checks?check_run_id=652398319#step:10:53.

Thank you for the review. I've updated the docs and submitted PR #772

webknjaz:
Just update this PR, it'll always work if they are together.

ssbarnea:
@vbotka Please close the doc update and include it here. This can be avoided by always running `tox -e lint` before any push.

vbotka:
Done. "tox -e lint" passed before the push. Thank you for comments.

vbotka:
Fro the record: Local "tox -e lint' at branch patch-1 passed.
--
The build crashed:
The HTML pages are in _build/html.
make[1]: Leaving directory '/home/runner/work/ansible-lint/ansible-lint/docs/docsite'
ERROR: Building docs changed tracked files, include them in commit to avoid this failure.
--

webknjaz:
Looks green to me

PR #772, opened 2020-05-07
Update default_rules.rst

vbotka:
[Fixed] ERROR: Building docs changed tracked files, include them in commit to avoid this failure.
Update ComparisonToEmptyStringRule.py #770


webknjaz:
Closing as this should be in #770

Issue #744, opened 2020-04-20
Reduce McCabe complexity


webknjaz:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

N/A

# Desired Behaviour

The source code should be refactored so that flake8 doesn't fail.

# Actual Behaviour (Bug report only)

* [ ] `lib/ansiblelint/__main__.py`:32:1: C901 'main' is too complex (13) (@dpendolino -- #931)
* [x] ~`lib/ansiblelint/utils.py`:229`:1: C901 '_taskshandlers_children' is too complex (12)~ (#903 / #862)
* [ ] `lib/ansiblelint/utils.py`:413:1: C901 'normalize_task_v1' is too complex (11)
* [ ] `lib/ansiblelint/utils.py`:802:1: C901 'get_playbooks_and_roles' is too complex (17)
* [x] ~`lib/ansiblelint/rules/MetaMainHasInfoRule.py`:24:5: C901 'MetaMainHasInfoRule.matchplay' is too complex (13)~

To see the errors and fix them, one needs to remove `C901` ignores from the `per-file-ignores` setting in the `.flake8` config file and from the source code (noqa).

webknjaz:
@dpendolino if you want, you could take some of the subtasks here. While it's assigned to @ssbarnea I believe he hasn't had time to start solving this. OTOH since there's complexity in several places, the fixes may be submitted as separate PRs. Just comment here to announce your intent to work on something specific to avoid race conditions...

ssbarnea:
I removed myself as I do not expect to work on it. Indeed, there is a set of areas where it should be easy for anyone to make small contributions: increase typing coverage, one function at a time, removing QA skips like mccabe or docstrings.

Hint: do not try to fix too much at once, small batches are better than big.

webknjaz:
@dpendolino I assigned you here, try to select one function to improve for starters.

@tohtohro this may be great for you too, it's a relatively simple refactoring task. Since there's 3 places still needing fixing, you may take one of those. Just announce which one.

dpendolino:
@webknjaz I'll take the `main` complexity subtask. PR incoming =]

PR #774, opened 2020-05-07
Fix exclude_paths from get_playbooks_and_roles

ssbarnea:
Fix bug which made impossible to use relative exclude paths as
these where resolved while the file tested where not.

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -712,9 +712,16 @@ def get_playbooks_and_roles(options=None):
 
     for p in map(Path, files):
 
-        if any(str(p).startswith(file_path) for file_path in options.exclude_paths):
+        try:
+            for file_path in options.exclude_paths:
+                if str(p.resolve()).startswith(str(file_path)):
+                    raise FileNotFoundError(
+                        f'File {file_path} matched exclusion entry: {p}')
+        except FileNotFoundError as e:
+            _logger.debug('Ignored %s due to: %s', p, e)
             continue
-        elif (next((i for i in p.parts if i.endswith('playbooks')), None) or
+
+        if (next((i for i in p.parts if i.endswith('playbooks')), None) or
                 'playbook' in p.parts[-1]):
             playbooks.append(normpath(p))
             continue
--- a/test/TestUtils.py
+++ b/test/TestUtils.py
@@ -257,3 +257,15 @@ def test_cli_auto_detect(capfd):
         "[E401] Git checkouts must contain explicit version" in out
     # assures that our .ansible-lint exclude was effective in excluding github files
     assert "Unknown file type: .github/" not in out
+
+
+def test_auto_detect_exclude(monkeypatch):
+    """Verify that exclude option can be used to narrow down detection."""
+    options = cli.get_config(['--exclude', 'foo'])
+
+    def mockreturn(options):
+        return ['foo/playbook.yml', 'bar/playbook.yml']
+
+    monkeypatch.setattr(utils, 'get_yaml_files', mockreturn)
+    result = utils.get_playbooks_and_roles(options)
+    assert result == ['bar/playbook.yml']


lib/ansiblelint/utils.py
```diff
@@ -712,7 +712,7 @@ def get_playbooks_and_roles(options=None):
 
     for p in map(Path, files):
 
-        if any(str(p).startswith(file_path) for file_path in options.exclude_paths):
+        if any(str(p.resolve()).startswith(file_path) for file_path in options.exclude_paths):
```
webknjaz, line 715: Looks like @cans also tried to fix it @ https://github.com/ansible/ansible-lint/pull/749/files#diff-8a54e9ace4c09bec09fd356284e1853eR859

webknjaz: Maybe you could get some inspiration there?

ssbarnea: We need to start introduction of type checks in order to avoid so many casts in our code but this should be done in separated PR.

cans:
Why do you keep making competing changes ? 
Similar changes were made in #749. Also your changes won't help with issue #744.
Should I close #749 ?

ssbarnea:
@cans That mention PR is huge and contains lots of unrelated changes. A PR should be kept atomic, addressing one issue, so it can easily be reviewed, merged and eventually identified if a regression occurs. 

Also the code complexity is unrelated to this bug and should be addressed separately.

webknjaz:
@cans I'll be glad to review the rest of the changes when submitted separately. It so happened that another PR fixed the issue and was easier to review and accept (because unrelated changes often have problems that block the whole PR even if the related changes are fine because it's hard to identify which part of the change is acceptable to merge already).

@ssbarnea I'd like to encourage you to let PR authors complete their changes if they are not urgent and received updates during the last two weeks. This way we can focus on non-conflicting changes and work in more improvement directions w/o dismissing the hard work that other folks try to contribute.

lib/ansiblelint/utils.py
```diff
@@ -712,9 +716,15 @@ def get_playbooks_and_roles(options=None):
 
     for p in map(Path, files):
 
-        if any(str(p).startswith(file_path) for file_path in options.exclude_paths):
+        try:
+            for file_path in options.exclude_paths:
+                if str(p.resolve()).startswith(str(file_path)):
```
webknjaz, line 721: Since you dropped py3.5 this could probably be simplified.

test/TestUtils.py
```diff
@@ -257,3 +257,16 @@ def test_cli_auto_detect(capfd):
...
+    """Verify that exclude option can be used to narrow down detection."""
+    options = cli.get_config(['--exclude', 'foo'])
+
+    def mockreturn(options):
+        return ['foo/playbook.yml', 'bar/playbook.yml']
+
+    with monkeypatch.context() as m:
+        m.setattr(utils, 'get_yaml_files', mockreturn)
+        result = utils.get_playbooks_and_roles(options)
+        assert result == ['bar/playbook.yml']
```
webknjaz, line 272: Let's put this out of context
```suggestion
    assert result == ['bar/playbook.yml']
```

test/TestUtils.py
```diff
@@ -257,3 +257,16 @@ def test_cli_auto_detect(capfd):
...
+
+
+def test_auto_detect_exclude(monkeypatch):
+    """Verify that exclude option can be used to narrow down detection."""
+    options = cli.get_config(['--exclude', 'foo'])
+
+    def mockreturn(options):
+        return ['foo/playbook.yml', 'bar/playbook.yml']
+
+    with monkeypatch.context() as m:
```
webknjaz, line 269: It looks like you don't really need this CM because you patch the entire body anyway.

test/TestUtils.py
webknjaz, line 272: It's cleaner this way:
```suggestion
    monkeypatch.setattr(utils, 'get_yaml_files', mockreturn)
    discovered_playbooks = utils.get_playbooks_and_roles(options)

    assert discovered_playbooks == ['bar/playbook.yml']
```

lib/ansiblelint/utils.py
```diff
@@ -57,6 +57,10 @@
 _logger = logging.getLogger(__name__)
 
 
+class FilePathIgnored(Exception):
+    pass
+
+
```
webknjaz, line 63: As per comments below:
```suggestion
```

lib/ansiblelint/utils.py
ssbarnea: I look and it seems that not.

PR #775, opened 2020-05-07
Require Python 3.6 or newer

ssbarnea:
This change removes support for Python 3.6 starting with 4.3.x version.

Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/452


ssbarnea:
recheck

PR #777, opened 2020-05-08
Adopt mypy static type checking

ssbarnea:


.pre-commit-config.yaml
```diff
@@ -42,4 +42,9 @@ repos:
     additional_dependencies:
     - flake8-2020>=1.6.0
     - flake8-docstrings
+    - flake8-mypy
     - flake8-pytest-style>=1.0.0
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v0.770  # Use the sha / tag you want to point at
```
webknjaz, line 48:
```suggestion
  rev: v0.770
```

lib/ansiblelint/generate_docs.py
```diff
@@ -4,8 +4,8 @@
 import importlib
 from inspect import getmembers, ismodule, isclass
 import logging
-import rules
 from ansiblelint import AnsibleLintRule
+from ansiblelint import rules
```
webknjaz, line 8: I prefer relative imports but this would belong to a separate PR.

lib/ansiblelint/rules/PlaybookExtension.py
```diff
@@ -1,8 +1,8 @@
 # Copyright (c) 2016, Tsukinowa Inc. <info@tsukinowa.jp>
 # Copyright (c) 2018, Ansible Project
-
+# from typing import List
```
webknjaz, line 3:
```suggestion

```

lib/ansiblelint/rules/PlaybookExtension.py
```diff
@@ -1,8 +1,8 @@
...
 from ansiblelint import AnsibleLintRule
-
+from typing import List
```
webknjaz, line 5: I'd not have stdlib imports after first-party

lib/ansiblelint/rules/PlaybookExtension.py
```diff
@@ -12,7 +12,8 @@ class PlaybookExtension(AnsibleLintRule):
     description = 'Playbooks should have the ".yml" or ".yaml" extension'
     severity = 'MEDIUM'
     tags = ['formatting']
-    done = []  # already noticed path list
+    # already noticed path list
+    done = []  # type: List
```
webknjaz, line 16: Can this be
```suggestion
    done = []  # type: List  # already noticed path list
```

setup.cfg
```diff
@@ -84,3 +84,11 @@ console_scripts =
...
 [options.packages.find]
 where = lib
+
+[mypy_config]
+color_output = True
+error_summary = True
+
+; ignore_missing_imports=False
+; disallow_untyped_calls=True
+; warn_redundant_casts=True
```
webknjaz, line 94: This must go to a mypy config. I intentionally keep configs of different tools separate.

.pre-commit-config.yaml
```diff
@@ -43,3 +43,9 @@ repos:
     - flake8-2020>=1.6.0
     - flake8-docstrings
     - flake8-pytest-style>=1.0.0
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v0.770
+  hooks:
+  - id: mypy
+    # empy args needed in order to match mypy cli behavior
```
webknjaz, line 50:
```suggestion
    # empty args needed in order to match mypy CLI behavior
```

.gitignore
```diff
@@ -38,3 +38,7 @@ pip-log.txt
 coverage.xml
 pip-wheel-metadata
 .test-results/
+
+#mypy
```
webknjaz, line 42:
```suggestion
# mypy
```

mypy.ini
```diff
@@ -0,0 +1,35 @@
+[mypy]
+python_version = 3.6
+color_output = True
+error_summary = True
+# TODO(ssbarnea): Remove ignores below:
```
webknjaz, line 5:
```suggestion
# TODO(ssbarnea): Uncomment ignores below:
```

lib/ansiblelint/generate_docs.py
```diff
@@ -47,20 +48,21 @@ def main():
...
         if module == '__init__.py' or module[-3:] != '.py':
             continue
         module = 'rules.{}'.format(module[:-3])
         importlib.import_module(module)
 
 
-def get_serialized_rules():
+def get_serialized_rules() -> List[Dict]:
     mod_list = [m[1] for m in getmembers(rules) if ismodule(m[1])]
+    print(mod_list)
```
albinvass, line 65: These prints are for debugging only right?

.gitignore
```diff
@@ -38,3 +38,7 @@ pip-log.txt
 coverage.xml
 pip-wheel-metadata
 .test-results/
+
+# mypy
+.mypy_cache
+out
```
webknjaz, line 44: What is this?

lib/ansiblelint/__main__.py
```diff
@@ -29,20 +29,21 @@
...
+def main() -> int:
     cwd = pathlib.Path.cwd()
 
     options = cli.get_config(sys.argv[1:])
 
     initialize_logger(options.verbosity)
     _logger.debug("Options: %s", options)
 
-    formatter_factory = formatters.Formatter
+    formatter_factory: Any = formatters.Formatter
```
webknjaz, line 46: Why is it any?

lib/ansiblelint/utils.py
```diff
@@ -57,7 +58,7 @@
 _logger = logging.getLogger(__name__)
 
 
-def initialize_logger(level=0):
+def initialize_logger(level=0) -> None:
```
webknjaz, line 61:
```suggestion
def initialize_logger(level: int = 0) -> None:
```

lib/ansiblelint/generate_docs.py
```diff
@@ -47,19 +48,19 @@ def main():
 
     filename = '../../docs/docsite/rst/rules/default_rules.rst'
     with open(filename, 'w') as file:
-        file.write(make_table(grid))
+        file.write(make_table(grid))  # type: ignore
```
webknjaz, line 51: Why?

.gitignore
ssbarnea: both folders are created by mypy.

webknjaz: Is it configurable? Can you point it elsewhere?

lib/ansiblelint/generate_docs.py
```diff
@@ -4,9 +4,10 @@
 import importlib
 from inspect import getmembers, ismodule, isclass
 import logging
-import rules
+import rules  # type: ignore
```
webknjaz, line 7: why do you need to ignore it?

ssbarnea: Yep, because it was an invalid import, correctly identified by mypy. Still, this will go away once we merge https://github.com/ansible/ansible-lint/pull/783

.gitignore
ssbarnea: TBH, I did not bother to investigate, I know that the first one is configurable but it needs to be kept inside repository as it cannot work as a shared cache (so basically no benefits from custmizing it). The 2nd one, not sure exactly what/when it appears but I can have a look. 
I can remove it as CI/CD will not complain about it, but probably you will not be very happy to discover it appearing later ;)

lib/ansiblelint/__main__.py
ssbarnea: That is a hack to make mypy happy about allowing other classes to be assigned as formatters. Otherwise, it will complain on assignment. I was not able to find another workaround yet. Is one of the known issues with python logging and mypy. Even the stdlib has only partial stubs.

.gitignore
webknjaz: At least add a comment hinting why it shouldn't be removed from gitignore during refactoring

ssbarnea: I removed it and I will add it back once I get the same error again, mostly because I want to find out who is really generating this folder.

PR #779, opened 2020-05-09
Refactor generate docs

ssbarnea:
- allow generate docs to be called from any directory
- adopts modern module discovery method
- includes docs regeneration in linting in order to avoid accidents
  where people forget to update the file.
- Minor grammar fix


ssbarnea:
@webknjaz I think I may have hit a bug in pre-commit, as described at https://github.com/pre-commit/pre-commit/issues/1445 - maybe you have an idea about what causes it.

In case it was obvious, the scope of this change is to avoid CI/CD doc building errors which may confuse new contributors. I seen at least two recent cases on PRs, in addition to hitting it myself in the past.

.pre-commit-config.yaml
```diff
@@ -18,6 +18,12 @@ repos:
     language: system
     files: >-
       ^setup\.py$
+  - id: docs-rebuild
+    name: Checks that generated docs are in-sync with commit
+    entry: python -m ansiblelint.generate_docs
+    language: python
+    pass_filenames: false
```
webknjaz, line 25: Have you tried this?
```suggestion
    pass_filenames: false
    additional_dependencies:
    - .
```

.pre-commit-config.yaml
```diff
@@ -18,6 +18,12 @@ repos:
     language: system
     files: >-
       ^setup\.py$
+  - id: docs-rebuild
+    name: Checks that generated docs are in-sync with commit
+    entry: python -m ansiblelint.generate_docs
```
webknjaz, line 23: This looks heavy. I'd move it further, after most of the linters, just before flake8.

lib/ansiblelint/generate_docs.py
```diff
@@ -45,18 +45,19 @@ def main():
...
 
 def import_all_rules():
-    for module in list(os.walk('rules'))[0][2]:
-        if module == '__init__.py' or module[-3:] != '.py':
-            continue
-        module = 'rules.{}'.format(module[:-3])
-        importlib.import_module(module)
+    prefix = 'ansiblelint.rules.'
+    for importer, modname, ispkg in pkgutil.iter_modules(rules.__path__, prefix):
+        _logger.info("Found submodule %s (is a package: %s)", modname, ispkg)
```
webknjaz, line 59: Did you mean subpackage? A module is a py-file while a package is a dir.

lib/ansiblelint/generate_docs.py
```diff
@@ -45,18 +45,19 @@ def main():
...
+        __import__(modname)
```
webknjaz, line 60: I'm pretty sure that `__import__` is not recommended and using `importlib` is better.

tox.ini
```diff
@@ -98,7 +98,7 @@ deps =
   pre-commit>=1.20.0
 skip_install = true
 commands =
-  python -m pre_commit run {posargs:--all-files}
+  python -m pre_commit run -v {posargs:--all-files}
```
webknjaz, line 101: Keep it overridable.
```suggestion
  python -m pre_commit run {posargs:--all-files -v}
```

lib/ansiblelint/generate_docs.py
```diff
@@ -45,18 +45,19 @@ def main():
         id_text = id_link.format(d['id'], d['classname'])
         grid.append([id_text, d['version_added'], d['shortdesc'], d['description']])
 
-    filename = '../../docs/docsite/rst/rules/default_rules.rst'
+    filename = Path(__file__).resolve().parents[2].joinpath(
+        "docs/docsite/rst/rules/default_rules.rst")
```
webknjaz, line 49: I think the canonical way is
```suggestion
    filename = (
        Path(__file__).resolve().parents[2] / "docs" /
        "docsite" / "rst" / "rules" / "default_rules.rst"
    )
```

.pre-commit-config.yaml
ssbarnea: I does not work, please see https://github.com/pre-commit/pre-commit/issues/1445 -- I am still looking for a workaround for this issue. Maybe you can find one?

webknjaz: I don't see this specific option set in that issue snippets. Also, I think I use this elsewhere and it works.

docs/docsite/Makefile
```diff
@@ -35,7 +35,7 @@ docs: clean htmldocs
 generate_rst: staticmin
 
 htmldocs: generate_rst
-	cd ../../lib/ansiblelint; python ./generate_docs.py
+	python -m ansiblelint.generate_docs
```
webknjaz, line 38: I'd rather stick it directly into tox.ini. But this could be done in a separate PR

tox.ini
```diff
@@ -98,14 +98,15 @@ deps =
...
-  python -m pre_commit run {posargs:--all-files}
+  python -m pre_commit run {posargs:--all-files -v}
 passenv =
   {[testenv]passenv}
   PRE_COMMIT_HOME
 
 [testenv:docs]
 whitelist_externals = make
 deps = sphinx
+usedevelop = true
```
webknjaz, line 109: I don't think this is a good idea. I prefer using things as installed.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,21 @@
...
+      )$
+  - id: mixed-line-ending
+  - id: check-byte-order-marker
+  - id: check-executables-have-shebangs
+  - id: check-merge-conflict
+  - id: debug-statements
+    language_version: python3
 - repo: local
   hooks:
   - id: immutable-setup-py
```
webknjaz, line 21: I'd like to keep this hook on top because it's quite light-weight.

.pre-commit-config.yaml
```diff
@@ -18,22 +34,14 @@ repos:
...
-      )$
-  - id: mixed-line-ending
-  - id: check-byte-order-marker
-  - id: check-executables-have-shebangs
-  - id: check-merge-conflict
-  - id: debug-statements
-    language_version: python3
+  - id: docs-rebuild
+    name: Checks that generated docs are in-sync with commit
+    # Editable mode is key in order to find targed doc directory.
```
webknjaz, line 39: Why is it a problem?

.pre-commit-config.yaml
```diff
@@ -18,22 +34,14 @@ repos:
...
-        examples/example.yml
-      )$
-  - id: mixed-line-ending
-  - id: check-byte-order-marker
-  - id: check-executables-have-shebangs
-  - id: check-merge-conflict
-  - id: debug-statements
-    language_version: python3
+  - id: docs-rebuild
+    name: Checks that generated docs are in-sync with commit
```
webknjaz, line 38: Also add a `description: ` key

.pre-commit-config.yaml
```diff
@@ -43,3 +43,13 @@ repos:
...
+- repo: local
+  hooks:
+  - id: docs-rebuild
+    name: Checks that generated docs are in-sync with commit
+    description: >-
+      Regenerate rules documentation from source and report failure if the
+      the the result is modified.
+    entry: sh -c "pip install . && python -m ansiblelint -L -f rst > docs/docsite/rst/rules/default_rules.rst"
+    language: python
+    pass_filenames: false
```
webknjaz, line 55:
```suggestion
    pass_filenames: false
    stages:
    - merge-commit
    - push
```

.pre-commit-config.yaml
```diff
@@ -43,3 +43,13 @@ repos:
...
     - flake8-docstrings
     - flake8-pytest-style>=1.0.0
+- repo: local
+  hooks:
+  - id: docs-rebuild
+    name: Checks that generated docs are in-sync with commit
+    description: >-
+      Regenerate rules documentation from source and report failure if the
+      the the result is modified.
+    entry: sh -c "pip install . && python -m ansiblelint -L -f rst > docs/docsite/rst/rules/default_rules.rst"
```
webknjaz, line 53: can we multiline this?
```suggestion
    entry: >-
      sh -c "
        pip install . &&
        python -m ansiblelint -L -f rst
        >
        docs/docsite/rst/rules/default_rules.rst
      "
```

lib/ansiblelint/__main__.py
```diff
@@ -62,7 +63,10 @@ def main():
     rules = RulesCollection(rulesdirs)
 
     if options.listrules:
-        print(rules)
+        if options.format == 'plain':
+            print(rules)
+        else:
+            print(rules_as_rst(rules))
```
webknjaz, line 69:
```suggestion
        formatted_rules = print(rules if options.format == 'plain' else rules_as_rst(rules)
        print(formatted_rules)
```

lib/ansiblelint/cli.py
```diff
@@ -92,6 +92,10 @@ def get_cli_parser():
 
     parser.add_argument('-L', dest='listrules', default=False,
                         action='store_true', help="list all the rules")
+    parser.add_argument('-f', dest='format', default='plain',
+                        nargs='?',
```
webknjaz, line 96: Are you sure this is needed?
```suggestion
```

.pre-commit-config.yaml
ssbarnea: This a really bad idea, will prevent the hook from running from normal pre-commit runs (`pre-commit run -a`) and also with `tox -e linters`. I is a premature-optimization that basically renders the hook useless. Very few people are using its git hooks.

ssbarnea:
> Waiting on the reaction for nargs question.

Oops. I missed that one. The answer is nope, is not mandatory and I removed it. It allowed it to work without passing an argument, like `ansible-lint -L -f`, something that is not really needed here.

Issue #780, opened 2020-05-11
RuntimeError: Failed to find required 'name' key in include_role


ssbarnea:
##### Summary
Bug found on master code, where ansible-lint may produce an exception like:
```python-traceback
RuntimeError: Failed to find required 'name' key in include_role
```

Feel free to assign but to me.

##### Issue Type

- Bug Report

##### STEPS TO REPRODUCE
```yaml
- hosts: all
  tasks:
    - name: foo
      include_role:
        role: bar
```

##### Desired Behaviour
* No exception raised
* If error occurs the filename must be visible, and if possible even line number. At this moment even when running in debug mode we are unable to find the name of the file.

##### Actual Behaviour
```python-traceback
Traceback (most recent call last):
File "/Users/ssbarnea/.pyenv/versions/3.8.2/bin/ansible-lint", line 11, in <module>
load_entry_point('ansible-lint', 'console_scripts', 'ansible-lint')()
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/__main__.py", line 91, in main
matches.extend(runner.run())
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/__init__.py", line 278, in run
for child in ansiblelint.utils.find_children(arg, self.playbook_dir):
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/utils.py", line 183, in find_children
for child in play_children(basedir, item, playbook[1], playbook_dir):
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/utils.py", line 235, in play_children
return delegate_map[k](basedir, k, v, parent_type)
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/utils.py", line 264, in _taskshandlers_children
raise RuntimeError(
RuntimeError: Failed to find required 'name' key in include_role
```

webknjaz:
Assuming it's a regression.

ssbarnea:
@webknjaz The same exception happens with 4.2.0 and 4.1.0 so it does not really count as a regression.

ssbarnea:
As a note, while Ansible itself accepts the current syntax, that is not the one documented inside https://docs.ansible.com/ansible/latest/modules/include_role_module.html so I am inclined to say that would be enough to make this failure non-fatal, so the user can fix the syntax to the documented one.

I already raised a PR to implement this but we need to look at it as it requires some additional code refactoring.

PR #781, opened 2020-05-11
Convert failed to find required 'name' key in include_role into a match

ssbarnea:
Convert `Failed to find required 'name' key in include_role` fatal error into a rule match, so the linting can continue with other files.

Also included with this change, as needed:
* move of errors/exceptions to their own module, so we can avoid cyclic-imports
* minor condition change to avoid hitting mccable limit (11)

Fixes: #780


lib/ansiblelint/errors.py
```diff
@@ -0,0 +1,32 @@
+class Match(object):
```
webknjaz, line 1: Explicit inheritance from object is unnecessary in py3
```suggestion
class Match:
```

webknjaz: And while you're here plz add a class docstring.

lib/ansiblelint/__init__.py
```diff
@@ -138,6 +139,15 @@ def matchyaml(self, file, text):
         return matches
 
 
+class LoadingFailureRule(AnsibleLintRule):
```
webknjaz, line 142: This should probably go into its own rule module just like other rules.

webknjaz, line 142: And `AnsibleLintRule` could go into `rules/__init__.py` to avoid circular imports.

webknjaz: ^

docs/docsite/rst/rules/default_rules.rst
```diff
@@ -39,6 +39,8 @@ ID
...
+`E500 <https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/LoadingFailureRule.py>`_                                                                                                                                                                                                    v4.3.0                                                                                                                                                                                                                                                                                                       Failed to load or parse file                                                                                                                                                                                                                                                                                 Linter failed to process a YAML file, possible not an Ansible file.
```
webknjaz, line 43: @ssbarnea why is this duplicated and why is it under 4xx section?

lib/ansiblelint/errors.py
```diff
@@ -17,3 +17,21 @@ def __repr__(self):
         formatstr = u"[{0}] ({1}) matched {2}:{3} {4}"
         return formatstr.format(self.rule.id, self.message,
                                 self.filename, self.linenumber, self.line)
+
+
+class MatchError(Exception):
```
webknjaz, line 22: Would it be semantically better to inherit something a bit more concrete?
```suggestion
class MatchError(ValueError):
```

lib/ansiblelint/rules/LoadingFailureRule.py
```diff
@@ -0,0 +1,10 @@
+from ansiblelint import AnsibleLintRule
```
webknjaz, line 1:
```suggestion
"""Rule definition for a failure to load a file."""

from ansiblelint import AnsibleLintRule
```

lib/ansiblelint/rules/LoadingFailureRule.py
```diff
@@ -0,0 +1,10 @@
...
+
+
+class LoadingFailureRule(AnsibleLintRule):
```
webknjaz, line 4:
```suggestion
class LoadingFailureRule(AnsibleLintRule):
    """File loading failure."""

```

lib/ansiblelint/utils.py
```diff
@@ -181,19 +182,24 @@ def find_children(playbook, playbook_dir):
...
+                        break
+                    valid_tokens.append(token)
+                path = ' '.join(valid_tokens)
+                results.append({
+                    'path': path_dwim(basedir, path),
+                    'type': child['type']
+                })
+        except MatchError as e:
+            e.filename = playbook[0]
+            raise e
```
webknjaz, line 201: This try/except block is too broad. It should contain only one instruction inside. Since the exception is happening in `play_children()`, it should be either moved there or you could create a decorator-style wrapper that would encapsulate this.

lib/ansiblelint/utils.py
```diff
@@ -181,19 +182,24 @@ def find_children(playbook, playbook_dir):
...
-                if '=' in token:
-                    break
-                valid_tokens.append(token)
-            path = ' '.join(valid_tokens)
-            results.append({
-                'path': path_dwim(basedir, path),
-                'type': child['type']
-            })
+        try:
+            for child in play_children(basedir, item, playbook[1], playbook_dir):
```
webknjaz, line 186: Here's how you could solve this:
```suggestion
            def rebind_match_filename(filename, func):
                def func_wrapper(*args, **kwargs):
                    try:
                        return func(*args, **kwargs)
                    except MatchError as e:
                        e.filename = filename
                        raise e
                return func_wrapper
            for child in rebind_match_filename(playbook[0], play_children)(basedir, item, playbook[1], playbook_dir):
```

lib/ansiblelint/utils.py
```diff
@@ -181,19 +182,33 @@ def find_children(playbook, playbook_dir):
...
-                if '=' in token:
-                    break
-                valid_tokens.append(token)
-            path = ' '.join(valid_tokens)
-            results.append({
-                'path': path_dwim(basedir, path),
-                'type': child['type']
-            })
+        try:
+            def rebind_match_filename(filename, func):
```
webknjaz, line 186: This can be moved to the root level of the module.

lib/ansiblelint/utils.py
```diff
@@ -181,19 +182,33 @@ def find_children(playbook, playbook_dir):
...
-            for token in split_args(child['path']):
-                if '=' in token:
-                    break
-                valid_tokens.append(token)
-            path = ' '.join(valid_tokens)
-            results.append({
-                'path': path_dwim(basedir, path),
-                'type': child['type']
-            })
+        try:
```
webknjaz, line 185: try/except can now be removed in this place.

docs/docsite/rst/rules/default_rules.rst
```diff
@@ -56,4 +56,10 @@ ID
...
+
+**E9xx - core**
+`E901 <https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/LoadingFailureRule.py>`_                                                                                                                                                                                                    v4.3.0                                                                                                                                                                                                                                                                                                       Failed to load or parse file                                                                                                                                                                                                                                                                                 Linter failed to process a YAML file, possible not an Ansible file.
+
+**E9xx - core**
```
ssbarnea, line 63: This is caused by the fact that the class is loaded twice, is a bug docs generate logic, wait for https://github.com/ansible/ansible-lint/pull/779 to merge and we should not see it anymore.

PR #783, opened 2020-05-12
Move AnsibleLintRule inside rules module

ssbarnea:
Moved code to ease reusability and avoid cyclic imports.

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -22,121 +22,13 @@
 from collections import defaultdict
 import logging
 import os
-import re
 
+from ansiblelint.rules import AnsibleLintRule  # noqa F401: exposing public API
 import ansiblelint.utils
 import ansiblelint.skip_utils
-from ansiblelint.errors import Match
 
 default_rulesdir = os.path.join(os.path.dirname(ansiblelint.utils.__file__), 'rules')
 _logger = logging.getLogger(__name__)
-
-
-class AnsibleLintRule(object):
-
-    def __repr__(self):
-        """Return a AnsibleLintRule instance representation."""
-        return self.id + ": " + self.shortdesc
-
-    def verbose(self):
-        return self.id + ": " + self.shortdesc + "\n  " + self.description
-
-    match = None
-    matchtask = None
-    matchplay = None
-
-    @staticmethod
-    def unjinja(text):
-        return re.sub(r"{{[^}]*}}", "JINJA_VAR", text)
-
-    def matchlines(self, file, text):
-        matches = []
-        if not self.match:
-            return matches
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/__init__.py
```diff
@@ -22,123 +22,15 @@
 from collections import defaultdict
 import logging
 import os
-import re
 
+from ansiblelint.rules import AnsibleLintRule  # noqa
```
webknjaz, line 26: Please avoid `noqa` unless there's a serious reason to do so.

ssbarnea: That is keeping the API compatibility with older versions. Any custom rules would import the base class from here, as that was also included inside the docs.

AnsibleLintRule is likely the only effectively public API for the linter, others stuff are likely to be considered private.

lib/ansiblelint/__init__.py
```diff
@@ -22,123 +22,15 @@
 from collections import defaultdict
 import logging
 import os
-import re
 
+from ansiblelint.rules import AnsibleLintRule # noqa F401 backwards compatibility
```
webknjaz, line 26: I'm pretty sure that the correct syntax would be
```suggestion
from ansiblelint.rules import AnsibleLintRule  # noqa: F401 
 # exposing the public API
```

lib/ansiblelint/__init__.py
ssbarnea: I improved the exception, explaining it.

lib/ansiblelint/generate_docs.py
```diff
@@ -60,15 +58,10 @@ def import_all_rules():
...
-    class_list = []
-    for mod in mod_list:
-        class_temp = [m[1] for m in getmembers(mod) if isclass(m[1])]
-        class_temp = [c for c in class_temp if c is not AnsibleLintRule]
-        class_list.extend(class_temp)
+    # called after all classes are loaded
 
     all_rules = []
-    for c in class_list:
+    for c in AnsibleLintRule.__subclasses__():
```
webknjaz, line 64: I'm not sure if this'd work because you'd have to import all the modules in this runtime first.

webknjaz: So I'd leave this change out of the PR.

ssbarnea: This is `what import_all_rules` does just before calling `get_serialized_rules`. The proof that it works is that the docs are not affected by this change.

I do have many questions regarding how the  generate_docs.py was originally implemented but that would be subject of other changes. For the moment I did only what was needed in order to allow me to move AnsibleLintRule definition inside.

PR #784, opened 2020-05-12
docstrings: narrowed down Dxxx ignores

ssbarnea:
- bumped linters
- resolved few Dxxx ignores and moved them to file level


.flake8
```diff
@@ -37,11 +37,6 @@ exclude =
 
 # Completely and unconditionally ignore the following errors:
 ignore =
-  # TODO(ssbarnea): gradually address exceptions below:
-  D100  # D100 Missing docstring in public module
-  D101  # D101 Missing docstring in public class
-  D102  # D102 Missing docstring in public method
-  D103  # D103 Missing docstring in public function
   # W503 is incompatible with W504
   W504  # W504 line break after binary operator
```
webknjaz, line 42: Looks like this section can be removed completely. Both W503 and W504 are in default_ignore.

docs/docsite/rst/conf.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+"""Documentation Configuration."""
```
webknjaz, line 2: FTR in such cases it's usually fine to put a docstring after the whole commented block

examples/rules/TaskHasTag.py
```diff
@@ -1,13 +1,17 @@
+"""TaskHasTag."""
```
webknjaz, line 1:
```suggestion
"""Example implementation of a rule requiring tasks to have tags set."""
```

lib/ansiblelint/__main__.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+"""Command line implementation."""
```
webknjaz, line 2: Move this under the comment.

lib/ansiblelint/__main__.py
```diff
@@ -35,6 +35,7 @@
 
 
 def main():
+    """Linter cli main."""
```
webknjaz, line 38:
```suggestion
    """Linter CLI entry point."""
```

lib/ansiblelint/cli.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+"""CLI methods."""
```
webknjaz, line 2:
```suggestion
"""CLI parser setup and helpers."""
```

lib/ansiblelint/errors.py
```diff
@@ -1,4 +1,8 @@
+"""Code related to errors and exceptions."""
```
webknjaz, line 1:
```suggestion
"""Exceptions and error representations."""
```

lib/ansiblelint/errors.py
```diff
@@ -1,4 +1,8 @@
...
+
+
 class Match(object):
+    """A Match is a Rule failure detected during linting."""
```
webknjaz, line 5:
```suggestion
    """Rule violation detected during linting."""
```

lib/ansiblelint/utils.py
```diff
@@ -1,3 +1,4 @@
+"""Utilities."""
```
webknjaz, line 1: Plz put this under the comment.
```suggestion
"""Generic utility helpers."""
```

.flake8
ssbarnea: Is not, removing them would raise ~20 W504 errors.

webknjaz: This is weird: https://github.com/PyCQA/pycodestyle/blob/1797af2/pycodestyle.py#L84

webknjaz: btw it looks like it's recommended to use `extend-ignore` instead of `ignore`: https://gitlab.com/pycqa/flake8/-/issues/466#note_111487130

lib/ansiblelint/__main__.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
```
webknjaz, line 2: Let's not touch this line.

lib/ansiblelint/__main__.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
```
webknjaz, line 1:
```suggestion
#!/usr/bin/env python

```

.flake8
webknjaz: @ssbarnea locally, removing it doesn't cause violations. Did you just remove the option or the whole setting? I can imaging that removing an option will redefine default ignores to an empty string which is not what we want.

PR #793, opened 2020-05-19
Added reproducer for become in blocks

ssbarnea:
Reproduces bug #705 with an xfail test, so we can fix it in a
follow-up.

Related: #705
Closes: #706

### DIFF
--- a/test/TestSkipInsideYaml.py
+++ b/test/TestSkipInsideYaml.py
@@ -89,7 +89,6 @@ ROLE_TASKS_WITH_BLOCK_BECOME = '''
     - name: foo
       become: true
       block:
-
         - name: bar
           become_user: jonhdaa
           command: "/etc/test.sh"
--- a/test/TestSkipInsideYaml.py
+++ b/test/TestSkipInsideYaml.py
@@ -105,20 +105,25 @@ def test_role_tasks_with_block(default_text_runner):
     assert len(results) == 4
 
 
-def test_playbook(default_text_runner):
-    results = default_text_runner.run_playbook(PLAYBOOK)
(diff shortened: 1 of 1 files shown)


test/TestSkipInsideYaml.py
```diff
@@ -99,3 +114,12 @@ def test_playbook(default_text_runner):
 def test_role_meta(default_text_runner):
     results = default_text_runner.run_role_meta_main(ROLE_META)
     assert len(results) == 0
+
+
+@pytest.mark.xfail(
+    reason="Bug: https://github.com/ansible/ansible-lint/issues/705",
+    strict=True,
```
webknjaz, line 121: This is not really necessary because it's already strict by default: https://github.com/ansible/ansible-lint/blob/fd1bb1c/pytest.ini#L63.

test/TestSkipInsideYaml.py
```diff
@@ -99,3 +114,12 @@ def test_playbook(default_text_runner):
...
+)
+def test_block_become_inheritance(default_text_runner):
```
webknjaz, line 123: It's better to merge this with `test_playbook` with param+xfail.

test/TestSkipInsideYaml.py
```diff
@@ -80,6 +83,18 @@
   license: MIT
 '''
 
+ROLE_TASKS_WITH_BLOCK_BECOME = '''
+- hosts:
+  tasks:
+    - name: foo
+      become: true
+      block:
+
```
webknjaz, line 92: Is this empty line purposeful?
```suggestion
```

PR #799, opened 2020-05-26
Improve artifact collection

ssbarnea:
- Assure we collect failures too
- Use wildcards to collect logs from all tox envs

### DIFF
--- a/.github/workflows/tox-tests.yml
+++ b/.github/workflows/tox-tests.yml
@@ -68,7 +68,8 @@ jobs:
       env:
         TOXENV: ansible${{ matrix.ansible-version }}
     - name: Archive logs
-      uses: actions/upload-artifact@v1
+      uses: actions/upload-artifact@v2
+      if: ${{ always() }}
       with:
-        name: logs.zip
-        path: .tox/ansible${{ matrix.ansible-version }}/log/
+        name: logs-${{ matrix.python-version }}-${{ matrix.os }}.zip
+        path: .tox/**/log/


.github/workflows/tox-tests.yml
```diff
@@ -68,7 +68,8 @@ jobs:
       env:
         TOXENV: ansible${{ matrix.ansible-version }}
     - name: Archive logs
-      uses: actions/upload-artifact@v1
+      uses: actions/upload-artifact@v2
+      if: ${{ always() }}
       with:
         name: logs.zip
```
webknjaz, line 74: This will still conflict with other jobs in the matrix.

.github/workflows/tox-tests.yml
```diff
@@ -68,7 +68,8 @@ jobs:
       env:
         TOXENV: ansible${{ matrix.ansible-version }}
     - name: Archive logs
-      uses: actions/upload-artifact@v1
+      uses: actions/upload-artifact@v2
+      if: ${{ always() }}
```
webknjaz, line 72: The idiomatic way to having if-clauses is as follows:
```suggestion
      if: always()
```

.github/workflows/tox-tests.yml
```diff
@@ -68,7 +68,8 @@ jobs:
...
       with:
-        name: logs.zip
-        path: .tox/ansible${{ matrix.ansible-version }}/log/
+        name: logs-${{ matrix.python-version }}-${{ matrix.os }}.zip
```
webknjaz, line 74: This would still have conflicts, you need to add ansible version factor into the name too:
```suggestion
        name: logs-${{ matrix.ansible-version }}-${{ matrix.python-version }}-${{ matrix.os }}.zip
```

Issue #778, opened 2020-05-08
auto-detect modules with collection layouts


ssbarnea:
Ansible-lint ability to just work without any extra tuning can be improved by making it automatically define `ANSIBLE_LIBRARY=plugins/modules` when the variable is not already defined.

This would follow the [official collection repository layout](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections.html) and make ansible-lint more likely work without extra configuration.

This feature should be enabled only with auto-detection mode because it needs to know what is the repository root location. That is because the tool could be called from any subdirectory and we do expect to give the same kind of results.

When implement this feature should allow people to remove extra code added to files like `tox.ini` or `.pre-commit-config.yaml` that define ANSIBLE_LIBRARY in order to be able perform the linting.

ssbarnea:
@webknjaz While trying to fix this I discovered one problem inside __main__, where I tried to alter ANSIBLE_LIBRARY in order to make it discoverable.

By the time __main__ runs, we already imported ansible and altering it will no longer work. That is happening because we do import from ansiblelint.utils which imports from ansible.

Fixing this is not very easy because we host `initialize_logger()` inside utils, which needs to run ASAP, clearly before our detection code.

webknjaz:
@ssbarnea I think this could be done by splitting the utils.

ssbarnea:
While spliting utils may move us into a good direction it will not be enough to address it.

Mainly we are affected by https://github.com/ansible/ansible/issues/69758 and we need to find a way to config Ansible to rediscover its modules after some ansible module imports already happened.

PR #801, opened 2020-05-27
Enable stub/fake modules

ssbarnea:
Provides a way to define a list of ansible modules that would be
mocked by the linter in order to allow Ansible to load the linted
files.

This should fix a very common problem which hindered linting on
various repositories, where the modules where hard to install inside
the linting jobs.

This changes also auto-adds `plugins/modules` if the folder is found,
as this is the official location for putting modules.

Fixes: #778

### DIFF
--- a/examples/roles/bobbins/tasks/main.yml
+++ b/examples/roles/bobbins/tasks/main.yml
@@ -1,3 +1,7 @@
 ---
 - name: test tasks
   action: git a=b c=d
+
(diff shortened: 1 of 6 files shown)


ssbarnea:
@webknjaz @cloudnull @avass Any ideas on how to improve this?

ssbarnea:
Closing in favour of https://github.com/ansible/ansible-lint/pull/804 which is much cleaner and avoids altering the environment unless is needed.

PR #803, opened 2020-05-28
Remove implementation from __init__

ssbarnea:
- Refactored code out of __init__ by moving last two classes out of it.
- Documents the new imports for custom rule writers.
- In order to avoid a circular import, import_plugins function is moved from utils to rules, the only place where it was used.
- This move enables us to import only needed bits of ansiblelint package, without being forced to load almost everything due to the code present inside __init__.py file.
- Keeps a temporary backwards compatibility import for AnsibleLintRule

Fixes: #812
Related: #801


ssbarnea:
Closing in favour of https://github.com/ansible/ansible-lint/pull/804 which is much cleaner and avoids altering the environment unless is needed.

lib/ansiblelint/rules/AlwaysRunRule.py
```diff
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-from ansiblelint import AnsibleLintRule
+from ansiblelint.rules import AnsibleLintRule
```
webknjaz, line 21: I'd like to see relative imports in rules instead:
```suggestion
from . import AnsibleLintRule
```

lib/ansiblelint/__init__.py
```diff
@@ -19,167 +19,9 @@
...
-                "Examining %s of type %s",
-                ansiblelint.utils.normpath(file['path']),
-                file['type'])
-            matches.extend(self.rules.run(file, tags=set(self.tags),
-                           skip_list=self.skip_list))
-        # update list of checked files
-        self.checked_files.update([x['path'] for x in files])
-
-        return matches
+default_rulesdir = os.path.join(os.path.dirname(__file__), 'rules')
```
webknjaz, line 27: It seems weird that this var is declared right here.

lib/ansiblelint/__init__.py
```diff
@@ -19,167 +19,9 @@
...
 import os
 
-from ansiblelint.rules import AnsibleLintRule  # noqa F401: exposing public API
-import ansiblelint.utils
-import ansiblelint.skip_utils
-from ansiblelint.errors import MatchError
-from ansiblelint.rules.LoadingFailureRule import LoadingFailureRule
-from typing import List, Set
+# package should not do any imports, especially those that would have side
+# effect of importing ansible.
```
webknjaz, line 25: We should postpone removing the public API imports.

lib/ansiblelint/runner.py
```diff
@@ -0,0 +1,94 @@
+"""Runner Module."""
```
webknjaz, line 1: It should describe that's inside, not just recite the file name and things that are already known. This is hard but we could do at least a bit better:
```suggestion
"""Runner implementation."""
```

lib/ansiblelint/runner.py
```diff
@@ -0,0 +1,94 @@
...
+
+import ansiblelint.utils
+import ansiblelint.skip_utils
+from ansiblelint.errors import MatchError
+from ansiblelint.rules.LoadingFailureRule import LoadingFailureRule
+from typing import List, Set
+
+
+class Runner(object):
+    """Runner Class."""
```
webknjaz, line 12:
```suggestion
class Runner:
    """Tasks runner."""
```

lib/ansiblelint/runner.py
```diff
@@ -0,0 +1,94 @@
+"""Runner Module."""
+import os
+
+import ansiblelint.utils
+import ansiblelint.skip_utils
+from ansiblelint.errors import MatchError
+from ansiblelint.rules.LoadingFailureRule import LoadingFailureRule
+from typing import List, Set
```
webknjaz, line 8: It's an stdlib import.

webknjaz, line 8:
```suggestion
import os
from typing import List, Set

import ansiblelint.utils
import ansiblelint.skip_utils
from .errors import MatchError
from .rules.LoadingFailureRule import LoadingFailureRule
```

lib/ansiblelint/runner.py
```diff
@@ -0,0 +1,94 @@
...
+                    matches.append(e.get_match())
+                visited.add(arg)
+
+        # remove duplicates from files list
+        files = [value for n, value in enumerate(files) if value not in files[:n]]
+
+        # remove files that have already been checked
+        files = [x for x in files if x['path'] not in self.checked_files]
+        for file in files:
+            ansiblelint.utils._logger.debug(
```
webknjaz, line 85: The global logger mustn't be used.

lib/ansiblelint/rules/AlwaysRunRule.py
ssbarnea: Why? Can you please show where is the use of relative imports indicated as a better coding practice?

Also, why to change import types as part of this change? This looks outside the scope of current change.

lib/ansiblelint/__init__.py
ssbarnea: @webknjaz I agree with you but that is also outside the scope. Likely is kept here in order to allow some consumers to override it. We can think about changing this in different PR.

lib/ansiblelint/__init__.py
ssbarnea: AFAIK that is not possible. Breaking the API is mandatory as part of cleaning the   `__init__py` file. It is not possible to keep it compatible and avoid running undesired imports chain.

We need to document the change well, allowing custom rule users. Also that is the kind of change that cannot be made in a minor release, so 4.3.0 is the right kind of release to do it.

I challenge you to find a way to avoid id. I am quite curious as dislike breaking API too. The reality is that AnsibleLintRule class loading cannot happen without loading ansible modules. If this happens in ``__init__``, we loose ability to run ``__main__`` module w/o already having ansible imported.

If we move the bootstrapping our of __main__ to init we loose the ability to perform boostrapping using config options or logger, both needed.

lib/ansiblelint/runner.py
ssbarnea: mypy == typing, not really stdlib but close enough.

lib/ansiblelint/rules/AlwaysRunRule.py
webknjaz: Absolute imports may hit places outside of this project while relative imports point to the target modules explicitly. It's in the scope of this PR because you added a bunch of imports all around the place.

ssbarnea: It seems that you missed to spot that use of relative imports will break [load_plugins](https://github.com/ansible/ansible-lint/blob/7cffde295e238f13f563ebdeb1d4722589353fcb/lib/ansiblelint/utils.py#L129) and triger something like:
```
File "/Users/ssbarnea/.pyenv/versions/3.8.2/bin/ansible-lint", line 11, in <module>
load_entry_point('ansible-lint', 'console_scripts', 'ansible-lint')()
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/__main__.py", line 66, in main
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/rules/__init__.py", line 136, in __init__
self.extend(ansiblelint.utils.load_plugins(rulesdir))
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/utils.py", line 137, in load_plugins
spec.loader.exec_module(module)
File "<frozen importlib._bootstrap_external>", line 783, in exec_module
File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/rules/RoleNames.py", line 23, in <module>
from .roles import AnsibleLintRule
ImportError: attempted relative import with no known parent package
```

So I would defer switching from absolute imports to relative ones for a different change. At this moment this keeps the change at minimum: only points to the newer location of the AnsibleLintRule class.

In fact due to the introduction of the shim function, the code would run even without touching the import but it would raise a DeprecationWarning.

webknjaz: Agreed. I didn't remember that we have our own import logic there.

ssbarnea:
@webknjaz Any updates on this?

lib/ansiblelint/__init__.py
```diff
@@ -18,168 +18,15 @@
...
-                "Examining %s of type %s",
-                ansiblelint.utils.normpath(file['path']),
-                file['type'])
-            matches.extend(self.rules.run(file, tags=set(self.tags),
-                           skip_list=self.skip_list))
-        # update list of checked files
-        self.checked_files.update([x['path'] for x in files])
 
-        return matches
+def AnsibleLintRule(*args, **kwargs):
```
webknjaz, line 27: This will break `isinstance` checks.

I was thinking maybe add import level `__getattr__()` for python 3.7+ and just have a normal import for 3.6.

ssbarnea: The only use-case for AnsibleLintRule was to create your own classes, like we have on our examples. 

Do we have any real world example of this breaking? I am sure not in our docs, examples or tests, which are mainly the only contract.

Anyone encountering the deprecation warning should update their code ASAP to avoid other surprises.

AFAIK it makes no sense to introduce conditional behavior for py36 alone, as it would only complicate the code and not solving the issue.

Doing this would make impossible to fix py36 ansible imports, shortly we need to be sure we do not endup import ansible.* from ansiblelint.__init__ on *any* supported platforms (py,ansible,os).

On the other hand, we can even follow Ansible team practice and state that, if [not explicitly documented is not supported](https://github.com/ansible/ansible/issues/69758).

webknjaz: Oh, and it'll break the inheritance too. This one is supported and will explode when somebody will try to inherit a class from a function:

```python


class ANewRule(AnsibleLintRule):  # 💣💣💣 Oops! 💣💣💣
    ...
```

ssbarnea: Dropped any backwards compatibility attempts as they cause more problems than solving.

webknjaz: Then keep the imports and don't break the API for now.

ssbarnea: Keeping the imports is not possible, it would create cyclic imports. That is one of the major issues with putting code inside __init__.

test/TestRunner.py
```diff
@@ -20,8 +20,7 @@
 import os
 
 import pytest
-
-from ansiblelint import formatters
+import ansiblelint.formatters as formatters
```
webknjaz, line 23: Local project imports should be separate:
```suggestion

import ansiblelint.formatters as formatters
```

lib/ansiblelint/rules/__init__.py
```diff
@@ -1,11 +1,34 @@
...
+
+    for pluginfile in glob.glob(os.path.join(directory, '[A-Za-z]*.py')):
+
+        pluginname = os.path.basename(pluginfile.replace('.py', ''))
+        spec = importlib.util.spec_from_file_location(pluginname, pluginfile)
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        obj = getattr(module, pluginname)()
+        # We load only rules with a valid id
+        if issubclass(obj.__class__, AnsibleLintRule):
```
webknjaz, line 27: This changes the implementation and so it belongs to a separate PR.

lib/ansiblelint/rules/__init__.py
```diff
@@ -15,6 +38,9 @@ def __repr__(self):
     def verbose(self):
         return self.id + ": " + self.shortdesc + "\n  " + self.description
 
+    id = ""
+    shortdesc = ""
+    description = ""
```
webknjaz, line 43: This PR is for moving things from the package init, not other things.

lib/ansiblelint/rules/__init__.py
```diff
@@ -120,3 +146,70 @@ def matchyaml(self, file, text):
...
+
+class RulesCollection(object):
+
+    def __init__(self, rulesdirs=None):
+        """Initialize a RulesCollection instance."""
+        if rulesdirs is None:
+            rulesdirs = []
+        self.rulesdirs = ansiblelint.utils.expand_paths_vars(rulesdirs)
+        self.rules = []
+        for rulesdir in self.rulesdirs:
```
webknjaz, line 159: Logger call got lost here

PR #804, opened 2020-05-28
WIP: Enable stub/fake modules

ssbarnea:
Contains #803 which should be merged fist.

Fixes: https://github.com/ansible/ansible-lint/issues/806 allow module_stubs
Fixes: https://github.com/ansible/ansible-lint/issues/778 load plugins/modules


lib/ansiblelint/__main__.py
```diff
@@ -20,23 +20,98 @@
...
+    # add plugins/modules to module paths if it exists
+    p = pathlib.Path.cwd() / "plugins" / "modules"
+    if p.exists() and str(p) not in lib_paths:
+        _logger.info(f"Adding {str(p)} to module path")
+        lib_paths.append(str(p))
+        changed_module_path = True
+
+    if options.stub_modules:
+        # creates temporary directory for stub modules (auto-cleaned)
+        tmp_dir = tempfile.TemporaryDirectory()
```
webknjaz, line 99: I'm pretty sure this should be used as a context manager.

lib/ansiblelint/__main__.py
```diff
@@ -20,23 +20,98 @@
...
+        _logger.info(f"Adding {tmp_dir.name} to module path")
+        changed_module_path = True
+
+        for m in options.stub_modules:
+            with (pathlib.Path(tmp_dir.name) / f"{m}.py") as f:
+                _logger.debug(f"Generating fake module stub: {f.name}")
+                f.write_text(MODULE_STUB)
+
+    if changed_module_path:
+        os.environ['ANSIBLE_LIBRARY'] = ":".join(lib_paths)
```
webknjaz, line 110: This is unnecessary. The only thing needed is
```suggestion
        from ansible.plugins.loader import module_loader
        module_loader.add_directory(tmp_dir.name)
```

In fact, it's already used in other places: https://github.com/ansible/ansible-lint/blob/0cf285b/lib/ansiblelint/utils.py#L292-L294.

Issue #540, opened 2019-06-16
Unable to find files (but not really)


Thulium-Drake:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
$ ansible --version
ansible 2.8.0
  config file = None
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/ansible/2.8.0/local/lib/python2.7/site-packages/ansible
  executable location = /opt/ansible/2.8.0/bin/ansible
  python version = 2.7.13 (default, Sep 26 2018, 18:42:22) [GCC 6.3.0 20170516]
$ ansible-lint --version
ansible-lint 4.1.0
```
- ansible installation method: pip
- ansible-lint installation method: pip

# Actual Behaviour (Bug report only)

When running ansible-lint (via Molecule or without) I get the following warnings about it being unable to find some files:

```
$ ansible-lint .
WARNING: Couldn't open /root/roles/acme_ssl/tasks/{u'file': - No such file or directory
WARNING: Couldn't open /root/roles/acme_ssl/tasks/{u'file': u'acme_http01_create.yml'} - No such file or directory
WARNING: Couldn't open /root/roles/acme_ssl/tasks/{u'file': u'acme_dns01_create.yml'} - No such file or directory
WARNING: Couldn't open /root/roles/acme_ssl/tasks/{u'file': u'acme_http01_cleanup.yml'} - No such file or directory
WARNING: Couldn't open /root/roles/acme_ssl/tasks/{u'file': u'acme_dns01_cleanup.yml'} - No such file or directory
```

When running with -vvv, you can see it _IS_ able to find and lint the files, but wants to do something with what seems a Python list interpreted quite litterally.

```
$ ansible-lint . -vvv
Examining /root/roles/acme_ssl/tasks/acme_dns01_create.yml of type tasks
Examining /root/roles/acme_ssl/tasks/main.yml of type tasks
Examining /root/roles/acme_ssl/tasks/acme_create_account.yml of type tasks
Examining /root/roles/acme_ssl/tasks/acme_http01_create.yml of type tasks
Examining /root/roles/acme_ssl/tasks/acme_dns01_cleanup.yml of type tasks
Examining /root/roles/acme_ssl/tasks/acme_http01_cleanup.yml of type tasks
Examining /root/roles/acme_ssl/tasks/{u'file': of type tasks
[...]

PR #807, opened 2020-06-02
Fixed the search method when the file path not exists

cahlchang:
##### SUMMARY
Fixed the search method when the file path was not exists.

##### ISSUE TYPE
Bugfix Pull Request

##### Related
https://github.com/ansible/ansible-lint/issues/540

##### INFORMATION

Ansible-lint uses a relative path to search again when the path of an include is not found.
https://github.com/ansible/ansible-lint/blob/654331d7c1ae43fbf224b924d6e1b0fb41074e98/lib/ansiblelint/utils.py#L259-L260

The `path_dwim` function refers to the `DataLoader.path_dwim` of ansible.
https://github.com/ansible/ansible-lint/blob/654331d7c1ae43fbf224b924d6e1b0fb41074e98/lib/ansiblelint/utils.py#L92-L94
https://github.com/ansible/ansible/blob/dc63b365011a583b9e9bcd60d1fad6fb10b553c7/lib/ansible/parsing/dataloader.py#L181-L195

The path is passed to the unfrackpath function and the relative path is stripped out.

https://github.com/ansible/ansible/blob/ba87c225cd13343c35075fe7fc15b4cf1343fed6/lib/ansible/utils/path.py#L31-L45

That's why the path to `include_task` will fail to resolve.In order to match the path search, it is better to move up the path of the basedir one level.And this search need not be limited to the `tasks` directory.This function should also be reexamined in the `handlers` directory, for example.


cahlchang:
@ssbarnea 
Thank you for reviewing it !!

I think your point is spot on.
But...
In fact, I focused my attention on the following test.
https://github.com/ansible/ansible-lint/blob/master/test/TestTaskIncludes.py.

but the test didn't really work.
The test seems to return OK, even if you put an obviously bad value in the path.

https://github.com/ansible/ansible-lint/blob/master/test/taskincludes_2_4_style.yml.

Probably, this is because the output when an invalid path is returned is warning, so it is not trapped by pytest.
I felt that to change this, I would need to make a lot of changes to the runner in ansible-lint.
That's why we've included an explanation of the logic of the change in our PR.

If necessary, you can change the test code so that it becomes an error when you include it, and you can add test.
If this test code seems to be valid in the future for ansible-lint, let's add the test code.

cahlchang:
Ah...I may have overlooked a few things.
Let's check the test code a bit.
Thank you.

cahlchang:
@ssbarnea 
Added a test.
Please review.

ssbarnea:
@webknjaz This seems to be working fine. The added tests should prevent future regressions on this area.

ssbarnea:
I marked it as draft as I does not pass tests now. If not fixed for a long time I may close it.

cahlchang:
Ah, I see.
I mistakenly thought there was no problem with this test case failure.
I'll get to work on solving that test case.

Thanks for letting me know.

cahlchang:
There may be some tests that need to be corrected by comparing them with the current master.
I'll address this soon, thanks.

cahlchang:
@ssbarnea 
I made sure it passed all the tests.
Could you please review it again?

PR #811, opened 2020-06-04
Refactor default_rulesdir

ssbarnea:
- Simplify -r/-R documentation by avoiding repetitions
- Correct grammar (avoid two plural nouns in a row)
- Reduce internal imports
- Avoid use of ``__init__``
- Add debugging message about effective directories used to load rules

## Before

>   -r RULESDIR           specify one or more rules directories using one or
>                         more -r arguments. Any -r flags override the default
>                         rules in /home/ssbarnea/.local/lib/python3.6/site-
>                         packages/ansiblelint/rules, unless -R is also used.
>   -R                    Use default rules in
>                         /home/ssbarnea/.local/lib/python3.6/site-
>                         packages/ansiblelint/rules in addition to any extra
>                         rules directories specified with -r. There is no need
>                         to specify this if no -r flags are used
> 

## After

>   -r RULESDIR           Specify custom rule directories. Add -R to keep using embedded rules from /Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/rules
>   -R                    Keep using default rules when using -r

Partial: #812

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -26,7 +26,8 @@ import pathlib
 import sys
 
 import ansiblelint.formatters as formatters
-from ansiblelint import cli, default_rulesdir, RulesCollection, Runner
+from ansiblelint import cli, RulesCollection, Runner
+from ansiblelint.constants import DEFAULT_RULESDIR
 from ansiblelint.utils import get_playbooks_and_roles
 from ansiblelint.utils import normpath, initialize_logger
 from ansiblelint.generate_docs import rules_as_rst
@@ -58,9 +59,9 @@ def main():
     formatter = formatter_factory(cwd, options.display_relative_path)
 
     if options.use_default_rules:
-        rulesdirs = options.rulesdir + [default_rulesdir]
+        rulesdirs = options.rulesdir + [DEFAULT_RULESDIR]
     else:
-        rulesdirs = options.rulesdir or [default_rulesdir]
+        rulesdirs = options.rulesdir or [DEFAULT_RULESDIR]
     rules = RulesCollection(rulesdirs)
 
     if options.listrules:
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -9,8 +9,9 @@ import sys
 import yaml
 from typing import NamedTuple
 
-import ansiblelint
+from ansiblelint.constants import DEFAULT_RULESDIR
 from ansiblelint.version import __version__
+import ansiblelint.utils
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


lib/ansiblelint/__main__.py
```diff
@@ -23,17 +23,19 @@
...
-from ansiblelint import cli, default_rulesdir, RulesCollection, Runner
+from ansiblelint import cli, RulesCollection, Runner
 from ansiblelint.utils import get_playbooks_and_roles
 from ansiblelint.utils import normpath, initialize_logger
 from ansiblelint.generate_docs import rules_as_rst
 from typing import Any, Set
 
 
 _logger = logging.getLogger(__name__)
+default_rulesdir = os.path.join(os.path.dirname(__file__), 'rules')
```
webknjaz, line 38: shouldn't this live somewhere in utils?

lib/ansiblelint/cli.py
```diff
@@ -110,17 +110,12 @@ def get_cli_parser() -> argparse.ArgumentParser:
                         help="parseable output including severity of rule")
     parser.add_argument('-r', action=AbspathArgAction, dest='rulesdir',
                         default=[], type=Path,
-                        help="specify one or more rules directories using "
-                             "one or more -r arguments. Any -r flags override "
-                             "the default rules in %s, unless -R is also used."
```
webknjaz, line 115: I don't like this being removed. Why can't the var just be in some utils module instead?

lib/ansiblelint/__main__.py
ssbarnea: As someone else said it before, presence of "utils" is another anti-pattern as that is likely to become the kitchensink of the package.

Default rules directory is specific only to the command line, so it makes sense to use either `__main__` or `cli` modules.

Take a look at https://hackmd.io/@42/ansible-lint-structure -- i build it in order to help us decide where to move things.

webknjaz: So put it in CLI then. This way you'd not have to change the help strings.

ssbarnea: If you want I can even add _ prefix to it, its only used internally.

webknjaz: yeah, that'd be fine

lib/ansiblelint/cli.py
ssbarnea: @webknjaz If you really want we could follow Ansible model and create a constants module where to put bits like this (also exit codes).

Regarding printing the internal include path with --help, i have some doubts that this is really needed. If I remember well, I was the one that added this but the reality is that there is not much use to it.

Use of logging provides a far better feedback for those wanting to understand the internals.

webknjaz: This is not the same as logging. More comprehensive help output improves the UX of folks who are trying to figure out what CLI args to use even before running the thing.

lib/ansiblelint/constants.py
```diff
@@ -0,0 +1,4 @@
+import os
```
webknjaz, line 1: Maybe even `_constants.py`?

lib/ansiblelint/cli.py
ssbarnea: I added it back to help, but without mentioning it twice. Help needed rephrasing anyway. I also created constants module, I will move other constants to it in follow-ups.

lib/ansiblelint/constants.py
ssbarnea: @webknjaz When Ansible team will start doing that, we can do it too. ;)

webknjaz, line 1:
```suggestion
import os.path
```

lib/ansiblelint/cli.py
```diff
@@ -110,17 +111,12 @@ def get_cli_parser() -> argparse.ArgumentParser:
...
+                        help="Specify custom rule directories. Add -R "
+                             f"to keep using embedded rules from {DEFAULT_RULESDIR}")
     parser.add_argument('-R', action='store_true',
                         default=False,
                         dest='use_default_rules',
-                        help="Use default rules in %s in addition to any extra "
-                             "rules directories specified with -r. There is "
-                             "no need to specify this if no -r flags are used"
-                             % ansiblelint.default_rulesdir)
+                        help="Keep using default rules when using -r")
```
albinvass, line 119: "Keep default rules when using -r"

lib/ansiblelint/constants.py
```diff
@@ -0,0 +1,5 @@
+"""Constants used by AnsibleLint."""
+import os.path
```
webknjaz, line 2: Future note: I'd migrate this to pathlib but it's out of scope.

Issue #813, opened 2020-06-04
Ansible lint should not crash with an empty config file


ssbarnea:
##### Summary
```
ansible-lint -c /dev/null
```

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
Traceback (most recent call last):
File "/Users/ssbarnea/.pyenv/versions/3.8.2/bin/ansible-lint", line 11, in <module>
load_entry_point('ansible-lint', 'console_scripts', 'ansible-lint')()
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/__main__.py", line 45, in main
options = cli.get_config(sys.argv[1:])
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/cli.py", line 204, in get_config
config = load_config(options.config_file)
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/cli.py", line 78, in load_config
expand_to_normalized_paths(config, config_dir)
File "/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/cli.py", line 41, in expand_to_normalized_paths
if paths_var not in config:
TypeError: argument of type 'NoneType' is not iterable
```

Desired behavior: run the same way as if no config would exist, avoiding a crash.

ssbarnea:
@webknjaz FYI, if user mentions an invalid filename, the tool will end with an error, like below:
```
$ ansible-lint -c /dev/sdsds -v                                                                                                                     Config file not found '/dev/sdsds'
FAIL: 2
```

This change only allows running with empty existing config files, which are still valid YAML files. Basically /dev/null was used just as an example.

PR #814, opened 2020-06-04
Allow null config file

ssbarnea:
Avoids an ugly exception when the config file was empty. Enables
ignoring of configuration file by using something like: `-c /dev/null`

Fixes: #813

### DIFF
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -36,6 +36,9 @@ def abspath(path, base_dir):
 
 
 def expand_to_normalized_paths(config, base_dir=None):
+    # config can be None (-c /dev/null)
+    if not config:
+        return
     base_dir = base_dir or os.getcwd()
     for paths_var in _PATH_VARS:
         if paths_var not in config:


ssbarnea:
@webknjaz @awcrosby low hanging!

webknjaz:
I think we should special-case `/dev/null`. Tolerating typos in paths is bad UX.

lib/ansiblelint/cli.py
```diff
@@ -36,6 +36,9 @@ def abspath(path, base_dir):
 
 
 def expand_to_normalized_paths(config, base_dir=None):
+    # config can be None (-c /dev/null)
+    if not config:
```
webknjaz, line 40: If it can be `None`, why not check for it?

ssbarnea: Comparision with None would not be pythonic and would rely on implementation details. I could as well introduce a bug where and empty dictionary (or similar structure) would be considered "as having content".

Is it clear to everyone that this change does NOT introduce toleration of broken file paths? They will continue to be fatal errors. That is only about loading empty config files, where empty loose concept (no config items specified, it could still be a non empty file, like one with comments).

PR #819, opened 2020-06-08
Changed AnsibleLintRule imports

ssbarnea:
Change all imports of AnsibleLintRule to point to its future location.

This change does not move implementation, that will be done in a
follow-up in order to make it easier to review.

Related: #801

### DIFF
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -22,7 +22,6 @@
 from collections import defaultdict
 import logging
 
-from ansiblelint.rules import AnsibleLintRule  # noqa F401: exposing public API
 import ansiblelint.utils
 import ansiblelint.skip_utils
 from typing import List
@@ -97,3 +96,6 @@ class RulesCollection(object):
         for tag in sorted(tags):
             results.append("{0} {1}".format(tag, tags[tag]))
         return "\n".join(results)
+
+
+from ansiblelint.rules import AnsibleLintRule  # noqa F401: exposing public API
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -120,3 +120,5 @@ class AnsibleLintRule(object):
                 matches.append(Match(linenumber,
                                      section, file['path'], self, message))
         return matches
+
+from ansiblelint import RulesCollection  # noqa F401: new class location
--- a/test/TestRunner.py
+++ b/test/TestRunner.py
@@ -20,8 +20,7 @@
 import os
 
 import pytest
-
-from ansiblelint import formatters
+import ansiblelint.formatters as formatters
 from ansiblelint.runner import Runner
 from ansiblelint.cli import abspath
 
--- a/test/TestRunner.py
+++ b/test/TestRunner.py
@@ -20,6 +20,7 @@
 import os
 
 import pytest
+
 import ansiblelint.formatters as formatters
 from ansiblelint.runner import Runner
 from ansiblelint.cli import abspath
--- a/README.rst
+++ b/README.rst
@@ -541,7 +541,7 @@ An example rule using ``match`` is:
 
 .. code-block:: python
 
-    from ansiblelint import AnsibleLintRule
+    from ansiblelint.rules import AnsibleLintRule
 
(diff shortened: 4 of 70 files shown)


lib/ansiblelint/__init__.py
```diff
@@ -22,7 +22,6 @@
 from collections import defaultdict
 import logging
 
-from ansiblelint.rules import AnsibleLintRule  # noqa F401: exposing public API
```
webknjaz, line 25: Let's avoid relocating this to the bottom.

lib/ansiblelint/rules/__init__.py
```diff
@@ -120,3 +120,5 @@ def matchyaml(self, file, text):
                 matches.append(Match(linenumber,
                                      section, file['path'], self, message))
         return matches
+
+from ansiblelint import RulesCollection  # noqa F401: new class location
```
webknjaz, line 124: Let's not do imports at the bottom.

test/TestRunner.py
```diff
@@ -20,8 +20,7 @@
 import os
 
 import pytest
-
-from ansiblelint import formatters
+import ansiblelint.formatters as formatters
```
webknjaz, line 23: Keep the separation between project imports and third parties.
```suggestion

import ansiblelint.formatters as formatters
```

lib/ansiblelint/rules/__init__.py
ssbarnea: @webknjaz This is why I asked you to test yourself, that was needed in order to avoid the cyclic-import. This code has a cyclic import but it avoids a runtime failure due to this late-import.

lib/ansiblelint/__init__.py
ssbarnea: That is by design because otherwise you will get circular imports failure:
```
ImportError while loading conftest '/Users/ssbarnea/c/os/ansible-lint/test/conftest.py'.
test/__init__.py:7: in <module>
lib/ansiblelint/__init__.py:28: in <module>
    from ansiblelint.rules import AnsibleLintRule  # noqa F401: exposing public API
lib/ansiblelint/rules/__init__.py:124: in <module>
    from ansiblelint import RulesCollection  # noqa F401: new class location
E   ImportError: cannot import name 'RulesCollection' from partially initialized module 'ansiblelint' (most likely due to a circular import) (/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/__init__.py)
FAIL: 4
```

ssbarnea:
I will post it here to be clear: imports at the end of the file are needed in order to avoid cyclic import failures like below:

```
ImportError while loading conftest '/Users/ssbarnea/c/os/ansible-lint/test/conftest.py'.
test/__init__.py:7: in <module>
lib/ansiblelint/__init__.py:28: in <module>
    from ansiblelint.rules import AnsibleLintRule  # noqa F401: exposing public API
lib/ansiblelint/rules/__init__.py:124: in <module>
    from ansiblelint import RulesCollection  # noqa F401: new class location
E   ImportError: cannot import name 'RulesCollection' from partially initialized module 'ansiblelint' (most likely due to a circular import) (/Users/ssbarnea/c/os/ansible-lint/lib/ansiblelint/__init__.py)
FAIL: 4
```

So, before asking to move imports, see if that is possible.

ssbarnea:
> If there's cyclic imports, they should be solved properly instead of applying a band-aid.
> In this case, it looks like the first PR should've been relocating `RulesCollection` so that it doesn't cause loops in the first place.

Sorry but after almost two weeks since I initiated the [__init__ refactorings](https://github.com/ansible/ansible-lint/pull/801), the only thing I get is impossible requests, suggestions that do not even pass project own tests. Mainly the message I get is: "Yes, we need to clean init but don't touch anything". 

As we already investigated multiple approaches in https://github.com/ansible/ansible-lint/pull/803 iterations the band-aid was suggested by you when you asked "keep the import for backward compatibility" (which in fact translates to blocking the full implementation of #803)

If you know better ways of achieving what is described at https://github.com/ansible/ansible-lint/pull/801 please propose them and I will happy to test and review them.

ssbarnea:
Should be covered by  https://github.com/ansible/ansible-lint/pull/803/

PR #821, opened 2020-06-09
Moved exit codes to constants

ssbarnea:
Moves two more constants to their specific module, causing simplification of  internal module dependency graph.

### DIFF
--- a/lib/ansiblelint/constants.py
+++ b/lib/ansiblelint/constants.py
@@ -3,3 +3,6 @@ import os.path
 
 
 DEFAULT_RULESDIR = os.path.join(os.path.dirname(__file__), 'rules')
+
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/constants.py
```diff
@@ -3,3 +3,6 @@
 
 
 DEFAULT_RULESDIR = os.path.join(os.path.dirname(__file__), 'rules')
+
+INVALID_CONFIG_RC = 2
+ANSIBLE_FAILURE_RC = 3
```
webknjaz, line 8: It'd be better if this was in Enum.

ssbarnea: I have mixed feelings about that. Lets see how it grows and we can refactor it later.

PR #822, opened 2020-06-10
Make testing ansible-2.10 compatible

ssbarnea:
- avoids use of ec2 and chocolately modules in tests, as they do not longer
  ship with ansible in 2.10
- avoids installing extra galaxy collections are they take too much time and randomly fail

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -31,9 +31,6 @@ deps =
 commands =
   # safety measure to assure we do not accidentaly run tests with broken dependencies
   python -m pip check
-  # Collections outside of ACD that config/routing.yml points to in Core:
-  ansibledevel: ansible-galaxy collection install --collections-path "{envtmpdir}" amazon.aws
-  ansibledevel: ansible-galaxy collection install --collections-path "{envtmpdir}" chocolatey.chocolatey
   {envpython} -m pytest \
   --cov "{envsitepackagesdir}/ansiblelint" \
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -49,6 +46,8 @@ setenv =
   ANSIBLE_COLLECTIONS_PATHS = {envtmpdir}
   COVERAGE_FILE = {env:COVERAGE_FILE:{toxworkdir}/.coverage.{envname}}
   PIP_DISABLE_PIP_VERSION_CHECK = 1
+  ansibledevel: PIP_EXTRA_INDEX_URL = https://toshio.fedorapeople.org/ansible/acd/
+  ansibledevel: PIP_PRE = 1
```
webknjaz, line 50: This does nothing since the deps target full URLs.

tox.ini
```diff
@@ -32,9 +32,6 @@ deps =
 commands =
   ansibledevel: sh -c "{envpython} -m pip show ansible | >/dev/null grep '^Version: 2.10.0$'"
   ansibledevel: sh -c "{envpython} -m pip show ansible-base | >/dev/null grep '^Version: 2.10.0.dev0$'"
-  # Collections outside of ACD that config/routing.yml points to in Core:
```
webknjaz, line 35: This doesn't fix anything currently because the problem is on the install stage, not in commands.

PR #824, opened 2020-06-11
Disable ansible devel testing

ssbarnea:
As ansible-devel proved to be break our jobs far too often, we will
stop testing it until a working pre-release is published on
pypi.org.

This will allow the project to move-on without being blocked or
slowed down by ansible internal works.

This changes should be reverted once we have an official pre-release.


webknjaz:
Hold on. I think I have a workaround for this already.

PR #825, opened 2020-06-11
🚑 Install amazon.aws from GitHub

webknjaz:
This works around the broken command:

    $ ansible-galaxy collection install amazon.aws

Resolves #824.


webknjaz:
This addresses one of the issues we have. One more to go.

Issue #823, opened 2020-06-10
proposal: reduce number of CI/CD involved with ansible-lint


ssbarnea:
##### Summary
Currently ansible-lint project is configured to use not less than 4 major CI/CD systems: travis, azure-pipelines, github-actions and zuul. In addition to these it also have at least one extra bot (LGTM).

> At one moment in time there were more than 100 checks running on this project, something I never seen on any other github project. Not even big and sensitive projects like libpod or openstack do run so many jobs. And most of them where only instances of the same set of tox environments configured on different CI/CD systems.

##### Issue Type

- Feature Idea

##### Additional Information

That is not the first time I raise this question but it seems that I need to do it again as nothing happened.

This current concoction of different external systems is causing lots of false failures, mainly related to either issues internal to these CI/CD systems or just networking.

The amount of checks running on this projects makes is very confusing for the user, also their errors are far from easy to read.

As non core developer is impossible for me to retrigger these jobs without closing and reopening the PR. The only exception is Zuul where a "recheck" comment does this, still that one seems not to be the noisy one of of the all 4.

As Zuul CI is run and supported by Ansible Networking team with plans to extend it even further to cover more Ansible project, we can rule it out of the "removal" list.

I propose dropping:
- azure-pipelines (seems to be the source of the most failures, leading by far)
- travis (degraded QoS year over year)

We need to be sure we still cover: macos, windows and linux. Linux ones are covered for sure by zuul jobs.

Before arguing that "this CI has a feature that I like that is missing from another", please remember that we cannot make a kitchensink out of ansible-lint project and resilience is more important than features.

Resiliency cannot be obtained by increasing entropy, we need to simplify as much as possible.

ssbarnea:
![](https://github.com/ansible/ansible-lint/pull/822/files)

PR #830, opened 2020-06-15
Remove azure-pipelines integration

ssbarnea:
Removes Azure Pipelines CI/CD integration as the entire testing matrix
is already already fully covered by GitHub Actions.

Partial: #823


ssbarnea:
> I don't like getting rid of useful features not present elsewhere like, well-visualized test run/cov reports.

I think that the added complexity of having azure-pipelines enabled heights more than the perceived benefits of nice reporting. We should remember that these gimmicks can be implemented on the other two, zuul or github actions. I know for sure that passing coverage results between jobs and combining them and publishing  is possible with zuul (just that nobody bothered to implement them).

We can think from another perspective: how many bugs did the consolidated coverage report prevented us from introducing in the last year? The CI/CD role is to prevent bugs, not to generate nice reports.

Lets see what other think about it. IMHO, is better to invest more effort in better maintaining a lower number of CI/CD systems.

There are also two aspects we forgot to consider:
* we could consider dropping github-workflows instead of azure-pipelines if the reports are so important
* costs: how do our decisions impact CI/CD costs and how does this play with Ansible plans regarding CI/CD testing in general (like increasing investing in
one or reducing investing in another one). If we are on free tiers on both, that is less important.

webknjaz:
@ssbarnea suggesting zuul is unreasonable because its UI is the worst and you know it. As for the CI costs, all current CIs are on respective free plans. They are all already well-configured and need next to zero maintenance.

albinvass:
> @ssbarnea suggesting zuul is unreasonable because its UI is the worst and you know it. As for the CI costs, all current CIs are on respective free plans. They are all already well-configured and need next to zero maintenance.

Can we please try to have a serious discussion about issues?
Saying things like '...its UI is the worst and you know it.' is not constructive in any way and does not add to the conversation.

Can you please try to describe what problems you are experiencing instead, and how other systems are more user friendly?

webknjaz:
Sorry, I don't have the energy for this right now.

awcrosby:
since the tests are covered by GitHub Actions, I support reducing the complexity of the CI

webknjaz:
Honestly, I'd also drop travis but that would require some help from @gundalow (because we'd need new secrets).

gundalow:
If this part (dropping AZ pipelines) is good, could we please merge. Then as a follow up we can review Travis?

Let's be mindful not to try and solve all problems at once.

PR #831, opened 2020-06-15
WIP: simplify github workflows

ssbarnea:
- make workflow job name identical with tox environment, so user
  could easily figure out how to reproduce failure locally
- simplify workflow definition

### DIFF
--- a/.github/workflows/tox-tests.yml
+++ b/.github/workflows/tox-tests.yml
@@ -1,4 +1,4 @@
-name: 👷
+name: unit
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/tox-tests.yml
```diff
@@ -12,16 +12,19 @@ on:
...
-        - 3.8
-        - 3.7
-        - 3.6
+        PYTHON:
+          - VERSION: 3.8
+            VERSION_COMPACT: py38
+          - VERSION: 3.7
+            VERSION_COMPACT: py37
+          - VERSION: 3.6
+            VERSION_COMPACT: py36
```
webknjaz, line 27: plz don't use UPPERCASE, these are not ENV_VARS

ssbarnea:
> This would need to include changes to dist metadata and readme + no env-var style non-env-vars plz.

I am still working on this but as low-priority, so it make take a while to polish it. I used as surse of inspiration https://github.com/pyca/cryptography/actions/runs/107452460/workflow

I wonder if we can eliminate the workflow-name from the full name displayed as is kinda bogus. I know we have two workflows but we do not really need two, as we can use a single file. Still, I am not sure if that would remove the extra prefix.

Do you happen to know if there is a way to group the checks in some way in order to avoid having so many "jobs" in the end?

webknjaz:
> I wonder if we can eliminate the workflow-name from the full name displayed as is kinda bogus.

It's not possible. That's why I used a single-char name there.



> I used as surse of inspiration [pyca/cryptography/actions/runs/107452460/workflow](https://github.com/pyca/cryptography/actions/runs/107452460/workflow)

That two-var setup is unreasonable duplication, you can easily generate one from another.

> I know we have two workflows but we do not really need two, as we can use a single file. Still, I am not sure if that would remove the extra prefix.

No, the prefix will remain. Also, there's a reason behind having a separate workflow: if you have a lot of jobs, it's easier to consume them when linters are filtered out as a separate category + they don't influence test runs (on low-resources CIs like travis it may be important to cancel tests if linters don't pass but with GHA there's enough resources to not care about this).

> Do you happen to know if there is a way to group the checks in some way in order to avoid having so many "jobs" in the end?

No. They will all show up. So it's easier to keep them in separate workflows. As for the PR summary block, I use Refined GitHub which collapses them if they all pass.

PR #833, opened 2020-06-16
Remove dependency on ansible.utils.color

ssbarnea:
Removes dependency on internal Ansible coloring implementation. This
was a problem for multiple reasons:
- use of private Ansible APIs (color already broke linter once)
- licensing: Ansible is GPL, linter is MIT
- premature initialization of Ansible before our cli loads

The same colors are kept but now we set them using styles, making easier to restyle it if needed and to assure consistency.

### DIFF
--- a/lib/ansiblelint/color.py
+++ b/lib/ansiblelint/color.py
@@ -1,0 +1,18 @@
+"""Console coloring support."""
+from enum import Enum
+
+
+class Color(Enum):
+    """Color styles."""
+
+    reset = "0"
+    error_code = "1;31"  # bright red
+    error_title = "0;31"  # red
+    filename = "0;34"  # blue
+    linenumber = "0;36"  # cyan
+    line = "0;35"  # purple
+
+
+def colorize(text: str, color: Color):
+    """Return ANSI formated string."""
+    return f"\u001b[{color.value}m{text}\u001b[{Color.reset.value}m"
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -1,11 +1,7 @@
 """Output formatters."""
 import os
 from pathlib import Path
-
-try:
-    from ansible import color
-except ImportError:
-    from ansible.utils import color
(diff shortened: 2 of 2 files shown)


lib/ansiblelint/color.py
```diff
@@ -0,0 +1,18 @@
+"""Console coloring support."""
+from enum import Enum
+
+
+class C(Enum):
```
webknjaz, line 5: `C` is a bad name. It doesn't mean anything in code. Is it a constant? A config value? Some magic name?

ssbarnea: Is it constant-ish and I wanted a short name to ease its re-use in various places. Which name would you like? I do not have strong feelings towards "C".

webknjaz: Just `Color`. Because ansible/ansible uses C for configs, for example. Transferring the concept here would be confusing.

lib/ansiblelint/color.py
```diff
@@ -0,0 +1,18 @@
...
+    """Color styles."""
+
+    reset = "0"
+    error_code = "1;31"  # bright red
```
webknjaz, line 9: This is problematic because it'll only work in terminals that support misusing the bold setting for bright. Standard-compliant terminals don't work like that.

ssbarnea: My change does not touch current behavior. ANSI support is a can of worms due to the variance in implementations. I am ok to investigate drop of bold but in a different change, also after looking at what other linters are doing with output (is likely that ansible-lint would be used in parallel with others too).

lib/ansiblelint/color.py
```diff
@@ -0,0 +1,18 @@
...
+    error_title = "0;31"  # red
+    filename = "0;34"  # blue
+    linenumber = "0;36"  # cyan
+    line = "0;35"  # purple
+
+
+def stringc(text: str, color: Color):
```
webknjaz, line 16: Since it's a function, it should be a verb:
```suggestion
def colorize(text: str, color: Color):
```

PR #836, opened 2020-06-16
Test against newer Pythons (3.9) in GHA

webknjaz:


.github/workflows/tox-tests.yml
```diff
@@ -42,10 +52,24 @@ jobs:
       run: |
         git fetch --prune --unshallow
         git fetch --depth=1 origin +refs/tags/*:refs/tags/*
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
+    - name: Set up stock Python ${{ matrix.python-version }} from GitHub
+      if: >-
+        !endsWith(matrix.python-version, '-dev')
```
albinvass, line 57: A bit nitty but this doesn't look like it needs to be multi-line. But won't stop this PR because of it.

webknjaz: I often prefer multiline because shorter lines read better. But in this case, it's because it allows not to wrap the value with extra double quotes (because `!` is special in YAML and would break it when unwrapped).

PR #840, opened 2020-06-18
Migrate publishing to PyPI from Travis CI to GHA

webknjaz:


webknjaz:
(decided that dropping travis can be done separately from moving the publish procedure)

.github/workflows/tox-tests.yml
```diff
@@ -101,3 +102,62 @@ jobs:
       with:
         name: logs.zip
         path: .tox/ansible${{ matrix.ansible-version }}/log/
+
+  publish-dists:
+    name: Publish 🐍📦 to (Test)PyPI
```
ssbarnea, line 107: Please avoid emoji an use plain english, please.

webknjaz: No, I want the text to fit in GH UI.

webknjaz:
> * assure that it does not run if linters are failing (you used condition only on test-matrix, which does not include linters).

This is out of scope, plus I don't think that linters should block the release. That's why they are in a separate job.

ssbarnea:
When I say linters, I mean `.github/workflows/tox-linters.yml` file which includes execution of all jobs that are not unit-tests, including: lint,build-dists,metadata-validation. Clearly that we want all to pass.

Regarding emoji, lets not impose our personal preferences and see what other have to say about it. I am aware of lots of projects that forbid use of emoji in commit messages or code (openstack is one of them).

.github/workflows/tox-tests.yml
ssbarnea: Well `Publish to (test.)pypi.org" does fit an is shorter, no need to mention what. Just removing tee unicode-emoji part would keep the text clear.

webknjaz:
> Regarding emoji, lets not impose our personal preferences and see what other have to say about it.

Exactly, it's not worth blocking a good change. You didn't suggest a better solution anyway and this one already conforms my requirements to that name.
It's funny that you request your preference to be fulfilled while saying that we shouldn't impose our personal preferences. Either way, it follows the existing style and as such, is out of the scope of this PR. You know I hate unrelated changes in PRs so I'm not going to debate it here anymore.

> Clearly that we want all to pass.

They will still show up on PRs and you're not going to be tagging PR branches, right? What if something changes in the env and a perfectly valid and green, at the moment of merging, commit publishing will ruin the publish just because some linter changed or something silly broke? I've put them separately for a reason and I exercise this set up in many places and it works fine. The only reason for linters to be run early in Travis is to fail early, preventing the tests from running, not publishing. That's because it has way more limited resources that we wanted to consume moderately.

webknjaz:
> including: lint,build-dists,metadata-validation. Clearly that we want all to pass.

FTR build-dists and metadata-validation are in the deploy job already. A and `lint` shouldn't be a blocker.

.github/workflows/tox-tests.yml
gundalow:
```suggestion
    name: Publish to (Test)PyPI
```

webknjaz: I guess so. Also, not PyPi but PyPI.
My point is that this is out of scope anyway because it follows the existing style. It's okay to debate style change in a separate PR and @ssbarnea already has one for that. I don't want to mix different kinds of changes in the middle of random unrelated PRs.
I currently follow the style that exists here and in other projects where I maintain GHA. Blocking a PR just because it follows the rest of the style is nonsense.

gundalow: I don't like doing this, though I'm just going to pull rank here.
Let's get the functional changes in so we can drop Travis. Then we can address this in a follow up PR.

@webknjaz thanks for the clarification that it's `PyPI`, I wasnt aware of that.

gundalow:
I don't like doing this, though I'm just going to pull rank here.
Let's get the functional changes in so we can drop Travis. Then we can address emoji in a follow up PR. Thanks for your understanding.

webknjaz:
Thanks @gundalow, adding an approval should do. Whoever does this, please feel free to merge it yourself since I'm mostly afk this weekend + PTO on Mon.

gundalow:
I can't review this from a technical level, so I'll leave that in the good hands of @awcrosby @ssbarnea and @albinvass

PR #842, opened 2020-06-19
Add support for ansible 2.10

ssbarnea:
- switch to use official pre-release from pypi
- fix bug where newer ModuleArgsParser no longer accepts our custom
  attributes when loading tasks
- address parsing failure with now removed "always_run"
- avoid duplicate match errors at runtime

Fixes: #843

### DIFF
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -19,19 +19,35 @@ class Match(object):
                                 self.filename, self.linenumber, self.line)
 
 
+class AnsibleLintInternalRule():
+    """Internal error likely caused by parsing failures."""
+
+    id = "999"
+    shortdesc = "Parsing Error"
+    description = "Ansible failed to file."
+
+
 class MatchError(ValueError):
     """Exception that would end-up as linter rule match."""
 
-    def __init__(self, message, rule=None):
+    def __init__(self, message, rule=AnsibleLintInternalRule, filename="", linenumber=0, line=None):
         """Initialize a MatchError instance."""
         super().__init__(message)
 
         self.message = message
-        self.linenumber = 0
-        self.line = None
-        self.filename = None
+        self.linenumber = linenumber
+        self.line = line
+        self.filename = filename
         self.rule = rule
 
     def get_match(self) -> Match:
         """Return a Match instance."""
         return Match(self.linenumber, self.line, self.filename, self.rule, self.message)
+
+    def __eq__(self, other):
+        """Identify duplicate matches."""
+        return self.__hash__() == other.__hash__()
+
+    def __hash__(self):
+        """Perform hash of matches."""
+        return hash((self.message, self.rule, self.filename, self.linenumber))
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -365,18 +365,27 @@ def _kv_to_dict(v):
 def normalize_task_v2(task):
     """Ensure tasks have an action key and strings are converted to python objects."""
     result = dict()
-    mod_arg_parser = ModuleArgsParser(task)
+    clean_task = task.copy()
+    # task is an AnsibleMapping which inherits from OrderedDict, so we need
+    # to use `del` to remove unwanted keys.
+    for k in ['skipped_rules', FILENAME_KEY, LINE_NUMBER_KEY]:
+        if k in clean_task:
+            del clean_task[k]
+    if 'always_run' in task:
+        raise MatchError(
+            "Ansible 2.10 and later does not support `always_run`.",
+            filename=task[FILENAME_KEY],
+            linenumber=task[LINE_NUMBER_KEY])
+    mod_arg_parser = ModuleArgsParser(clean_task)
     try:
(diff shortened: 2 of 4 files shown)


webknjaz:
LGTM. Side note: it looks like we forgot to clean up some devel related parts in testenv definition but that's a topic for a separate PR, of course.

P.S. GitHub seems to be drunk today so if GHA get canceled, it's probably because they're firefighting so we need to restart jobs...

webknjaz:
OTOH these failures look legit: https://github.com/ansible/ansible-lint/pull/842/checks?check_run_id=788289763

lib/ansiblelint/utils.py
```diff
@@ -365,18 +366,27 @@ def _kv_to_dict(v):
...
-    mod_arg_parser = ModuleArgsParser(task)
+    clean_task = task.copy()
+    # task is an AnsibleMapping which inherits from OrderedDict, so we need
+    # to use `del` to remove unwanted keys.
+    for k in ['skipped_rules', FILENAME_KEY, LINE_NUMBER_KEY]:
+        if k in clean_task:
+            del clean_task[k]
+    if 'always_run' in task:
+        raise MatchError(
+            "Ansible 2.10 would raise an error if always_run is present.",
```
albinvass, line 377: Ansible 2.10 and later does not support 'always_run'

lib/ansiblelint/errors.py
```diff
@@ -19,19 +19,35 @@ def __repr__(self):
                                 self.filename, self.linenumber, self.line)
 
 
+class AnsibleLintInternalRule():
+    """Internal error likely cause by parsing failures."""
+
+    id = "999"
+    shortdesc = "Parsing Error"
+    description = "Ansible failed to parse that file."
```
albinvass, line 27: 'Failed to parse file.' is good enough.
Anyway it's not ansible that fails to parse the fail it's ansible-lint that errors.

lib/ansiblelint/utils.py
```diff
@@ -241,7 +241,8 @@ def _taskshandlers_children(basedir, k, v, parent_type):
             module = th['action']['__ansible_module__']
             if "name" not in th['action']:
                 raise MatchError(
-                    "Failed to find required 'name' key in %s" % module)
+                    "Failed to find required 'name' key in %s" % module,
+                    AnsibleLintInternalRule)
```
albinvass, line 245: no need to explicitly assign the default

lib/ansiblelint/errors.py
```diff
@@ -19,19 +19,35 @@ def __repr__(self):
                                 self.filename, self.linenumber, self.line)
 
 
+class AnsibleLintInternalRule():
+    """Internal error likely caused by parsing failures."""
+
+    id = "999"
+    shortdesc = "Parsing Error"
+    description = "Ansible failed to file."
```
albinvass, line 27: I think you missed a word. 'parse' :)

lib/ansiblelint/errors.py
```diff
@@ -19,19 +19,35 @@ def __repr__(self):
...
+
+
 class MatchError(ValueError):
     """Exception that would end-up as linter rule match."""
 
-    def __init__(self, message, rule=None):
+    def __init__(self, message, rule=AnsibleLintInternalRule, filename="", linenumber=0, line=None):
```
webknjaz, line 33: Please don't use mutable defaults like a class or a string. Use `None` instead.

lib/ansiblelint/errors.py
```diff
@@ -19,19 +19,35 @@ def __repr__(self):
                                 self.filename, self.linenumber, self.line)
 
 
+class AnsibleLintInternalRule():
```
webknjaz, line 22:
```suggestion
class AnsibleLintInternalRule:
```

lib/ansiblelint/errors.py
```diff
@@ -19,19 +19,35 @@ def __repr__(self):
...
+    """Internal error likely caused by parsing failures."""
```
webknjaz, line 23: I think this belongs in the same place as rule 901.

lib/ansiblelint/errors.py
ssbarnea: Current implementation relies on `rule` being an instance of a rule and checks for specific attributes. We would require several changes in order to make it work without a rule.

I think we can keep the code complexity lower by assuring that a match always has a rule, even if is a placeholder one. Also better for type checking.

webknjaz: Still, you should use a None-idiom instead. Mutables in func/meth args is straight bad.

ssbarnea:
Closing in favour of https://github.com/ansible/ansible-lint/pull/878/files

lib/ansiblelint/errors.py
SimonHeimberg: maybe this is an option:
```python
def __init__(self, message, rule=None, filename="", linenumber=0, line=None):
    # ...
    if rule is None:
        rule = AnsibleLintInternalRule
    # ...
```

PR #846, opened 2020-06-21
Refactor MetaMainHasInfoRule

ssato:
This change may reduce the McCabe complexity of lib/ansiblelint/rules/MetaMainHasInfoRule.py and fixes the part of the issue #744.

 - export the parts of the matchplay method into a few other functions
 - define some constantants give meta data name outside of the class,
   MetaMainHasInfoRule
- replace repeated {'meta/main.yml': data} with a var, meta
- replace some imperative style list modification with generators

### DIFF
--- a/lib/ansiblelint/rules/MetaMainHasInfoRule.py
+++ b/lib/ansiblelint/rules/MetaMainHasInfoRule.py
@@ -34,12 +34,12 @@ def _galaxy_info_errors_itr(galaxy_info,
         ginfo = galaxy_info.get(info, False)
         if ginfo:
             if info in str_info_list and not isinstance(ginfo, str):
-                yield '{} should be a string'.format(info)
+                yield '{info} should be a string'.format(info=info)
             elif info == 'platforms':
                 for err in _platform_info_errors_itr(ginfo):
                     yield err
         else:
-            yield 'Role info should contain {}'.format(info)
+            yield 'Role info should contain {info}'.format(info=info)
 
 
 class MetaMainHasInfoRule(AnsibleLintRule):
--- a/lib/ansiblelint/rules/MetaMainHasInfoRule.py
+++ b/lib/ansiblelint/rules/MetaMainHasInfoRule.py
@@ -4,15 +4,15 @@
 from ansiblelint.rules import AnsibleLintRule
 
 
-META_STR_INFO = [
+META_STR_INFO = (
     'author',
-    'description',
-]
-META_INFO = META_STR_INFO + [
+    'description'
+)
+META_INFO = tuple(list(META_STR_INFO) + [
     'license',
     'min_ansible_version',
     'platforms',
-]
+])
 
 
 def _platform_info_errors_itr(platforms):
--- a/lib/ansiblelint/rules/MetaMainHasInfoRule.py
+++ b/lib/ansiblelint/rules/MetaMainHasInfoRule.py
@@ -4,16 +4,49 @@
 from ansiblelint.rules import AnsibleLintRule
 
 
+META_STR_INFO = [
+    'author',
+    'description',
+]
+META_INFO = META_STR_INFO + [
+    'license',
+    'min_ansible_version',
+    'platforms',
+]
+
+
+def _platform_info_errors_itr(platforms):
+    if not isinstance(platforms, list):
+        yield 'Platforms should be a list of dictionaries'
+        return
+
+    for platform in platforms:
+        if not isinstance(platform, dict):
+            yield 'Platforms should be a list of dictionaries'
+        elif 'name' not in platform:
+            yield 'Platform should contain name'
+
+
+def _galaxy_info_errors_itr(galaxy_info,
+                            info_list=META_INFO,
+                            str_info_list=META_STR_INFO):
+    for info in info_list:
+        ginfo = galaxy_info.get(info, False)
+        if ginfo:
+            if info in str_info_list and not isinstance(ginfo, str):
+                yield '{} should be a string'.format(info)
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/rules/MetaMainHasInfoRule.py
```diff
@@ -4,16 +4,49 @@
...
+        elif 'name' not in platform:
+            yield 'Platform should contain name'
+
+
+def _galaxy_info_errors_itr(galaxy_info,
+                            info_list=META_INFO,
+                            str_info_list=META_STR_INFO):
+    for info in info_list:
+        ginfo = galaxy_info.get(info, False)
+        if ginfo:
```
webknjaz, line 35: you can easily dedent the whole block of code by just investing this check to be a guard expression:
```suggestion
        if not ginfo:
            yield 'Role info should contain 
            continue
```

lib/ansiblelint/rules/MetaMainHasInfoRule.py
```diff
@@ -4,16 +4,49 @@
...
+    for platform in platforms:
+        if not isinstance(platform, dict):
+            yield 'Platforms should be a list of dictionaries'
+        elif 'name' not in platform:
+            yield 'Platform should contain name'
+
+
+def _galaxy_info_errors_itr(galaxy_info,
+                            info_list=META_INFO,
+                            str_info_list=META_STR_INFO):
```
webknjaz, line 32: Having mutable defaults is a dangerous thing. You should either use a None-idiom or turn them into tuples.

lib/ansiblelint/rules/MetaMainHasInfoRule.py
```diff
@@ -25,40 +58,10 @@ def matchplay(self, file, data):
...
-                results.append(({'meta/main.yml': data},
-                                'Platform should contain name'))
+        meta = {'meta/main.yml': data}
+        galaxy_info = data.get('galaxy_info', False)
+        if galaxy_info:
+            return [(meta, err) for err
+                    in _galaxy_info_errors_itr(galaxy_info)]
 
-        return results
+        return [(meta, "No 'galaxy_info' found")]
```
webknjaz, line 67: How about
```suggestion
        errors = _galaxy_info_errors_itr(galaxy_info) if galaxy_info else ("No 'galaxy_info' found", )
        return [(meta, err) for err in errors]
```

lib/ansiblelint/rules/MetaMainHasInfoRule.py
```diff
@@ -4,16 +4,49 @@
...
+
+
+def _galaxy_info_errors_itr(galaxy_info,
+                            info_list=META_INFO,
+                            str_info_list=META_STR_INFO):
+    for info in info_list:
+        ginfo = galaxy_info.get(info, False)
+        if ginfo:
+            if info in str_info_list and not isinstance(ginfo, str):
+                yield '{} should be a string'.format(info)
```
webknjaz, line 37: explicit is more readable
```suggestion
                yield '{info} should be a string'.format(info=info)
```

webknjaz:
@ssato if this eliminates the complexity, please also remove the ignore line from `.flake8` file as a part of this PR.

lib/ansiblelint/rules/MetaMainHasInfoRule.py
ssato: Thanks for reviewing and your comment!

> you can easily dedent the whole block of code by just investing this check to be a guard expression:

I cannot remember its exact reason but maybe I prefer to avoid explicit call of 'continue' and the negating conditional statement.

Anyway, I don't have strong opinions about this and if it's one of requirements for merge, I'll change that.

lib/ansiblelint/rules/MetaMainHasInfoRule.py
ssato: Absolutely right. Thanks for your advice!

I usually use a None-idiom you indicated but I feel it's a bit too much so I committed and pushed to make them into tuples.

lib/ansiblelint/rules/MetaMainHasInfoRule.py
ssato: Absolutely right. Thanks for your advice!

I pushed the fix for this.

ssato:
> @ssato if this eliminates the complexity, please also remove the ignore line from `.flake8` file as a part of this PR.

Thanks for your suggestion! I also pushed this change.

lib/ansiblelint/rules/MetaMainHasInfoRule.py
webknjaz: I usually prefer guard-expressions because they simplify the code improving the readability. But I guess we can keep this for now.

webknjaz:
@ssato I noticed that your commits are not connected to your GitHub account. I think you have misconfigured Git on your computer. You can fix this by either adding the email in committer field to your GH account or you could fix the local setup via `git config --global user.email YOUR@ACTUAL.EMAIL` and then updating your commits with `--reset-author`.

Do you want to do this before I merge the PR?

ssato:
> @ssato I noticed that your commits are not connected to your GitHub account. I think you have misconfigured Git on your computer. You can fix this by either adding the email in committer field to your GH account or you could fix the local setup via `git config --global user.email YOUR@ACTUAL.EMAIL` and then updating your commits with `--reset-author`.
> 
> Do you want to do this before I merge the PR?

No,I don't because the mail address shown in my github account profile which contains +github is just an alias. I'm OK with my real mail address in the commits. Excuse me for the confusion.

PR #847, opened 2020-06-22
Skip running always_run rule with ansible 2.10

ssbarnea:
On Ansible 2.10, presence of `always_run` generates a parsing error so we need to skip running that
particular test.

### DIFF
--- a/test/TestAlwaysRunRule.py
+++ b/test/TestAlwaysRunRule.py
@@ -2,8 +2,15 @@ import unittest
 from ansiblelint.rules import RulesCollection
 from ansiblelint.runner import Runner
 from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule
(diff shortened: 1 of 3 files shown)


test/TestAlwaysRunRule.py
```diff
@@ -2,8 +2,15 @@
...
 from ansiblelint.runner import Runner
 from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule
+from test import ANSIBLE_MAJOR_VERSION
+import pytest
 
 
+@pytest.mark.skipif(
+    ANSIBLE_MAJOR_VERSION > (2, 9),
+    reason='Ansible 2.10 removed always_run attribute.',
+    raises=SystemExit, strict=True,
```
webknjaz, line 12: @ssbarnea did you confuse this with xfail?

Issue #544, opened 2019-07-08
ansible-lint erroneously detects pipes in shell when that is template


ignatenkobrain:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
⋊> ~/P/g/ansible on master ◦ ansible --version                                                                                                                              13:30:37
ansible 2.8.1
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/brain/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.7.4rc2 (default, Jul  3 2019, 01:30:40) [GCC 9.1.1 20190605 (Red Hat 9.1.1-2)]
⋊> ~/P/g/ansible on master ◦ ansible-lint --version                                                                                                                         13:30:54
ansible-lint 4.1.0
```

- ansible installation method: OS package
- ansible-lint installation method: OS package

# Desired Behaviour

ansible-lint should not find this false positive…

# Actual Behaviour (Bug report only)

```yaml
---
- name: Test
  hosts: localhost
  gather_facts: false
  tasks:
  - name: Test
    shell: echo
      {% for i in range(nameservers | length) %}
      {{ nameservers[i] }}
      {% endfor %}
    changed_when: false
  vars:
    nameservers:
    - 8.8.8.8
    - 4.4.4.4
```

And ansible-lint returns:

```
[306] Shells that use pipes should set the pipefail option
test.yml:6
Task/Handler: Test
```

I did not check code, but I think it just checks whether pipe is present in shell string, but shell is actually jinja2 template, so in this case it is false-positive.

willthames:
That code should fail a "no `{%`" in playbooks check!

You're quite correct though, it's a simple regex. Checking templates is hard, not sure what the easy fix to this one is.

Checking whether the pipe is inside templating might be the simplest - it wouldn't fix your scenario but it would fix `echo {{ nameservers | length }}`.

ignatenkobrain:
Hello @willthames,

> That code should fail a "no {%" in playbooks check!

Can you give me link to this definition of a check please? I'm curious what is the better way to construct commandline out of the template.

willthames:
@ignatenkobrain there is no such check, I just think there ought to be!

[...]

PR #853, opened 2020-06-23
Improve unjinja function

ssbarnea:
Corrects unjinja implementation in order to assure it also removes
statements and comments, not only expressions.

Fixes: #544

### DIFF
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -36,7 +36,10 @@ class AnsibleLintRule(object):
 
     @staticmethod
     def unjinja(text):
-        return re.sub(r"{{[^}]*}}", "JINJA_VAR", text)
+        text = re.sub(r"{{.+?}}", "JINJA_EXPRESSION", text)
+        text = re.sub(r"{%.+?%}", "JINJA_STATEMENT", text)
+        text = re.sub(r"{#.+?#}", "JINJA_COMMENT", text)
+        return text
 
     def matchlines(self, file, text) -> List[Match]:
         matches: List[Match] = []
--- a/test/TestAnsibleLintRule.py
+++ b/test/TestAnsibleLintRule.py
@@ -1,0 +1,7 @@
+from ansiblelint.rules import AnsibleLintRule
+
+
+def test_unjinja():
+    input = "{{ a }} {% b %} {# try to confuse parsing inside a comment { {{}} } #}"
+    output = "JINJA_EXPRESSION JINJA_STATEMENT JINJA_COMMENT"
+    assert AnsibleLintRule.unjinja(input) == output


lib/ansiblelint/rules/__init__.py
```diff
@@ -36,7 +36,10 @@ def verbose(self) -> str:
 
     @staticmethod
     def unjinja(text):
-        return re.sub(r"{{[^}]*}}", "JINJA_VAR", text)
+        text = re.sub(r"{{[^}]*}}", "JINJA_EXPRESSION", text)
```
ssbarnea, line 39: replacement text is not used anywhere, so we can use variations of it.

albinvass, line 39: Should this not replace jinja looking like "{{ test_var | combine({'key': 'value'}) }}" with "JINJA_EXPRESSION" for some reason?

Because the current regex won't do that.

test/TestRuleClass.py
```diff
@@ -1,3 +1,6 @@
+from ansiblelint.rules import AnsibleLintRule
```
webknjaz, line 1: Why did you rename it?

lib/ansiblelint/utils.py
```diff
@@ -225,7 +225,7 @@ def _include_children(basedir, k, v, parent_type):
     return [{'path': result, 'type': parent_type}]
 
 
-def _taskshandlers_children(basedir, k, v, parent_type):
+def _taskshandlers_children(basedir, k, v, parent_type):  # noqa: C901
```
webknjaz, line 228: This and similar changes don't belong here. Use a separate PR for moving ignores from the flake8 config to files.

lib/ansiblelint/rules/__init__.py
```diff
@@ -36,7 +36,10 @@ def verbose(self) -> str:
 
     @staticmethod
     def unjinja(text):
-        return re.sub(r"{{[^}]*}}", "JINJA_VAR", text)
+        text = re.sub(r"{{.+?}}", "JINJA_EXPRESSION", text)
```
webknjaz, line 39: Is this the only place that is relevant to the PR?

test/TestRuleClass.py
ssbarnea: File contains tests related to base rule class, previous name was referring only to attributes.

lib/ansiblelint/rules/__init__.py
ssbarnea: Yes.

test/TestAnsibleLintRule.py
```diff
@@ -0,0 +1,7 @@
+from ansiblelint.rules import AnsibleLintRule
+
+
+def test_unjinja():
```
webknjaz, line 4: I'd prefer this to have a docstring but low priority

webknjaz: Also, ideally, there could be multiple cases to document more behaviors as separate tests.

PR #855, opened 2020-06-23
Fix README badges

ssbarnea:
- Correct GitHub Action badges to match the new workflow name
- Shorten CoC badge name to lower changes of wrapping the badge line

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,7 +6,7 @@
    :target: https://docs.ansible.com/ansible-lint/rules/default_rules.html
    :alt: Ansible-lint rules explanation
 
-.. image:: https://img.shields.io/badge/Code%20of%20Conduct-Ansible-silver.svg
+.. image:: https://img.shields.io/badge/Code%20of%20Conduct-black.svg
    :target: https://docs.ansible.com/ansible/latest/community/code_of_conduct.html
    :alt: Ansible Code of Conduct
 
@@ -14,13 +14,9 @@
    :target: https://docs.ansible.com/ansible/latest/community/communication.html#mailing-list-information
    :alt: Ansible mailing lists
 
-.. image:: https://github.com/ansible/ansible-lint/workflows/%F0%9F%91%B7/badge.svg
-   :target: https://github.com/ansible/ansible-lint/actions?query=workflow%3A%F0%9F%91%B7
-   :alt: 👷 GitHub Actions CI/CD build status — tests
-
-.. image:: https://github.com/ansible/ansible-lint/workflows/%F0%9F%9A%A8/badge.svg
-   :target: https://github.com/ansible/ansible-lint/actions?query=workflow%3A%F0%9F%9A%A8
-   :alt: 🚨 GitHub Actions CI/CD build status — linters
+.. image:: https://github.com/ansible/ansible-lint/workflows/gh/badge.svg
+   :target: https://github.com/ansible/ansible-lint/actions?query=workflow%3Agh+branch%3Amaster+event%3Apush
+   :alt: GitHub Actions CI/CD
 
 .. image:: https://img.shields.io/lgtm/grade/python/g/ansible/ansible-lint.svg?logo=lgtm&logoWidth=18
    :target: https://lgtm.com/projects/g/ansible/ansible-lint/context:python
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -6,21 +6,17 @@
    :target: https://docs.ansible.com/ansible-lint/rules/default_rules.html
    :alt: Ansible-lint rules explanation
 
-.. image:: https://img.shields.io/badge/Code%20of%20Conduct-Ansible-silver.svg
+.. image:: https://img.shields.io/badge/Code%20of%20Conduct-silver.svg
```
webknjaz, line 9: I don't think this is correct. It breaks the badge and also I wanted to indicate that it's Ansible's CoC, not a separate one. FWIW this is up to @gundalow to decide.

README.rst
```diff
@@ -6,21 +6,17 @@
...
 
-.. image:: https://github.com/ansible/ansible-lint/workflows/%F0%9F%91%B7/badge.svg
-   :target: https://github.com/ansible/ansible-lint/actions?query=workflow%3A%F0%9F%91%B7
-   :alt: 👷 GitHub Actions CI/CD build status — tests
-
-.. image:: https://github.com/ansible/ansible-lint/workflows/%F0%9F%9A%A8/badge.svg
-   :target: https://github.com/ansible/ansible-lint/actions?query=workflow%3A%F0%9F%9A%A8
-   :alt: 🚨 GitHub Actions CI/CD build status — linters
+.. image:: https://github.com/ansible/ansible-lint/workflows/gh/badge.svg
+   :target: https://github.com/ansible/ansible-lint/actions?query=workflow%3Agh
```
webknjaz, line 18: How about adding a branch filter (`&branch=master`) and maybe event (`push`)? I think this would filter out some builds that may be confusing to folks who are unfamiliar with the UI.

webknjaz, line 18: Oh, and I'd also do this in `setup.cfg`. WDYT?

webknjaz:
> * Shorten CoC badge name to lower changes of wrapping the badge line

Instead, I'd prefer to rearrange the order of badges so that ones you want in one line would fit. There's still two lines occupied by the badges (at least it's rendered like that for me).

README.rst
ssbarnea: This works fine and is already used by other projects, see https://pypi.org/project/molecule/ -- I think that I asked @gundalow at that time but we can definitely wait for his input.

In the end, what matters is to have a link towards the CoC document, where it is listed and how it is formatted is more of an implementation detail.

README.rst
ssbarnea: Really good point, both. It is likely that a visitor may want to see CI results from push actions made to master branch only.

README.rst
gundalow: I'm OK with this

webknjaz: Then maybe consider changing the color scheme. White text on silver is barely visible. Before, it was just one word and now it's the important phrase that becomes "invisible".

ssbarnea: Well, is the same color it used to be for Ansible. Still, I made it black to avoid lack of contrast.

setup.cfg
```diff
@@ -9,7 +9,7 @@ name = ansible-lint
 url = https://github.com/ansible/ansible-lint
 project_urls =
   Bug Tracker = https://github.com/ansible/ansible-lint/issues
-  CI: GitHub = https://github.com/ansible/ansible-lint/actions
+  CI: GitHub = https://github.com/ansible/ansible-lint/actions?query=workflow%%3Agh+branch%%3Amaster+event%%3Apush
```
webknjaz, line 12: @ssbarnea I see that you escaped `%` as `%%` but I think Python 3 doesn't need that. There was some difference of behaviors wrt this in Python 2.

I prefer not to urlencode this at all because it'd look more straightforward and predictable on PyPI.

```suggestion
  CI: GitHub = https://github.com/ansible/ansible-lint/actions?query=workflow:gh+branch:master+event:push
```

Proof that it works well w/o escaping: https://github.com/ansible/pylibssh/blob/6a59f7b/setup.cfg#L11 + https://pypi.org/project/ansible-pylibssh/

PR #857, opened 2020-06-23
Prevent failing on duplicate pushes to TestPyPI @ GHA

webknjaz:
This change addresses the case when there's two workflow runs
for the same commit triggered via different events. Like `push`
and `create`. In this case, publishing to PyPI may fail because
both would generate dists with the same version causing a
conflict on upload.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,8 +1,15 @@
 name: gh
 
 on:
-  create:  # branch or tag created, need to filter on publish
-  push:
+  create:  # is used for publishing to PyPI and TestPyPI
+    tags:  # any tag regardless of its name, no branches
+  push:  # only publishes pushes to the main branch to TestPyPI
+    branches:  # any branch but not tag
+    - >-
+      **
+    tags-ignore:
+    - >-
+      **
   pull_request:
   schedule:
   - cron: 1 0 * * *  # Run daily at 0:01 UTC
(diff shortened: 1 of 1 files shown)


.github/workflows/tox.yml
```diff
@@ -225,6 +225,7 @@ jobs:
       with:
         password: ${{ secrets.testpypi_password }}
         repository_url: https://test.pypi.org/legacy/
+        skip_existing: true
```
ssbarnea, line 228: My impression is that this is only hiding a mistake we made configuring what triggers a publish job. We already test the package building as part of the "linters" group so we should trigger "publish" only on tags and pushes. tags would go pypi and non-tags to testpypi. This logic should avoid needing a skip.

I am inclined to avoid skip as it hides such logic issues, and it may take some time to discover them. What do you think?

ssbarnea:
I had a similar issue today and I think I found a better way to avoid duplicates. Also look at setup.py

* https://github.com/ansible-community/sphinx_ansible_theme/blob/master/.github/workflows/tox.yml#L4-L14
* https://github.com/ansible-community/sphinx_ansible_theme/blob/master/setup.py#L33-L41

.github/workflows/tox.yml
webknjaz: tagged releases should go to both PyPI and TestPyPI

PR #858, opened 2020-06-23
Refactor building docs

webknjaz:
This change is only a first (small) step towards making the docs setup better.
* It introduces an explicitly defined RTD config.
* It gets rid of Makefiles moving everything under one roof with other processes (tox).

Everything else is kept out intentionally: small changes are easier to reason about and accepting them fast allows us to minimize conflicts with related changes.


webknjaz:
@ssbarnea I agree that we need to get the theme out of the repo but, as I see it, it's easier to go it step-by-step with small incremental changes like this one.

ssbarnea:
@webknjaz I am ok with the incremental approach but there is one critical issue: the RTD build is broken and this build also breaks the zuul publishing due to changed path, practically making impossible to review the outcome of the change.

If RTD would render fine, I would merge it right now.

webknjaz:
I think you could change Zuul config to improve this. Right now I don't want to overwhelm it with too many changes.

webknjaz:
@ssbarnea here I've adjusted the var to point to the changed location https://github.com/ansible/ansible-zuul-jobs/pull/576

PR #861, opened 2020-06-26
Add a shared docs build env dependencies location

webknjaz:
* `pip install -r filename` accepts any filenames regardless of whether it's txt or in.
  I'd like to [integrate `pip-tools`](https://hynek.me/articles/python-app-deps-2018/#pip-tools--everything-old-is-new-again) in the future in order to have a locked build env.
  It has a canonical way of naming files and uses `.in` for direct deps and `.txt` for the
  emitted constraints files: https://github.com/jazzband/pip-tools#output-file.
* I'd also like to enable dependabot at some point which supports pip-tools but only for `.in`+`.txt` file
  pairs. It doesn't pick up non-in files. If we already have a file named like this, no extra commits would be
  required.
* We specify this file in our dev setup (tox) and any other tools needed and so nobody will really have to face
  it + it's easy to follow the breadcrumbs from tox.ini to the file.
* Using extras for docs is semantically incorrect because extras are for enabling additional runtime features
  and what I'm declaring here is build *environment* deps, not runtime lib deps, these shouldn't go
  into dist meta.


webknjaz:
Description added.

> * does not include RTD config change to make it use it

There's no RTD config in master, adding it is out of scope and so it doesn't make sense to even mention it.
When it'll be there, we'll update it too.

PR #864, opened 2020-06-27
Expose package version

ssbarnea:
Export ``ansiblelint.__version__`` which is a a standard way to expose version of a python module.


lib/ansiblelint/__init__.py
```diff
@@ -19,5 +19,11 @@
 # THE SOFTWARE.
 """Main ansible-lint package."""
 
-# deprecated backwards compability:
-from ansiblelint.rules import AnsibleLintRule  # noqa: F401 api backwards compability
+from ansiblelint.rules import AnsibleLintRule
+from ansiblelint.version import __version__
```
webknjaz, line 23: honestly, I'm not sure if we want to expose it like this. I understand that it's common but earlier you advocated for removing everything from the package init. We need to have some sort of a decision of what we actually want to do with this.

lib/ansiblelint/__init__.py
```diff
@@ -19,5 +19,11 @@
...
+
+
+__all__ = [
+    "__version__",
+    "AnsibleLintRule"  # deprecated, import it directly from rules
+]
```
webknjaz, line 29: pylint has a rule for enforcing `__all__` to be immutable and I pretty much agree with it.
```suggestion
__all__ = (
    "__version__",
    "AnsibleLintRule",  # deprecated, import it directly from rules
)
```

PR #866, opened 2020-06-29
Add pylint to pre-commit

ssbarnea:
Adds default config with specific ignores required in order to pass with current code.


.pre-commit-config.yaml
```diff
@@ -75,3 +75,7 @@ repos:
   - id: mypy
     # empty args needed in order to match mypy cli behavior
     args: []
+- repo: https://github.com/pre-commit/mirrors-pylint
+  rev: 'v2.5.3'  # Use the sha / tag you want to point at
```
webknjaz, line 79:
```suggestion
  rev: v2.5.3
```

.pylintrc
```diff
@@ -0,0 +1,627 @@
...
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
```
webknjaz, line 25:
```suggestion
jobs=0
```

.pylintrc
```diff
@@ -0,0 +1,628 @@
...
+        dict-keys-not-iterating,
+        dict-values-not-iterating,
+        deprecated-operator-function,
+        deprecated-urllib-function,
+        xreadlines-attribute,
+        deprecated-sys-function,
+        exception-escape,
+        comprehension-escape,
+        # TODO(ssbarnea): remove temporary skips adding during initial adoption:
+        C0103,  # Variable name "r" doesn't conform to snake_case naming style (invalid-name)
```
webknjaz, line 143: Please use name aliases instead of numbers

.pylintrc
```diff
@@ -0,0 +1,628 @@
...
+
+# List of qualified module names which can have objects that can redefine
+# builtins.
+redefining-builtins-modules=six.moves,past.builtins,future.builtins,builtins,io
+
+
+[FORMAT]
+
+# Expected format of line ending, e.g. empty (any line ending), LF or CRLF.
+expected-line-ending-format=
```
webknjaz, line 350:
```suggestion
expected-line-ending-format=LF
```

.pylintrc
```diff
@@ -0,0 +1,628 @@
...
+
+# Number of spaces of indent required inside a hanging or continued line.
+indent-after-paren=4
+
+# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
+# tab).
+indent-string='    '
+
+# Maximum number of characters on a single line.
+max-line-length=100
```
webknjaz, line 363:
```suggestion
# (accessibility/large fonts and PEP8 unfriendly)
max-line-length=100
```

.pylintrc
```diff
@@ -0,0 +1,628 @@
...
+# Regular expression matching correct function names. Overrides function-
+# naming-style.
+#function-rgx=
+
+# Good variable names which should always be accepted, separated by a comma.
+good-names=i,
+           j,
+           k,
+           ex,
+           Run,
```
webknjaz, line 464: This is not good
```suggestion
```

.pylintrc
```diff
@@ -0,0 +1,628 @@
...
+           _
+
+# Good variable names regexes, separated by a comma. If names match any regex,
+# they will always be accepted
+good-names-rgxs=
+
+# Include a hint for the correct naming format with invalid-name.
+include-naming-hint=no
```
webknjaz, line 472:
```suggestion
include-naming-hint=yes
```

.pylintrc
```diff
@@ -0,0 +1,628 @@
...
+# Regular expression matching correct variable names. Overrides variable-
+# naming-style.
+#variable-rgx=
+
+
+[STRING]
+
+# This flag controls whether inconsistent-quotes generates a warning when the
+# character used as a quote delimiter is used inconsistently within a module.
+check-quote-consistency=no
```
webknjaz, line 520:
```suggestion
check-quote-consistency=yes
```

lib/ansiblelint/skip_utils.py
```diff
@@ -26,7 +26,7 @@
 import ruamel.yaml
 from typing import Any, Generator, List, Sequence
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import Literal  # pylint: disable-msg=E0611
```
webknjaz, line 29: I think the proper fix would be try-except style

ssbarnea: @webknjaz This is not possible in this case, is mostly a bug in pylint which fails because it goes inside this branch with an older version of python. I found that online. 

Python itself will not fail with this import, only pylint would complain.

ssbarnea:
> It is recommended to have a full pylint config.

I really do not want to add a boilerplate config. Even ansible itself does not have `.pylintrc` file and I do not want to encourage addition of extra custom preferences for the defaults. If we copy the defaults it will be very hard to know what was customized for internal use.

Instead of having a stand-off on this subject, I would say is better to merge it as is, mainly because it hardens QA and makes harder to introduce bad code. We can always consider adding it if we have proofs that that is needed. My goal is to avoid customization if possible, not to impose new more or less ad-hoc customizations or exceptions. Thus is why I mentioned clearly that these skips where put these during introduction of the tool, and not because we do not want to follow them.

lib/ansiblelint/skip_utils.py
albinvass: Well, I have to agree with @webknjaz here. Looks like the proper fix would probably be to use a try-except, or would pylint complain about that?

PR #868, opened 2020-06-29
Specify `basepython` as `python3` in non-test envs

ssbarnea:


tox.ini
```diff
@@ -54,9 +55,6 @@ whitelist_externals =
   ansibledevel: sh
 
 [testenv:.dev-env]
-#basepython = python3
-basepython = /home/wk/.pyenv/versions/ansible-lint-py3.8.0-pyenv-venv/bin/python3
-#{[testenv]deps}
```
webknjaz, line 59: Let's not touch this for now. I've been meaning to make `--dev-env` support work but it's currently broken anyway.

tox.ini
```diff
@@ -84,15 +84,8 @@ commands =
     --out-dir {toxinidir}/dist/ \
     {toxinidir}
 
-# deprecated: use more generic 'lint' instead
```
webknjaz, line 87: This is not related to enforcing basepython

PR #870, opened 2020-06-30
Flatten docs structure

ssbarnea:
We have less than ten documentation pages and keeping them in a flat layout makes it easier to maintain them.

This change also fixes regression introduced by the replacement on the custom theme.


webknjaz:
@ssbarnea what was the regression? Also, let's remove `rst/` too. All could be right under `docs/`.

PR #872, opened 2020-06-30
Generate initial docs env lock file with pip-tools

webknjaz:


docs/requirements.txt
```diff
@@ -0,0 +1,149 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --generate-hashes --output-file=docs/requirements.txt docs/requirements.in
```
webknjaz, line 5: @ssbarnea lockfile update command

webknjaz:
This will enable dependabot to identify in+txt files pair and should let it upgrade the dependencies properly.

webknjaz:
@ssbarnea that would be for the dependabot config, we'd still pip-compile lock-files that can be generated manually, of course.

webknjaz:
Alright, requirements appeared in the list, for starters: https://github.com/ansible/ansible-lint/network/dependencies.

webknjaz:
It works! https://github.com/ansible/ansible-lint/network/updates (no PR because the lock-file has just been generated but also there's no error like before)

PR #874, opened 2020-06-30
Enable dependabot for docs build env pins

webknjaz:


gundalow:
Docs if others are interested https://help.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates

gundalow:
Seems to be running https://github.com/ansible/ansible-lint/network/updates

webknjaz:
> Seems to be running [ansible/ansible-lint/network/updates](https://github.com/ansible/ansible-lint/network/updates)

Also seems to be broken. The "legacy" dependabot wouldn't error out. It looks like we'll have to add the GitHub App after all. https://github.com/apps/dependabot-preview

ssbarnea:
@webknjaz That is why I propose to use the sphinx_ansible_theme project as a testing ground for any new tools/bots/gadgets. If they prove to work fine for a some time there, we can copy the working setup to ansible-lint, molecule, or whatever more popular project we are maintaining.

webknjaz:
To be fair, they mark it as beta and promise the migration during the nearest months. As of the GitHub Apps version, it's proven to work so there shouldn't be any problems.

webknjaz:
Found the root cause: https://github.com/dependabot/dependabot-core/issues/2272.

webknjaz:
Will be fixed by https://github.com/ansible/ansible-lint/pull/872.

Issue #839, opened 2020-06-17
Fix RTD renderring


webknjaz:
In order to make the rendering the same across different envs, we need to:
1. ~Get rid of Makefiles~ (https://github.com/ansible/ansible-lint/pull/858)
2. ~Get rid of unnecessary docs nestings~ (https://github.com/ansible/ansible-lint/pull/870)
3. ~Add an RTD config~ (https://github.com/ansible/ansible-lint/pull/858)
4. Maybe adjust some options
5. ~Embed external generators of rst into sphinx via extensions~ (https://github.com/ansible/ansible-lint/pull/877)

ssbarnea:
Does this involve duplicating a considerable amount of files from ansible repository? I did POC on adopting the ansible-theme for molecule and I discovered that the theme is quite outdated (by >4y years) in comparison with current rtd themes.

I am not sure what is better/worse: an outdated Ansible theme or a current RTD theme and avoiding having to copy customizations.

Ideally Ansible docs team should have packaged the theme. Still, that idea is more than an year old and without any progress being done.

We need to publish the docs before end of June 2020, somewhere.

PR #877, opened 2020-07-01
Move default rules table generator to Sphinx ext

webknjaz:
Ref: https://github.com/ansible/ansible-lint/issues/839


webknjaz:
@ssbarnea could you please switch Zuul docs job to use Python 3.8?

ssbarnea:
@webknjaz At this moment this is not possible but it will once https://github.com/ansible-network/windmill-config/pull/708 lands.

Still, **it does not make sense to enforce use of py38 for any contribution**. Stuff like linters/docs/packaging but pass with all supported python versions.

BTW, I see that our jobs magically doubled in number again, now I see push and pull_request versions.

PR #878, opened 2020-07-02
Fix broken always_run rule on Ansible 2.10

ssbarnea:
- address parsing failure with now removed "always_run"
- remove skipif for 2.10 as it works correctly now

### DIFF
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -69,7 +69,9 @@ class AnsibleLintRule(object):
             matches.append(m)
         return matches
 
-    def matchtasks(self, file: str, text: str) -> List[MatchError]:
+    # TODO(ssbarnea): Reduce mccabe complexity
+    # https://github.com/ansible/ansible-lint/issues/744
+    def matchtasks(self, file: str, text: str) -> List[MatchError]:  # noqa: C901
         matches: List[MatchError] = []
         if not self.matchtask:
             return matches
@@ -83,7 +85,12 @@ class AnsibleLintRule(object):
 
         yaml = append_skipped_rules(yaml, text, file['type'])
 
-        for task in ansiblelint.utils.get_normalized_tasks(yaml, file):
+        try:
+            tasks = ansiblelint.utils.get_normalized_tasks(yaml, file)
+        except MatchError as e:
+            return [e]
+
+        for task in tasks:
             if self.id in task.get('skipped_rules', ()):
                 continue
 
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -363,7 +363,11 @@ def _kv_to_dict(v):
 
 
 def _sanitize_task(task):
-    """Remove internally used data from task arguments."""
+    """Return a stripped-off task structure compatible with new Ansible.
+
+    This helper takes a copy of the incoming task and drops
+    any internally used keys from it.
+    """
     result = task.copy()
     # task is an AnsibleMapping which inherits from OrderedDict, so we need
     # to use `del` to remove unwanted keys.
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -362,21 +362,41 @@ def _kv_to_dict(v):
     return dict(__ansible_module__=command, __ansible_arguments__=args, **kwargs)
 
 
+def _sanitize_task(task):
+    """Remove internally used data from task arguments."""
+    result = task.copy()
+    # task is an AnsibleMapping which inherits from OrderedDict, so we need
+    # to use `del` to remove unwanted keys.
+    for k in ['skipped_rules', FILENAME_KEY, LINE_NUMBER_KEY]:
+        if k in result:
+            del result[k]
+    return result
+
+
 def normalize_task_v2(task):
     """Ensure tasks have an action key and strings are converted to python objects."""
     result = dict()
-    mod_arg_parser = ModuleArgsParser(task)
+    if 'always_run' in task:
+        # FIXME(ssbarnea): Delayed import to avoid circular import
+        # See https://github.com/ansible/ansible-lint/issues/880
+        from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule  # noqa: # pylint:disable=cyclic-import,import-outside-toplevel
+
+        raise MatchError(
+            rule=AlwaysRunRule,
+            filename=task[FILENAME_KEY],
+            linenumber=task[LINE_NUMBER_KEY])
+
+    sanitized_task = _sanitize_task(task)
+    mod_arg_parser = ModuleArgsParser(sanitized_task)
     try:
         action, arguments, result['delegate_to'] = mod_arg_parser.parse()
     except AnsibleParserError as e:
         try:
             task_info = "%s:%s" % (task[FILENAME_KEY], task[LINE_NUMBER_KEY])
-            del task[FILENAME_KEY]
-            del task[LINE_NUMBER_KEY]
         except KeyError:
             task_info = "Unknown"
         pp = pprint.PrettyPrinter(indent=2)
-        task_pprint = pp.pformat(task)
+        task_pprint = pp.pformat(sanitized_task)
 
         _logger.critical("Couldn't parse task at %s (%s)\n%s", task_info, e.message, task_pprint)
         raise SystemExit(ANSIBLE_FAILURE_RC)
--- a/test/TestAlwaysRunRule.py
+++ b/test/TestAlwaysRunRule.py
@@ -1,19 +1,10 @@
 import unittest
 
-import pytest
-
 from ansiblelint.rules import RulesCollection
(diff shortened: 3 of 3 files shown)


lib/ansiblelint/utils.py
```diff
@@ -365,18 +365,29 @@ def _kv_to_dict(v):
...
     result = dict()
-    mod_arg_parser = ModuleArgsParser(task)
+    clean_task = task.copy()
+    # task is an AnsibleMapping which inherits from OrderedDict, so we need
+    # to use `del` to remove unwanted keys.
+    for k in ['skipped_rules', FILENAME_KEY, LINE_NUMBER_KEY]:
+        if k in clean_task:
+            del clean_task[k]
+    if 'always_run' in task:
+        from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule
```
webknjaz, line 375:
```suggestion
        from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule  # FIXME: https://....issue/..
```

lib/ansiblelint/utils.py
```diff
@@ -365,18 +365,31 @@ def _kv_to_dict(v):
 def normalize_task_v2(task):
     """Ensure tasks have an action key and strings are converted to python objects."""
     result = dict()
-    mod_arg_parser = ModuleArgsParser(task)
+    clean_task = task.copy()
+    # task is an AnsibleMapping which inherits from OrderedDict, so we need
+    # to use `del` to remove unwanted keys.
+    for k in ['skipped_rules', FILENAME_KEY, LINE_NUMBER_KEY]:
+        if k in clean_task:
+            del clean_task[k]
```
webknjaz, line 373: Move this after the if-block. This is doing unnecessary work in cases when we'll just raise an exception.

lib/ansiblelint/utils.py
```diff
@@ -365,18 +365,31 @@ def _kv_to_dict(v):
...
     except AnsibleParserError as e:
         try:
             task_info = "%s:%s" % (task[FILENAME_KEY], task[LINE_NUMBER_KEY])
-            del task[FILENAME_KEY]
-            del task[LINE_NUMBER_KEY]
         except KeyError:
             task_info = "Unknown"
         pp = pprint.PrettyPrinter(indent=2)
-        task_pprint = pp.pformat(task)
+        task_pprint = pp.pformat(clean_task)
```
webknjaz, line 392: Why should it be clean?

tox.ini
```diff
@@ -15,10 +15,7 @@ description =
 deps =
   ansible28: ansible>=2.8,<2.9
   ansible29: ansible>=2.9,<2.10
-  # TODO(ssbarnea): Replace replace the two tarbals created by @abadger1999 once
-  # we official (pre-)releases are available on pypi.org
-  ansibledevel: ansible-base @ https://toshio.fedorapeople.org/ansible/acd/ansible-base/ansible-base-2.10.0.dev1.tar.gz#egg=ansible-base
-  ansibledevel: ansible @ https://toshio.fedorapeople.org/ansible/acd/ansible/ansible-2.10.0.tar.gz#egg=ansible
+  ansibledevel: ansible>=2.10.0a1,<2.11
```
webknjaz, line 18: I feel like this change should be submitted separately. It doesn't look related to the change and it'd be important to see how tests behave with just this swap.

lib/ansiblelint/errors.py
```diff
@@ -22,16 +23,33 @@ def __repr__(self):
 class MatchError(ValueError):
     """Exception that would end-up as linter rule match."""
 
-    def __init__(self, message, rule=None):
+    def __init__(
+            self,
+            message: str = "",
+            rule: Any = None,
```
webknjaz, line 29: How about
```suggestion
            rule: 'AnsibleLintRule' = None,
```
and import it under `if TYPE_CHECKING`.

lib/ansiblelint/errors.py
```diff
@@ -22,16 +23,33 @@ def __repr__(self):
...
+            filename="",
```
webknjaz, line 30: arg defaults mustn't be mutable.

lib/ansiblelint/errors.py
```diff
@@ -22,16 +23,33 @@ def __repr__(self):
...
         """Return a Match instance."""
         return Match(self.linenumber, self.line, self.filename, self.rule, self.message)
+
+    def __eq__(self, other):
+        """Identify duplicate matches."""
+        return self.__hash__() == other.__hash__()
+
+    def __hash__(self):
+        """Perform hash of matches."""
+        return hash((self.message, self.rule, self.filename, self.linenumber))
```
webknjaz, line 55: This looks like a separate feature. Why not submit it separately?

lib/ansiblelint/errors.py
```diff
@@ -1,4 +1,5 @@
 """Exceptions and error representations."""
+from typing import Any
```
webknjaz, line 2:
```suggestion
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ansiblelint.rules import AnsibleLintRule
```

lib/ansiblelint/utils.py
ssbarnea: Ansible 2.10 no longer accepts extra/unknown arguments, like the ones we where using for filename and line number. We need to be sure we remove them when loading the task with ansible. Older versions did allow them.

lib/ansiblelint/utils.py

tox.ini
ssbarnea: We have different PR for this now, one this we should merge before this change.

lib/ansiblelint/rules/__init__.py
```diff
@@ -72,10 +72,7 @@ def matchlines(self, file, text) -> List[MatchError]:
 
     def matchtasks(self, file: str, text: str) -> List[MatchError]:
         matches: List[MatchError] = []
-        if not self.matchtask:
-            return matches
-
-        if file['type'] == 'meta':
+        if not self.matchtask or file['type'] == 'meta':
```
webknjaz, line 75: I don't like this refactoring. Simple is better than complex.

lib/ansiblelint/utils.py
```diff
@@ -364,18 +364,32 @@ def _kv_to_dict(v):
 def normalize_task_v2(task):
     """Ensure tasks have an action key and strings are converted to python objects."""
     result = dict()
-    mod_arg_parser = ModuleArgsParser(task)
+    clean_task = task.copy()
```
webknjaz, line 367: Mind moving the copy after the if-block too? There's no use for it when it enters the raise-branch. So why waste memory?

lib/ansiblelint/rules/__init__.py
ssbarnea: There is a real reason for doing this, otherwise linting would fail as we would exceed mcabe complexity for this function. See https://github.com/ansible/ansible-lint/issues/744

I tried to avoid it and reduce complexity for this function but this brings a long chain of changes, something more intrusive than expected. I plan to still do it and this is why I assigned the bug to me.

I think it would be better to merge this patch first as it fixes a critical issue with 2.10.

webknjaz: Okay. Point is that it should be done separately.

lib/ansiblelint/utils.py
```diff
@@ -361,21 +361,40 @@ def _kv_to_dict(v):
     return dict(__ansible_module__=command, __ansible_arguments__=args, **kwargs)
 
 
+def _sanitize_task(task):
```
webknjaz, line 364: :+1:

lib/ansiblelint/utils.py
```diff
@@ -361,21 +361,40 @@ def _kv_to_dict(v):
...
+    """Remove internally used data from task arguments."""
```
webknjaz, line 365: This docstring is a bit misleading because it implies mutating the object while in fact, a new one is returned. 
```suggestion
    """Return a stripped-off task structure compatible with new Ansible.

    This helper takes a copy of the incoming task and drops
    any internally used keys from it.
    """
```

lib/ansiblelint/utils.py
```diff
@@ -361,21 +361,40 @@ def _kv_to_dict(v):
...
+        # FIXME(ssbarnea): Delayed import to avoid circular import
+        # See https://github.com/ansible/ansible-lint/issues/880
+        from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule  # noqa: # pylint:disable=import-outside-toplevel
+
+        raise MatchError(
+            rule=AlwaysRunRule,
+            filename=task[FILENAME_KEY],
+            linenumber=task[LINE_NUMBER_KEY])
+
+    mod_arg_parser = ModuleArgsParser(_sanitize_task(task))
```
webknjaz, line 388: `_sanitize_task(task)` result should be put in a var and reused to prevent double computation. And also it'd make it more obvious that the same data is used in two places.

lib/ansiblelint/utils.py
webknjaz: @ssbarnea ^

PR #879, opened 2020-07-02
Avoid sorting failure with matches without an id

ssbarnea:
If a match did contain rules with None class, sorting them would have failed. This prevents such an error.

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -115,7 +115,11 @@ def main() -> int:
                         options.verbosity, checked_files)
         matches.extend(runner.run())
 
-    matches.sort(key=lambda x: (normpath(x.filename), x.linenumber, x.rule.id))
+    matches.sort(
+        key=lambda x: (
+            normpath(x.filename),
+            x.linenumber,
+            getattr(x.rule, 'id', 0)))
 
     for match in matches:
         print(formatter.format(match, options.colored))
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/__main__.py
```diff
@@ -115,7 +115,11 @@ def main() -> int:
                         options.verbosity, checked_files)
         matches.extend(runner.run())
 
-    matches.sort(key=lambda x: (normpath(x.filename), x.linenumber, x.rule.id))
+    matches.sort(
+        key=lambda x: (
+            normpath(x.filename),
+            x.linenumber,
+            x.rule.id if hasattr(x.rule, 'id') else 0))
```
webknjaz, line 122:
```suggestion
            getattr(x.rule, 'id', 0),
        ),
    )
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -42,14 +42,15 @@ class Formatter(BaseFormatter):
 
     def format(self, match, colored=False):
         formatstr = u"{0} {1}\n{2}:{3}\n{4}\n"
+        _id = getattr(match.rule, 'id', '000')
```
webknjaz, line 45: Maybe it makes sense to add a property to the rule base or match that would do this instead of us special-casing all the uses?

ssbarnea: I plan to rework this logic and introduce full typing checks, as these bugs would not have happened with them. Still, fixing 2.10 compatibility takes priority, so we should delay other refactorings until we pass tests with 2.10.

PR #881, opened 2020-07-03
Improve MatchError class

ssbarnea:
- allow comparison of MatchError instances
- avoid printing duplicated MatchErrors
- allow creation of MatchErrors without a rule
- allow creation of MatchErrors without a message (only a rule)
- improve type checking on MatchError

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -29,7 +29,7 @@ from typing import Any, Set
 from ansiblelint import cli
 from ansiblelint.constants import DEFAULT_RULESDIR
 from ansiblelint.generate_docs import rules_as_rst
-from ansiblelint.utils import normpath, get_playbooks_and_roles
+from ansiblelint.utils import get_playbooks_and_roles
 import ansiblelint.formatters as formatters
 from ansiblelint.runner import Runner
 from ansiblelint.rules import RulesCollection
@@ -115,13 +115,7 @@ def main() -> int:
                         options.verbosity, checked_files)
         matches.extend(runner.run())
 
-    matches.sort(
-        key=lambda x: (
-            normpath(x.filename),
-            x.linenumber,
-            getattr(x.rule, 'id', 0)))
-
-    for match in matches:
+    for match in sorted(matches):
         print(formatter.format(match, options.colored))
 
     if len(matches):
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -52,4 +52,6 @@ class MatchError(ValueError):
 
     def __eq__(self, other):
         """Identify duplicate matches."""
+        if not isinstance(other, self.__class__):
+            raise NotImplementedError
         return self.__hash__() == other.__hash__()
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -1,6 +1,9 @@
 """Exceptions and error representations."""
+import functools
+from ansiblelint.file_utils import normpath
 
 
+@functools.total_ordering
 class MatchError(ValueError):
     """Rule violation detected during linting.
 
@@ -12,10 +15,14 @@ class MatchError(ValueError):
         """Initialize a MatchError instance."""
         super().__init__(message)
 
+        if not (message or rule):
+            self.linenumber = 0
+            raise RuntimeError("Calling MatchError requires either a message or a rule.")
+
         self.message = message or getattr(rule, 'shortdesc', "")
         self.linenumber = linenumber
         self.line = line
-        self.filename = filename
+        self.filename = normpath(filename) if filename else None
         self.rule = rule
 
     def __repr__(self):
@@ -27,3 +34,19 @@ class MatchError(ValueError):
 
         return formatstr.format(_id, self.message,
                                 self.filename, self.linenumber, self.line)
+
+    @property
+    def _hash_key(self):
+        return self.filename, self.linenumber, self.message, getattr(self.rule, 'id', 0)
+
+    def __lt__(self, other):
+        """Enable sorting of MatchError instances."""
+        return self._hash_key < other._hash_key
+
+    def __hash__(self):
+        """Perform hash of matches."""
+        return hash(self._hash_key)
+
+    def __eq__(self, other):
+        """Identify duplicate matches."""
+        return self.__hash__() == other.__hash__()
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -17,7 +17,10 @@ class MatchError(ValueError):
 
         if not (message or rule):
             self.linenumber = 0
-            raise RuntimeError("Calling MatchError requires either a message or a rule.")
+            raise TypeError(
+                f'{self.__class__.__name__}() missing a '
+                "required argument: one of 'message' or 'rule'",
+            )
 
         self.message = message or getattr(rule, 'shortdesc', "")
         self.linenumber = linenumber
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -47,7 +47,7 @@ class MatchError(ValueError):
         return self._hash_key < other._hash_key
 
     def __hash__(self):
-        """Perform hash of matches."""
+        """Return a hash value of the MatchError instance."""
         return hash(self._hash_key)
 
     def __eq__(self, other):
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -51,7 +51,7 @@ class MatchError(ValueError):
         return hash(self._hash_key)
 
     def __eq__(self, other):
-        """Identify duplicate matches."""
+        """Identify whether the other object represents the same rule match."""
         if not isinstance(other, self.__class__):
             raise NotImplementedError
         return self.__hash__() == other.__hash__()
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -39,7 +39,7 @@ class MatchError(ValueError):
 
     @property
     def _hash_key(self):
-        return self.filename, self.linenumber, self.message, getattr(self.rule, 'id', 0)
+        return self.filename, self.linenumber, str(getattr(self.rule, 'id', 0)), self.message
 
     def __lt__(self, other):
         """Return whether the current object is less than the other."""
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -16,7 +16,6 @@ class MatchError(ValueError):
         super().__init__(message)
 
         if not (message or rule):
-            self.linenumber = 0
             raise TypeError(
                 f'{self.__class__.__name__}() missing a '
                 "required argument: one of 'message' or 'rule'",
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -43,7 +43,7 @@ class MatchError(ValueError):
         return self.filename, self.linenumber, self.message, getattr(self.rule, 'id', 0)
 
     def __lt__(self, other):
-        """Enable sorting of MatchError instances."""
+        """Return whether the current object is less than the other."""
         return self._hash_key < other._hash_key
 
     def __hash__(self):
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -9,9 +9,18 @@ class MatchError(ValueError):
 
     It can be raised as Exception but also just added to the list of found
     rules violations.
+
+    Note that line argument is not considered when building hash of an
+    instance.
     """
 
-    def __init__(self, message=None, linenumber=0, line=None, filename=None, rule=None) -> None:
+    def __init__(
+            self,
+            message=None,
+            linenumber=0,
+            line: dict = None,
+            filename=None,
+            rule=None) -> None:
         """Initialize a MatchError instance."""
         super().__init__(message)
 
@@ -39,6 +48,7 @@ class MatchError(ValueError):
 
     @property
     def _hash_key(self):
+        # line attr is knowingly excluded, as dict is not hashable
         return self.filename, self.linenumber, str(getattr(self.rule, 'id', 0)), self.message
 
     def __lt__(self, other):
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -44,6 +44,8 @@ class MatchError(ValueError):
 
     def __lt__(self, other):
         """Return whether the current object is less than the other."""
+        if not isinstance(other, self.__class__):
+            raise NotImplementedError
         return self._hash_key < other._hash_key
 
     def __hash__(self):
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -221,7 +221,7 @@ class RulesCollection(object):
                     matches.extend(rule.matchlines(playbookfile, text))
                     matches.extend(rule.matchtasks(playbookfile, text))
                     matches.extend(rule.matchyaml(playbookfile, text))
-        return list(set(matches))
+        return matches
 
     def __repr__(self) -> str:
         """Return a RulesCollection instance representation."""
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -221,6 +221,7 @@ class RulesCollection(object):
                     matches.extend(rule.matchlines(playbookfile, text))
                     matches.extend(rule.matchtasks(playbookfile, text))
                     matches.extend(rule.matchyaml(playbookfile, text))
+
         return matches
 
     def __repr__(self) -> str:
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -221,8 +221,7 @@ class RulesCollection(object):
                     matches.extend(rule.matchlines(playbookfile, text))
                     matches.extend(rule.matchtasks(playbookfile, text))
                     matches.extend(rule.matchyaml(playbookfile, text))
-
-        return matches
+        return list(set(matches))
 
     def __repr__(self) -> str:
         """Return a RulesCollection instance representation."""
--- a/test/TestMatchError.py
+++ b/test/TestMatchError.py
@@ -26,5 +26,6 @@ def test_matcherror_sort(a, b):
 
 def test_matcherror_invalid():
     """Ensure that MatchError requires message or rule."""
-    with pytest.raises(RuntimeError):
+    expected_err = '^MatchError\(\) missing a required argument: one of 'message' or 'rule'$"
+    with pytest.raises(TypeError, match=expected_err):
         MatchError()
--- a/test/TestMatchError.py
+++ b/test/TestMatchError.py
@@ -70,14 +70,14 @@ class TestMatchErrorCompare:
 @pytest.mark.parametrize(
     'operation',
     (
-        '__eq__',
-        '__ne__',
-        '__le__',
-        '__gt__',
+        operator.eq,
+        operator.ne,
+        operator.le,
+        operator.gt,
     ),
     ids=['eq', 'ne', 'le', 'gt']
 )
 def test_matcherror_compare_invalid(other, operation):
     """Check that MatchError comparison with other types raises."""
     with pytest.raises(NotImplementedError):
-        getattr(MatchError("foo"), operation)(other)
+        operation(MatchError("foo"), other)
--- a/test/TestMatchError.py
+++ b/test/TestMatchError.py
@@ -6,13 +6,19 @@ from ansiblelint.rules.BecomeUserWithoutBecomeRule import BecomeUserWithoutBecom
 import pytest
 
 
-def test_matcherror_compare():
-"""Check that MatchError instances with similar attrs are equivalent."""
-    assert MatchError("foo") == MatchError("foo")
+@pytest.mark.parametrize(
+    ('a', 'b'), (
+        (MatchError("foo"), MatchError("foo")),
+        # line is ignored in comparisions, even if it may be different
+        (MatchError("a", line={}), MatchError("a", line={"a": "b"}))
+    ))
+def test_matcherror_compare(a, b):
+    """Check that MatchError instances with similar attrs are equivalent."""
+    assert a == b
 
 
 class AnsibleLintRuleWithStringId(AnsibleLintRule):
-    id = "200"
+    id = "ANSIBLE200"
 
 
 def test_matcherror_invalid():
@@ -33,7 +39,7 @@ def test_matcherror_invalid():
         # rule id "200" > rule id 101
         (MatchError(rule=AnsibleLintRuleWithStringId), MatchError(rule=AlwaysRunRule))
     ))
-class TestMatchErrorCompare():
+class TestMatchErrorCompare:
 
     def test_lt(self, a, b):
         assert b < a
@@ -41,5 +47,5 @@ class TestMatchErrorCompare():
     def test_gt(self, a, b):
         assert a > b
 
-    def test_eq(self, a, b):
+    def test_ne(self, a, b):
         assert a != b
--- a/test/TestMatchError.py
+++ b/test/TestMatchError.py
@@ -59,12 +59,12 @@ class TestMatchErrorCompare:
 
 
 @pytest.mark.parametrize(
-    ('other'),
+    'other',
     (
-        (None),
-        ("foo"),
-        (42),
-        (Exception("foo")),
+        None,
+        "foo",
+        42,
+        Exception("foo"),
     ),
 )
 @pytest.mark.parametrize(
--- a/test/TestMatchError.py
+++ b/test/TestMatchError.py
@@ -68,12 +68,12 @@ class TestMatchErrorCompare:
     ),
 )
 @pytest.mark.parametrize(
-    ('operation'),
+    'operation',
     (
-        ('__eq__'),
-        ('__ne__'),
-        ('__le__'),
-        ('__gt__')
+        '__eq__',
+        '__ne__',
+        '__le__',
+        '__gt__',
     ),
     ids=['eq', 'ne', 'le', 'gt']
 )
--- a/test/TestMatchError.py
+++ b/test/TestMatchError.py
@@ -32,7 +32,7 @@ def test_matcherror_invalid():
 
 
 @pytest.mark.parametrize(
-    ('a', 'b'), (
+    ('left_match_error', 'right_match_error'), (
         # sorting by message
         (MatchError("z"), MatchError("a")),
         # filenames takes priority in sorting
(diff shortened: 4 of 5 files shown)


lib/ansiblelint/errors.py
```diff
@@ -22,16 +23,33 @@ def __repr__(self):
 class MatchError(ValueError):
     """Exception that would end-up as linter rule match."""
 
-    def __init__(self, message, rule=None):
+    def __init__(
+            self,
+            message: str = "",
```
webknjaz, line 28: The default value should be immutable.

lib/ansiblelint/errors.py
```diff
@@ -22,16 +23,33 @@ def __repr__(self):
...
+            rule: Any = None,
+            filename="",
```
webknjaz, line 30: The default value should be immutable.

pabelanger:
recheck

lib/ansiblelint/errors.py
ssbarnea: strings are imutable

lib/ansiblelint/errors.py
```diff
@@ -18,20 +22,52 @@ def __repr__(self):
         return formatstr.format(self.rule.id, self.message,
                                 self.filename, self.linenumber, self.line)
 
+    def __lt__(self, other):
+        """Enable sorting of MatchError instances."""
+        return (
+                self.filename, self.linenumber, self.message, getattr(self.rule, "id", 0)
+            ) < (
+            other.filename, other.linenumber, other.message, getattr(other.rule, "id", 0))
```
webknjaz, line 30: I'd put this into a property like `_unique_key` (the actual name could be better), then it'd be way easier to reason about:
```suggestion
        return self._unique_key < other._unique_key

    @property
    def _unique_key(self):
        return (
                self.filename, self.linenumber, self.message, getattr(self.rule, "id", 0)
            )
```

lib/ansiblelint/errors.py
```diff
@@ -18,20 +22,52 @@ def __repr__(self):
...
-        self.linenumber = 0
-        self.line = None
-        self.filename = None
+        if not (message or rule):
+            raise RuntimeError("Calling MatchError requires either a message or a rule.")
+
+        self.message = message or getattr(rule, "description", "")
+        self.linenumber = linenumber
+        self.line = line
+        self.filename = normpath(str(filename)) if filename else filename
```
webknjaz, line 53: no need to stringify as it's done internally anyway
```suggestion
        self.filename = normpath(filename) if filename else filename
```

lib/ansiblelint/errors.py
```diff
@@ -18,20 +22,52 @@ def __repr__(self):
         return formatstr.format(self.rule.id, self.message,
```
webknjaz, line 22: How come `self.rule.id` is expected to always exist here but not in other places?

lib/ansiblelint/errors.py
```diff
@@ -18,20 +22,52 @@ def __repr__(self):
...
+    def __hash__(self):
+        """Perform hash of matches."""
+        return hash((self.message, self.rule, self.filename, self.linenumber))
+
+    def __lt__(self, other):
+        """Enable sorting of MatchError instances."""
+        return (
+                self.filename, self.linenumber, self.message, getattr(self.rule, "id", 0)
+            ) < (
+            other.filename, other.linenumber, other.message, getattr(other.rule, "id", 0))
```
webknjaz, line 73:
```suggestion
        return hash(self._hash_key)

    def __lt__(self, other):
        """Enable sorting of MatchError instances."""
        return self._hash_key < other._hash_key

    @property
    def _hash_key(self):
        return self.filename, self.linenumber, self.message, getattr(self.rule, "id", 0)
```

lib/ansiblelint/__main__.py
```diff
@@ -115,11 +115,7 @@ def main() -> int:
...
         matches.extend(runner.run())
 
-    matches.sort(
-        key=lambda x: (
-            normpath(x.filename),
-            x.linenumber,
-            getattr(x.rule, 'id', 0)))
+    matches.sort()
 
     for match in matches:
```
webknjaz, line 120:
```suggestion
    for match in sorted(matches):
```

lib/ansiblelint/errors.py
ssbarnea: One of the reasons why I proposed https://github.com/ansible/ansible-lint/pull/884

lib/ansiblelint/errors.py
```diff
@@ -12,10 +15,14 @@ def __init__(self, message=None, linenumber=0, line=None, filename=None, rule=No
         """Initialize a MatchError instance."""
         super().__init__(message)
 
+        if not (message or rule):
+            self.linenumber = 0
+            raise RuntimeError("Calling MatchError requires either a message or a rule.")
```
webknjaz, line 20: Technically, it's not "calling" but initializing (or raising).
Also, I'd like to follow Python's own conventions and raise the same type of exception that is `TypeError`:
```python
In [1]: def some_func(arg): ...                                               

In [2]: some_func(1)                                                          

In [3]: some_func()                                                           
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-3d9630d20b3d> in <module>
----> 1 some_func()

TypeError: some_func() missing 1 required positional argument: 'arg'


In [4]: def some_func_w_kwarg_only(arg, *, kwarg): ...                        

In [6]: some_func_w_kwarg_only(1, kwarg=2)                                    

In [7]: some_func_w_kwarg_only(1)                                             
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-861031c984e2> in <module>
----> 1 some_func_w_kwarg_only(1)

TypeError: some_func_w_kwarg_only() missing 1 required keyword-only argument: 'kwarg'

In [8]: some_func_w_kwarg_only(kwarg=2)                                       
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-bca0418a0e53> in <module>
----> 1 some_func_w_kwarg_only(kwarg=2)

TypeError: some_func_w_kwarg_only() missing 1 required positional argument: 'arg'

In [9]: class Demo: 
   ...:     def __init__(self, arg, *, kwarg): ... 
   ...:                                                                       

In [10]: Demo(1, kwarg=2)                                                     
Out[10]: <__main__.Demo at 0x7fe29886c2e8>

In [11]: Demo(kwarg=2)                                                        
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-5b16a6841949> in <module>
----> 1 Demo(kwarg=2)

TypeError: __init__() missing 1 required positional argument: 'arg'

In [12]: Demo(1)                                                              
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-12-69c1ae1e2087> in <module>
----> 1 Demo(1)

TypeError: __init__() missing 1 required keyword-only argument: 'kwarg'
```

And so the better exception would be
```suggestion
            raise TypeError(
                f'{self.__class__.__name__}() missing a '
                "required argument: one of 'message' or 'rule'",
            )
```

lib/ansiblelint/errors.py
```diff
@@ -27,3 +34,19 @@ def __repr__(self):
 
         return formatstr.format(_id, self.message,
                                 self.filename, self.linenumber, self.line)
+
+    @property
+    def _hash_key(self):
+        return self.filename, self.linenumber, self.message, getattr(self.rule, 'id', 0)
```
webknjaz, line 40: Side note: should we cache this?

lib/ansiblelint/errors.py
```diff
@@ -27,3 +34,19 @@ def __repr__(self):
...
+
+    def __lt__(self, other):
+        """Enable sorting of MatchError instances."""
+        return self._hash_key < other._hash_key
+
+    def __hash__(self):
+        """Perform hash of matches."""
```
webknjaz, line 47: This is not a comment but a docstring: it shouldn't repeat the same thing that is written just one line below.
```suggestion
        """Return a hash value of the MatchError instance."""
```

lib/ansiblelint/errors.py
```diff
@@ -27,3 +34,19 @@ def __repr__(self):
...
+        return hash(self._hash_key)
+
+    def __eq__(self, other):
+        """Identify duplicate matches."""
```
webknjaz, line 51:
```suggestion
        """Identify whether the other object represents the same rule match."""
```

lib/ansiblelint/errors.py
```diff
@@ -27,3 +34,19 @@ def __repr__(self):
...
+        return self.__hash__() == other.__hash__()
```
webknjaz, line 52: Hash equality is not enough. We should check if the type is compatible too. And if not, we should give a chance to the other object to check it.
```suggestion
        if not isinstance(other, self.__class__):
            raise NotImplementedError
        return self.__hash__() == other.__hash__()
```

lib/ansiblelint/errors.py
```diff
@@ -27,3 +34,19 @@ def __repr__(self):
...
         return formatstr.format(_id, self.message,
                                 self.filename, self.linenumber, self.line)
+
+    @property
+    def _hash_key(self):
+        return self.filename, self.linenumber, self.message, getattr(self.rule, 'id', 0)
+
+    def __lt__(self, other):
+        """Enable sorting of MatchError instances."""
+        return self._hash_key < other._hash_key
```
webknjaz, line 44:
```suggestion
        if not isinstance(other, self.__class__):
            raise NotImplementedError
        return self._hash_key < other._hash_key
```

lib/ansiblelint/errors.py
```diff
@@ -27,3 +34,19 @@ def __repr__(self):
 
         return formatstr.format(_id, self.message,
                                 self.filename, self.linenumber, self.line)
+
+    @property
+    def _hash_key(self):
+        return self.filename, self.linenumber, self.message, getattr(self.rule, 'id', 0)
+
+    def __lt__(self, other):
+        """Enable sorting of MatchError instances."""
```
webknjaz, line 43: This docstring doesn't explain what the method does, it only refers to a side-effect and therefore is incorrect.
```suggestion
        """Return whether the current object is less than the other."""
```

test/TestMatchError.py
```diff
@@ -0,0 +1,29 @@
...
+        (MatchError(rule=BecomeUserWithoutBecomeRule), MatchError(rule=AlwaysRunRule))
+    ))
+def test_matcherror_sort(a, b):
+    assert b < a
+    assert a > b
+    assert a != b
+
+
+def test_matcherror_invalid():
+    with pytest.raises(RuntimeError):
```
webknjaz, line 28: You should always use `match=` for such tests to avoid false-positive results.
```suggestion
    expected_err = '^MatchError\(\) missing a required argument: one of 'message' or 'rule'$"
    with pytest.raises(TypeError, match=expected_err):
```

test/TestMatchError.py
```diff
@@ -0,0 +1,29 @@
...
+        # rule id 501 > rule id 101
+        (MatchError(rule=BecomeUserWithoutBecomeRule), MatchError(rule=AlwaysRunRule))
+    ))
+def test_matcherror_sort(a, b):
+    assert b < a
+    assert a > b
+    assert a != b
+
+
+def test_matcherror_invalid():
```
webknjaz, line 27:
```suggestion
def test_matcherror_invalid():
    """Ensure that MatchError requires message or rule."""
```

test/TestMatchError.py
```diff
@@ -0,0 +1,29 @@
...
+        (MatchError("z"), MatchError("a")),
+        # filenames takes priority in sorting
+        (MatchError("a", filename="b"), MatchError("a", filename="a")),
+        # rule id 501 > rule id 101
+        (MatchError(rule=BecomeUserWithoutBecomeRule), MatchError(rule=AlwaysRunRule))
+    ))
+def test_matcherror_sort(a, b):
+    assert b < a
+    assert a > b
+    assert a != b
```
webknjaz, line 24: A test must contain just one assert. This will ensure that the tests show up separately and early failures don't prevent other asserts from running.

test/TestMatchError.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@pytest.mark.parametrize(
+    ('a', 'b'), (
+        # sorting by message
+        (MatchError("z"), MatchError("a")),
+        # filenames takes priority in sorting
+        (MatchError("a", filename="b"), MatchError("a", filename="a")),
+        # rule id 501 > rule id 101
+        (MatchError(rule=BecomeUserWithoutBecomeRule), MatchError(rule=AlwaysRunRule))
```
webknjaz, line 19: Also, add a test case with a file having a non-numeric ID.

lib/ansiblelint/errors.py
```diff
@@ -12,10 +15,17 @@ def __init__(self, message=None, linenumber=0, line=None, filename=None, rule=No
         """Initialize a MatchError instance."""
         super().__init__(message)
 
+        if not (message or rule):
+            self.linenumber = 0
```
webknjaz, line 19: @ssbarnea could please explain why this is in the branch that will explode anyway?

lib/ansiblelint/errors.py
ssbarnea: No. If we do, it will count as premature optimization. Lets keep optimizations where it really affects total execution time.

lib/ansiblelint/errors.py
webknjaz: Should it be just removed? (judging from the logic I see)
```suggestion
```

test/TestMatchError.py
```diff
@@ -0,0 +1,31 @@
...
+    ))
+def test_matcherror_sort(a, b):
+    assert b < a
+    assert a > b
+    assert a != b
+
+
+def test_matcherror_invalid():
+    """Ensure that MatchError requires message or rule."""
+    expected_err = '^MatchError\(\) missing a required argument: one of 'message' or 'rule'$"
```
webknjaz, line 29: Oh, this probably needs
```suggestion
    expected_err = r'^MatchError\(\) missing a required argument: one of 'message' or 'rule'$"
```

test/TestMatchError.py
ssbarnea: I do not think that it makes sense to split this into multiple tests, especially as it is already parametrized.

Doing a ``b < a and a < b and a != b`` would also use a single assert but it would clearly not produce better errors if the test fails.

Can you please provide some links that sustain the use of multiple asserts is inherently wrong? IMHO, that is one case where they are ok. 

I doubt anyone pays us by number of tests written, so no need to boost their number ;)

test/TestMatchError.py
ssbarnea: Done

test/TestMatchError.py
ssbarnea: Already done.

test/TestMatchError.py
albinvass: I agree with @ssbarnea here, I don't think it makes sense to split this into several tests just to keep a 'one assert per test' policy.

ssbarnea: I found a way to fix it!

webknjaz: Well, the idea behind the "single assert" is not that only one keyword is allowed syntactically but that one single behavior should be tested and there should be one reason for the test to fail.

https://dev.to/donut87/one-assert-per-test---what-is-that-about-4l75

test/TestMatchError.py
```diff
@@ -0,0 +1,44 @@
...
+        # sorting by message
+        (MatchError("z"), MatchError("a")),
+        # filenames takes priority in sorting
+        (MatchError("a", filename="b"), MatchError("a", filename="a")),
+        # rule id 501 > rule id 101
+        (MatchError(rule=BecomeUserWithoutBecomeRule), MatchError(rule=AlwaysRunRule)),
+        # rule id "200" > rule id 101
+        (MatchError(rule=AnsibleLintRuleWithStringId), MatchError(rule=AlwaysRunRule))
+    ))
+class TestMatchErrorCompare():
```
webknjaz, line 35: You shouldn't use empty braces in class definitions.

test/TestMatchError.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_lt(self, a, b):
+        assert b < a
+
+    def test_gt(self, a, b):
+        assert a > b
+
+    def test_eq(self, a, b):
+        assert a != b
```
webknjaz, line 44: I don't think that using classes is a good idea. We have modules for scoping.

test/TestMatchError.py
```diff
@@ -0,0 +1,44 @@
...
+    ))
+class TestMatchErrorCompare():
+
+    def test_lt(self, a, b):
+        assert b < a
+
+    def test_gt(self, a, b):
+        assert a > b
+
+    def test_eq(self, a, b):
```
webknjaz, line 43: eq is `==`, not `!=`

test/TestMatchError.py
webknjaz, line 44:
```suggestion
parametrize_match_error = pytest.mark.parametrize(
    ('left_match_error', 'right_match_error'),
    (
        # sorting by message
        (MatchError("z"), MatchError("a")),

        # filename takes priority in sorting
        (MatchError("a", filename="b"), MatchError("a", filename="a")),

        # rule id 501 > rule id 101
        (MatchError(rule=BecomeUserWithoutBecomeRule), MatchError(rule=AlwaysRunRule)),

        # rule id "200" > rule id 101
        (MatchError(rule=AnsibleLintRuleWithStringId), MatchError(rule=AlwaysRunRule)),

        # line should be taken into account
        (MatchError("a", line="b"), MatchError("a", line="a")),
    ),
)


@parametrize_match_error
def test_match_error_less_than(left_match_error, right_match_error):
    """Check 'less than' protocol implementation in MatchError."""
    assert left_match_error < right_match_error


@parametrize_match_error
def test_match_error_greater_than(left_match_error, right_match_error):
    """Check 'greater than' protocol implementation in MatchError."""
    assert left_match_error > right_match_error


@parametrize_match_error
def test_match_error_not_equal(left_match_error, right_match_error):
    """Check 'not equals' protocol implementation in MatchError."""
    assert left_match_error != right_match_error
```

test/TestMatchError.py
```diff
@@ -0,0 +1,44 @@
...
+@pytest.mark.parametrize(
+    ('a', 'b'), (
+        # sorting by message
+        (MatchError("z"), MatchError("a")),
+        # filenames takes priority in sorting
+        (MatchError("a", filename="b"), MatchError("a", filename="a")),
+        # rule id 501 > rule id 101
+        (MatchError(rule=BecomeUserWithoutBecomeRule), MatchError(rule=AlwaysRunRule)),
+        # rule id "200" > rule id 101
+        (MatchError(rule=AnsibleLintRuleWithStringId), MatchError(rule=AlwaysRunRule))
```
webknjaz, line 33: Based on the latest comment, there should be a param with different "line" arg.

test/TestMatchError.py
```diff
@@ -0,0 +1,44 @@
+from ansiblelint.errors import MatchError
+from ansiblelint.rules import AnsibleLintRule
+from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule
+from ansiblelint.rules.BecomeUserWithoutBecomeRule import BecomeUserWithoutBecomeRule
+
+import pytest
+
+
+def test_matcherror_compare():
```
webknjaz, line 9:
```suggestion
def test_matcherror_compare():
"""Check that MatchError instances with similar attrs are equivalent."""
```

webknjaz, line 9: This needs to be parametrized with all of the args because it doesn't catch a bug with `self.line` missing from hash.

test/TestMatchError.py
```diff
@@ -7,6 +7,7 @@
 
 
 def test_matcherror_compare():
+"""Check that MatchError instances with similar attrs are equivalent."""
```
webknjaz, line 10: Looks like I missed the indentation here
```suggestion
    """Check that MatchError instances with similar attrs are equivalent."""
```

test/TestMatchError.py
ssbarnea: Not really, see last change where I document that line dictionary! is not part of the hashing logic.

webknjaz: Then it should be added to tests showing that it's expected.

lib/ansiblelint/errors.py
```diff
@@ -27,3 +36,23 @@ def __repr__(self):
 
         return formatstr.format(_id, self.message,
                                 self.filename, self.linenumber, self.line)
+
+    @property
+    def _hash_key(self):
+        return self.filename, self.linenumber, str(getattr(self.rule, 'id', 0)), self.message
```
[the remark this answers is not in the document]
webknjaz: This needs a comment with the motivation behind leaving `self.line` out.

test/TestMatchError.py
ssbarnea: done

test/TestMatchError.py
```diff
@@ -0,0 +1,38 @@
...
+
+import pytest
+
+
+def test_matcherror_compare():
+    assert MatchError("foo") == MatchError("foo")
+
+
+class AnsibleLintRuleWithStringId(AnsibleLintRule):
+    id = "200"
```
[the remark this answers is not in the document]
ssbarnea: done

test/TestMatchError.py
ssbarnea: renamed

lib/ansiblelint/errors.py
```diff
@@ -27,3 +45,24 @@ def __repr__(self):
 
         return formatstr.format(_id, self.message,
                                 self.filename, self.linenumber, self.line)
+
+    @property
+    def _hash_key(self):
+        # line attr is knowingly excluded, as dict is not hashable
```
webknjaz, line 51: Have you considered turning dict into a tuple of tuples for the hashing purposes?

test/TestMatchError.py
```diff
@@ -0,0 +1,51 @@
...
+
+@pytest.mark.parametrize(
+    ('a', 'b'), (
+        (MatchError("foo"), MatchError("foo")),
+        # line is ignored in comparisions, even if it may be different
+        (MatchError("a", line={}), MatchError("a", line={"a": "b"}))
+    ))
+def test_matcherror_compare(a, b):
+    """Check that MatchError instances with similar attrs are equivalent."""
+    assert a == b
```
webknjaz, line 17: I prefer better names here
```suggestion
    ('left_match_error', 'right_match_error'),
    (
        (MatchError("foo"), MatchError("foo")),
        # line is ignored in comparisions, even if it may be different
        (MatchError("a", line={}), MatchError("a", line={"a": "b"})),
    ),
)
def test_matcherror_compare(left_match_error, right_match_error):
    """Check that MatchError instances with similar attrs are equivalent."""
    assert left_match_error == right_match_error
```

lib/ansiblelint/errors.py
ssbarnea: Not really because it would be outside the current change scope. I do not want to rewrite half of the ansible-lint to make this change happen.

I am not against such idea, but I know for sure that the only way to progress is in small steps, not big rewrites that never end being merged.

The scope of a review is to assure that now new bugs are introduced and that the overall code quality is improved, not that is "perfect" or that it also fixes any other per-existent issue.

Shortly, I would try to contain changes and minimize them. We should raise bugs for found issues, maybe even assign them.

ssbarnea:
@webknjaz Any other changes needed?

webknjaz:
> @webknjaz Any other changes needed?

Yes, I wanted you to apply https://github.com/ansible/ansible-lint/pull/881#discussion_r451135446

test/TestMatchError.py
webknjaz: @ssbarnea ^

lib/ansiblelint/errors.py
webknjaz: I totally support the idea of small incremental changes. The reason I'm raising this question is that I suspect that in the current state it'll negatively influence the filtering logic. That is when there's two different matches, it'll randomly remove one of them as a duplicate. Right? I'd consider that a bug.

lib/ansiblelint/rules/__init__.py
```diff
@@ -221,8 +221,7 @@ def run(self, playbookfile, tags=set(), skip_list=frozenset()) -> List:
                     matches.extend(rule.matchlines(playbookfile, text))
                     matches.extend(rule.matchtasks(playbookfile, text))
                     matches.extend(rule.matchyaml(playbookfile, text))
-
-        return matches
+        return list(set(matches))
```
webknjaz, line 224: This is what I'm concerned about here https://github.com/ansible/ansible-lint/pull/881#discussion_r451631163

test/TestMatchError.py
ssbarnea: I see no compelling reason not to use classes, especially as this is not the first place to do so.

Can we please stop the nitpicking and stick to things that really matter?

For the dimension of this change it received an overwhelming amount of change requests, many of them even contradictory.

The review process is not about forcing the OP to write the change and exactly how the reviewer would have written it. All this nitpicking is draining and discouraging me from contributing.

webknjaz: I don't think that having classes is a good idea. Also, "not the first place" is a bad excuse. Those things are legacy unittest-based modules mostly.

FWIW, if you want to keep the class, feel free but you must add docstrings and proper variable names. I won't approve a/b.
Also, add a param with lines (see the code suggestion above). I requested this to be present in tests several times already.

lib/ansiblelint/errors.py
webknjaz: @ssbarnea ^

ssbarnea: I removed the set() part, so we no longer remove duplicates. We can analyze and make a decision in another change. The best is to do this when we have at least one example or two of duplicated results.

lib/ansiblelint/rules/__init__.py
ssbarnea: Sorted

test/TestMatchError.py
ssbarnea: No longer needed as we do not actively use the hashing anymore.

webknjaz: No, it's still needed because the hash key is explicitly used for the things you test here.

test/TestMatchError.py
webknjaz: ^ the request still stands

test/TestMatchError.py
ssbarnea: We should not test stuff for which we do not have a determined behavior. We test only documented features/behaviors.

test/TestMatchError.py
ssbarnea: Implemented var rename plus docstrings. even if I doubt the var names really add value for final reports or readability of the two-line test functions.

test/TestMatchError.py
```diff
@@ -0,0 +1,83 @@
...
+    ('other'),
+    (
+        (None),
+        ("foo"),
+        (42),
+        (Exception("foo")),
+    ),
+)
+@pytest.mark.parametrize(
+    ('operation'),
```
webknjaz, line 71:
```suggestion
    'operation',
```

test/TestMatchError.py
```diff
@@ -0,0 +1,83 @@
...
+        """Check 'greater than' protocol implementation in MatchError."""
+        assert left_match_error > right_match_error
+
+    def test_match_error_not_equal(self, left_match_error, right_match_error):
+        """Check 'not equals' protocol implementation in MatchError."""
+        assert left_match_error != right_match_error
+
+
+@pytest.mark.parametrize(
+    ('other'),
```
webknjaz, line 62:
```suggestion
    'other',
```

test/TestMatchError.py
```diff
@@ -0,0 +1,83 @@
...
+    (
+        (None),
+        ("foo"),
+        (42),
+        (Exception("foo")),
+    ),
```
webknjaz, line 68:
```suggestion
    'other',
    (
        None,
        "foo",
        42,
        Exception("foo"),
    ),
```

test/TestMatchError.py
```diff
@@ -0,0 +1,83 @@
...
+)
+@pytest.mark.parametrize(
+    ('operation'),
+    (
+        ('__eq__'),
+        ('__ne__'),
+        ('__le__'),
+        ('__gt__')
```
webknjaz, line 76:
```suggestion
    'operation',
    (
        '__eq__',
        '__ne__',
        '__le__',
        '__gt__',
```

test/TestMatchError.py
```diff
@@ -0,0 +1,83 @@
...
+        '__ne__',
+        '__le__',
+        '__gt__',
+    ),
+    ids=['eq', 'ne', 'le', 'gt']
+)
+def test_matcherror_compare_invalid(other, operation):
+    """Check that MatchError comparison with other types raises."""
+    with pytest.raises(NotImplementedError):
+        getattr(MatchError("foo"), operation)(other)
```
webknjaz, line 83:
```suggestion
    'operation',
    (
        operator.eq,
        operator.ne,
        operator.le,
        operator.gt,
    ),
    ids=['eq', 'ne', 'le', 'gt']
)
def test_matcherror_compare_invalid(other, operation):
    """Check that MatchError comparison with other types raises."""
    with pytest.raises(NotImplementedError):
        operation(MatchError("foo"), other)
```

test/TestMatchError.py
```diff
@@ -0,0 +1,83 @@
+from ansiblelint.errors import MatchError
```
webknjaz, line 1:
```suggestion
"""Tests for MatchError."""

import operator

from ansiblelint.errors import MatchError
```

PR #884, opened 2020-07-04
Combine MatchError into Match

ssbarnea:
Combine two very similar classes in order to avoid extra complexity
in processing rule validates.

From now on a Match is an Exception so it can be raised, captured,
and annotated in various places in the call stack.

### DIFF
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -1,37 +1,29 @@
 """Exceptions and error representations."""
 
 
-class Match(object):
-    """Rule violation detected during linting."""
+class Match(ValueError):
+    """Rule violation detected during linting.
 
-    def __init__(self, linenumber, line, filename, rule, message=None) -> None:
+    It can be raised as Exception but also just added to the list of found
+    rules violations.
+    """
+
+    def __init__(self, message=None, linenumber=0, line=None, filename=None, rule=None) -> None:
         """Initialize a Match instance."""
+        super().__init__(message)
+
+        self.message = message or getattr(rule, 'shortdesc', "")
         self.linenumber = linenumber
         self.line = line
         self.filename = filename
         self.rule = rule
-        self.message = message or rule.shortdesc
 
     def __repr__(self):
         """Return a Match instance representation."""
         formatstr = u"[{0}] ({1}) matched {2}:{3} {4}"
-        return formatstr.format(self.rule.id, self.message,
+        # note that `rule.id` can be int, str or even missing, as users
+        # can defined their own custom rules.
+        _id = getattr(self.rule, "id", "000")
+
+        return formatstr.format(_id, self.message,
                                 self.filename, self.linenumber, self.line)
-
-
-class MatchError(ValueError):
-    """Exception that would end-up as linter rule match."""
-
-    def __init__(self, message, rule=None):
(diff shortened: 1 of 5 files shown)


lib/ansiblelint/errors.py
```diff
@@ -1,37 +1,27 @@
 """Exceptions and error representations."""
 
 
-class Match(object):
-    """Rule violation detected during linting."""
+class Match(Exception):
```
webknjaz, line 4: We should keep
```suggestion
class Match(ValueError):
```

lib/ansiblelint/errors.py
```diff
@@ -1,37 +1,27 @@
...
     def __repr__(self):
         """Return a Match instance representation."""
         formatstr = u"[{0}] ({1}) matched {2}:{3} {4}"
-        return formatstr.format(self.rule.id, self.message,
-                                self.filename, self.linenumber, self.line)
-
-
-class MatchError(ValueError):
-    """Exception that would end-up as linter rule match."""
+        _id = getattr(self.rule, "id", "000")
```
webknjaz, line 24: Rules have int ids, no need to fall-back to a string. Besides, you can easily use the number padding specifier:
```suggestion
        formatstr = "[{0:03d}] ({1}) matched {2}:{3} {4}"
        _id = getattr(self.rule, "id", 0)
```

ssbarnea: `rule.id` can be int, str or even missing, @albinvass wrote some custom rules for zuul starting with "ZUUL" prefix and I do not want to break them.

albinvass: @webknjaz yes, that's even part of the documentation for creating custom rules:
https://docs.ansible.com/ansible-lint/rules/rules.html#creating-custom-rules

lib/ansiblelint/errors.py
```diff
@@ -1,37 +1,29 @@
 """Exceptions and error representations."""
 
 
-class Match(object):
-    """Rule violation detected during linting."""
+class Match(ValueError):
```
webknjaz, line 4: I still think that it should be
```suggestion
class MatchError(ValueError):
```
https://github.com/ansible/ansible-lint/pull/884#discussion_r449854967

test/TestFormatter.py
```diff
@@ -33,13 +33,13 @@ def setUp(self):
...
         self.formatter.format(match, True)
 
     def test_unicode_format_string(self):
-        match = Match(1, "hello", "filename.yml", self.rule, u'\U0001f427')
+        match = Match(u'\U0001f427', 1, "hello", "filename.yml", self.rule)
         self.formatter.format(match, False)
 
     def test_dict_format_line(self):
-        match = Match(1, {'hello': 'world'}, "filename.yml", self.rule, "xyz")
+        match = Match("xyz", 1, {'hello': 'world'}, "filename.yml", self.rule,)
```
webknjaz, line 44: Accidental comma at the end:
```suggestion
        match = MatchError("xyz", 1, {'hello': 'world'}, "filename.yml", self.rule)
```

lib/ansiblelint/errors.py
ssbarnea: done

ssbarnea:
zuul issue is infra related, reported.

PR #888, opened 2020-07-10
Add ansible210 testing

ssbarnea:
Instead of using devel branch which is now 2.11, switch back to pypi versions for 2.10 testing.

Related: https://github.com/ansible/ansible/issues/70705

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -157,7 +157,7 @@ jobs:
       run: |
         python -m tox --parallel auto --parallel-live --notest -vv
       env:
-        TOXENV: ansible28,ansible29,ansibledevel
+        TOXENV: ansible28,ansible29,ansible210,ansibledevel
     - name: "Test with tox: ansible28"
       run: |
         python -m tox
@@ -169,6 +169,11 @@ jobs:
         python -m tox
       env:
         TOXENV: ansible29
+    - name: "Test with tox: ansible210"
+      run: |
+        python -m tox
+      env:
+        TOXENV: ansible210
     - name: "Test with tox: ansibledevel"
       run: |
         python -m tox
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.16.1
-envlist = lint,py{38,37,36}-ansible{29,28,devel}
+envlist = lint,py{38,37,36}-ansible{29,28,210,devel}
 isolated_build = true
 requires =
   setuptools >= 41.4.0
@@ -15,10 +15,11 @@ description =
 deps =
   ansible28: ansible>=2.8,<2.9
   ansible29: ansible>=2.9,<2.10
-  # TODO(ssbarnea): Replace replace the two tarbals created by @abadger1999 once
-  # we official (pre-)releases are available on pypi.org
-  ansibledevel: ansible-base @ https://toshio.fedorapeople.org/ansible/acd/ansible-base/ansible-base-2.10.0.dev1.tar.gz#egg=ansible-base
-  ansibledevel: ansible @ https://toshio.fedorapeople.org/ansible/acd/ansible/ansible-2.10.0.tar.gz#egg=ansible
+  ansible210: ansible>=2.10.0a1,<2.11
+  # Be sure we do not install old ansible from default deps
+  # https://github.com/ansible/ansible/issues/70705
+  ansibledevel: ansible>=2.10.0a1,<2.11
+  ansibledevel: ansible-base @ git+https://github.com/ansible/ansible.git
   ruamel.yaml==0.16.5  # NOTE: 0.15.34 has issues with py37
   flake8
   pep8-naming


tox.ini
```diff
@@ -15,10 +15,7 @@ description =
 deps =
   ansible28: ansible>=2.8,<2.9
   ansible29: ansible>=2.9,<2.10
-  # TODO(ssbarnea): Replace replace the two tarbals created by @abadger1999 once
-  # we official (pre-)releases are available on pypi.org
-  ansibledevel: ansible-base @ https://toshio.fedorapeople.org/ansible/acd/ansible-base/ansible-base-2.10.0.dev1.tar.gz#egg=ansible-base
-  ansibledevel: ansible @ https://toshio.fedorapeople.org/ansible/acd/ansible/ansible-2.10.0.tar.gz#egg=ansible
+  ansibledevel: ansible>=2.10.0a1,<2.11
```
webknjaz, line 18: I'd call it 
```suggestion
  ansible210: ansible>=2.10.0a1,<2.11
```
because it's not really devel.

ssbarnea: In that case lots of other changes would be required, so we add all the 210 jobs.

webknjaz: Yep. But it's not devel. devel of ansible-base is on github.

tox.ini
```diff
@@ -15,10 +15,10 @@ description =
 deps =
   ansible28: ansible>=2.8,<2.9
   ansible29: ansible>=2.9,<2.10
+  ansible210: ansible>=2.10.0a1,<2.11
   # TODO(ssbarnea): Replace replace the two tarbals created by @abadger1999 once
   # we official (pre-)releases are available on pypi.org
-  ansibledevel: ansible-base @ https://toshio.fedorapeople.org/ansible/acd/ansible-base/ansible-base-2.10.0.dev1.tar.gz#egg=ansible-base
-  ansibledevel: ansible @ https://toshio.fedorapeople.org/ansible/acd/ansible/ansible-2.10.0.tar.gz#egg=ansible
+  ansibledevel: ansible @ git+https://github.com/ansible/ansible#egg=ansible
```
webknjaz, line 21: This is incorrect. What this is pointing at is base:
```suggestion
  ansibledevel: ansible-base @ git+https://github.com/ansible/ansible.git
```

webknjaz: I'm curious if @abadger has any plans to build nightlies for ACD...

abadger: I don't at the moment.

webknjaz: @abadger could you point to the ACD build script? maybe we could generate something for ourselves.

abadger: https://github.com/ansible-community/antsibull/blob/main/docs/build-ansible.rst#building-ansible

For 2.10, we use the build-single strategy (not build-multiple or build-collection). The instructions have one problem due to a bug in the way the script assigns default values to the deps-file. Leave out `--deps-file ansible-build-data/2.10/acd-2.10.0.deps` when you build-single and it should work.  The deps file will end up in the same directory as the tarball then.

webknjaz: > python -m pip install --user --upgrade --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple ansible

@abadger FYI this doesn't make much sense: this does not set preference for indexes. Using just `--extra-index-url https://test.pypi.org/simple/` would give the same result.

.github/workflows/tox.yml
```diff
@@ -82,15 +82,17 @@ jobs:
     name: >-
       py${{ matrix.python-version }}@${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
```
webknjaz, line 85:
```suggestion
    continue-on-error: >
      ${{
        matrix.experimental
        && true
        || false
      }}
```

.github/workflows/tox.yml
```diff
@@ -82,15 +82,17 @@ jobs:
...
-      # fail-fast: false
+      fail-fast: false
       # max-parallel: 5
       # The matrix testing goal is to cover the *most likely* environments
       # which are expected to be used by users in production. Avoid adding a
       # combination unless there are good reasons to test it, like having
       # proof that we failed to catch a bug by not running it. Using
       # distribution should be prefferred instead of custom builds.
       matrix:
+        experimental: [false]
```
webknjaz, line 95: this shouldn't be necessary

.github/workflows/tox.yml
```diff
@@ -169,11 +177,11 @@ jobs:
         python -m tox
       env:
         TOXENV: ansible29
-    - name: "Test with tox: ansibledevel"
+    - name: "Test with tox: ansible210"
```
webknjaz, line 180: I'm curious if
```suggestion
    - name: Test with tox: ${{ env.TOXENV }}
```
will render the var from the outer scope of the step-specific one...

.github/workflows/tox.yml
```diff
@@ -82,8 +82,14 @@ jobs:
...
     runs-on: ${{ matrix.os }}
+    continue-on-error: >-
+      ${{
+        matrix.experimental
+        && true
+        || false
+      }}
     strategy:
-      # fail-fast: false
+      fail-fast: false
```
webknjaz, line 92: @ssbarnea could you explain why this change is needed?

ssbarnea: Not sure if needed, it was in docs example around using experimental jobs. I removed it in last patch.

PR #891, opened 2020-07-12
LRU Cache for frequently called functions

ragne:
#### Measurements
For my 10kloc of yaml ansible repo:
 * `ansible-lint 4.2.0`: `time ansible-lint *.yml -q` returns `real    0m56.920`
 * version from PR returns: `real    0m8.039s` for the same command (same environment, just used setup.py to install)

OS: macOS 10.15.4
CPU: 2,3 GHz 8-Core Intel Core i9
Python: Python 3.7.8

### DIFF
--- a/lib/ansiblelint/skip_utils.py
+++ b/lib/ansiblelint/skip_utils.py
@@ -19,6 +19,7 @@
 # THE SOFTWARE.
 
 """Utils related to inline skipping of rules."""
+from functools import lru_cache
 from itertools import product
 import logging
 import sys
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/skip_utils.py
```diff
@@ -72,10 +73,22 @@ def append_skipped_rules(pyyaml_data: str, file_text: str, file_type: FileType):
     return yaml_skip
 
 
+@lru_cache(maxsize=128)
```
ssbarnea, line 76: Why 128? Do we really need a default? I know for sure many repos with more than 128 yaml files. Also, I doubt we would run out of memory without a cache.

Make it 1024, but put a comment before to explaining that default is 128, which may not be big enough for all known use cases.

ragne: This should be around number of total rules; Consider the following: this function is called from either `matchyaml` or `matchtasks` in `AnsibleLintRule`, which called basically in the following:

```python
for playbook in playbooks:
    for rule in rules:
        rule.run(...)
```

Basically LRU cache should live only for each iteration of inner loop (e.g after inner loop completed, we don't care anymore about content of `playbook`; 
Are my assumptions wrong?

lib/ansiblelint/utils.py
```diff
@@ -533,6 +534,7 @@ def get_normalized_tasks(yaml, file):
     return res
 
 
+@lru_cache(maxsize=128)
```
ragne, line 537: This is the same as above. Called only in `matchyaml` or `matchtasks`

ssbarnea: The only reason why I did not approve it is that I did not had time to test it myself. I want to fix 2.10 testing first and right after that this one.

PR #894, opened 2020-07-13
Allow backticks in shell commands

turettn:
```yaml
- name: create rabbitmq-admin-creds creds
  shell: kubectl --context {{ k8s_context }} \
                 -n {{ namespace }} create secret generic rabbitmq-admin-creds \
                 --from-literal=password=`openssl rand -base64 12` \
                 --from-literal=cookie=`uuidgen`
  register: rbt_secret
  changed_when: rbt_secret.rc == 0
  failed_when: rbt_secret.rc != 0 and 'AlreadyExists' not in rbt_secret.stderr
```

Ansible-lint did not like it:

```
[305] Use shell only when shell functionality is required
roles/secrets/tasks/main.yml:29
Task/Handler: create rabbitmq-admin-creds creds
```

I switched it to a command, and it failed because the backticks were not expanded by the shell.

This PR prevents 305 from being thrown if a backtick is present in the command.  A unit test is also included.

### DIFF
--- a/lib/ansiblelint/rules/UseCommandInsteadOfShellRule.py
+++ b/lib/ansiblelint/rules/UseCommandInsteadOfShellRule.py
@@ -42,4 +42,4 @@ class UseCommandInsteadOfShellRule(AnsibleLintRule):
             else:
                 unjinjad_cmd = self.unjinja(
                     ' '.join(task["action"].get("__ansible_arguments__", [])))
-            return not any([ch in unjinjad_cmd for ch in '&|<>;$\n*[]{}?'])
+            return not any([ch in unjinjad_cmd for ch in '&|<>;$\n*[]{}?`'])
--- a/test/command-instead-of-shell-success.yml
+++ b/test/command-instead-of-shell-success.yml
@@ -26,6 +26,9 @@
   - name: use shell generator
     shell: ls foo{.txt,.xml}
 
+  - name: use backticks
+    shell: ls `ls foo*`
+
   - name: use shell with cmd
     shell:
       cmd: |


webknjaz:
> ```yaml
>                  --from-literal=cookie=`uuidgen`
> ```

FWIW backticks are considered deprecated in favor of `$()` and I'd suggest you to use


```diff
 - name: create rabbitmq-admin-creds creds
   shell: kubectl --context {{ k8s_context }} \
                  -n {{ namespace }} create secret generic rabbitmq-admin-creds \
-                 --from-literal=password=`openssl rand -base64 12` \
+                 --from-literal=password=`openssl rand -base64 12` \
-                 --from-literal=cookie=$(uuidgen)
+                 --from-literal=cookie=$(uuidgen)
   register: rbt_secret
   changed_when: rbt_secret.rc == 0
   failed_when: rbt_secret.rc != 0 and 'AlreadyExists' not in rbt_secret.stderr
```

http://mywiki.wooledge.org/BashFAQ/082

turettn:
@webknjaz Thanks for your response!  To get my CI chain to stop complaining in the short term, I did exactly as you suggested and swapped in `$(...)` instead of the backticks.

While definitely not best practice, I don't think bash ever deprecated backticks (in the sense of "knock it off, these will not be supported in a future version").  The manual doesn't say anything about them being deprecated that I can find, and still lists them on the [command substitution page](https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html).

If there is a consensus that people should not be using backticks via the shell command in Ansible, could we split that into a separate check, in the spirit of [ShellWithoutPipefail](https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/ShellWithoutPipefail.py)?  That way, the implied fix (use command, not shell) will not break their code.

webknjaz:
Yeah, having a separate check sounds reasonable. I'm even thinking that maybe there should be a check that invokes some external binary like `shellcheck` that would do all the magic. OTOH this would probably be too heavy for the ansible-lint core.

PR #887, opened 2020-07-10
Enable isort

ssbarnea:
Introduce isort with minimal configuration needed in order to make it work (not make mistake).

The only files manually modified here the `.pre-config-config.yml` ones, the other ones where automatically fixed by isort.

Note that I am against including "preference" like configuration to isort as they are objective. We already had long debates with pylint and before and I will stick to minimalism, just enough to make
it produce decent output (does not even have to be perfect). The amount of changes done highlights
the number of human mistakes made at sorting and grouping dependencies.

Unless there is a clear bug like import present in wrong section, please do not comment regarding
isort decision to split or join imports, accept it and move on. We do have far more important changes
to review and merge.

PR #895, opened 2020-07-14
Standardize test imports

ssbarnea:
Adopt standard stdlib, 3rd-party, 1st-party sectioning for imports
inside tests.

Reduces size of #887 review.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,8 +1,8 @@
 """Test suite for ansible-lint."""
 
-import tempfile
-import shutil
 import os
+import shutil
+import tempfile
 
 from ansible import __version__ as ansible_version_str
 
--- a/test/TestAlwaysRunRule.py
+++ b/test/TestAlwaysRunRule.py
@@ -1,9 +1,12 @@
 import unittest
-from ansiblelint.rules import RulesCollection
-from ansiblelint.runner import Runner
-from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule
-from test import ANSIBLE_MAJOR_VERSION
+
 import pytest
+
+from ansiblelint.rules import RulesCollection
+from ansiblelint.rules.AlwaysRunRule import AlwaysRunRule
+from ansiblelint.runner import Runner
+
+from . import ANSIBLE_MAJOR_VERSION
 
 
 @pytest.mark.skipif(
--- a/test/TestBecomeUserWithoutBecome.py
+++ b/test/TestBecomeUserWithoutBecome.py
@@ -1,7 +1,8 @@
 import unittest
+
 from ansiblelint.rules import RulesCollection
-from ansiblelint.runner import Runner
 from ansiblelint.rules.BecomeUserWithoutBecomeRule import BecomeUserWithoutBecomeRule
+from ansiblelint.runner import Runner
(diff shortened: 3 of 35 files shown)


test/TestImportIncludeRole.py
```diff
@@ -2,7 +2,6 @@
 
 from ansiblelint.runner import Runner
 
-
```
webknjaz, line 5: Let's not do this here.

test/TestImportIncludeRole.py
```diff
@@ -2,7 +2,6 @@
 
 from ansiblelint.runner import Runner
```
webknjaz, line 4:
```suggestion


```

test/TestImportWithMalformed.py
```diff
@@ -4,7 +4,6 @@
 
 from ansiblelint.runner import Runner
```
webknjaz, line 6:
```suggestion


```

test/TestIncludeMissingFileRule.py
```diff
@@ -4,7 +4,6 @@
 
 from ansiblelint.runner import Runner
```
webknjaz, line 6:
```suggestion


```

test/TestLineNumber.py
```diff
@@ -20,7 +20,6 @@
 
 from ansiblelint.rules.SudoRule import SudoRule
```
webknjaz, line 22:
```suggestion


```

test/TestNestedJinjaRule.py
```diff
@@ -27,7 +27,6 @@
 
 from ansiblelint.runner import Runner
```
webknjaz, line 29:
```suggestion


```

test/TestSkipInsideYaml.py
```diff
@@ -1,6 +1,5 @@
 import pytest
```
webknjaz, line 2:
```suggestion


```

test/TestSkipPlaybookItems.py
```diff
@@ -1,6 +1,5 @@
 import pytest
```
webknjaz, line 2:
```suggestion


```

test/__init__.py
```diff
@@ -1,14 +1,13 @@
...
 
-import tempfile
-import shutil
 import os
+import shutil
+import tempfile
 
 from ansible import __version__ as ansible_version_str
 
 from ansiblelint.runner import Runner
```
webknjaz, line 10:
```suggestion


```

ssbarnea:
@webknjaz Maybe it was not obvious but the single line between imports and constants was what isort did, not my own addition. I think that the logic is to have double newline before "code" (class/function), but variable assignment being considered "before-code".

Anyway, I manually added a 2nd line as you requested.

webknjaz:
@ssbarnea my point is that isort mustn't conflict with the current convention and that's why I suggested the correct config for it earlier.

webknjaz:
And yes, that is the default isort mode but I think that it'd be better to keep two-line separator regardless w/o any branchy conditions. This when you write code, it's easier to do so w/o triggering isort to reformat things unnecessarily.

Issue #850, opened 2020-06-22
ansible-lint give no output and just a python error


imranshaik5155:
##### Summary
I am running the ansible-lint on all of our ansible code with 500+ roles, and it is showing no output and failing with python error.
How can I find where it is actually failing and with what reason.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible 2.9.9
  config file = None
  configured module search path = ['/Users/imranshaik/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.7.6 (default, Dec 30 2019, 19:38:26) [Clang 11.0.0 (clang-1100.0.33.16)]

ansible-lint --version
ansible-lint 4.2.0

```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
mac OS

```python-traceback
Traceback (most recent call last):
  File "/usr/local/bin/ansible-lint", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/__main__.py", line 187, in main
    matches.extend(runner.run())
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/__init__.py", line 287, in run
    skip_list=self.skip_list))
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/__init__.py", line 178, in run
    matches.extend(rule.matchyaml(playbookfile, text))
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/__init__.py", line 119, in matchyaml
    yaml = ansiblelint.utils.append_skipped_rules(yaml, text, file['type'])
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/utils.py", line 596, in append_skipped_rules
    yaml_skip = _append_skipped_rules(pyyaml_data, file_text, file_type)
  File "/usr/local/lib/python3.7/site-packages/ansiblelint/utils.py", line 635, in _append_skipped_rules
    if pyyaml_task.get('name') != ruamel_task.get('name'):
AttributeError: 'str' object has no attribute 'get'
```

ssbarnea:
Bug form asks for using latest version from pypi or master, which are 4.3.x ones. Please use them and you are unlikely to get the same error.

imranshaik5155:
[...]

PR #899, opened 2020-07-17
Avoid failure with playbooks having tasks key a null value

ssbarnea:
Avoid creating a stacktrace when encountering playbooks where tasks is defined but has a null value.

Fixes: #850

### DIFF
--- a/lib/ansiblelint/rules/IncludeMissingFileRule.py
+++ b/lib/ansiblelint/rules/IncludeMissingFileRule.py
@@ -23,7 +23,9 @@ class IncludeMissingFileRule(AnsibleLintRule):
     def matchplay(self, file, data):
         absolute_directory = file.get('absolute_directory', None)
         results = []
-        for task in data.get('tasks', []):
+
+        # avoid failing with a playbook having tasks: null
+        for task in (data.get('tasks', []) or []):
 
             # check if the id of the current rule is not in list of skipped rules for this play
             if self.id in task['skipped_rules']:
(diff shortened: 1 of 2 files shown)


ssbarnea:
> Tests needed.

Done.

test/TestNullTasks.py
```diff
@@ -0,0 +1,10 @@
+SUCCESS_PLAYBOOK = '''
```
webknjaz, line 1: Needs a docstring

test/TestNullTasks.py
```diff
@@ -0,0 +1,10 @@
...
+- hosts: all
+  tasks:
+'''
+
+
+def test_null_tasks(default_text_runner):
+    """Assure we do not fail when encountering null tasks."""
+    results = default_text_runner.run_playbook(SUCCESS_PLAYBOOK)
+    assert len(results) == 0
```
webknjaz, line 10:
```suggestion
    assert not results
```

test/TestNullTasks.py
ssbarnea: @webknjaz Are you joking? None of the constants used in our tests have docstrings attached to them and there is no need to have any. That kind of nitpicking assures that we fix bugs slower, delay releases and discourage contributions.

webknjaz: I'm talking about the module docstrings

test/TestNullTasks.py
```diff
@@ -0,0 +1,8 @@
+def test_null_tasks(default_text_runner):
```
webknjaz, line 1: Module docstring plz

webknjaz: After https://github.com/ansible/ansible-lint/pull/900 it'll be linted automatically

PR #902, opened 2020-07-20
Guard target tox interpreter presence in GHA

webknjaz:


.github/workflows/tox.yml
```diff
@@ -65,8 +65,14 @@ jobs:
...
         Initialize tox envs
-      run: |
-        python -m tox --parallel auto --parallel-live --notest -vv
+      run: >-
+        python -m
+        tox
+        --parallel auto
+        --parallel-live
+        --notest
+        --skip-missing-interpreters false
```
webknjaz, line 74: Note: this is a technique that tox itself employs in their CI. I guess I forgot to transfer it here earlier.

webknjaz:
No, it shouldn't be in tox.ini. It should be explicitly set in each CI config. Local tox runs mustn't be affected.

ssbarnea:
> No, it shouldn't be in tox.ini. It should be explicitly set in each CI config. Local tox runs mustn't be affected.

Until a couple of months ago I had the same opinion as you regarding the skip setting, that is better to be relaxed. Still, after problems I seen with openstack projects and discussions with other folks, I realised that it was a too dangerous option to have. Maybe you are not aware of what happens with it when you try to tell tox which python interpreter to use (--discover). After that I told myself that I will never enable skip again.

I will fight for this one because I not not want a guard specific for GHA, I want a generic one. Also I happen to know how many people where involved in that decision on openstack projects. If you check the link you will see that >100 repositories have this in. There are only ~10 having the opposite and the only reason why they have it is because nobody bothered to fix it these (not being mainstream repos).

webknjaz:
This is not OpenStack. And I'd prefer to follow the trusted example of tox itself. Besides, there's no conversation about such a decision in OpenStack linked so let's not speculate on it.

.github/workflows/tox.yml
ssbarnea: I still consider that we need to set skip-missing-interpreters=False inside tox.ini but I do not want to delay this even further as nobody else cared to review it. I will merge as it but I hope you will support a change on tox.ini when zuul will fail to use right python version.

webknjaz: I cannot support it without seeing any evidence

PR #907, opened 2020-07-21
Make newlines after imports  consistent

ssbarnea:
Our current code-base uses a random mix of 1-2 newlines after imports and before variable assignment.

This change **normalize** number of newlines after imports to 1 newline
as this seems to be the default setting current preferred by tools
like isort/black/pylint.

Included changes are result of cherry picking newlines changes from running isort from https://github.com/ansible/ansible-lint/pull/887 so not much credit to claim on my side. Still, is needed in order to cleanup isort enable itself.

isort [documentation](https://github.com/timothycrosley/isort/wiki/isort-Settings) seems to clear about the rule, and user does not even need to remember it as isort will reformat when needed:

> lines_after_imports: Forces a certain number of lines after the imports and before the first line of functional code. By default this is 2 lines if the first line of code is a class or function. Otherwise it's 1.

Do not confuse newlines after import with newlines before function or class definition, which remains 2. When there are no variables-section the "2" value takes precedence, which seems natural.

Note: if someone would try to implement the opposite and force "2" newlines
after imports, we would have to touch ~40 files instead of 20, and also diverge from default settings, requiring us to add custom options to isort.

### DIFF
--- a/docs/rules_table_generator_ext.py
+++ b/docs/rules_table_generator_ext.py
@@ -13,7 +13,6 @@ from ansiblelint.constants import DEFAULT_RULESDIR
 from ansiblelint.generate_docs import rules_as_rst
 from ansiblelint.rules import RulesCollection
 
-
 DEFAULT_RULES_RST = (Path(__file__).parent / 'default_rules.rst').resolve()
 
 
--- a/lib/ansiblelint/__init__.py
+++ b/lib/ansiblelint/__init__.py
@@ -22,7 +22,6 @@
 from ansiblelint.rules import AnsibleLintRule
 from ansiblelint.version import __version__
 
-
 __all__ = (
     "__version__",
     "AnsibleLintRule"  # deprecated, import it directly from rules
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -33,7 +33,6 @@ from ansiblelint.rules import RulesCollection
 from ansiblelint.runner import Runner
 from ansiblelint.utils import get_playbooks_and_roles
 
-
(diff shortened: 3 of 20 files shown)


ssbarnea:
> I don't like this sort of inconsistency because it does not allow to easily guess what style to use when adding something new. I'd like 2-line default.

Please step back and remember that reviews are not about what *you* want or prefer as an individual. The idea is to adopt what most developers use/want. We have to avoid subjectivity as much as possible.

I often accept that consistency is more important that (my) personal preference. That is why he have standards and not "custom preferences".

webknjaz:
But this is exactly about consistency. I'd argue that two or one lines after imports depending on some condition is inconsistent.

awcrosby:
I think adopting black would be preferable... and thinking about contributors, having 1 or 2 lines based on a condition they may or may not remember will make it less easy.

webknjaz:
> I think adopting black would be preferable

I'd rather not have to fight all the ugliness it brings...

awcrosby:
well regarding this, im not sure this change is necessary (if we adopt an auto-formatter elsewhere it can do these particular changes per config)

ssbarnea:
@awcrosby I missed to mention that this change is a snip-off from https://github.com/ansible/ansible-lint/pull/887 -- introduction of isort. We all want to introduce isort, but we must decide between letting it do its default behavior or forcing it not stick to a different settings.

Updating newlines is mandatory before enabling isort as we do not want to add noise to that change.

ssbarnea:
Closing as we apparently agreed to use default settings for newlines so this will be part of https://github.com/ansible/ansible-lint/pull/887

Issue #829, opened 2020-06-15
proposal: use release-drafter to build draft release notes


ssbarnea:
##### Summary
I would like to propose use of two github apps [release-drafter](https://github.com/marketplace/actions/release-drafter) which can help a lot with maintenance chores.

Please note that this is not a request to change the release process, only to enable the tools that help us "draft" the release notes, mainly by keeping the next release notes in sync with recent merges, w/o asking for contributors to do anything.

There are multiple projects in Ansible using it, notable example https://github.com/ansible-community/molecule/releases -- as seen there the drafter updates the description with information from merged PRs, being able to skip mentioning changes that do not affect the final product, using categories, mentioning contributors.

When the maintainer decides to make a release he can still fine tune the already drafter release notes.

##### Issue Type

- Feature Idea

##### Additional Information

An example config which is used in unmodified form by 10+ projects from ansible-community can be seen at https://github.com/ansible-community/molecule/blob/master/.github/release-drafter.yml -- it is simple and effective. The number of categories is kept low to avoid cluttering with major changes listed on top.

I propose enabling this as experimental without having to update any development/release workflows. Once we have it running for a while we can see if we want to update these (like putting a to github releases pages as a place to read the release notes).

webknjaz:
I don't like relying on PR titles because people don't know how to write commit messages already. It's better to enforce something like towncrier change notes like most of the Python projects do.

ssbarnea:
Implemented.

PR #908, opened 2020-07-22
Run release-drafter as action

ssbarnea:
Run [release-drafter](https://github.com/marketplace/actions/release-drafter) as an action on push instead of having to depend on app version which needs github admin approval.

Workflow was named "push" as we may want to run other activities in the future.

Related: #904
Related: #829

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,19 @@
+name: push
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


gundalow:
I think I've setup the App now

ssbarnea:
Closing as app should do the trick.

PR #911, opened 2020-07-22
Enable github annotations parsable format

ssbarnea:
Detects if is run as a github action/workflow and adapts its output
accordingly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,6 +117,15 @@ The following is the output from ``ansible-lint --help``, providing an overview 
                             ".ansible-lint"
       --version             show program's version number and exit
 
+CI/CD
+-----
+
+If executed under `Github Actions`_ is detected via the presence of
+``GITHUB_ACTIONS=true`` and ``GITHUB_WORFLOW=...`` variables, the linter will
+also print errors using their `annotation`_ format.
+
+.. _GitHub Actions: https://github.com/features/actions
+.. _annotation: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message
 
 Linting Playbooks and Roles
 ---------------------------
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -22,6 +22,7 @@
 
 import errno
 import logging
+import os
 import pathlib
 import sys
 from typing import Any, Set
@@ -114,8 +115,17 @@ def main() -> int:
                         options.verbosity, checked_files)
         matches.extend(runner.run())
 
-    for match in sorted(set(matches)):
+    # Assure we do not print duplicates and the order is consistent
+    matches = sorted(set(matches))
+
+    for match in matches:
         print(formatter.format(match, options.colored))
+
+    # If run under GitHub Actions we also want to emit output recognized by it.
+    if os.getenv('GITHUB_ACTIONS') == 'true' and os.getenv('GITHUB_WORKFLOW'):
+        formatter = formatters.AnnotationsFormatter(cwd, True)
+        for match in matches:
+            print(formatter.format(match))
 
     if matches:
         return 2
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -19,7 +19,7 @@ class BaseFormatter:
         display_relative_path (bool): whether to show path as relative or absolute
     """
 
-    def __init__(self, base_dir: Union[str, Path], display_relative_path: str) -> None:
+    def __init__(self, base_dir: Union[str, Path], display_relative_path: bool) -> None:
         """Initialize a BaseFormatter instance."""
         if isinstance(base_dir, str):
             base_dir = Path(base_dir)
@@ -92,6 +92,46 @@ class ParseableFormatter(BaseFormatter):
                                     match.linenumber,
                                     "E" + match.rule.id,
                                     match.message)
+
+
+class AnnotationsFormatter(BaseFormatter):
+    # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message
+    """Formatter for emitting violations as GitHub Workflow Commands.
+
+    These commands trigger the GHA Workflow runners platform to post violations
+    in a form of GitHub Checks API annotations that appear rendered in pull-
+    request files view.
+
+    ::debug file={name},line={line},col={col}::{message}
+    ::warning file={name},line={line},col={col}::{message}
+    ::error file={name},line={line},col={col}::{message}
+
+    Supported levels: debug, warning, error
+    """
+
+    def format(self, match, colored=False) -> str:
+        """Prepare a match instance for reporting as a GitHub Actions annotation."""
+        if colored:
+            raise ValueError('The colored mode is not supported.')
+
+        level = self._severity_to_level(match.rule.severity)
+        file_path = self._format_path(match.filename)
+        line_num = match.linenumber
+        rule_id = match.rule.id
+        violation_details = match.message
+        return (
+            f"::{level} file={file_path},line={line_num}"
+            f"::[E{rule_id}] {violation_details}"
+        )
+
+    @staticmethod
+    def _severity_to_level(severity: str) -> str:
+        if severity in ['VERY_LOW', 'LOW']:
+            return 'warning'
+        elif severity in ['INFO']:
+            return 'debug'
+        # ['MEDIUM', 'HIGH', 'VERY_HIGH'] or anything else
+        return 'error'
 
 
 class ParseableSeverityFormatter(BaseFormatter):
--- a/test/TestCliRolePaths.py
+++ b/test/TestCliRolePaths.py
@@ -82,3 +82,21 @@ class TestCliRolePaths(unittest.TestCase):
 
         result = run_ansible_lint(cwd=cwd, role_path=role_path)
         assert '106 Role name invalid-name does not match' in str(result)
+
+
+def test_run_playbook_github():
+    """Call ansible-lint simulating GitHub Actions environment."""
+    cwd = os.path.dirname(os.path.realpath(__file__))
+    role_path = '../examples/example.yml'
+
+    env = {
+        "GITHUB_ACTIONS": "true",
+        "GITHUB_WORKFLOW": "foo"
+    }
+
+    result = run_ansible_lint(cwd=cwd, role_path=role_path, env=env)
+    expected = (
+        '::error file=../examples/example.yml,line=47::[E101] '
+        'Deprecated always_run'
+    )
+    assert expected in str(result)


lib/ansiblelint/formatters/__init__.py
```diff
@@ -87,6 +87,26 @@ def format(self, match, colored=False):
...
+    ::warning file={name},line={line},col={col}::{message}
+    """
+
+    def format(self, match, colored=False):
+        formatstr = u"::warning file={0},line={1}::[{2}] {3}"
+        if colored:
+            return formatstr.format(colorize(self._format_path(match.filename), Color.filename),
+                                    colorize(str(match.linenumber), Color.linenumber),
+                                    colorize(u"E{0}".format(match.rule.id), Color.error_code),
+                                    colorize(u"{0}".format(match.message), Color.error_title))
```
webknjaz, line 102: I'm pretty sure that colorizing will break this feature.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -87,6 +87,26 @@ def format(self, match, colored=False):
                                     match.message)
 
 
+class AnnotationsFormatter(BaseFormatter):
+    """Emits errors in format that is recognized by Github Annotation parsing.
+
+    ::warning file={name},line={line},col={col}::{message}
+    """
+
+    def format(self, match, colored=False):
```
webknjaz, line 96: This should probably print both to console and to workflow commands. Because the log must still remain readable.

lib/ansiblelint/__main__.py
```diff
@@ -71,7 +72,10 @@ def main() -> int:
         formatter_factory = formatters.QuietFormatter
 
     if options.parseable:
-        formatter_factory = formatters.ParseableFormatter
+        if os.environ.get('GITHUB_WORKFLOW'):
```
webknjaz, line 75:
```suggestion
        if os.getenv('GITHUB_WORKFLOW'):
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -87,6 +87,20 @@ def format(self, match, colored=False):
...
+        formatstr = u"::warning file={0},line={1}::[{2}] {3}"
```
webknjaz, line 97: I think this shouldn't be only `warning`. Can we use `error` based on severity?

README.rst
```diff
@@ -117,6 +117,15 @@ The following is the output from ``ansible-lint --help``, providing an overview
...
 
+CI/CD
+-----
+
+If execution under `Github Actions`_ is detected via the presence of
+``GITHUB_WORKFLOW`` variable, the linter will also print errors using their
+`annotation`_ format.
+
+.. _GitHub Actions: https://github.com/features/actions
+.. _annotation: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message
```
webknjaz, line 128: Here's how you can match the length of the paragraph above
```suggestion
.. _annotation:
   https://docs.github.com/en/actions/reference
   /workflow-commands-for-github-actions#setting-an-error-message
```

lib/ansiblelint/__main__.py
```diff
@@ -118,6 +119,12 @@ def main() -> int:
     for match in sorted(matches):
         print(formatter.format(match, options.colored))
 
+    # If run under GitHub Actions we also want to emit output recognized by it.
+    if os.getenv('GITHUB_WORKFLOW'):
```
webknjaz, line 123: This doc https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables suggests that `GITHUB_ACTIONS` var will always equal `true` in such envs. So let's rely on this property to avoid false positives in some artificial test envs:
```suggestion
    if os.getenv('GITHUB_ACTIONS') == 'true' and os.getenv('GITHUB_WORKFLOW'):
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -15,7 +15,7 @@ class BaseFormatter:
         display_relative_path (bool): whether to show path as relative or absolute
     """
 
-    def __init__(self, base_dir, display_relative_path):
+    def __init__(self, base_dir=str, display_relative_path=True) -> None:
```
webknjaz, line 18: typo?
```suggestion
    def __init__(self, base_dir: str, display_relative_path=True) -> None:
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -87,6 +87,33 @@ def format(self, match, colored=False):
                                     match.message)
 
 
+class AnnotationsFormatter(BaseFormatter):
+    # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message
+    """Emits errors in format that is recognized by Github Annotation parsing.
+
+    ::warning file={name},line={line},col={col}::{message}
```
webknjaz, line 94: How about mentioning all the possibilities?
```suggestion
    ::debug file={name},line={line},col={col}::{message}
    ::error file={name},line={line},col={col}::{message}
    ::warning file={name},line={line},col={col}::{message}
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -87,6 +87,33 @@ def format(self, match, colored=False):
...
+    def format(self, match, colored=False):
+        formatstr = u"::{0} file={1},line={2}::[{3}] {4}"
+        return formatstr.format(
+            self.severity_to_level(match.rule.severity),
+            self._format_path(match.filename),
+            match.linenumber,
+            "E" + match.rule.id,
+            match.message)
+
+    def severity_to_level(self, severity: str) -> str:
```
webknjaz, line 108: I think this shouldn't be embedded into the class (it does nothing with the instance. But if you want to keep it, turn into a `@staticmethod` at least, and add a leading underscore.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -87,6 +87,33 @@ def format(self, match, colored=False):
...
+    """
+
+    def format(self, match, colored=False):
+        formatstr = u"::{0} file={1},line={2}::[{3}] {4}"
+        return formatstr.format(
+            self.severity_to_level(match.rule.severity),
+            self._format_path(match.filename),
+            match.linenumber,
+            "E" + match.rule.id,
+            match.message)
```
webknjaz, line 106: Let's use named placeholders, numbers in templates are confusing.
```suggestion
        workflow_command = self.severity_to_level(match.rule.severity)
        file_path = self._format_path(match.filename)
        line_num = match.linenumber
        rule_id = match.rule.id
        violation_details = match.message

        return (
            f"::{workflow_command} file={file_path},line={line_num}"
            f"::[E{rule_id}] {violation_details}"
        )
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -87,6 +87,33 @@ def format(self, match, colored=False):
...
+class AnnotationsFormatter(BaseFormatter):
+    # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message
+    """Emits errors in format that is recognized by Github Annotation parsing.
+
+    ::warning file={name},line={line},col={col}::{message}
+
+    Supported levels: debug, warning, error
+    """
+
+    def format(self, match, colored=False):
```
webknjaz, line 99: Let's explicitly state the invocation contract:
```suggestion
    def format(self, match, colored=False):
        """Prepare a match instance for reporting as a GitHub Actions annotation."""
        if colored:
            raise ValueError('The colored mode is not supported.')

```

README.rst
ssbarnea: If our linters jobs is not complaining lets not go into this. URLs are known to cause problems
when wrapped and I do not want to risk.

tox.ini
```diff
@@ -31,6 +31,7 @@ deps =
 commands =
   # safety measure to assure we do not accidentaly run tests with broken dependencies
   python -m pip check
+  -ansible-lint -p examples/example.yml
```
webknjaz, line 34: don't forget to remove this before merging
```suggestion
```

lib/ansiblelint/__main__.py
```diff
@@ -117,6 +118,12 @@ def main() -> int:
     for match in sorted(set(matches)):
         print(formatter.format(match, options.colored))
 
+    # If run under GitHub Actions we also want to emit output recognized by it.
+    if os.getenv('GITHUB_ACTIONS') == 'true' and os.getenv('GITHUB_WORKFLOW'):
+        formatter = formatters.AnnotationsFormatter(cwd, options.display_relative_path)
```
webknjaz, line 123: This should always use a relative path. Otherwise, the annotations won't be mapped to things in the project:
```suggestion
        formatter = formatters.AnnotationsFormatter(cwd, True)
```

tox.ini
```diff
@@ -42,6 +42,7 @@ install_command =
     {packages}
 passenv =
   CURL_CA_BUNDLE  # https proxies, https://github.com/tox-dev/tox/issues/1437
+  GITHUB_*
```
webknjaz, line 45: I don't think this belongs here. We don't really rely on it in tests atm.

webknjaz:
@ssbarnea I'd like some tests for this new formatter. Also, we need tests for the default auto-selection inside and outside of GHA (probably parametrized + `monkeypatch.setenv()`).

ssbarnea:
> @ssbarnea I'd like some tests for this new formatter. Also, we need tests for the default auto-selection inside and outside of GHA (probably parametrized + `monkeypatch.setenv()`).

I added test, so we would also cover this condition.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -15,7 +16,7 @@ class BaseFormatter:
         display_relative_path (bool): whether to show path as relative or absolute
     """
 
-    def __init__(self, base_dir, display_relative_path):
+    def __init__(self, base_dir: Union[str, Path], display_relative_path=True) -> None:
```
webknjaz, line 19: Why does this need to default to `True`?

lib/ansiblelint/__main__.py
```diff
@@ -117,6 +118,12 @@ def main() -> int:
     for match in sorted(set(matches)):
```
webknjaz, line 118: @ssbarnea this is `sorted(set())` and the one below is just `sorted()`.
1) I think that both should be `sorted(set())`
2) Also, it'd be better to just safe this in a variable and reuse it in two places

lib/ansiblelint/formatters/__init__.py
```diff
@@ -90,6 +91,42 @@ def format(self, match, colored=False):
                                     match.message)
 
 
+class AnnotationsFormatter(BaseFormatter):
+    # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message
+    """Emits errors in format that is recognized by Github Annotation parsing.
```
webknjaz, line 96: This mustn't start with a verb because classes represent objects and don't do anything by themselves, the actions are in methods and methods should have imperative strings.

```suggestion
    """Formatter for emitting violations as GitHub Workflow Commands.

    These commands trigger the GHA Workflow runners platform to post violations
    in a form of GitHub Checks API annotations that appear rendered in pull-
    request files view.
```

test/conftest.py
```diff
@@ -19,4 +21,19 @@ def default_text_runner(default_rules_collection):
     return RunFromText(default_rules_collection)
 
 
-# vim: et:sw=4:syntax=python:ts=4:
```
webknjaz, line 22: Keep this.

test/conftest.py
```diff
@@ -19,4 +21,19 @@ def default_text_runner(default_rules_collection):
...
+def run_ansible_lint(cwd, role_path=None, bin=None, env=None):
```
webknjaz, line 24: Conftest is not the best location for such helpers. It should have pytest-related things like fixtures and hooks. Put this into `__init__.py` along with other helpers instead of having them spread across multiple locations w/o any structure.

Also, this change is unrelated to GHA annotations and must be submitted separately.

test/TestCliRolePaths.py
```diff
@@ -90,12 +73,30 @@ def test_run_playbook(self):
...
+
+def test_run_playbook_github():
+    """Call ansible-lint simulating GitHub Actions environment."""
+    cwd = os.path.dirname(os.path.realpath(__file__))
+    role_path = '../examples/example.yml'
+
+    env = {
+        "GITHUB_ACTIONS": "true",
+        "GITHUB_WORKFLOW": "foo"
+    }
```
webknjaz, line 95: We need a negative test where these values don't match the condition. You could use parameterized to keep them all in one test function.

test/TestCliRolePaths.py
```diff
@@ -90,12 +73,30 @@ def test_run_playbook(self):
...
+
+    result = run_ansible_lint(cwd=cwd, role_path=role_path, env=env)
+    expected = (
+        '::error file=../examples/example.yml,line=47::[E101] '
```
webknjaz, line 99: It is bad if the formatter really emits paths like this, with `../`. They should be resolved and relative to the git repo root. Otherwise, GHA won't match the annotations with files in the repo.

ssbarnea: Not really true. By design ansilble lint emits relative paths relative to CWD when command was called. Same behavior can be seen from other linters too. The "git" repo part is defined for all use-cases.

This path is like this because we call it from inside `./test`, but we can easily avoid it. Still, is not a bug.

webknjaz: It is a bug because it breaks the main purpose of the feature.

ssbarnea:
Closing in favor of #926 but I will address the test comment.

PR #912, opened 2020-07-23
Assure we do not produce duplicated matches

ssbarnea:
Due to the way we parse files it is possible to get duplicated matches.

This change avoids printing duplications and also assures no regressions are made.

Our `examples/example.yml` was not used during testing and was a good example that caused duplicates. Now we test it and assure we find the exact number of matches that we were expecting.

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -115,7 +115,7 @@ def main() -> int:
                         options.verbosity, checked_files)
         matches.extend(runner.run())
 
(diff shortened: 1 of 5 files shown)


ssbarnea:
> This is missing tests for cases with MatchError being the same except for different `MatchError.line`. Per [#881 (files)](https://github.com/ansible/ansible-lint/pull/881/files#r451486472).
> 
> I think this is introducing a regression without that.

Refactoring to details removed this issue. Now details matters and different value will be counted.

PR #914, opened 2020-07-24
Add linter branding for docs

ssbarnea:
- Add linter logo and favicon that renders ok with light/dark themes
- Corrects documentation title


ssbarnea:
Note to reviewers, client the docs job and look at browser favicon, you should see a gray shield with Ansible "A" on top of it, a design that works with both day/night themes.

webknjaz:
@ssbarnea why isn't `logo_invert.png` SVG?

ssbarnea:
> @ssbarnea why isn't `logo_invert.png` SVG?

Because filename is hardcoded in the theme. We can look into making it configurable in the future.

webknjaz:
> > @ssbarnea why isn't `logo_invert.png` SVG?
> 
> Because filename is hardcoded in the theme. We can look into making it configurable in the future.

Then, I'd postpone it until we do so. Adding+removing binaries from Git make the repo bigger unnecessarily.

ssbarnea:
> > > @ssbarnea why isn't `logo_invert.png` SVG?
> > 
> > 
> > Because filename is hardcoded in the theme. We can look into making it configurable in the future.
> 
> Then, I'd postpone it until we do so. Adding+removing binaries from Git make the repo bigger unnecessarily.

We are talking about a 5kb image, not 5mb one. It is smaller than most of our source codes. If would have being >200kb it would have made sense.

Issue #893, opened 2020-07-13
adopt full type checking


ssbarnea:
While we recently introduced mypy, only a small amount of code has type checks and I am already aware of several places where this could have uncovered real bugs.

I will assign  this to me as I plan to address this soon, as it blocks other changes which require touching code w/ and w/o type checks.

PR #915, opened 2020-07-24
Added more type information

ssbarnea:
Adds type information to few more functions. Moves `FileType` to constants in order to be able to re-use in other places.

This change has no runtime effects, it only improves typing coverage.

Partial: #893

### DIFF
--- a/lib/ansiblelint/constants.py
+++ b/lib/ansiblelint/constants.py
@@ -1,7 +1,17 @@
 """Constants used by AnsibleLint."""
 import os.path
+import sys
+
+# mypy needs this special condition or it may fail to run
+# https://github.com/python/typeshed/issues/3500#issuecomment-560958608
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 DEFAULT_RULESDIR = os.path.join(os.path.dirname(__file__), 'rules')
 
 INVALID_CONFIG_RC = 2
 ANSIBLE_FAILURE_RC = 3
+
+FileType = Literal["playbook", "pre_tasks", "post_tasks"]
--- a/lib/ansiblelint/constants.py
+++ b/lib/ansiblelint/constants.py
@@ -2,10 +2,10 @@
 import os.path
 import sys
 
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/constants.py
```diff
@@ -1,7 +1,15 @@
 """Constants used by AnsibleLint."""
 import os.path
+import sys
+
+if sys.version_info >= (3, 8):
+    from typing import Literal  # pylint: disable-msg=E0611
```
webknjaz, line 6: plz use readable message identifiers instead of some numbers

webknjaz: also there's no `disable-msg` in docs, only `disable`

lib/ansiblelint/constants.py
```diff
@@ -1,7 +1,15 @@
...
+else:
+    from typing_extensions import Literal
```
webknjaz, line 8: The proper fallback idiom is
```suggestion

try:
    from typing import Literal  # pylint: disable-msg=E0611
except ImportError:
    from typing_extensions import Literal
```

ssbarnea: Have you tried it locally? I suspect your approach may upset one of our linters. I am used to the same idiom, and if I remember well, there was an issue that forced me to do it the ugly way.

lib/ansiblelint/constants.py
```diff
@@ -1,7 +1,17 @@
 """Constants used by AnsibleLint."""
 import os.path
+import sys
+
+# mypy needs this special condition or it may fail to run
+# https://github.com/python/typeshed/issues/3500#issuecomment-560958608
```
webknjaz, line 6: This must mention py3.6 specifically because the link only shows the solution and not motivation.

ssbarnea: Maybe you want me to also include instructions about pressing the left mouse button over the link and reading the entire page, maybe the reader will not know that to do?!

webknjaz: link != motivation

PR #916, opened 2020-07-24
Added more type information (v2)

ssbarnea:
Added more type information, covering formatters and skip_utils.

Partial: #893

### DIFF
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -1,8 +1,10 @@
 """Output formatters."""
 import os
 from pathlib import Path
+from typing import Union
 
 from ansiblelint.color import Color, colorize
+from ansiblelint.errors import MatchError
 
 
 class BaseFormatter:
@@ -15,7 +17,7 @@ class BaseFormatter:
         display_relative_path (bool): whether to show path as relative or absolute
     """
 
-    def __init__(self, base_dir, display_relative_path):
+    def __init__(self, base_dir: Union[str, Path], display_relative_path: str) -> None:
         """Initialize a BaseFormatter instance."""
         if isinstance(base_dir, str):
             base_dir = Path(base_dir)
@@ -28,7 +30,7 @@ class BaseFormatter:
 
         self._base_dir = base_dir if display_relative_path else None
 
-    def _format_path(self, path):
+    def _format_path(self, path: Union[str, Path]) -> str:
         # Required 'cause os.path.relpath() does not accept Path before 3.6
         if isinstance(path, Path):
             path = str(path)  # Drop when Python 3.5 is no longer supported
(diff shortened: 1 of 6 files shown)


pabelanger:
recheck

pabelanger:
recheck

lib/ansiblelint/formatters/__init__.py
```diff
@@ -41,69 +43,72 @@ def _format_path(self, path):
...
-                                    colorize(match.message, Color.error_title),
-                                    colorize(self._format_path(match.filename), Color.filename),
-                                    colorize(str(match.linenumber), Color.linenumber),
-                                    colorize(u"{0}".format(match.details), Color.line))
+            return formatstr.format(
+                colorize(u"[{0}]".format(_id), Color.error_code),
+                colorize(match.message, Color.error_title),
+                colorize(self._format_path(match.filename or ""), Color.filename),
+                colorize(str(match.linenumber), Color.linenumber),
+                colorize(u"{0}".format(match.details), Color.line))
```
webknjaz, line 55: looks unrelated

lib/ansiblelint/formatters/__init__.py
```diff
@@ -41,69 +43,72 @@ def _format_path(self, path):
...
+    def format(self, match: MatchError, colored: bool = False) -> str:
         formatstr = u"{0} {1}:{2}"
         if colored:
-            return formatstr.format(colorize(u"[{0}]".format(match.rule.id), Color.error_code),
-                                    colorize(self._format_path(match.filename), Color.filename),
-                                    colorize(str(match.linenumber), Color.linenumber))
+            return formatstr.format(
+                colorize(u"[{0}]".format(match.rule.id), Color.error_code),
+                colorize(self._format_path(match.filename or ""), Color.filename),
+                colorize(str(match.linenumber), Color.linenumber))
```
webknjaz, line 72: unrelated?

lib/ansiblelint/formatters/__init__.py
```diff
@@ -41,69 +43,72 @@ def _format_path(self, path):
...
         if colored:
-            return formatstr.format(colorize(self._format_path(match.filename), Color.filename),
-                                    colorize(str(match.linenumber), Color.linenumber),
-                                    colorize(u"E{0}".format(match.rule.id), Color.error_code),
-                                    colorize(u"{0}".format(match.message), Color.error_title))
+            return formatstr.format(
+                colorize(self._format_path(match.filename or ""), Color.filename),
+                colorize(str(match.linenumber), Color.linenumber),
+                colorize(u"E{0}".format(match.rule.id), Color.error_code),
+                colorize(u"{0}".format(match.message), Color.error_title))
```
webknjaz, line 87: unrelated?

lib/ansiblelint/formatters/__init__.py
ssbarnea: **** it is related! If you would try to add some typing information you will see it. Typing complains about not allowing None, so you add `or ""` part and the line becomes too long, so you need to wrap it to pass the linters job.

Also, colorize requires color Literals, not strings but we forgot to update some calls when we did it. Now mypy does slap us, forcing to fix the bug.

You will see more of these. That is why I said is better to add typing info in chunks that keep the review size decent (ideally under 50 lines).

lib/ansiblelint/formatters/__init__.py
ssbarnea: ditto, not trying to slip anything not required.

lib/ansiblelint/formatters/__init__.py
ssbarnea: ditto, not trying to slip anything not required.

lib/ansiblelint/formatters/__init__.py
webknjaz: I guess all this formatting change makes it look like a formatting change. You shouldn't force reviewers to force to compare the diff byte-by-byte hoping to find something legit in the middle for the reformatting.

PR #917, opened 2020-07-24
Only render deps tree in sphinx builds

webknjaz:
This change attempts to make the contributing file well-rendered
on GitHub while injecting a deps list on the corresponding RTD page.


ssbarnea:
Did we ever had a requirement to render doc pages on github? AFAIK, rst rendering in github is shaking anyway and especially when it comes to sphinx, it will never be able to render docs well in general, as they can easily use features needing compilation.

The title of the PR is at least misleading as basically it converts a MD file to RST without mentioning that it does that.

I am not sure that we really have to fix the fact that dependency graph renders as an empty block on github at https://github.com/ansible/ansible-lint/blob/master/docs/CONTRIBUTING.md

The only benefit it would be that we would have a single file format for the entire docs, but considering that CommonMark(MD) is the official docs format for collections and that it seems to be far more popular and easier to use that RST, i would wonder if it would not make sense to do the opposite direction and slowly migrate to MD only, obviously keeping Sphinx.

I would really like to get some feedback from the docs team on that aspect. Having some insights from them and the longer term plans could make us avoid some rework.

webknjaz:
@ssbarnea it's not about the sphinx docs, it's about the other file that GH links in the UI and points users to

webknjaz:
It's those bits that GH actively links that should not contain sphinx-specific directives, but the contents of `docs/` is fine.

.github/CONTRIBUTING.rst
```diff
@@ -0,0 +1,79 @@
+Contributing to Ansible-lint
+============================
+
+To contribute to ansible-lint, please use pull requests on a branch
```
bmillemathias, line 4: To contribute code

PR #918, opened 2020-07-25
Avoid displaying Null with missing filenames

ssbarnea:
Consider Null filename the same as empty, as some errors may not be assigned to specific files.

### DIFF
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -45,15 +45,16 @@ class Formatter(BaseFormatter):
         formatstr = u"{0} {1}\n{2}:{3}\n{4}\n"
         _id = getattr(match.rule, 'id', '000')
         if colored:
-            return formatstr.format(colorize(u"[{0}]".format(_id), Color.error_code),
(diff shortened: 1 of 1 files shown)


lib/ansiblelint/formatters/__init__.py
```diff
@@ -45,15 +45,16 @@ def format(self, match, colored=False):
         formatstr = u"{0} {1}\n{2}:{3}\n{4}\n"
         _id = getattr(match.rule, 'id', '000')
         if colored:
-            return formatstr.format(colorize(u"[{0}]".format(_id), Color.error_code),
-                                    colorize(match.message, Color.error_title),
-                                    colorize(self._format_path(match.filename), Color.filename),
-                                    colorize(str(match.linenumber), Color.linenumber),
-                                    colorize(u"{0}".format(match.details), Color.line))
+            return formatstr.format(
```
webknjaz, line 48: Could you please not modify unrelated lines?

ssbarnea: That is not an unrelated line, adding ` or ""` makes the entire line > 100 which requires reformatting of the entire block to keep each line under 100.

PR #921, opened 2020-07-26
Add more type information

ssbarnea:
Partial: #893

### DIFF
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -1,8 +1,12 @@
 """Output formatters."""
 import os
 from pathlib import Path
+from typing import TYPE_CHECKING, Union
 
 from ansiblelint.color import Color, colorize
+
+if TYPE_CHECKING:
+    from ansiblelint.errors import MatchError
 
 
 class BaseFormatter:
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,8 +1,10 @@
 """Output formatters."""
 import os
 from pathlib import Path
+from typing import Union
 
 from ansiblelint.color import Color, colorize
+from ansiblelint.errors import MatchError
```
webknjaz, line 7: Since it's not necessary in runtime:
```suggestion
if TYPE_CHECKING:
    from ansiblelint.errors import MatchError
```

lib/ansiblelint/runner.py
```diff
@@ -1,11 +1,12 @@
 """Runner implementation."""
 import logging
 import os
-from typing import List, Set
+from typing import Any, FrozenSet, List, Set
 
 import ansiblelint.file_utils
 import ansiblelint.skip_utils
 import ansiblelint.utils
+from ansiblelint.rules import RulesCollection
```
webknjaz, line 9: This should be under type checking too

lib/ansiblelint/runner.py
```diff
@@ -1,23 +1,33 @@
...
 
 
 class Runner(object):
     """Runner class performs the linting process."""
 
-    def __init__(self, rules, playbook, tags, skip_list, exclude_paths,
-                 verbosity=0, checked_files=None) -> None:
+    def __init__(
+            self,
+            rules: "RulesCollection",
```
webknjaz, line 24: I'm pretty sure that stringifying the type is not necessary, is it?

ssbarnea: It is, pylint would complain if not. Is not the mypy itself but pylint needs it.

PR #923, opened 2020-07-26
Enforce type on AnsibleLintRule.match method

ssbarnea:
Adds type information on match() method.

Partial: #893

### DIFF
--- a/lib/ansiblelint/file_utils.py
+++ b/lib/ansiblelint/file_utils.py
@@ -1,8 +1,20 @@
 """Utility functions related to file operations."""
 import os
+import sys
+from typing import TYPE_CHECKING, Optional, Union
+
+from ansiblelint.constants import FileType
+
+if TYPE_CHECKING:
+    from pathlib import Path
+# idiom: https://github.com/python/typeshed/issues/3500#issuecomment-560958608
+if sys.version_info >= (3, 8):
+    from typing import TypedDict  # pylint: disable=no-name-in-module,ungrouped-imports
+else:
+    from typing_extensions import TypedDict
 
 
-def normpath(path) -> str:
+def normpath(path: Union[str, "Path"]) -> str:
     """
     Normalize a path in order to provide a more consistent output.
 
@@ -11,3 +23,10 @@ def normpath(path) -> str:
     """
     # convertion to string in order to allow receiving non string objects
     return os.path.relpath(str(path))
+
+
+class TargetFile(TypedDict):
+    """TargetFile identifies a file or path to be linted."""
+
+    path: str
+    type: Optional[FileType]
--- a/lib/ansiblelint/rules/ComparisonToEmptyStringRule.py
+++ b/lib/ansiblelint/rules/ComparisonToEmptyStringRule.py
@@ -2,8 +2,13 @@
 # Copyright (c) 2018, Ansible Project
 
 import re
+from typing import TYPE_CHECKING, List
 
+from ansiblelint.errors import MatchError
 from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import TargetFile
 
 
 class ComparisonToEmptyStringRule(AnsibleLintRule):
@@ -19,5 +24,8 @@ class ComparisonToEmptyStringRule(AnsibleLintRule):
 
     empty_string_compare = re.compile("[=!]= ?(\"{2}|'{2})")
 
-    def match(self, file, line):
-        return self.empty_string_compare.search(line)
+    def match(self, file: "TargetFile", line: str = "") -> List["MatchError"]:
+        result = []
+        if self.empty_string_compare.search(line):
+            result.append(MatchError(filename=file['path'], details=line, rule=self))
+        return result
--- a/lib/ansiblelint/rules/ComparisonToLiteralBoolRule.py
+++ b/lib/ansiblelint/rules/ComparisonToLiteralBoolRule.py
@@ -2,8 +2,13 @@
 # Copyright (c) 2018, Ansible Project
 
 import re
+from typing import TYPE_CHECKING, List
 
+from ansiblelint.errors import MatchError
 from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import TargetFile
 
 
 class ComparisonToLiteralBoolRule(AnsibleLintRule):
@@ -19,5 +24,7 @@ class ComparisonToLiteralBoolRule(AnsibleLintRule):
 
     literal_bool_compare = re.compile("[=!]= ?(True|true|False|false)")
 
-    def match(self, file, line):
-        return self.literal_bool_compare.search(line)
+    def match(self, file: "TargetFile", line: str = "") -> List["MatchError"]:
+        if self.literal_bool_compare.search(line):
+            return [MatchError(rule=self)]
+        return []
--- a/lib/ansiblelint/rules/LineTooLongRule.py
+++ b/lib/ansiblelint/rules/LineTooLongRule.py
@@ -1,7 +1,12 @@
 # Copyright (c) 2016, Will Thames and contributors
 # Copyright (c) 2018, Ansible Project
+from typing import TYPE_CHECKING, List
 
+from ansiblelint.errors import MatchError
 from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import TargetFile
 
 
 class LineTooLongRule(AnsibleLintRule):
@@ -15,5 +20,10 @@ class LineTooLongRule(AnsibleLintRule):
     tags = ['formatting']
     version_added = 'v4.0.0'
 
-    def match(self, file, line):
-        return len(line) > 160
+    def match(self, file: "TargetFile", line: str = "") -> List["MatchError"]:
+        if len(line) > 160:
+            return [MatchError(
+                filename=file['path'],
+                details=line,
+                rule=self)]
+        return []
--- a/lib/ansiblelint/rules/NoTabsRule.py
+++ b/lib/ansiblelint/rules/NoTabsRule.py
@@ -1,7 +1,12 @@
 # Copyright (c) 2016, Will Thames and contributors
 # Copyright (c) 2018, Ansible Project
+from typing import TYPE_CHECKING, List
 
+from ansiblelint.errors import MatchError
 from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import TargetFile
 
 
 class NoTabsRule(AnsibleLintRule):
@@ -12,5 +17,8 @@ class NoTabsRule(AnsibleLintRule):
     tags = ['formatting']
     version_added = 'v4.0.0'
 
-    def match(self, file, line):
-        return '\t' in line
+    def match(self, file: "TargetFile", line: str = "") -> List["MatchError"]:
+        result = []
+        if '\t' in line:
+            result.append(MatchError(rule=self))
+        return result
--- a/lib/ansiblelint/rules/PlaybookExtension.py
+++ b/lib/ansiblelint/rules/PlaybookExtension.py
@@ -2,9 +2,13 @@
 # Copyright (c) 2018, Ansible Project
 
 import os
-from typing import List
+from typing import TYPE_CHECKING, List
 
+from ansiblelint.errors import MatchError
 from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import TargetFile
 
 
 class PlaybookExtension(AnsibleLintRule):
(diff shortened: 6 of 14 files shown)


lib/ansiblelint/file_utils.py
```diff
@@ -1,8 +1,19 @@
 """Utility functions related to file operations."""
 import os
+import sys
+from pathlib import Path
+from typing import Optional, Union
```
webknjaz, line 5: The idiom for imports that are unnecessary in runtime is conditional imports of those things:
```suggestion
from typing import Optional, TYPE_CHECKING, Union

if TYPE_CHECKING:
    from pathlib import Path
```

lib/ansiblelint/rules/ComparisonToEmptyStringRule.py
```diff
@@ -2,7 +2,9 @@
 # Copyright (c) 2018, Ansible Project
 
 import re
+from typing import Optional, Union
 
+from ansiblelint.file_utils import AFile
 from ansiblelint.rules import AnsibleLintRule
```
webknjaz, line 8: same as above
```suggestion
from typing import Optional, TYPE_CHECKING, Union

from ansiblelint.rules import AnsibleLintRule

if TYPE_CHECKING:
    from ansiblelint.file_utils import AFile
```

lib/ansiblelint/file_utils.py
```diff
@@ -11,3 +22,10 @@ def normpath(path) -> str:
     """
     # convertion to string in order to allow receiving non string objects
     return os.path.relpath(str(path))
+
+
+class AFile(TypedDict):
```
webknjaz, line 27: This name should be improved. Currently "a file" does not communicate the intent.
Maybe
```suggestion
class LintedFile(TypedDict):
```

webknjaz: Or
```suggestion
class TargetFile(TypedDict):
```

lib/ansiblelint/rules/ComparisonToLiteralBoolRule.py
```diff
@@ -2,7 +2,9 @@
 # Copyright (c) 2018, Ansible Project
 
 import re
+from typing import Optional, Union
 
+from ansiblelint.file_utils import AFile
```
webknjaz, line 7: make this conditional too

lib/ansiblelint/rules/LineTooLongRule.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) 2016, Will Thames and contributors
 # Copyright (c) 2018, Ansible Project
+from typing import Optional, Union
 
+from ansiblelint.file_utils import AFile
```
webknjaz, line 5: conditional

lib/ansiblelint/rules/NoTabsRule.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) 2016, Will Thames and contributors
 # Copyright (c) 2018, Ansible Project
+from typing import Optional, Union
 
+from ansiblelint.file_utils import AFile
```
webknjaz, line 5: conditional

lib/ansiblelint/rules/PlaybookExtension.py
```diff
@@ -2,8 +2,9 @@
 # Copyright (c) 2018, Ansible Project
 
 import os
-from typing import List
+from typing import List, Optional, Union
 
+from ansiblelint.file_utils import AFile
```
webknjaz, line 7: conditional

lib/ansiblelint/rules/RoleNames.py
```diff
@@ -20,8 +20,9 @@
 # THE SOFTWARE.
 
 import re
-from typing import List
+from typing import List, Optional, Union
 
+from ansiblelint.file_utils import AFile
```
webknjaz, line 25: conditional

lib/ansiblelint/rules/TaskNoLocalAction.py
```diff
@@ -1,6 +1,9 @@
 # Copyright (c) 2016, Tsukinowa Inc. <info@tsukinowa.jp>
 # Copyright (c) 2018, Ansible Project
 
+from typing import Optional, Union
+
+from ansiblelint.file_utils import AFile
```
webknjaz, line 6: conditional

lib/ansiblelint/rules/TrailingWhitespaceRule.py
```diff
@@ -17,7 +17,9 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
+from typing import Optional, Union
 
+from ansiblelint.file_utils import AFile
```
webknjaz, line 22: conditional

lib/ansiblelint/rules/VariableHasSpacesRule.py
```diff
@@ -1,8 +1,9 @@
 # Copyright (c) 2016, Will Thames and contributors
 # Copyright (c) 2018, Ansible Project
-
 import re
+from typing import Optional, Union
 
+from ansiblelint.file_utils import AFile
```
webknjaz, line 6: conditional

lib/ansiblelint/rules/__init__.py
```diff
@@ -6,10 +6,11 @@
 import re
 from collections import defaultdict
 from importlib.abc import Loader
-from typing import List
+from typing import List, Optional, Union
 
 import ansiblelint.utils
 from ansiblelint.errors import MatchError
+from ansiblelint.file_utils import AFile
```
webknjaz, line 13: conditional

test/rules/EMatcherRule.py
```diff
@@ -8,5 +11,5 @@ class EMatcherRule(AnsibleLintRule):
     shortdesc = 'The letter "e" is present'
     tags = {'fake', 'dummy', 'test1'}
 
-    def match(self, filename, line):
+    def match(self, file: Optional[AFile] = None, line: str = "") -> Union[bool, str]:
```
webknjaz, line 14: conditional

test/rules/UnsetVariableMatcherRule.py
```diff
@@ -8,5 +11,5 @@ class UnsetVariableMatcherRule(AnsibleLintRule):
                   'post templating that still contain {{'
     tags = {'fake', 'dummy', 'test2'}
 
-    def match(self, filename, line):
+    def match(self, file: Optional[AFile] = None, line: str = "") -> Union[bool, str]:
```
webknjaz, line 14: conditional

lib/ansiblelint/rules/VariableHasSpacesRule.py
```diff
@@ -17,8 +20,8 @@ class VariableHasSpacesRule(AnsibleLintRule):
     variable_syntax = re.compile(r"{{.*}}")
     bracket_regex = re.compile(r"{{[^{' -]|[^ '}-]}}")
 
-    def match(self, file, line):
+    def match(self, file: Optional["TargetFile"] = None, line: str = "") -> Union[bool, str]:
```
webknjaz, line 23: looks like this only returns bool

lib/ansiblelint/rules/PlaybookExtension.py
```diff
@@ -2,22 +2,25 @@
...
     severity = 'MEDIUM'
     tags = ['formatting']
-    done = []  # type: List  # already noticed path list
+    done: List[str] = []
     version_added = 'v4.0.0'
 
-    def match(self, file, text):
-        if file['type'] != 'playbook':
+    def match(self, file: Optional["TargetFile"] = None, line: str = "") -> Union[bool, str]:
+        if not file or file['type'] != 'playbook':
```
webknjaz, line 23: This feels like an invasion into the logic that is out of the scope of this PR. Could you show places where `file` is passed as `None`?

lib/ansiblelint/rules/TrailingWhitespaceRule.py
```diff
@@ -29,6 +33,6 @@ class TrailingWhitespaceRule(AnsibleLintRule):
     tags = ['formatting', 'ANSIBLE0002']
     version_added = 'historic'
 
-    def match(self, file, line):
+    def match(self, file: Optional["TargetFile"] = None, line: str = "") -> Union[bool, str]:
```
webknjaz, line 36: I think if the implementation does not actually return strings, it should be just
```suggestion
    def match(self, file: Optional["TargetFile"] = None, line: str = "") -> bool:
```

lib/ansiblelint/rules/__init__.py
```diff
@@ -28,10 +31,12 @@ def verbose(self) -> str:
     tags: List[str] = []
     shortdesc: str = ""
     description: str = ""
-    match = None
     matchtask = None
     matchplay = None
 
+    def match(self, file: Optional["TargetFile"] = None, line: str = "") -> Union[bool, str]:
```
webknjaz, line 37: What are the cases when `file` is passed in as `None`?

lib/ansiblelint/rules/ComparisonToEmptyStringRule.py
```diff
@@ -19,5 +23,5 @@ class ComparisonToEmptyStringRule(AnsibleLintRule):
 
     empty_string_compare = re.compile("[=!]= ?(\"{2}|'{2})")
 
-    def match(self, file, line):
-        return self.empty_string_compare.search(line)
+    def match(self, file: "TargetFile", line: str = "") -> Union[bool, str]:
```
webknjaz, line 26: I think per our earlier discussion, the return value should be bool and any strings would need to be attached to an exception raised. Or alternatively, the return type should be a specialized object.

ssbarnea: I was considering returning `List[MatchError]` but this may count as a potential breaking API change.

Using this makes it future proof by allowing to return multiple matches with different messages.

The breakage risk is low because we happen to control the likely single place where this is used, matchlines() from based class. 

@webknjaz What do you think? -- List[MatchError] is already the return type of matchlines().

webknjaz: Why list? Do you mean that it'd return the same violation on multiple lines? Why not implement a list in the `MatchError` then?

ssbarnea: Errors can be at different lines and if you check which attributes MatchError has you will see that it cannot be used to host multiple messages or multiple lines.

Some rules are more dynamic with their errors, for example galaxy one can identify multiple issues.

ssbarnea:
@albinvass What do you think about the return type of match()?

Now is some kind of `Optional[str, bool]` but without any guarantee that all rule writes did not went more creative on return types.

I did consider `List[MatchError]`, `Generator[MatchError]` or just `Optional[MatchError]` -- this last one is probably the least intrusive to implement mostly because no return still evaluates as false and that a MatchError can also be auto-converted to string.

AFAIK, match is used only on https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/__init__.py#L56 in core (unless custom rules call it again if they override matchlines). 

Are we ever going to have single rule that would return multiple occurrences on a single line? I can think of cases like "space missing before curly braces" as being such an example but that was never supported. 

Apparently we need to make a decision: either keep implementation relatively simple and allow only one rule failure per line or make a little bit bigger change and make it future proof and allow it to raise multiple matches.

albinvass:
I'm not sure if a rule would ever need to return multiple matches instead of splitting the rule into multiple rules. Is anyone requesting that kind of functionality? If not I believe that could probably wait until someone absolutely wants it.
But I agree that returning some kind of object is probably better and more easily handled than returning true/false + error message. Though I'm not sure how many projects are creating their own custom rules and how much it would affect them to do so.
It should be easy to update Zuul Jobs to do that instead at least.

ssbarnea:
> I'm not sure if a rule would ever need to return multiple matches instead of splitting the rule into multiple rules. Is anyone requesting that kind of functionality? If not I believe that could probably wait until someone absolutely wants it.

I can think of an hypothetical example like:

```yaml
# this is a file
name: this is gisdsfds sdsdsds!
```

Assuming that we have `SpellcheckRule`, that could produce
```
examples/example.yml:2:15: [Exxx] Spellcheck failed on gisdsfds
examples/example.yml:2:24: [Exxx] Spellcheck failed on sdsdsds
```

As we do already support column, what are we supposed to do when we have multiple problems on the same line? Should we list only first column? Also each error is of different nature.

There is also the aspect of editors/IDEs, which may want to annotate error, some of them even support start and end column for annotations. Take a look at https://developer.github.com/v3/checks/runs/#annotations-object -- which can have start/end line and start/end column.

For me is much easier to just return a single `MatchError` instead of a `List[MatchError]`, but I am trying to think longer term, especially as breaking API before 4.3.0 goes is released would ok, but later it will have to wait till 4.4.0.

lib/ansiblelint/rules/__init__.py
```diff
@@ -57,15 +62,20 @@ def matchlines(self, file, text) -> List[MatchError]:
             if not result:
                 continue
             message = None
-            if isinstance(result, str):
+            if isinstance(result, list):
+                matches.extend(result)
+            elif isinstance(result, str):
+                # warnings.warn("deprecated", DeprecationWarning)
```
greg-hellings, line 68: Why commented?

ssbarnea:
This got outdated. I will looking into rewriting it from scratch.

Issue #678, opened 2020-02-13
Make it possible to integrate `pytest-xdist`


webknjaz:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

devel/master

# Desired Behaviour

It should be possible to run tests in parallel within pytest using `-n auto` arg when `pytest-xdist` is installed.

# Actual Behaviour (Bug report only)

There's some percentage of the red tests appearing in the parallel mode. I assume that some of the tests use a shared context that causes race conditions. This is what has to be investigated and fixed.

ssbarnea:
Considering that running tests takes ~40s, I doubt that running in parallel would produce any real benefits, maybe quite the opposite from the point of view of how they are displayed.

I propose closing this ticket (cleanup backlog) with the remark that if someone wants to propose it, they are still welcomed.

webknjaz:
Real benefits are:
1) No race conditions and tests depending on some shared state (now they clearly collide)
2) New tests won't slow the whole thing down
3) It's not 40s in GHA. It's 50 + 50 + 95 (because you made test envs sequential). And it's occasionally slower in slow envs like macOS.

PR #925, opened 2020-07-26
Enable xdist when running tests

ssbarnea:
- removes hardocde verbose mode from pytest.ini as it does not
  produce desirable effects with parallel run
- enables "-n auto" mode on CI as we do not want to enforce developers
  to run with parallel mode.

Fixes: #678

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -10,7 +10,6 @@ addopts =
     # --testmon
 
     --durations=10
-    -v
     -ra
     --showlocals
     --doctest-modules
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,8 @@ commands =
   {envpython} -m pytest \
   --cov "{envsitepackagesdir}/ansiblelint" \
   --junitxml "{toxworkdir}/junit.{envname}.xml" \
-  {posargs:}
+  # -n auto kept as default to avoid it when running tox with posargs
+  {posargs:-n auto}
 install_command =
   {envpython} -m \
     pip install \


ssbarnea:
@webknjaz Very weird behavior from Github as it failed twice at the same place: tox init envs on macos. This step should not even be affected by our change, as that one is expected to affect only test runs. Quite weird. And I asure it works on macos as that is my main dev platform.

.github/workflows/tox.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         - TOXENV: build-dists,metadata-validation
     env:
       TOX_PARALLEL_NO_SPINNER: 1
+      PYTEST_ADDOPTS: -n auto
```
webknjaz, line 38: Move it to the pytest config. Everyone needs it

ssbarnea: That is not true, if I put it inside pytest.ini it will be impossible for me to run tests without parallel mode, within or outside tox. Only settings that are really universal should go there, and ability to decide between serial and parallel is not one of them.

If you want, there is another option I used in the past, to add it to tox.ini like:
```
setenv =
    PYTEST_ADDOPTS={env:PYTEST_ADDOPTS:-n auto}
```

This means tox will run with auto unless user defines his own set of pytest preferences, which would override it, allowing a developer to bypass it **without altering tracked files**.

Still, AFAIK, `pytest.ini` settings do not allow much flexibility

webknjaz: > That is not true, if I put it inside pytest.ini it will be impossible for me to run tests without parallel mode, within or outside tox.

This is incorrect. I do it all the time when I need to. `tox` has `{posargs}` set correctly and so it's super easy to run one test with parallelism using the following command:

```console
$ tox -e ansible29-py38 -- test/TestMatchError.py::test_matcherror_compare_invalid -n 0
```

I just copied it from my zsh history, this is what I use very often. The trick is to remember that whatever you pass after `--` gets placed in place of `{posargs}` and since it's in the end of pytest invocation command, anything you pass there will override any previous options, regardless of whether they are passed via CLI, `PYTEST_ADDOPTS` env var or `addopts` config option.

webknjaz: Pytest itself has it set up like this: https://github.com/pytest-dev/pytest/blob/master/tox.ini#L22. Admittedly, their config is more complicated than we need but this is an idiom that I see and rely on probably in every project that I ever saw on GitHub or elsewhere using tox+pytest.

webknjaz: More examples:

- https://github.com/pypa/setuptools/blob/master/tox.ini#L24
- https://github.com/pypa/twine/blob/master/tox.ini#L18
- https://github.com/pypa/virtualenv/blob/master/tox.ini#L37

ssbarnea: Maybe I was not clear: as a developer, I do not want to be forced to add `-n 0` when I run pytest on ansible-lint in order to get serial execution of tests.

I am aware that `pytest {posargs:-n auto}` inside tox.ini is another alternative, one that I used in other projects too.

As we are again blocked around personal preferences, I would like to get another opinion in order to make a decision and move on.

webknjaz: FWIW what you added does not work.

webknjaz: Besides, the point of this is to make developers case about tests not influencing each other's state and clean things up correctly.

ssbarnea:
> Blocking because it does not actually enable xdist

Updated it to avoid enabling it only for github.

webknjaz:
> Updated it to avoid enabling it only for github.

You haven't enabled it on GitHub earlier. You only set an env var that wasn't propagated inside tox and so it was no-op.

pytest.ini
```diff
@@ -10,7 +10,6 @@ addopts =
     # --testmon
 
     --durations=10
-    -v
```
webknjaz, line 13: Revert this. We need to see the details when something breaks.

tox.ini
```diff
@@ -35,7 +35,8 @@ commands =
   {envpython} -m pytest \
   --cov "{envsitepackagesdir}/ansiblelint" \
   --junitxml "{toxworkdir}/junit.{envname}.xml" \
-  {posargs:}
+  # -n auto kept as default to avoid it when running tox with posargs
+  {posargs:-n auto}
```
webknjaz, line 39: I still think it belongs to pytest config.

pytest.ini
ssbarnea: You will see detail if it breaks. "-v with -n auto" produce garbage-like on screen, not something to be trusted.

webknjaz: It fulfills its purpose of showing the progress well enough.

PR #926, opened 2020-07-27
Enable github actions parsable format

ssbarnea:
Detects if is run as a github action/workflow and adapts its output accordingly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,6 +117,15 @@ The following is the output from ``ansible-lint --help``, providing an overview 
                             ".ansible-lint"
       --version             show program's version number and exit
 
+CI/CD
+-----
+
+If executed under `Github Actions`_ is detected via the presence of
+``GITHUB_ACTIONS=true`` and ``GITHUB_WORFLOW=...`` variables, the linter will
+also print errors using their `annotation`_ format.
+
+.. _GitHub Actions: https://github.com/features/actions
+.. _annotation: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message
 
 Linting Playbooks and Roles
 ---------------------------
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -22,6 +22,7 @@
 
 import errno
 import logging
+import os
 import pathlib
 import sys
 from typing import Any, Set
@@ -114,8 +115,17 @@ def main() -> int:
                         options.verbosity, checked_files)
         matches.extend(runner.run())
 
-    for match in sorted(set(matches)):
+    # Assure we do not print duplicates and the order is consistent
+    matches = sorted(set(matches))
+
+    for match in matches:
         print(formatter.format(match, options.colored))
+
+    # If run under GitHub Actions we also want to emit output recognized by it.
+    if os.getenv('GITHUB_ACTIONS') == 'true' and os.getenv('GITHUB_WORKFLOW'):
+        formatter = formatters.AnnotationsFormatter(cwd, True)
+        for match in matches:
+            print(formatter.format(match))
 
     if matches:
         return 2
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -19,7 +19,7 @@ class BaseFormatter:
         display_relative_path (bool): whether to show path as relative or absolute
     """
 
-    def __init__(self, base_dir: Union[str, Path], display_relative_path: str) -> None:
+    def __init__(self, base_dir: Union[str, Path], display_relative_path: bool) -> None:
         """Initialize a BaseFormatter instance."""
         if isinstance(base_dir, str):
             base_dir = Path(base_dir)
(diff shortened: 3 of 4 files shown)


lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,7 +1,6 @@
 """Output formatters."""
 import os
 from pathlib import Path
-from typing import TYPE_CHECKING, Union
```
webknjaz, line 4: is this for test?

ssbarnea: Do https://github.com/ansible/ansible-lint/pull/927 first and after i rebase this one, it will be easy to understand what/why.

test/TestCliRolePaths.py
```diff
@@ -82,3 +82,21 @@ def test_run_role_name_invalid(self):
...
+
+def test_run_playbook_github():
+    """Call ansible-lint simulating GitHub Actions environment."""
+    cwd = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+    role_path = 'examples/example.yml'
+
+    env = {
+        "GITHUB_ACTIONS": "true",
+        "GITHUB_WORKFLOW": "foo"
+    }
```
webknjaz, line 95: https://github.com/ansible/ansible-lint/pull/911#discussion_r460552675

README.rst
```diff
@@ -117,6 +117,15 @@ The following is the output from ``ansible-lint --help``, providing an overview
                             ".ansible-lint"
       --version             show program's version number and exit
 
+CI/CD
+-----
+
+If executed under `Github Actions`_ is detected via the presence of
```
webknjaz, line 123:
```suggestion
If execution under `Github Actions`_ is detected via the presence of
```

README.rst
```diff
@@ -117,6 +117,15 @@ The following is the output from ``ansible-lint --help``, providing an overview
...
+``GITHUB_ACTIONS=true`` and ``GITHUB_WORFLOW=...`` variables, the linter will
+also print errors using their `annotation`_ format.
```
webknjaz, line 125:
```suggestion
also print errors using their `Checks API annotations <annotation_>`_ format.
```

lib/ansiblelint/formatters/__init__.py
```diff
@@ -94,6 +94,46 @@ def format(self, match: "MatchError", colored: bool = False) -> str:
...
+    request files view.
+
+    ::debug file={name},line={line},col={col}::{message}
+    ::warning file={name},line={line},col={col}::{message}
+    ::error file={name},line={line},col={col}::{message}
+
+    Supported levels: debug, warning, error
+    """
+
+    def format(self, match, colored=False) -> str:
```
webknjaz, line 112: Is this right?
```suggestion
    def format(self, match: "MatchError", colored: bool = False) -> str:
```

test/TestCliRolePaths.py
ssbarnea: Updated code to assure these are added only when vars are present.

lib/ansiblelint/formatters/__init__.py
webknjaz: @ssbarnea ^ ignored

test/TestCliRolePaths.py
webknjaz: This still needs negative tests.

test/TestCliRolePaths.py
```diff
@@ -82,3 +82,21 @@ def test_run_role_name_invalid(self):
 
         result = run_ansible_lint(cwd=cwd, role_path=role_path)
         assert '106 Role name invalid-name does not match' in str(result)
+
+
+def test_run_playbook_github():
+    """Call ansible-lint simulating GitHub Actions environment."""
+    cwd = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
webknjaz, line 89: I'd prefer `pathlib` to be used here but it's not a blocker.

lib/ansiblelint/__main__.py
```diff
@@ -114,9 +115,18 @@ def main() -> int:
...
+    matches = sorted(set(matches))
+
+    for match in matches:
         print(formatter.format(match, options.colored))
 
+    # If run under GitHub Actions we also want to emit output recognized by it.
+    if os.getenv('GITHUB_ACTIONS') == 'true' and os.getenv('GITHUB_WORKFLOW'):
+        formatter = formatters.AnnotationsFormatter(cwd, True)
+        for match in matches:
+            print(formatter.format(match))
```
webknjaz, line 128: I just realized that there's no mechanism to disable this. I think it's important to have a CLI flag that would prevent annotations from being produced.

ssbarnea: There is no req to implement flags for this. If someone wants to avoid it, they can easily do `export GITHUB_WORKFLOW=` or `unset GITHUB_WORKFLOW` before calling it. Lets avoid adding feature nobody will ever need.

webknjaz: How is relying on undocumented way of using implementation details is a good thing?

test/TestCliRolePaths.py
ssbarnea: I does have them, I forgot to push the last version.

lib/ansiblelint/__main__.py
ssbarnea: I think that section added to README.rst is more than enough to document the behavior and it does not need to tell user how to avoid passing environment variables. It is outside our scope to teach user about how shell works, or how shell evaluates variables.

test/TestCliRolePaths.py
```diff
@@ -82,3 +83,24 @@ def test_run_role_name_invalid(self):
...
+
+    result_gh = run_ansible_lint(cwd=cwd, role_path=role_path, env=env)
+    result_no_gh = run_ansible_lint(cwd=cwd, role_path=role_path)
+
+    expected = (
+        '::error file=examples/example.yml,line=47::[E101] '
+        'Deprecated always_run'
+    )
+    assert expected in str(result_gh)
+    assert expected not in str(result_no_gh)
```
webknjaz, line 106: This looks like two distinct tests, they should show up in the report separately. You could use parametrize here.

test/TestCliRolePaths.py
```diff
@@ -82,3 +85,24 @@ def test_run_role_name_invalid(self):
...
+    cwd = str(Path(__file__).parent.parent.resolve())
+    role_path = 'examples/example.yml'
+
+    result_gh = run_ansible_lint(cwd=cwd, role_path=role_path, env=env)
+
+    expected = (
+        '::error file=examples/example.yml,line=47::[E101] '
+        'Deprecated always_run'
+    )
+    assert (expected in str(result_gh)) == result
```
webknjaz, line 108: You should use identity check for None/True/False, not equality:
```suggestion
    assert (expected in str(result_gh)) is result
```

PR #927, opened 2020-07-27
Make run_ansible_lint reusable across tests

ssbarnea:
Moves implementation of run_ansible_lint function to root module of
tests in order to make it easy to reuse.

Related: #926

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -2,6 +2,8 @@
 
 import os
 import shutil
+import subprocess
+import sys
 import tempfile
 
 from ansible import __version__ as ansible_version_str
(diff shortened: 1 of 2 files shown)


test/__init__.py
```diff
@@ -49,3 +51,22 @@ def run_role_meta_main(self, meta_main_text):
         results = self._call_runner(role_path)
         shutil.rmtree(role_path)
         return results
+
+
+def run_ansible_lint(cwd, role_path=None, bin=None, env=None):
```
webknjaz, line 56: @ssbarnea anything stopping us from making a fixture out of it?

ssbarnea: Probably not

PR #928, opened 2020-07-27
Enable pytest-xdist

### LINKED ISSUE #678: Make it possible to integrate `pytest-xdist`

webknjaz:
Resolves #678

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,9 +1,7 @@
 [pytest]
 addopts =
     # `pytest-xdist`:
-    # a few tests currently fail with this:
-    # -n auto  # we need to uncomment it once it's fixed
-    # ^ this speeds up the test suite from ~40s to ~10s
+    -n auto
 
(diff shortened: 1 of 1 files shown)


jokajak:
This change is hard to perform a significant review because the previous issue didn't document which tests failed. Given that all tests now succeed it looks good and will help speed up testing.

webknjaz:
@jokajak yeah, but since no tests fail with xdist anymore (neither locally, nor in CI), it is safe to assume that they are already fixed in master and the only thing to do is enable xdist globally to help keep them healthy.

PR #929, opened 2020-07-28
Improve run_ansible_lint implementation

ssbarnea:
- add type information
- allow cwd to be None
- allow cwd to be "." which points to repository root

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -5,6 +5,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
+from typing import Dict
 
 from ansible import __version__ as ansible_version_str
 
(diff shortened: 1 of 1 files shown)


test/__init__.py
```diff
@@ -53,12 +54,22 @@ def run_role_meta_main(self, meta_main_text):
         return results
 
 
-def run_ansible_lint(cwd, role_path=None, bin=None, env=None):
-    """Run ansible-lint on a given path and returns its output."""
+def run_ansible_lint(
+        cwd: os.PathLike = None,
+        role_path: os.PathLike = None,
+        bin: str = None,
+        env: Dict = None):
```
webknjaz, line 61: I'm pretty sure it's
```suggestion
        env: Dict[str, str] = None):
```

test/__init__.py
```diff
@@ -53,12 +54,22 @@ def run_role_meta_main(self, meta_main_text):
...
+    """Run ansible-lint on a given path and returns its output.
+
+    If "." is passed to cwd, the root of the ansible-lint source repository will be used.
+    """
     command = '{} -v {}'.format(
         bin or (sys.executable + " -m ansiblelint"),
         role_path or "")
```
webknjaz, line 68: since you're already refactoring thinks, can we turn `command` into a list?

test/__init__.py
```diff
@@ -53,11 +54,24 @@ def run_role_meta_main(self, meta_main_text):
...
+    else:
+        command = [sys.executable, "-m", "ansiblelint"]
+    if role_path:
+        command.append(role_path)
+
+    if cwd == ".":
+        cwd = os.Path(__file__).parents[1]
 
     result, err = subprocess.Popen(
         [command],
```
webknjaz, line 77: Don't forget to fix this, otherwise, it's a nested list
```suggestion
        command,
```

webknjaz:
@ssbarnea any idea why the tests are failing?

test/__init__.py
```diff
@@ -53,14 +54,27 @@ def run_role_meta_main(self, meta_main_text):
...
+        cwd: os.PathLike = None,
+        role_path: os.PathLike = None,
+        bin: str = None,
+        env: Dict[str, str] = None):
+    """Run ansible-lint on a given path and returns its output.
+
+    If "." is passed to cwd, the root of the ansible-lint source repository will be used.
+    """
+    if bin:
+        command = [bin]
```
webknjaz, line 67: The original gode is verbose in both cases:
```suggestion
        command = [bin, "-v"]
```

webknjaz:
https://github.com/ansible/ansible-lint/pull/929/checks?check_run_id=919225801#step:11:277
Looks like special-casing `"."` breaks assumption in some tests (that expect it remain current cwd of the process rather than changed)...

PR #931, opened 2020-07-30
refactor: reduce complexity in main function

dpendolino:
Ref #931 - Refactor main function to reduce branching complexity ( and 
thus  quiet flake8 ) by moving the formatter_factory into a separate function.
Import argparse.Namespace for mypy static check requirement.

Signed-off-by: Daniel Pendolino <daniel@counterhack.com>

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -60,7 +60,7 @@ def initialize_logger(level: int = 0) -> None:
 
 
 def initalize_formatter_factory(options_list: Namespace):
-    """Apply options from command line arguments to the formatter_factory."""
+    """Select an output formatter based on the incoming command line arguments."""
     formatter_factory: Any = formatters.Formatter
     if options_list.quiet:
         formatter_factory = formatters.QuietFormatter
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -82,7 +82,7 @@ def main() -> int:
     initialize_logger(options.verbosity)
     _logger.debug("Options: %s", options)
 
-    formatter_factory = initalize_formatter_factory(options)
+    formatter_factory = choose_formatter_factory(options)
     formatter = formatter_factory(cwd, options.display_relative_path)
 
     if options.use_default_rules:
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -61,7 +61,7 @@ def initialize_logger(level: int = 0) -> None:
 
 def choose_formatter_factory(options_list: Namespace) -> Type[formatters.BaseFormatter]:
     """Select an output formatter based on the incoming command line arguments."""
-    formatter_factory: Any = formatters.Formatter
+    formatter_factory = formatters.Formatter
     if options_list.quiet:
         formatter_factory = formatters.QuietFormatter
 
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -59,7 +59,7 @@ def initialize_logger(level: int = 0) -> None:
     _logger.debug("Logging initialized to level %s", logging_level)
 
 
-def initalize_formatter_factory(options_list: Namespace):
+def choose_formatter_factory(options_list: Namespace) -> Type[formatters.BaseFormatter]:
     """Select an output formatter based on the incoming command line arguments."""
     formatter_factory: Any = formatters.Formatter
     if options_list.quiet:
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -24,8 +24,10 @@ import errno
 import logging
 import pathlib
 import sys
-from argparse import Namespace
-from typing import Any, Set
+from typing import Any, Set, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from argparse import Namespace
 
 from ansiblelint import cli, formatters
 from ansiblelint.constants import DEFAULT_RULESDIR
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -24,7 +24,7 @@ import errno
 import logging
 import pathlib
 import sys
-from typing import TYPE_CHECKING, Set, Type, Union
+from typing import TYPE_CHECKING, Set, Type
 
 from ansiblelint import cli, formatters
 from ansiblelint.constants import DEFAULT_RULESDIR
@@ -61,17 +61,16 @@ def initialize_logger(level: int = 0) -> None:
 
 def choose_formatter_factory(
     options_list: "Namespace"
-) -> Union[Type[formatters.Formatter], Type[formatters.QuietFormatter],
-           Type[formatters.ParseableFormatter],
-           Type[formatters.ParseableSeverityFormatter]]:
+) -> Type[formatters.BaseFormatter]:
     """Select an output formatter based on the incoming command line arguments."""
+    r: Type[formatters.BaseFormatter] = formatters.Formatter
     if options_list.quiet:
-        return formatters.QuietFormatter
+        r = formatters.QuietFormatter
     elif options_list.parseable:
-        return formatters.ParseableFormatter
+        r = formatters.ParseableFormatter
     elif options_list.parseable_severity:
-        return formatters.ParseableSeverityFormatter
-    return formatters.Formatter
+        r = formatters.ParseableSeverityFormatter
+    return r
 
 
 def main() -> int:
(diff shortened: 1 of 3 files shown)


webknjaz:
@dpendolino please always add a link to the issue like "Ref #931" in the PR description. It'll generate a backref from that issue here too. And if the PR fully resolves the issue (not this case), put a special keyword like "closes", "resolves" or "fixes" in front of the number so that merging the PR would automatically close the related issue.

webknjaz:
Zuul CI check is failing in some envs but it seems to be not your fault. Though, if you want to re-trigger it, just leave a comment with text "recheck" and it'll restart Zuul jobs.

webknjaz:
recheck

lib/ansiblelint/__main__.py
```diff
@@ -24,6 +24,7 @@
 import logging
 import pathlib
 import sys
+from argparse import Namespace
 from typing import Any, Set
```
webknjaz, line 28: The typical mypy usage idiom is
```suggestion
from typing import Any, Set, TYPE_CHECKING

if TYPE_CHECKING:
    from argparse import Namespace
```

lib/ansiblelint/__main__.py
```diff
@@ -56,6 +57,20 @@ def initialize_logger(level: int = 0) -> None:
     _logger.debug("Logging initialized to level %s", logging_level)
 
 
+def initalize_formatter_factory(options_list: Namespace):
+    """Apply options from command line arguments to the formatter_factory."""
```
webknjaz, line 61: `formatter_factory` has no meaning on this level of abstraction, if you just look at the function input. The docstring should describe what the function's effect would be for the caller.

```suggestion
    """Select an output formatter based on the incoming command line arguments."""
```

lib/ansiblelint/__main__.py
```diff
@@ -56,6 +57,20 @@ def initialize_logger(level: int = 0) -> None:
     _logger.debug("Logging initialized to level %s", logging_level)
 
 
+def initalize_formatter_factory(options_list: Namespace):
```
webknjaz, line 60: This name is misleading. Classes have `__init__()` method and this is what's called an initializer because it performs the initialization of the class state post-creation (when you call `ClassName(...)` it first gets instantiated via `__new__()` constructor and then initialized via `__init__()`. Initializing is done after this function is called.

It would be appropriate to use something along the lines of "select", or "choose", or "pick" maybe...

Also, regarding `options_list`. It's probably not exactly a list so maybe invent some better var name too. Like `cli_args`.

```suggestion
def choose_formatter_factory(options_list: Namespace) -> Type[formatters.BaseFormatter]:
```

lib/ansiblelint/__main__.py
```diff
@@ -56,6 +57,20 @@ def initialize_logger(level: int = 0) -> None:
...
+    """Apply options from command line arguments to the formatter_factory."""
+    formatter_factory: Any = formatters.Formatter
```
webknjaz, line 62: This time declaration is probably unnecessary
```suggestion
    formatter_factory = formatters.Formatter
```

webknjaz: But if it complains, use something like
```suggestion
    formatter_factory: Type[formatters.BaseFormatter] = formatters.Formatter
```

webknjaz: (unless `TypeVar` is necessary, I haven't checked it)

lib/ansiblelint/__main__.py
```diff
@@ -65,16 +80,7 @@ def main() -> int:
...
-    if options.quiet:
-        formatter_factory = formatters.QuietFormatter
-
-    if options.parseable:
-        formatter_factory = formatters.ParseableFormatter
-
-    if options.parseable_severity:
-        formatter_factory = formatters.ParseableSeverityFormatter
-
+    formatter_factory = initalize_formatter_factory(options)
```
webknjaz, line 83:
```suggestion
    formatter_factory = choose_formatter_factory(options)
```

lib/ansiblelint/__main__.py
ssbarnea: If you use TYPE_CHECKING conditional you must put the var type (Napespace) in double quotes, or pylint with fail as it runs outside typchecking context.

webknjaz:
Looks like there's some linting needing to be fixed + you may optionally try to make commits in the branch nicer. I prefer to use "merge" when the branch is nice because it preserves the history well, otherwise we resort to just squash on merge (less nice IMHO because it often produces non-atomic commits).

dpendolino:
@webknjaz I had some trouble fixing the MyPy related type errors:
```
lib/ansiblelint/__main__.py:61: error: Incompatible return value type (got "Type[QuietFormatter]", expected "Type[Formatter]")
lib/ansiblelint/__main__.py:63: error: Incompatible return value type (got "Type[ParseableFormatter]", expected "Type[Formatter]")
lib/ansiblelint/__main__.py:65: error: Incompatible return value type (got "Type[ParseableSeverityFormatter]", expected "Type[Formatter]")
```

I think what I pushed is reasonable, but I'm not very familiar with the framework. If there is a better way, please let me know. Otherwise I'll clean up the commit history.

Thanks!

ssbarnea:
Take a look at https://github.com/ansible/ansible-lint/pull/934/files about how to do formatters type checking. I will be out for one week so feel free to inspire from it.

webknjaz:
@dpendolino I feel like I need to explain why unrelated changes (like reformatting) mustn't be mixed up with other things.

Maintaining a useful Git history may seem a difficult task and it is vital to keep it sane easy to review for the maintainers and other developers.

_Remember:_ both code and history are being read more often than written so care about your readers!

To help you with this task, there's certain best practices to follow. I think the most important ones are making good commit messages combined with atomic commits.

_TL;DR_ when you make one self-sufficient logical change, you should commit it separately + it's easier to describe it in a commit message when it's _really_ one change that achieves one goal. Reformatting code is by itself a separate task that if combined with anything else makes it read confusing — for example, when your PR title says "reduce complexity in **main** function" it's confusing when the same change also introduces changes outside of that function all over the place making the reviewers question their purpose and try to find the connection to the actual change. This sort of stuff increases the cognitive complexity of the patch (and will raise the same question when is in Git history and somebody does archeology while debugging things).

I have a few more extended pointers to what it takes to write great commits and keep changes atomic, it case you're up to doing some reading:
* https://chris.beams.io/posts/git-commit/
* https://twitter.com/nedbat/status/1283147492616556544
* https://dhwthompson.com/2019/my-favourite-git-commit
* https://alistapart.com/article/the-art-of-the-commit/
* https://frederickvanbrabant.com/posts/atomic-commits
* https://curiousprogrammer.io/blog/how-to-craft-your-changes-into-small-atomic-commits-using-git

P.S. PRs should also be atomic, just on a bit higher level.

webknjaz:
@dpendolino any updates?

dpendolino:
@webknjaz Sorry for the delay and thanks for all the help/suggestions. Would you prefer I start over or try to clean up this PR?

webknjaz:
No, iterating within the same PR is a normal process.

lib/ansiblelint/__main__.py
```diff
@@ -19,12 +19,11 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 """Command line implementation."""
-
```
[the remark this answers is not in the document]
dpendolino: Nope, sorry about that, I must have accidentally auto-formatted and didn't notice.

dpendolino:
@webknjaz I pushed some changes to revert the spurious formatting alterations and add those `mypy` updates from your example. The lint check seems to be failing now, but all `tox` checks pass on my local system, so I'm not sure what happened.

webknjaz:
@dpendolino you can run `tox -e lint` locally and it should show you the same errors

P.S. Come to `#ansible-galaxy` on Freenode if you need help with that.

lib/ansiblelint/__main__.py
```diff
@@ -56,6 +59,20 @@ def initialize_logger(level: int = 0) -> None:
...
+        r = formatters.QuietFormatter
+    elif options_list.parseable:
+        r = formatters.ParseableFormatter
+    elif options_list.parseable_severity:
+        r = formatters.ParseableSeverityFormatter
+    return r
+
+
 def main() -> int:
     """Linter CLI entry point."""
```
webknjaz, line 77: @dpendolino flake still thinks that main is too complex: https://github.com/ansible/ansible-lint/pull/931/checks?check_run_id=991487868#step:10:76

ssbarnea: Removing the complexity whitelisting needs multiple merges in order to reduce it from 13 to 10.

PR #933, opened 2020-07-31
Install pip if missing on zuul

ssbarnea:
As pip was removed from some nodepool images we now install it in
order to avoid failure to install tox itself.

### DIFF
--- a/tools/test-setup.sh
+++ b/tools/test-setup.sh
@@ -2,6 +2,16 @@
 set -euxo pipefail
 # Used by Zuul CI to perform extra bootstrapping
 
-PYTHON=$(command -v python3 python | head -n1)
+# sudo used only because currently zuul default tox_executable=tox instead of
+# "python3 -m tox"
+# https://zuul-ci.org/docs/zuul-jobs/python-roles.html#rolevar-ensure-tox.tox_executable
 
-sudo $PYTHON -m pip install -U tox
+# Install pip if not already install on the system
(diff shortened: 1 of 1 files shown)


tools/test-setup.sh
```diff
@@ -3,5 +3,8 @@ set -euxo pipefail
 # Used by Zuul CI to perform extra bootstrapping
 
 PYTHON=$(command -v python3 python | head -n1)
```
webknjaz, line 5: This appears unnecessary now

tools/test-setup.sh
```diff
@@ -3,5 +3,8 @@ set -euxo pipefail
...
-
-sudo $PYTHON -m pip install -U tox
+python3 -m pip --version || {
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+    sudo python3 get-pip.py
```
webknjaz, line 8: I'd install it under user instead:
```suggestion
    python3 get-pip.py --user
```

ssbarnea: @webknjaz Sadly, that is not true, please try it yourself, mainly because zuul tox role expect executable tox in user path, and --user does not add it to path. I tried for many months to get some changes to tox role but I lost hope.

tools/test-setup.sh
ssbarnea: True

PR #934, opened 2020-07-31
Refactored use of formatter factory

ssbarnea:
- reduce branching complexity by moving the formatter_factory into a separate function
- removed linting exception for main complexity
- add typing information to formatter classes

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -25,7 +25,7 @@ import logging
 import os
 import pathlib
 import sys
-from typing import Any, Set
+from typing import TYPE_CHECKING, Set, Type
 
 from ansiblelint import cli, formatters
 from ansiblelint.constants import DEFAULT_RULESDIR
@@ -33,6 +33,10 @@ from ansiblelint.generate_docs import rules_as_rst
 from ansiblelint.rules import RulesCollection
 from ansiblelint.runner import Runner
 from ansiblelint.utils import get_playbooks_and_roles
+
+if TYPE_CHECKING:
+    from argparse import Namespace
+
 
 _logger = logging.getLogger(__name__)
 
@@ -57,6 +61,20 @@ def initialize_logger(level: int = 0) -> None:
     _logger.debug("Logging initialized to level %s", logging_level)
 
 
+def choose_formatter_factory(
+    options_list: "Namespace"
+) -> Type[formatters.BaseFormatter]:
+    """Select an output formatter based on the incoming command line arguments."""
+    r: Type[formatters.BaseFormatter] = formatters.Formatter
+    if options_list.quiet:
+        r = formatters.QuietFormatter
+    elif options_list.parseable:
+        r = formatters.ParseableFormatter
+    elif options_list.parseable_severity:
+        r = formatters.ParseableSeverityFormatter
+    return r
+
+
 def main() -> int:
     """Linter CLI entry point."""
     cwd = pathlib.Path.cwd()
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/__main__.py
```diff
@@ -56,17 +57,16 @@ def initialize_logger(level: int = 0) -> None:
 
 def choose_formatter_factory(
     options_list: "Namespace"
-) -> Union[Type[formatters.Formatter], Type[formatters.QuietFormatter],
-           Type[formatters.ParseableFormatter],
-           Type[formatters.ParseableSeverityFormatter]]:
+) -> Type[formatters.BaseFormatter]:
```
webknjaz, line 60: I believe that the TypeVar approach would be more appropriate here.

lib/ansiblelint/__main__.py
```diff
@@ -56,6 +59,20 @@ def initialize_logger(level: int = 0) -> None:
...
+    """Select an output formatter based on the incoming command line arguments."""
+    r: Type[formatters.BaseFormatter] = formatters.Formatter
+    if options_list.quiet:
+        r = formatters.QuietFormatter
+    elif options_list.parseable:
+        r = formatters.ParseableFormatter
+    elif options_list.parseable_severity:
+        r = formatters.ParseableSeverityFormatter
+    return r
```
webknjaz, line 73: Extra var is unnecessary but instead using a TypeVar is needed as requested @ https://github.com/ansible/ansible-lint/pull/931/files#r463479262.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -1,15 +1,17 @@
...
 from ansiblelint.color import Color, colorize
 
 if TYPE_CHECKING:
     from ansiblelint.errors import MatchError
 
+T = TypeVar('T', bound='BaseFormatter')
 
-class BaseFormatter:
+
+class BaseFormatter(Generic[T]):
```
webknjaz, line 14: Not sure if it's necessary. It looks like it's needed to just use custom type @ `__main__`

lib/ansiblelint/formatters/__init__.py
```diff
@@ -42,6 +44,9 @@ def _format_path(self, path: Union[str, Path]) -> str:
         # Use os.path.relpath 'cause Path.relative_to() misbehaves
         return os.path.relpath(path, start=self._base_dir)
 
+    def format(self, match: "MatchError", colored: bool = False) -> str:
+        return str(match)
+
```
webknjaz, line 49: I don't think this is related to reducing the complexity in main

ssbarnea: It forces inherited objects to respect type, this change is as much about typing as reducing complexity. I will make it clear in description, as original commit message did not endup being copied and the subject line is too short.

lib/ansiblelint/__main__.py
ssbarnea: I tried different approaches and they did not work. If you know a better way to do it, one that is passing all our testing, lets do it in a follow-up.

lib/ansiblelint/formatters/__init__.py
ssbarnea: See my previous comment.

lib/ansiblelint/__main__.py
```diff
@@ -19,13 +19,12 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 """Command line implementation."""
-
```
webknjaz, line 22: Unrelated. Also it's better to have a visual separator.

lib/ansiblelint/__main__.py
webknjaz: This request still stands.

lib/ansiblelint/__main__.py
webknjaz: I can take a look in a week.

PR #943, opened 2020-08-11
Added MissingFilePermissionsRule

ssbarnea:
Makes missing file permissions a linting error in order to proactively detect security issues that were fixed via https://github.com/ansible/ansible/issues/67794

While that was fixed in last Ansible hotfix, it changed default value and thus it has a high risk
of breaking any usage without `mode`.

Related: https://github.com/ansible/ansible/issues/71200

### DIFF
--- a/lib/ansiblelint/rules/MissingFilePermissionsRule.py
+++ b/lib/ansiblelint/rules/MissingFilePermissionsRule.py
@@ -1,0 +1,52 @@
+# Copyright (c) 2020 Sorin Sbarnea <sorin.sbarnea@gmail.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+from ansiblelint.rules import AnsibleLintRule
+
+
+class MissingFilePermissionsRule(AnsibleLintRule):
+    id = "208"
+    shortdesc = 'File permissions not mentioned'
+    description = (
+        "Missing mode parameter can cause unexpected file permissions based "
+        "on version of Ansible being used. Be explicit, or if you still "
+        "want the default behavior you can use ``mode: preserve`` to avoid "
+        "hitting this rule. See "
+        "https://github.com/ansible/ansible/issues/71200"
+    )
+    severity = 'VERY_HIGH'
+    tags = ['unpredictability']
+    version_added = 'v4.3.0'
+
+    _modules = (
+        'copy',
+        'file',
+        'ini_file',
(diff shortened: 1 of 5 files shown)


lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -0,0 +1,52 @@
...
+# THE SOFTWARE.
+from ansiblelint.rules import AnsibleLintRule
+
+
+class MissingFilePermissionsRule(AnsibleLintRule):
+    id = "208"
+    shortdesc = 'File permissions not mentioned'
+    description = (
+        "Missing file permissions can cause unexpected behavior and is a "
+        "a security risk with older versions of Ansible. If still "
```
yoctozepto, line 28: nit: "If **you** still"

ssbarnea:
recheck

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -0,0 +1,52 @@
...
+        'template',
+        'unarchive',
+    )
+
+    def matchtask(self, file, task):
+        if task["action"]["__ansible_module__"] in self._modules:
+            mode = task['action'].get('mode', None)
+
+            if not isinstance(mode, str):
+                return True
```
webknjaz, line 52: Let's simplify this a bit
```suggestion
        if task["action"]["__ansible_module__"] not in self._modules:
            return False

        mode = task['action'].get('mode', None)

        return not isinstance(mode, str)
```

test/TestMissingFilePermissionsRule.py
```diff
@@ -0,0 +1,67 @@
...
+---
+- hosts: hosts
+  tasks:
+    - name: permissions missing
+      file:
+        path: foo
+'''
+
+
+class TestMissingFilePermissionsRule(unittest.TestCase):
```
webknjaz, line 48: https://github.com/ansible/ansible-lint/issues/725

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -0,0 +1,52 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+from ansiblelint.rules import AnsibleLintRule
+
+
+class MissingFilePermissionsRule(AnsibleLintRule):
+    id = "208"
+    shortdesc = 'File permissions not mentioned'
+    description = (
+        "Missing file permissions can cause unexpected behavior and is a "
```
MarkusTeufelberger, line 27:
```suggestion
        "Missing file permissions can cause unexpected behavior and is "
```

yoctozepto: Should not it be "are"?

MarkusTeufelberger: I'm not sure if the "behavior" or the "permissions" are referenced and while my English isn't that bad, I guess there are more qualified people than me out there to decide that. The double "a" is definitely an issue though.

test/conftest.py
```diff
@@ -19,4 +20,15 @@ def default_text_runner(default_rules_collection):
     return RunFromText(default_rules_collection)
 
 
+@pytest.fixture
+def rule_runner(request):
+    """Return runner for a specific rule class."""
+    rule_name = request.param
+    module = import_module(f"ansiblelint.rules.{rule_name}")
```
webknjaz, line 27: I don't think that imports should be delegated to the fixture.

ssbarnea: That is only for internal use and not doing that would make its use much more complex, especially if we start to migrate tests to reuse the runner. The hole point of this fixture was to avoid the old repeating setUp code.

webknjaz: Import errors should be happening in the test modules, not in the middle of the machinery.

test/conftest.py
```diff
@@ -19,4 +20,15 @@ def default_text_runner(default_rules_collection):
...
+    collection = RulesCollection()
+    rule = getattr(module, rule_name)
```
webknjaz, line 29: this should be just
```suggestion
    rule = request.param
    collection = RulesCollection()
```

ssbarnea: This would obviously not work as the rule needs to be loaded and instantiated before being added to the collection.

webknjaz: Yes, it's still on the next line. Look closer.

webknjaz: Just to reiterate: it's import that I'm against here, the rest of the fixture is great.

ssbarnea: So basically you want to pass the class object instead of class name to the fixture. Well, that would add one more line to where we use it but I am ok with it, changing it now.

webknjaz: Yes, it would add one more line in order to make what's happening explicit and thus more obvious to the test module reader. This fixture, though, may need some sort of input validation but it's not strictly necessary right now.

PR #945, opened 2020-08-11
Prefer pylint instead over unittest

webknjaz:
This change supports the goal of #725 to improve the pytest
compatibility of the existing tests and enforce it in the new tests.


webknjaz:
This is necessary so that people don't use features that are incompatible with pytest. Besides, we've been moving in this direction for a while now and this is one of the major reasons I'm still involved.

ssbarnea (approved):

Issue #947, opened 2020-08-12
proposal: make rule specific testing part of the rule definition


ssbarnea:
##### Summary
One serious barrier in receiving new rule contributions is related to the complexity of adding tests for the rule.

##### Desired Behaviour

If we make each rule **self-contained** in a single file, one with **implementation, documentation and testing** it will be much easier to receive contributions from new users, users that may know ansible very well but not master Python testing very well.

Note that this proposal is only about testing of particular rule, not about the generic functionality of linter, which is expected to remain the same.

One particular challenge we have for implementing this is that we do not want to add pytest as a runtime requirement and that we may need to tune its test discovery logic in order to make it detect tests that do reside inside the library. I am not even sure that is possible but we should investigate and see if others succeeded with this challenge.

PR #948, opened 2020-08-12
Use embedded rule testing for MetaTagValidRule

ssbarnea:
This is the first change from a series that will allow us to embed  testing of simple rules in the rule files, making it much easier to create and maintain.

As mentioned in the linked ticket the goal of this change is to make it as easy as possible for new contributors to write or edit new rules, even if they may not master python, pytest or the project internals. By keeping all the rule-specific code in one place: implementation, documentation and testing we enable them to edit a single file.

For example, if they want to write a new rule, they would only have to copy an existing one that is kinda similar and modify it. This will prevent a common issue where a new contributor manages to write a rule but fails to write tests for it.

Partial-Fix: #947
Related: https://github.com/pytest-dev/pytest/issues/6451


ssbarnea:
Apparently this approach break discovery with errors like:
```
______________________________ ERROR collecting lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py ______________________________
.tox/py36-ansible28/lib/python3.6/site-packages/py/_path/local.py:724: in pyimport
    raise self.ImportMismatchError(modname, modfile, self)
E   py._path.local.LocalPath.ImportMismatchError: ('ansiblelint.rules.UsingBareVariablesIsDeprecatedRule', '/home/ssbarnea/rmux/ansible-lint/.tox/py36-ansible28/lib/python3.6/site-packages/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py', local('/home/ssbarnea/rmux/ansible-lint/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py'))
        ensuresyspath = True
        ignore     = None
        issame     = False
        mod        = <module 'ansiblelint.rules.UsingBareVariablesIsDeprecatedRule' from '/home/ssbarnea/rmux/ansible-lint/.tox/py36-ansible28/lib/python3.6/site-packages/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py'>
        modfile    = '/home/ssbarnea/rmux/ansible-lint/.tox/py36-ansible28/lib/python3.6/site-packages/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py'
        modname    = 'ansiblelint.rules.UsingBareVariablesIsDeprecatedRule'
        names      = ['ansiblelint', 'rules', 'UsingBareVariablesIsDeprecatedRule']
        pkgpath    = local('/home/ssbarnea/rmux/ansible-lint/lib/ansiblelint')
        pkgroot    = local('/home/ssbarnea/rmux/ansible-lint/lib')
        self       = local('/home/ssbarnea/rmux/ansible-lint/lib/ansiblelint/rules/UsingBareVariablesIsDeprecatedRule.py')
```

pabelanger:
recheck

conftest.py
```diff
@@ -1,29 +1,36 @@
+"""PyTest Fixtures."""
```
webknjaz, line 1: I don't think conftest should be moved.

lib/ansiblelint/rules/MetaTagValidRule.py
```diff
@@ -57,3 +58,23 @@ def matchplay(self, file, data):
...
+
+META_TAG_VALID = '''
+galaxy_info:
+    galaxy_tags: ['database', 'my s q l', 'MYTAG']
+    categories: 'my_category_not_in_a_list'
+'''
+
+if "pytest" in sys.modules:
+
+    import pytest
```
webknjaz, line 71: Now this looks scary...

lib/ansiblelint/rules/MetaTagValidRule.py
```diff
@@ -57,3 +58,23 @@ def matchplay(self, file, data):
...
+
+    @pytest.mark.parametrize('rule_runner', (MetaTagValidRule, ), indirect=['rule_runner'])
+    def test_valid_tag_rule(rule_runner):
```
webknjaz, line 74: I don't like that this approach adds an extra indent. Maybe we could just have an extra file with RuleTest instead of Rule in the name?

conftest.py
ssbarnea: Again, that is not up to personal preferences. This file must be at the root of all tests, now that we have tests in test/ and also in lib/ansiblelib/rules/ we must put it at repository root.

lib/ansiblelint/rules/MetaTagValidRule.py
ssbarnea: It works. I can also ask few pytest cores for feeback if you really want. It is dead code in production and it is loaded when in testing. If you think a little bit it is kinda similar with how we do type checking.

lib/ansiblelint/rules/MetaTagValidRule.py
ssbarnea: The whole point of this move is to encapsulate testing of rules in a single file. This would easy contributions and also allow us to include tests as examples/documentation. We will be able to expose them in docs and command like.

Take a look at eslint docs, random rule documentation: https://eslint.org/docs/rules/no-import-assign -- it easy to read with good and bad usage examples, exactly what we do to test the rules.

conftest.py
webknjaz: Are you sure you cannot set its path as a pytest invocation arg?

webknjaz: Besides, pytest traverses the dir layout and will probably pick it up anyway

lib/ansiblelint/rules/MetaTagValidRule.py
```diff
@@ -1,6 +1,8 @@
+#!/usr/bin/env python
```
webknjaz, line 1: Why are you making it executable? I don't think that adding more way to do the same thing is a good idea.

conftest.py
```diff
@@ -1,29 +1,36 @@
...
 import os
+from test import RunFromText
```
webknjaz, line 3: This import is invalid in this place. It's not coming from stdlib.

conftest.py
```diff
@@ -1,29 +1,36 @@
...
+    """Add rule folder to test include for running embedded rule tests."""
+    config.args.append(os.path.dirname(rules.__file__))
 
 
 @pytest.fixture
 def default_rules_collection():
-    return RulesCollection(rulesdirs=[os.path.abspath(os.path.join('lib',
-                                                                   'ansiblelint',
-                                                                   'rules'))])
+    """Return default rule collection."""
```
webknjaz, line 17:
```suggestion
    """Return default rules collection."""
```

conftest.py
```diff
@@ -1,29 +1,36 @@
...
+    return rules.RulesCollection(rulesdirs=[os.path.abspath(os.path.join('lib',
+                                                                         'ansiblelint',
+                                                                         'rules'))])
 
 
 @pytest.fixture
 def default_text_runner(default_rules_collection):
+    """Return TextRunner."""
```
webknjaz, line 25:
```suggestion
    """Return TextRunner instance."""
```

conftest.py
```diff
@@ -1,29 +1,36 @@
...
 import pytest
 
-from ansiblelint.rules import RulesCollection
+import ansiblelint.rules as rules
 
-from . import RunFromText
+
+def pytest_configure(config):
+    """Add rule folder to test include for running embedded rule tests."""
+    config.args.append(os.path.dirname(rules.__file__))
```
webknjaz, line 12: When you run `pytest --help`, it explodes with
```python-traceback
  File "test/conftest.py", line 12, in pytest_configure
    config.args.append(os.path.dirname(rules.__file__))
AttributeError: 'Config' object has no attribute 'args'
```

conftest.py
ssbarnea: Hmm..., that is tricky because conftest.py is not really a python module so nobody is doing any from ansiblelint import conftest. These conftest files are special files used by pytest, which looks for all of them upwards. But anything defined in conftest files outside your paterns, will not be available to your test. This is why we need it at the root, to make them available for both the library and the tests, which are two different branches.

Probably we can get around by creating a symlink. Would that please you?

webknjaz: No, this should be added to isort as local/first-party import and that will work just fine.

conftest.py
webknjaz: Use `--pyargs` to avoid such problems:
https://github.com/ansible/ansible-lint/pull/1009/files#diff-513c9b6861fc981fb2c708db477f542cR17-R61. This will also make the setting more declarative.

pytest.ini
```diff
@@ -19,6 +19,9 @@ addopts =
     --cov-report term-missing:skip-covered
     --cov-report xml:.test-results/pytest/cov.xml
     --no-cov-on-fail
+
+    # mention where to look for tests
+    --pyargs test/ --pyargs ansiblelint.rules
```
webknjaz, line 24: I'm pretty sure double `--pyargs` does nothing and is unnecessary.

pytest.ini
```diff
@@ -54,5 +57,8 @@ python_files =
     # Ref: https://docs.pytest.org/en/latest/reference.html#confval-python_files
     # Needed to discover legacy nose test modules:
     Test*.py
-testpaths = test/
```
webknjaz, line 57: Keep this in testpaths. It's better when it's declarative.

docs/contributing.rst
```diff
@@ -12,3 +12,41 @@ without any warning.
...
+  correctly in them.
+* Assure linting is passing with ``tox -e lint``.
+* Make a commit that explains why the rule is added and that eventually
+  closing an existing bug by including ``Fixes: #123`` in its description.
+  Commit message should follow the 50/72 formatting rule.
+* Open a draft pull-request. When CI/CD reports green, move the draft PR from
+  draft to ready for review.
+
+.. _MetaTagValidRule: https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/MetaTagValidRule.py
+.. _virtual environment: https://docs.python.org/3/tutorial/venv.html
```
webknjaz, line 52: Python docs are already added via intersphinx so there's no need to use hardcoded links. Instead, use the sphinx's interlinking features.
This is how you can link it in text:
```rst
:doc:`virtual environment <python3:tutorial/venv>`
```

P.S. Since extracting linkable entries isn't exactly user-friendly, I've made a small static website the other day. It lists all intersphinx entries as a table and has an extra column with ready-to-use examples of how to link each thing: https://webknjaz.github.io/intersphinx-untangled.

docs/contributing.rst
```diff
@@ -12,3 +12,41 @@ without any warning.
...
+* Implement linting methods needed by you. Feel free to look on how similar
+  rules were implemented.
+* Update the tests. It must have at least one test and likely also a negative
+  match one. If the rule is task specific, you may want to test its use inside
+  blocks as well.
+* Run the rule specific tests ``pytest -k NewRule`` as this will run only
+  tests unique to that rule.
+* Now run ``tox`` in order to run all ansible-lint tests. Adding a new rule can
+  break some other tests. Update them if needed.
+* Run ``ansible-lint -L`` and check that the rule description renders nicely
```
webknjaz, line 40: Sphinx has `:command:` role for such cases
```suggestion
* Run :command:`ansible-lint -L` and check that the rule description renders nicely
```

docs/contributing.rst
```diff
@@ -12,3 +12,41 @@ without any warning.
...
+  and correct.
+* Build the docs using ``tox -e docs`` and check that the new rule is displayed
+  correctly in them.
+* Assure linting is passing with ``tox -e lint``.
+* Make a commit that explains why the rule is added and that eventually
+  closing an existing bug by including ``Fixes: #123`` in its description.
+  Commit message should follow the 50/72 formatting rule.
+* Open a draft pull-request. When CI/CD reports green, move the draft PR from
+  draft to ready for review.
```
webknjaz, line 49: This seems out of scope and should probably be separate (likely in a separate PR too).

pytest.ini
ssbarnea: I do want to avoid having the discovery in two places, it does confuse us. As long we do not have any bugs is better to use a single location.

pytest.ini
```diff
@@ -54,5 +57,8 @@ python_files =
...
+    # Needed to discover embedded Rule tests
+    *Rule.py
+# do not use testpaths due to https://github.com/pytest-dev/pytest/issues/6451
```
webknjaz, line 62: This link does not really explain the motivation. I think this comment should be updated.

pytest.ini
```diff
@@ -19,6 +19,9 @@ addopts =
     --cov-report term-missing:skip-covered
     --cov-report xml:.test-results/pytest/cov.xml
     --no-cov-on-fail
+
+    # mention where to look for tests
+    --pyargs test ansiblelint.rules
```
webknjaz, line 24: Putting these on the same line triggers a misleading suggestion that `test ansiblelint.rules` are the args to `--pyargs` option that is not true:
```console
  --pyargs              try to interpret all arguments as python packages.
```
So we must separate the test lookup targets:
```suggestion
    # interpret all the target args as importables:
    --pyargs

    # importable packages for test lookup:
    test ansiblelint.rules
```

webknjaz (approved):

PR #949, opened 2020-08-12
E208: Improve MissingFilePermissionsRule detection

ssbarnea:
Improve [rule 208](https://github.com/ansible/ansible-lint/pull/943) by avoiding to trigger it when `state: absent` is found or when any mode is present. This fixed bug where rule was triggered when mode was numeric.

### DIFF
--- a/lib/ansiblelint/rules/MissingFilePermissionsRule.py
+++ b/lib/ansiblelint/rules/MissingFilePermissionsRule.py
@@ -48,5 +48,8 @@ class MissingFilePermissionsRule(AnsibleLintRule):
         if task["action"]["__ansible_module__"] not in self._modules:
             return False
 
+        if task['action'].get('state', None) == "absent":
+            return False
+
         mode = task['action'].get('mode', None)
-        return not isinstance(mode, str)
+        return mode is None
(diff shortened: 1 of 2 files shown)


ssbarnea:
recheck

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -48,5 +48,8 @@ def matchtask(self, file, task):
         if task["action"]["__ansible_module__"] not in self._modules:
             return False
 
+        if task['action'].get('state', None) == "absent":
+            return False
+
         mode = task['action'].get('mode', None)
-        return not isinstance(mode, str)
+        return mode is None
```
webknjaz, line 55: It's probably cleaner to use
```suggestion
        try:
            task['action']['mode']
        except KeyError:
            return False
        else:
            return True
```
(because with `.get()` it is ambiguous whether it returns a fallback or the key exists and the value actually `is None`)

ssbarnea: Not in this case, I do still want to to fail if mode is null/empty inside YAML.

ssbarnea: Another alternative would be to return true only if mode is str or numeric, as that is the desired behavior.

pabelanger:
recheck

pabelanger:
recheck

ssbarnea:
recheck (zuul infra issue)

Issue #787, opened 2020-05-16
Perform Ansible syntax checking on linted playbooks


ssbarnea:
In order to include results from `ansible --syntax-check` we should make use of `PlaybookExecutor.run()` so users of the linter would not have to also run ansible check on the files.

A good start would be to identify a **problem that is reported by ansible own syntax check but not by ansible-lint**

PR #950, opened 2020-08-13
E911: Added AnsibleSyntaxCheckRule for running ansible syntax check

ssbarnea:
Fixes issue where passing task/var files to the linter would return false-positives. 
The linter reported the task files as playbooks but silently failed to lint them and passed.

From now, any argument that is not of playbook type will generate specific errors caused by ansible syntax-check fails.

This does affect performance, especially during our own testing, but the real benefit is that we now get confirmation from Ansible itself. We can look into improving performance in follow-ups if we find a way to instantiate ansible runtime only once instead of running the CLI tool every time. Still this comes with downsides too because the CLI interface is more stable. Another possible path to improve performance would be to run the syntax check async on multiple threads. Anyway, any approach will in **follow-up**, not as part of this initial implementation.

Fixes: #787

### DIFF
--- a/lib/ansiblelint/runner.py
+++ b/lib/ansiblelint/runner.py
@@ -4,12 +4,12 @@ import os
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, FrozenSet, Generator, List, Optional, Set, Union
 
-import ansiblelint.file_utils
 import ansiblelint.skip_utils
 import ansiblelint.utils
 from ansiblelint._internal.rules import LoadingFailureRule
 from ansiblelint.errors import MatchError
 from ansiblelint.file_utils import Lintable
+from ansiblelint.rules.AnsibleSyntaxCheckRule import AnsibleSyntaxCheckRule
 
 if TYPE_CHECKING:
     from ansiblelint.rules import RulesCollection
@@ -77,29 +77,37 @@ class Runner(object):
     def run(self) -> List[MatchError]:
         """Execute the linting process."""
         files: List[Lintable] = list()
+        matches: List[MatchError] = list()
+
         for playbook in self.playbooks:
             if self.is_excluded(str(playbook.path.resolve())) or playbook.kind == 'role':
                 continue
             files.append(playbook)
-        matches = set(self._emit_matches(files))
+            matches.extend(AnsibleSyntaxCheckRule._get_ansible_syntax_check_matches(playbook))
 
-        # remove duplicates from files list
-        files = [value for n, value in enumerate(files) if value not in files[:n]]
+        if not matches:  # do our processing only when ansible syntax check passed
 
-        # remove files that have already been checked
-        files = [x for x in files if x.path not in self.checked_files]
-        for file in files:
-            _logger.debug(
-                "Examining %s of type %s",
-                ansiblelint.file_utils.normpath(file.path),
-                file.kind)
-            matches = matches.union(
-                self.rules.run(file, tags=set(self.tags),
-                               skip_list=self.skip_list))
+            matches.extend(self._emit_matches(files))
+
+            # remove duplicates from files list
+            files = [value for n, value in enumerate(files) if value not in files[:n]]
+
+            # remove files that have already been checked
+            files = [x for x in files if x.path not in self.checked_files]
+            for file in files:
+                _logger.debug(
+                    "Examining %s of type %s",
+                    ansiblelint.file_utils.normpath(file.path),
+                    file.kind)
+
+                matches.extend(
+                    self.rules.run(file, tags=set(self.tags),
+                                   skip_list=self.skip_list))
+
         # update list of checked files
         self.checked_files.update([str(x.path) for x in files])
 
-        return sorted(matches)
+        return sorted(list(set(matches)))
 
     def _emit_matches(self, files: List[Lintable]) -> Generator[MatchError, None, None]:
         visited: Set = set()
--- a/lib/ansiblelint/text.py
+++ b/lib/ansiblelint/text.py
@@ -1,0 +1,15 @@
+"""Text utils."""
+import re
+from typing import Union
+
+
+def strip_ansi_escape(data: Union[str, bytes]) -> str:
+    """Remove all ANSI escapes from string or bytes.
+
+    If bytes is passed instead of string, it will be converted to string
+    using UTF-8.
+    """
+    if isinstance(data, bytes):
+        data = data.decode("utf-8")
+
+    return re.sub(r"\x1b[^m]*m", "", data)
--- a/lib/ansiblelint/rules/AnsibleSyntaxCheckRule.py
+++ b/lib/ansiblelint/rules/AnsibleSyntaxCheckRule.py
@@ -1,0 +1,101 @@
+"""Rule definition for ansible syntax check."""
+import re
+import subprocess
+import sys
+from typing import List
+
+from ansiblelint._internal.rules import BaseRule, RuntimeErrorRule
+from ansiblelint.errors import MatchError
+from ansiblelint.file_utils import Lintable
+from ansiblelint.rules import AnsibleLintRule
+from ansiblelint.text import strip_ansi_escape
+
+_ansible_syntax_check_re = re.compile(
+    r"^ERROR! (?P<title>[^\n]*)\n\nThe error appears to be in "
+    r"'(?P<filename>.*)': line (?P<line>\d+), column (?P<column>\d+)",
+    re.MULTILINE | re.S | re.DOTALL)
+
+
+class AnsibleSyntaxCheckRule(AnsibleLintRule):
+    """Ansible syntax check report failure."""
+
+    id = "911"
+    shortdesc = "Ansible syntax check failed"
+    description = "Running ansible-playbook --syntax-check ... reported an error."
+    severity = "VERY_HIGH"
+    tags = ["core"]
+    version_added = "v5.0.0"
+
+    @staticmethod
+    def _get_ansible_syntax_check_matches(lintable: Lintable) -> List[MatchError]:
+        """Run ansible syntax check and return a list of MatchError(s)."""
+        if lintable.kind != 'playbook':
+            return []
+
+        run = subprocess.run(
+            ['ansible-playbook', '--syntax-check', lintable.path],
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            shell=False,  # needed when command is a list
+            universal_newlines=True,
+            check=False
+        )
+        result = []
+        if run.returncode != 0:
+            message = None
+            filename = None
+            linenumber = 0
+            column = None
+
+            stderr = strip_ansi_escape(run.stderr)
+            stdout = strip_ansi_escape(run.stdout)
+            if stderr:
+                details = stderr
+                if stdout:
+                    details += "\n" + stdout
+            else:
+                details = stdout
+
+            m = _ansible_syntax_check_re.search(stderr)
+            if m:
+                message = m.groupdict()['title']
+                # Ansible returns absolute paths
+                filename = m.groupdict()['filename']
+                linenumber = int(m.groupdict()['line'])
+                column = int(m.groupdict()['column'])
+
+            if run.returncode == 4:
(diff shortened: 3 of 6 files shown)


lib/ansiblelint/runner.py
```diff
@@ -64,7 +66,7 @@ def is_excluded(self, file_path: str) -> bool:
         # not excluded.
         return any(file_path.startswith(path) for path in self.exclude_paths)
 
-    def run(self) -> List:
+    def run(self) -> List:  # noqa: C901
```
webknjaz, line 69: ouch

lib/ansiblelint/runner.py
```diff
@@ -101,6 +103,27 @@ def run(self) -> List:
...
+                    check=False
+                )
+                if result.returncode != 0:
+                    matches.append(MatchError(
+                        message="Ansible syntax-check failed, maybe file is not a playbook.",
+                        details="\n".join([result.stdout, result.stderr]),
+                        filename=file['path'],
+                        rule=AnsibleSyntaxCheckRule
+                    ))
+                    continue
```
webknjaz, line 125: the first step would be to do this
```suggestion
                try:
                    ansible_playbook_check_output = subprocess.check_output(
                        ['ansible-playbook', '--syntax-check', file['path']],
                        stderr=subprocess.STDOUT,
                        shell=False,  # needed when command is a list
                        universal_newlines=True,
                    )
                except subprocess.CalledProcessError:
                    matches.append(MatchError(
                        message="Ansible syntax-check failed, maybe file is not a playbook.",
                        details=ansible_playbook_check_output,
                        filename=file['path'],
                        rule=AnsibleSyntaxCheckRule,
                    ))
                    continue
```

test/TestDependenciesInMeta.py
```diff
@@ -3,6 +3,7 @@
 from ansiblelint.runner import Runner
 
 
+@pytest.mark.xfail(reason="Needs rework to pass syntax check")
```
webknjaz, line 6: add `raises=` and a link to the explanation

lib/ansiblelint/text.py
```diff
@@ -0,0 +1,76 @@
...
+    If bytes is passed instead of string, it will be converted to string
+    using UTF-8.
+    """
+    if isinstance(data, bytes):
+        data = data.decode("utf-8")
+
+    return re.sub(r"\x1b[^m]*m", "", data)
+
+
+def parse_ansible_syntax_check(run: CompletedProcess) -> List[MatchError]:
```
tadeboro, line 35: Why is a text-parsing function taking process as an input? It should just receive the Ansible output and parse it. Also, I am not sure if this is really a text-related function at all since it returns a list of error messages. I would expect this kind of function to be in some Ansible wrapper module.

lib/ansiblelint/text.py
```diff
@@ -0,0 +1,76 @@
...
+    """Convert ansible syntax check output into a list of MatchError(s)."""
+    result = []
+    if run.returncode != 0:
```
tadeboro, line 38: This is useless since this function should not be called when there is nothing wrong with the input. Replacing the `run` input parameter with some pure text would solve this issue as a side-effect.

tadeboro (changes requested):
I like the idea behind the proposed changes, but the implementation should be improved before this PR can be merged. Tightly coupling all sorts of unrelated stuff will not help us in the long run (we risk ending up in the same mess as with MatchError and Rule classes that now might just be the same class).

lib/ansiblelint/text.py
ssbarnea: Is not really useless as the result of running syntax check is not a single string value, is a tuple formed of exit code, stderr and stdout content. Differences between the stream are likely to be relevant in the future, when the warnings will finally become real warnings, maybe even machine parsable if we are lucky.

The exit code is also relevant because currently value 4 means syntax check failed, 0 success and other values indicate other Ansible errors.

lib/ansiblelint/text.py
ssbarnea: I am totally for isolating Ansible dependency on the linter but that is not really a case of code depending on Ansible code, is almost only text processing.

But after looking a bit more at how is used, I will make it a method of the runner and keep only the stripping inside text module (others methods will be moved there in the future, once we strip more out of the utils.py kitchensink)

lib/ansiblelint/runner.py
```diff
@@ -1,18 +1,28 @@
...
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, FrozenSet, Generator, List, Optional, Set
 
 import ansiblelint.file_utils
 import ansiblelint.skip_utils
-import ansiblelint.utils
-from ansiblelint._internal.rules import LoadingFailureRule
+from ansiblelint._internal.rules import BaseRule, LoadingFailureRule, RuntimeErrorRule
 from ansiblelint.errors import MatchError
+from ansiblelint.rules.AnsibleSyntaxCheckRule import AnsibleSyntaxCheckRule
```
tadeboro, line 13: Not relevant to this PR, but still: having files with camel-case names in the Python project is just plain bad. I had to have a look at what exactly is going on because this made no sense on the initial read.

test/test_runner.py
```diff
@@ -0,0 +1,36 @@
...
+be elsewhere in the file depending on the exact syntax problem.
+
+The offending line appears to be:
+
+
+    - name: always run
+      ^ here
+"""  # noqa: E501
+
+
```
tadeboro, line 21: We are missing at least a test for the use case where the exit code is not 4. It would be nice to also test stdout stuff, but that part is not that important.

test/test_runner.py
```diff
@@ -0,0 +1,36 @@
...
+The error appears to be in '/foo/example.yml': line 47, column 7, but may
+be elsewhere in the file depending on the exact syntax problem.
+
+The offending line appears to be:
+
+
+    - name: always run
+      ^ here
+"""  # noqa: E501
```
tadeboro, line 19: This should be part of the test case since nothing else uses it.

test/using-bare-variables-success.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - hosts: localhost
-  become: no
```
tadeboro, line 3: All changes in this file are not related to the changes implemented in the PR and should not be part of the changeset.

tadeboro (changes requested):
The high-level structure of changes is OK now, but there are still a few implementation details before this is ready for prime time.

lib/ansiblelint/errors.py
```diff
@@ -43,6 +44,7 @@ def __init__(
 
         self.message = message or getattr(rule, 'shortdesc', "")
         self.linenumber = linenumber
+        self.column = column
```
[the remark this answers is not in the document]
felixfontein: Also the implementation of `__eq__` is flawed, it should compare `_hash_key`, not `__hash__()`. Probably makes sense to fix it in another PR though, since it has already been broken in the past.

lib/ansiblelint/runner.py
```diff
@@ -100,6 +115,27 @@ def run(self) -> List[MatchError]:
...
+            if file['type'] == 'playbook':
+                result = subprocess.run(
+                    ['ansible-playbook', '--syntax-check', file['path']],
+                    stdin=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    shell=False,  # needed when command is a list
+                    universal_newlines=True,
+                    check=False
+                )
```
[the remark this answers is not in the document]
felixfontein: I think that it returns at most one error is an implementation detail (which could change), so keeping this code here flexible to allow multiple errors is preferrable IMO.

lib/ansiblelint/errors.py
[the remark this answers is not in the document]
tadeboro: Comparing hashes is OK in this case since if keys are the same, the hashes will also be the same. It is a bit inefficient, but still current as far as the end-result goes.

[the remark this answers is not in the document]
felixfontein: If you are assuming that hash collisions will not happen, it's correct. But Python's `hash` is not designed for strong collision resistance, so it should really not be used for this. Especially since most data structures which use hashes (like `set`, `dict`, etc.) assume that they can use equality to distinguish between objects which have the same hash value.

lib/ansiblelint/runner.py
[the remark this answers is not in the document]
tadeboro: As the code stands now, it is confusing and smells like premature optimization. Since this is a private API, breaking it is not a big deal once we have a use case for that. And we also gain a meaningful commit that explains why we changed from single to multiple returns.

We can also add a comment describing how we are preparing the function for future use. Still, I am pretty sure Ansible will not report multiple errors any time soon (continuing parsing after the first error is discovered can be quite hard and not worth the complications in scenarios such as Ansible).

lib/ansiblelint/errors.py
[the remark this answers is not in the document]
tadeboro: Ugh, that is true. Highly unlikely, but still true.

lib/ansiblelint/runner.py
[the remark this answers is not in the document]
ssbarnea: @tadeboro I am still working on refactoring the Rule class methods to introduce type checks. Two more methods to go. I will do these before introducing this new rule.

I will first finish adoption of the Lintable object use before rebasing this change, we still have few places where we pass the old `(filename, type)` tuple.

Please keep posting feedback as this helps me with preparatory work.

test/test_runner.py
webknjaz: While this sounds reasonable, there's a reason to keep it here: it's harder to preserve internal string nesting in nested code blocks. So keeping it outside improves maintainability.

lib/ansiblelint/errors.py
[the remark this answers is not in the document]
felixfontein: If you're using CPython on a 32-bit system, and manage to run ansible-lint on a large codebase with ~65536 errors, the chances of a collision is pretty high (birthday paradox). I agree that the chances of this happening are not that high, but if enough people run ansible-lint for enough many times, at some point it will happen. And since the code is more efficient without the hash, it's definitely worth getting rid of it :)

lib/ansiblelint/runner.py
ssbarnea: @tadeboro Shortly: historical reasons. As a note only rules are like this, all the other modules are lowercase. If you have time to look into a better solution I would not mind refactoring them.

lib/ansiblelint/runner.py
```diff
@@ -77,29 +77,37 @@ def is_excluded(self, file_path: str) -> bool:
     def run(self) -> List[MatchError]:
         """Execute the linting process."""
         files: List[Lintable] = list()
+        matches: List[MatchError] = list()
```
tadeboro, line 80: Nitpick: is there a reason why are we using a list here instead of a set? We do not want to have duplicates, so we might as well use set here and call `matches.union` instead of `matches.extend`.

lib/ansiblelint/runner.py
```diff
@@ -77,29 +77,37 @@ def is_excluded(self, file_path: str) -> bool:
...
+
+                matches.extend(
+                    self.rules.run(file, tags=set(self.tags),
+                                   skip_list=self.skip_list))
+
         # update list of checked files
         self.checked_files.update([str(x.path) for x in files])
 
-        return sorted(matches)
+        return sorted(list(set(matches)))
```
tadeboro, line 110: Nitpick: `sorted` know how to handle sets (or any other iterabel), so no need to convert to list before sorting.

tadeboro (approved):
I only found two nitpicks in the code. Feel free to ignore them and merge this PR as is.

lib/ansiblelint/runner.py
ssbarnea: The reason why I did kept list is because I wanted to avoid changing the signature. Anyway, I am considering more refactoring in that area and maybe adopting a container that can hide the boring parts like unique, sort and maybe even retrieving stats or subsets.

lib/ansiblelint/runner.py
ssbarnea: True, i will fix it in follow-up, I am sure that I will touch the same method very soon, before the release.

PR #951, opened 2020-08-13
Refactored import_playbook tests

ssbarnea:
Updated our import_playbook tests to match Ansible implementation of import_playbook and to avoid hitting Ansible syntax check failures with our own testing playbooks.

Ansible can only import playbooks at top level, not via tasks thus our testing was incorrect. 

Needed-By: #950


test/TestImportPlaybook.py
```diff
@@ -0,0 +1,20 @@
+"""Test ability to import playbooks."""
+from ansiblelint.runner import Runner
+
+
+def test_task_hook_import_playbook(default_rules_collection):
+    """Assures import_playbook includes are recognized."""
+    playbook_path = (
+        'test/playbook-import/playbook_parent.yml'.
+        format_map(locals())
```
awcrosby, line 9: what is the purpose of this since there are no variable placeholders in the path string?

ssbarnea: Thanks. Fixed it already, leftover from previous version.

test/playbook-import/playbook_imported.yml
```diff
@@ -3,7 +3,7 @@
   connection: local
   gather_facts: no
   tasks:
-  - command: echo "no name"  # noqa 502
+  - command: echo "no name"  # should generate 502
```
rafaelfolco, line 6: 502 expected check

test/playbook-include/playbook_post.yml
```diff
@@ -1,8 +0,0 @@
----
-- hosts: localhost
-  tasks:
-    - command: echo "no name"  # noqa 502
-  post_tasks:
```
rafaelfolco, line 5: no longer importing as post_tasks, hope this does not lose coverage

test/playbook-include/playbook_pre.yml
```diff
@@ -1,8 +0,0 @@
----
-- hosts: localhost
-  tasks:
-    - command: echo "no name"  # noqa 502
-  pre_tasks:
```
rafaelfolco, line 5: ditto

test/playbook-include/playbook_post.yml
ssbarnea: Yeah. This is what I wanted.

PR #952, opened 2020-08-13
Add missing hosts to test files

ssbarnea:
Ansible requires hosts to be present in playbooks and in order to be able pass ansible syntax check we need to add it to our tests.

Needed-By: #950

### DIFF
--- a/test/jinja2-when-success.yml
+++ b/test/jinja2-when-success.yml
@@ -3,8 +3,6 @@
     - name: test when
       debug: msg=text
       when: true
-
-- hosts: all
-  roles:
-    - role: role
+    - name: test when 2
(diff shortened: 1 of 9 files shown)


test/jinja2-when-success.yml
```diff
@@ -3,8 +3,6 @@
     - name: test when
       debug: msg=text
       when: true
-
-- hosts: all
```
rafaelfolco, line 7: switching from all to localhost, I assume this is the pattern now

ssbarnea: It does not matter as long we are using values that are valid with implicit inventory. We do not really run them.

PR #954, opened 2020-08-13
Reduce complexity of runner main loop

ssbarnea:
Make use of a generator for finding MatchErrors.

### DIFF
--- a/lib/ansiblelint/runner.py
+++ b/lib/ansiblelint/runner.py
@@ -1,7 +1,7 @@
 """Runner implementation."""
 import logging
 import os
-from typing import TYPE_CHECKING, Any, FrozenSet, List, Set
+from typing import TYPE_CHECKING, Any, FrozenSet, Generator, List, Set
 
 import ansiblelint.file_utils
 import ansiblelint.skip_utils
@@ -64,7 +64,7 @@ class Runner(object):
         # not excluded.
         return any(file_path.startswith(path) for path in self.exclude_paths)
 
-    def run(self) -> List:
+    def run(self) -> List[MatchError]:
         """Execute the linting process."""
         files = list()
         for playbook in self.playbooks:
@@ -75,21 +75,7 @@ class Runner(object):
                           # add an absolute path here, so rules are able to validate if
                           # referenced files exist
                           'absolute_directory': os.path.dirname(playbook[0])})
-        visited: Set = set()
-        matches = list()
-
-        while visited != self.playbooks:
-            for arg in self.playbooks - visited:
-                try:
-                    for child in ansiblelint.utils.find_children(arg, self.playbook_dir):
-                        if self.is_excluded(child['path']):
-                            continue
-                        self.playbooks.add((child['path'], child['type']))
-                        files.append(child)
-                except MatchError as e:
-                    e.rule = LoadingFailureRule
-                    matches.append(e)
-                visited.add(arg)
+        matches = set(self._emit_matches(files))
 
         # remove duplicates from files list
         files = [value for n, value in enumerate(files) if value not in files[:n]]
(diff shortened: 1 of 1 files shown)


lib/ansiblelint/runner.py
```diff
@@ -101,9 +87,25 @@ def run(self) -> List:
...
+            matches = matches.union(
+                self.rules.run(file, tags=set(self.tags),
+                               skip_list=self.skip_list))
         # update list of checked files
         self.checked_files.update([x['path'] for x in files])
 
-        return sorted(set(matches))
+        return sorted(matches)
+
+    def _yield_matches(self, files: List) -> Generator[MatchError, None, None]:
```
webknjaz, line 98: The name should be made better. yield matches only answered the question about the return type not providing any details oh how it does that, based on what etc. "yield" is an implementation detail that doesn't really improve the semantical reference...

P.S. I usually prefer using "emit" to signal that the function is a generator.

lib/ansiblelint/runner.py
```diff
@@ -101,9 +87,25 @@ def run(self) -> List:
...
+        visited: Set = set()
+        while visited != self.playbooks:
+            for arg in self.playbooks - visited:
+                try:
+                    for child in ansiblelint.utils.find_children(arg, self.playbook_dir):
+                        if self.is_excluded(child['path']):
+                            continue
+                        self.playbooks.add((child['path'], child['type']))
+                        files.append(child)
```
webknjaz, line 107: What exactly can throw an exception here? Just wrap that part with try/except...

lib/ansiblelint/runner.py
webknjaz: It seems like this doesn't run any checks except for loading-related ones. This could be reflected too.

lib/ansiblelint/runner.py
```diff
@@ -75,21 +75,7 @@ def run(self) -> List:
...
-                    for child in ansiblelint.utils.find_children(arg, self.playbook_dir):
-                        if self.is_excluded(child['path']):
-                            continue
-                        self.playbooks.add((child['path'], child['type']))
-                        files.append(child)
-                except MatchError as e:
-                    e.rule = LoadingFailureRule
-                    matches.append(e)
-                visited.add(arg)
+        matches = set(self._yield_matches(files))
```
webknjaz, line 78: :+1:

lib/ansiblelint/runner.py
ssbarnea: sure.

lib/ansiblelint/runner.py
```diff
@@ -101,9 +87,25 @@ def run(self) -> List:
...
+            matches = matches.union(
+                self.rules.run(file, tags=set(self.tags),
+                               skip_list=self.skip_list))
         # update list of checked files
         self.checked_files.update([x['path'] for x in files])
 
-        return sorted(set(matches))
+        return sorted(matches)
+
+    def _emit_matches(self, files: List) -> Generator[MatchError, None, None]:
```
webknjaz, line 98: Oh! It looks like the semantical goal of this block is loading the playbooks. So it'd probably be better to say
```suggestion
    def _emit_playbook_load_errors(self, files: List) -> Generator[MatchError, None, None]:
```

lib/ansiblelint/runner.py
ssbarnea: I kinda like the ability to raise MatchErrors from anywhere but more important: i do not want to change the logic with this change. I can consider changing the try/catch coverage in a follow-up.

I want to keep this change simple, and minimize risk of unexpected side-effects.

lib/ansiblelint/runner.py
ssbarnea: It would be incorrect load errors is just some of the possible MatchErrors, it would be misleading.

webknjaz: FWIW it is misleading now.

Issue #953, opened 2020-08-13
Make use of yamllint for generic YAML issues


ssbarnea:
##### Summary
Few weeks back I opened the subject on https://www.reddit.com/r/ansible/comments/hvo67m/should_ansiblelint_also_call_yamllint/ as I think our linter should not perform tasks that are already covered by YAML generic tools.

##### Issue Type

- Feature Idea

##### Additional Information

Instead of re-implementing rules that already exist in yamllint, like below, we should be able to just call yamllint and combine its reports with our own.

* https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/LineTooLongRule.py
* https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/NoTabsRule.py
* https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/TrailingWhitespaceRule.py

While we should have our own default YAML style settings, we should also respect any in-repository .yamllint configuration.

In order to avoid adding yamllint as a permanent dependency, this feature should be implemented as optional, if yamllint module is installed, it will be used. Even in this case we should allow user to to use skips.

For example I am aware of a very useful feature in yamllint the ability to detect duplicate map keys.

ansiblejunky:
We should consider the overall design strategy for ansible-lint. ansible-lint needs to detect various file types that are supported by Ansible (python, yaml, jinja) and even further detect types of yaml files (playbooks, tasks, variable files, meta, handlers).  Once we detect all of that, ansible-lint should allow using embedded rules (rules created and included with ansible-lint specifically to ensure ansible best practices are met).

Typically, when I need to run yamllint rules, I just add that to the CI pipeline or pre-commit configuration file and its done. I don't need ansible-lint configured to use yamllint.  So from this standpoint, I'm not sure ansible-lint should use yamllint. Rather let folks use yamllint if they want to, but ansible-lint should test ansible best practices.

Having these rules inside ansible-lint has the benefit of being in our control and perhaps even being "better" suited for ansible in some way. Having them external (managed in yamllint) and using default settings still requires testing and synchronizing between ansible-lint and yamllint.  In fact, as we branch out to use external linters it will also require alignment with other external tools like flask8 and jinjalint etc. 

You could argue that Ansible Engine itself was "batteries included" and it caused difficulties with orchestrating releases. I think adding dependencies on yamllint, jinjalint and flask8 is more or less adding the batteries.  

[...]

PR #955, opened 2020-08-14
Add rule violations from Yamllint

ssbarnea:
This detect yamllint rule violations when this is found as installed. This will allow us to remove a set of rules that
overlapped with general YAML syntax ones.

Fixes: #953
Release-Notes: https://github.com/ansible-community/ansible-lint/discussions/1150


lib/ansiblelint/rules/__init__.py
```diff
@@ -160,6 +161,19 @@ def matchyaml(self, file: str, text: str) -> List[MatchError]:
         return matches
 
 
+class YamllintRule(AnsibleLintRule):
```
greg-hellings, line 164: Why does it look like you've added this twice?

felixfontein:
I've tried this PR in a [role-only collection of mine](https://github.com/felixfontein/ansible-acme/), and found it ignores my .yamllint file.

ssbarnea:
@felixfontein Thanks for looking at it. Yep, I am aware that it ignores the in-repo yamllint config and that it uses a hardcoded configuration.

That is more or less by design as I wanted to fully control the results. I can see how this could get me into trouble and I guess I will have to make it use the embedded default only when a local config is not found.

One thing that you likely already know is that most Ansible content is not compatible with the yamllint-own-default-config and that the author has very strong opinions regarding what defaults should be.

Another feature that is currently missing is ability to fully disable this rule instead of ignoring or skipping it.

I do not expect to merge this very soon but getting early feedback is essential, as it would allow me to improve the implementation.

felixfontein:
@ssbarnea being able to disable that would be really helpful :) Does it make sense to make this configurable? I don't know if ansible-lint has other rules that can be configured (and not just disabled).

ssbarnea:
> @ssbarnea being able to disable that would be really helpful :) Does it make sense to make this configurable? I don't know if ansible-lint has other rules that can be configured (and not just disabled).

Now is configurable, it will respect user defined config. You can also skip entire rules with "YAML" tag, or specific ones with their named-tag reported by yamllint itself.

AFAIK, it is ready to be merged. I am just waiting for an approval. Once merged I will tag a pre-release in order to allow users to start consuming the newer changes.

.ansible-lint
```diff
@@ -1,2 +1,9 @@
+extends: default
+rules:
```
felixfontein, line 2: This looks a lot like a yamllint config. Who is reading and processing this part (both `extends` and `rules`)?

ssbarnea: @felixfontein Thanks, that was added by mistake, I removed it.

felixfontein (approved):

felixfontein:
@ssbarnea thanks for implementing this!

ssbarnea:
@felixfontein Thanks for reviewing it. I also have https://github.com/ansible-community/ansible-lint/pull/950/files ready for review, which is kinda similar: making use of ansible syntax check.

lib/ansiblelint/rules/YamllintRule.py
```diff
@@ -0,0 +1,94 @@
...
+  # 160 chars was the default used by old E204 rule, but
+  # you can easily change it or disable in your .yamllint file.
+  line-length:
+    max: 160
+"""
+
+DESCRIPTION = """\
+Rule violations reported by YamlLint when this is installed.
+
+You can fully disable all of them by adding 'YAML' to the 'skip_list'.
```
MarkusTeufelberger, line 35: Just to be sure: This is supposed to with `YAML` in the skip_list, not with `yaml`? To me it seems like the other way around?

ssbarnea: That is an error, correct key is 'yaml', I will fix it.

ssbarnea: Fixing via https://github.com/ansible-community/ansible-lint/pull/1335

PR #956, opened 2020-08-14
Ease creation of MatchError from inside rules

ssbarnea:
Adds utility method for creation of MatchError(s) from inside Rules, so we can easily reuse it. That is needed by #923 where we do want to start returning MatchError(s).

The expectation is to have 4-6 calls of the method once refactoring of all methods is done.

### DIFF
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -42,6 +42,20 @@ class AnsibleLintRule(object):
         text = re.sub(r"{#.+?#}", "JINJA_COMMENT", text)
         return text
 
+    def create_matcherror(
+            self,
+            message: str = None,
+            linenumber: int = 0,
+            details: str = "",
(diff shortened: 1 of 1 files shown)


webknjaz (changes requested):
I don't see a point of having another level of indirection. This is just a proxy-method and I can't imagine how I could justify merging this change.

ssbarnea:
> I don't see a point of having another level of indirection. This is just a proxy-method and I can't imagine how I could justify merging this change.

Updated description to mention reasoning. While it does not look as required for current use, it is needed for upcoming work. This change is a spinoff.

webknjaz:
I still don't see a point: if you want to instantiate a `MatchError` go and do it. Adding a proxy method that only bypasses the args to the initializer doesn't contribute to simplicity in any way.

greg-hellings:
It's already reduced code redundancy here. If this method ends up being used over and over in future work, then it continues to save redundancy each time.

PR #957, opened 2020-08-14
Reorganize test requirements

ssbarnea:
- Move test requirements from tox.ini to a test extra
- Removes unused test requirements
- Bumps test requirements to decrease change of running outdated
  version while developing. Avoids  https://github.com/pytest-dev/pytest-cov/issues/422

This enables running of tests w/o tox, with just `pip install -e .[test]` or similar, something that was not possible before.

Update: Please check https://discuss.python.org/t/is-it-preferable-to-add-test-dependensies-as-an-extra-require-instead-of-tox-ini/4984 thread if you have doubts if this approach may not be good.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,6 +78,14 @@ install_requires =
   # NOTE: per issue #509 0.15.34 included in debian backports
   typing-extensions; python_version < "3.8"
 
+[options.extras_require]
+test =
+  pytest >= 6.0.1
+  pytest-cov >= 2.10.1
+  pytest-xdist >= 2.1.0
+  # Needed to avoid DeprecationWarning errors in pytest:
+  setuptools >= 49.6.0
+
 [options.entry_points]
 console_scripts =
   ansible-lint = ansiblelint.__main__:main
--- a/tox.ini
+++ b/tox.ini
@@ -20,15 +20,8 @@ deps =
   # https://github.com/ansible/ansible/issues/70705
   ansibledevel: ansible>=2.10.0a1,<2.11
   ansibledevel: ansible-base @ git+https://github.com/ansible/ansible.git
-  ruamel.yaml==0.16.5  # NOTE: 0.15.34 has issues with py37
-  flake8
-  pep8-naming
-  pytest
-  pytest-cov
-  pytest-xdist
-  # Needed to avoid DeprecationWarning errors in pytest:
-  setuptools >= 40.4.3
-  wheel
+extras =
+  test
 commands =
   # safety measure to assure we do not accidentaly run tests with broken dependencies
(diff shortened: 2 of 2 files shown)


pabelanger:
recheck

ssbarnea:
recheck

setup.cfg
```diff
@@ -77,6 +77,16 @@ install_requires =
...
 
+[options.extras_require]
+test =
+  pytest >= 6.0.1
+  pytest-cov >= 2.10.0
+  # https://github.com/pytest-dev/pytest-cov/issues/422
+  pytest-xdist >= 1.34.0,<2.0.0
+  # Needed to avoid DeprecationWarning errors in pytest:
+  setuptools >= 40.4.3
+  wheel >= 0.35.0
```
webknjaz, line 88: This is weird.

ssbarnea: Not sure what is weird, clearly that was not really an actionable remark.

webknjaz: runtime dependency on wheel/setuptools (even optional) is weird.

webknjaz:
@greg-hellings we are not talking about any tools here. @ssbarnea is trying to shove tests execution virtualenv deps into the metadata of a pip-installable distribution that is completely unrelated to setting up the test environment.

webknjaz (changes requested):
Test deps should be (1) installable w/o the dist itself. They should be pinned (2). And those pins should be auto-updatable by tools like dependabot and similar. I don't see how adding them to dist metadata would help here. I'd say that it's harmful to add them to runtime deps. Besides, influencing `setuptools` in the virtualenv is not a good idea either.

ssbarnea:
> Test deps should be (1) installable w/o the dist itself. They should be pinned (2). And those pins should be auto-updatable by tools like dependabot and similar. I don't see how adding them to dist metadata would help here. I'd say that it's harmful to add them to runtime deps. Besides, influencing `setuptools` in the virtualenv is not a good idea either.

I already considered that after you introduced them to docs. Still I have some concerns that too tight control over them would translate into few additional issues:
- conflicts where a pinned set of reqs fails on specific platform (one not covered by CI)
- extra noise from dependabot updates -- i really do *not* want to see dependabot updating setuptools every other day. Now I want to use a pinned version of setuptools, as it would prevent us from discovering bugs with newer ones while testing.
- two extra files in repository root, a place already crowded

Note that I am not strongly against dependabot approach. I only stated my concerns and why I prefer using extras with range instead of pinning. Still, I am willing to accommodate other preferences. Lets see what others think, adding feedback-needed label.

ssbarnea:
Dropped in favour of #1011 alternative

PR #961, opened 2020-08-17
4.3.0 Release Notes

ssbarnea:
Copied release notes from github/releases page and documented
location where live versions can be read.

Removes outdated ROADMAP.rst file which was not included in the docs.


ssbarnea:
> LGTM unless there's something else that needs to be added that I don't know about :)

Not that I am aware of and I am sure that it will not take a lot of time to make a 4.3.1 after the release hits pypi. Still, I will try to avoid another 6 month gap between releases. I would personally aim to at least one release per month.

PR #968, opened 2020-08-18
Allow namespace dot delimiter in role names

nre-ableton:


ssbarnea:
Documentation is clear that dot is not allowed. See https://galaxy.ansible.com/docs/contributing/creating_role.html#role-names

Issue #936, opened 2020-08-02
Provide reason with rules


SISheogorath:
##### Summary

Provide a reason along the side with the explanation how to solve the linting error in order to allow users to make informed decisions about the linting error.

##### Issue Type

- Feature Idea

##### Additional Information

Currently the default rules simply state what is wrong and how to fix it. What they don't tell is why it's wrong or simply worth checking. For example: `403: Package installs should not use latest` simply states that `Package installs should use state=present with or without a version` but it doesn't tell why. This leaves me as a user in the position: Should I follow the linters advice or simply add an exception for the rule? If it would tell `Package installs should use state=present with or without a version because it can cause instability in the setup` or `Package installs should use state=present with or without a version because state=latest can harm makes the role no longer idempotent`.

I guess I simply miss a "because" section for the default rules and it would be great to get it, as it would help to transfer knowledge and provide the ability to make better informed decisions on why a linter rule exists and whether something is intended or unintended behaviour.

bendem:
This. Linting allows you to discover things you do wrong, if there is no reasoning, you can't learn why it's wrong, you can't make a decision, it becomes cargo-culting.

matkoniecz:
It would be very useful. For example I got

"Package installs should not use latest"

on role that have as whole point to upgrade `pip` and `youtube-dl`.

I am not sure what is supposed to be wrong with them. If the reason is "it can cause instability in the setup"  then I would happily ignore it - as in this case whole point is to introduce instability (AKA upgrade them as needed).

But if `latest` should not be used because it is buggy, misnamed or something then I would care about it.

Issue #966, opened 2020-08-18
E106: (Role name {} does not match pattern) should not be applied to non-collection roles


nre-ableton:
##### Summary

The explanation for violation 106 links to [this Ansible documentation page](https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#roles-directory), which details the requirements for _collection role_ names. However, [higher up in that same page](https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#galaxy-yml), a collection role is defined as:

> A collection must have a `galaxy.yml` file that contains the necessary information to build a collection artifact. See Collection Galaxy metadata structure for details.

It seems to me that rule 106 should not be enforced in a `galaxy.yml` file is not present in the role's root directory. If such a file is absent, then the role is just a regular role and not a collection role, and these naming requirements should not apply.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details
ansible 2.9.12
ansible-lint 4.3.0

- ansible installation method: pip
- ansible-lint installation method: pip

ssbarnea:
@nre-ableton You are welcomed to propose a PR for this. I personally did not bother because it was really easy to add exclusion.

I would even go so far to believe that this rule is benefit even for those that do not have yet a collection, as it would help them add new roles that could be packaged at some point, instead of creating more tech-debt.

Keep in mind: that applies to private roles too, not only those published to galaxy server.

nre-ableton:
@ssbarnea Yes, that's correct. I didn't mention this, but I discovered this bug with a project that has several private roles which are not published on Galaxy. I think it would be useful for this case as well.

I'll see if I can come up with a PR to solve it. I'm currently having a bit of trouble figuring out how to run the tests locally etc., but I'll try to sort it out when I have a spare moment.

geerlingguy:
Yeah... now I'm getting this error on every single one of my roles, for example: https://travis-ci.org/github/geerlingguy/ansible-role-solr/jobs/719707367#L489

I'm aware I can add a skip to this rule, but it seems like a very radical departure from this project's goal to, in one minor release, immediately cancel out the ability to use it with the 25,000+ roles that are on Galaxy and have them fail on a rule that's only supposed to apply to collection roles.

geerlingguy:
[...]

PR #971, opened 2020-08-18
Indicate how to skip rules on failures

ssbarnea:
Make it easier for users to disable rules that they do not find fit by displaying a sample configuration file after the execution.

example-output

Note that his change adds [rich](https://github.com/willmcgugan/rich) library as a dependency as we also plan to use it for improve terminal output in other areas like rule listing or more detailed explanation about why a rule was added (#936). Library is MIT and under good maintenance.

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -25,7 +25,10 @@ import logging
 import os
 import pathlib
 import sys
-from typing import TYPE_CHECKING, Set, Type
+from typing import TYPE_CHECKING, List, Set, Type
+
+from rich.console import Console
+from rich.markdown import Markdown
 
 from ansiblelint import cli, formatters
 from ansiblelint.constants import DEFAULT_RULESDIR
@@ -37,7 +40,10 @@ from ansiblelint.utils import get_playbooks_and_roles
 if TYPE_CHECKING:
     from argparse import Namespace
 
+    from ansiblelint.errors import MatchError
+
 _logger = logging.getLogger(__name__)
+console = Console()
 
 
 def initialize_logger(level: int = 0) -> None:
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/__main__.py
```diff
@@ -74,6 +80,22 @@ def choose_formatter_factory(
     return r
 
 
+def hint_about_skips(matches: List["MatchError"]):
+    """Display information about how to skip found rules."""
+    msg = """\
+Do **not** file bugs for rule violation just because they do not apply for your own
+use-case. You can skip rules using a configuration like below:
```
btravouillon, line 87: This message can be misleading. It might happen a new rule does not behave as expected. It's not a problem to skip a rule if it highlights a problem and the dev does not care.
It is a problem when a default rule reports false positive (see #966). Thus I would not discourage people to file bugs about rules implementation.

ssbarnea: @btravouillon Can you please suggest another sentence? The idea is indeed about discouraging users from creating bug reports right away. Obviosuly that genuine bugs are ok to file.

odyssey4me: I would suggest the following text:

``If you would prefer to skip this rule, then use the following configuration for ansible-lint:``

lib/ansiblelint/__main__.py
```diff
@@ -74,6 +80,21 @@ def choose_formatter_factory(
     return r
 
 
+def hint_about_skips(matches: List["MatchError"]):
+    """Display information about how to skip found rules."""
+    msg = """\
+If you would prefer to skip some rules, then use the following configuration:
```
acozine, line 86: How about this?
```suggestion
You can skip specific rules by adding them to the skip_list section of your configuration file:
```

PR #972, opened 2020-08-18
E106: Role name should be the parent dir of tasks

btravouillon:
For E106: Identify the rolename as the parent directory of tasks.

The `roles` directory is not reliable since it can be overridden with
`roles_path`. On the other hand, a role always have a `tasks`
subdirectory.

### DIFF
--- a/lib/ansiblelint/rules/RoleNames.py
+++ b/lib/ansiblelint/rules/RoleNames.py
@@ -48,8 +48,8 @@ class RoleNames(AnsibleLintRule):
     def match(self, file, text):
 
         path = file['path'].split("/")
-        if "roles" in path:
-            role_name = path[path.index("roles") + 1]
+        if "tasks" in path:
+            role_name = path[path.index("tasks") - 1]
             if role_name in self.done:
                 return False
             self.done.append(role_name)


ssbarnea (approved):

ssbarnea:
I will not merge this without extra eyes on it because I see the danger of identifying as roles folders that are not really roles. What about users using `include_tasks` with their standalone playbooks? 

Does this mean that the folder hosting the playbook will be falsely reported as a role?

ssbarnea (approved):

btravouillon:
> What about users using `include_tasks` with their standalone playbooks?
> Does this mean that the folder hosting the playbook will be falsely reported as a role?

That's a possibility, but in such case it should not be an issue to skip rule E106 which is about role names.

In a project I'm working with, they are using `roles_path` with a tree similar to: `roles/{product1,product2,product3}/any_role_name/`. I would rather use Collections, but it is what it is...

E106 must test `any_role_name`, not `productX`.

By the way, if any issue should be filed about this change, feel free to ping me and I'll look at it.

Thanks for the merge!

ssbarnea:
I am aware about nested layouts and other similar over-engineering pieces. I still work to flatten them and my plan is to add a new rule that enforces the collection layout, any role found not in `ROOT/roles` will trigger it.

Being standard means less bugs.

Issue #965, opened 2020-08-18
Print violation description in output


nre-ableton:
##### Summary

I find ansible-lint's output to be difficult to work with. In order to fix violations, I need to have two terminal windows open: one with `ansible-lint -L | less` and the other with the output of `ansible-lint` itself. I really wish that ansible-lint had an option to print out the violation summary in the output, so I wouldn't need to keep flipping back-and-forth between windows to look up the codes.

Apologies if this has been discussed before, I did a few searches of this project's issues and also manually scanned over currently open issues and didn't see anything. Also, apologies if I am missing something very obvious like a CLI switch which would provide this data. 😅 

##### Issue Type

- Feature Idea

ssbarnea:
I agree with that needs to be done but changing default output of `-L` is problematic as we do not want to lose ability to have a summary of all rules.

I also think that it is not of much usefulness to dump all details about rules, some will soon have quite long descriptions with reasoning, examples and workarounds in them.

How about displaying details of a single rule when the rule id is mentioned like `ansible -L 106` -- if doable with argparse.

Other ideas? The problem here is deciding the CLI API to use, implementation would be quite easy after.

nre-ableton:
Sorry, please forgive the noise on this issue. I didn't realize that the `quiet` switch was active in our `ansible-lint.cfg`, and that this suppressed the rule description in ansible-lint's output. 😅

PR #978, opened 2020-08-21
Made console rule listing rich

ssbarnea:
Adds 'rich' output format for listing rules (-L) and makes it default.

This format is more verbose and includes all known rule metadata in a way that makes it easier to read.

Old format is still supported and accessible using `ansible-lint -L -f plain`, like it was before.

screenshot

Fixes: #965

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -27,11 +27,11 @@ import pathlib
 import sys
 from typing import TYPE_CHECKING, List, Set, Type
 
-from rich.console import Console
 from rich.markdown import Markdown
 
 from ansiblelint import cli, formatters
-from ansiblelint.generate_docs import rules_as_rst
+from ansiblelint.color import console
+from ansiblelint.generate_docs import rules_as_rich, rules_as_rst
 from ansiblelint.rules import RulesCollection
 from ansiblelint.runner import Runner
 from ansiblelint.utils import get_playbooks_and_roles, get_rules_dirs
@@ -42,7 +42,12 @@ if TYPE_CHECKING:
     from ansiblelint.errors import MatchError
 
 _logger = logging.getLogger(__name__)
-console = Console()
+
+_rule_format_map = {
+    'plain': str,
+    'rich': rules_as_rich,
+    'rst': rules_as_rst
+}
 
 
 def initialize_logger(level: int = 0) -> None:
@@ -123,8 +128,9 @@ def main() -> int:
     rules = RulesCollection(rulesdirs)
 
     if options.listrules:
-        formatted_rules = rules if options.format == 'plain' else rules_as_rst(rules)
-        print(formatted_rules)
+        console.print(
+            _rule_format_map[options.format](rules),
+            highlight=False)
         return 0
 
     if options.listtags:
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -96,8 +96,8 @@ def get_cli_parser() -> argparse.ArgumentParser:
 
     parser.add_argument('-L', dest='listrules', default=False,
                         action='store_true', help="list all the rules")
-    parser.add_argument('-f', dest='format', default='plain',
-                        choices=['plain', 'rst'],
+    parser.add_argument('-f', dest='format', default='rich',
+                        choices=['rich', 'plain', 'rst'],
                         help="Format used rules output, (default: %(default)s)")
     parser.add_argument('-q', dest='quiet',
                         default=False,
--- a/lib/ansiblelint/color.py
+++ b/lib/ansiblelint/color.py
@@ -1,5 +1,16 @@
-"""Console coloring support."""
+"""Console coloring and terminal support."""
 from enum import Enum
+
+from rich.console import Console
+from rich.theme import Theme
+
+_theme = Theme({
+    "info": "cyan",
+    "warning": "dim yellow",
+    "danger": "bold red",
+    "title": "yellow"
+})
+console = Console(theme=_theme)
 
 
 class Color(Enum):
--- a/lib/ansiblelint/generate_docs.py
+++ b/lib/ansiblelint/generate_docs.py
@@ -1,5 +1,11 @@
 """Utils to generate rule table .rst documentation."""
 import logging
+from typing import Iterable
+
+from rich import box
+from rich.console import render_group
+from rich.markdown import Markdown
+from rich.table import Table
 
 from ansiblelint.rules import RulesCollection
 
@@ -41,3 +47,20 @@ def rules_as_rst(rules: RulesCollection) -> str:
         r += f"\n\n.. _{d.id}:\n\n{title}\n{'*' * len(title)}\n\n{d.description}"
 
     return r
+
+
+@render_group()
+def rules_as_rich(rules: RulesCollection) -> Iterable[Table]:
+    """Print documentation for a list of rules, returns empty string."""
+    for rule in rules:
+        table = Table(show_header=True, header_style="title", box=box.MINIMAL)
+        table.add_column(rule.id, style="dim", width=16)
+        table.add_column(Markdown(rule.shortdesc))
+        table.add_row("description", Markdown(rule.description))
+        if rule.version_added:
+            table.add_row("version_added", rule.version_added)
+        if rule.tags:
+            table.add_row("tags", ", ".join(rule.tags))
+        if rule.severity:
+            table.add_row("severity", rule.severity)
+        yield table
--- a/test/TestRulesCollection.py
+++ b/test/TestRulesCollection.py
@@ -25,6 +25,8 @@ import pytest
 
 from ansiblelint.rules import RulesCollection
 
+from . import run_ansible_lint
+
 
 @pytest.fixture
 def test_rules_collection():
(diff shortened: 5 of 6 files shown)


lib/ansiblelint/generate_docs.py
```diff
@@ -41,3 +45,19 @@ def rules_as_rst(rules: RulesCollection) -> str:
         r += f"\n\n.. _{d.id}:\n\n{title}\n{'*' * len(title)}\n\n{d.description}"
 
     return r
+
+
+def rules_as_rich(rules: RulesCollection) -> str:
+    """Print documentation for a list of rules, returns empty string."""
+    con = console.Console()
+    for d in rules:
+        table = Table(show_header=True, header_style="bold magenta", box=box.MINIMAL)
```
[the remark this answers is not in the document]
ssbarnea: Done with yellow.

ssbarnea:
> You might want to think about Rich [themes](https://rich.readthedocs.io/en/latest/style.html#style-themes) at some point. It can help if you have a number of common styles in the project that you can edit in one place.

Done!

lib/ansiblelint/__main__.py
```diff
@@ -19,19 +19,18 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 """Command line implementation."""
-
```
webknjaz, line 22: It's unrelated, besides it reads better when there's separators.

lib/ansiblelint/color.py
```diff
@@ -1,6 +1,17 @@
-"""Console coloring support."""
+"""Console coloring and terminal support."""
 from enum import Enum
 
+from rich.console import Console
+from rich.theme import Theme
+
+_theme = Theme({
+    "info": "cyan",
+    "warning": "dim yellow",
```
webknjaz, line 9: Have you tested it in standard-compliant terminal emulators like Kitty? I'm pretty sure this will explode (but needs checking).

lib/ansiblelint/generate_docs.py
```diff
@@ -41,3 +46,18 @@ def rules_as_rst(rules: RulesCollection) -> str:
...
+    """Print documentation for a list of rules, returns empty string."""
+    for d in rules:
+        table = Table(show_header=True, header_style="title", box=box.MINIMAL)
+        table.add_column(d.id, style="dim", width=16)
+        table.add_column(Markdown(d.shortdesc))
+        table.add_row("description", Markdown(d.description))
+        table.add_row("version_added", d.version_added)
+        table.add_row("tags", ", ".join(d.tags))
+        table.add_row("severity", d.severity)
+        console.print(table)
```
webknjaz, line 61: This looks like a hack. Can't you buffer the output and return it?

webknjaz (changes requested):
This needs some tests at least.

lib/ansiblelint/generate_docs.py
ssbarnea: Indeed, I asked on https://github.com/willmcgugan/rich/issues/253 to see if there is a nicer way of achieving that.

lib/ansiblelint/color.py
ssbarnea: Tested with kitty too, worked without any exploding kitterns.

ssbarnea:
I plan to rework this once I figure out how to use rich with the tables. Due the way terminal works is not possible to just return a string.

lib/ansiblelint/cli.py
```diff
@@ -96,8 +96,8 @@ def get_cli_parser() -> argparse.ArgumentParser:
 
     parser.add_argument('-L', dest='listrules', default=False,
                         action='store_true', help="list all the rules")
-    parser.add_argument('-f', dest='format', default='plain',
-                        choices=['plain', 'rst'],
+    parser.add_argument('-f', dest='format', default='rich',
```
greg-hellings, line 99: I'm wary of changing defaults without some sort of "shout it from the rooftop" behavior. Is there a compelling reason to make this part of a z-stream release, or are we expecting it to go with at least a minor version bump?

lib/ansiblelint/generate_docs.py
```diff
@@ -1,6 +1,13 @@
 """Utils to generate rule table .rst documentation."""
 import logging
+from typing import Iterable
 
+from rich import box
+from rich.console import render_group
+from rich.markdown import Markdown
+from rich.table import Table
+
+from ansiblelint.color import console
```
greg-hellings, line 10: You've got some lint on your code!

setup.cfg
```diff
@@ -72,7 +72,7 @@ setup_requires =
 install_requires =
   ansible >= 2.8
   pyyaml
-  rich
+  rich >= 6.0.0
```
greg-hellings, line 75: If you're restricting versions, it probably warrants a comment about why.

lib/ansiblelint/cli.py
ssbarnea: This does not count as a breaking chance as the format was not parseable. Basically changing format, adding colors is not much different than rephrasing the messages.

Another way to put it: if it does not modify existing tests, it does not need a major bump. Still, if others disagree, I can assure that the next version number will 4.4.0. Just add `major` label to the PR if you think is needed.

lib/ansiblelint/__main__.py
```diff
@@ -42,7 +42,12 @@
     from ansiblelint.errors import MatchError
 
 _logger = logging.getLogger(__name__)
-console = Console()
+
+_rule_format_map = {
+    'plain': lambda x: x,  # str upsets mypy
```
webknjaz, line 47: It would be useful to be more specific here

lib/ansiblelint/__main__.py
```diff
@@ -123,8 +128,8 @@ def main() -> int:
     rules = RulesCollection(rulesdirs)
 
     if options.listrules:
-        formatted_rules = rules if options.format == 'plain' else rules_as_rst(rules)
-        print(formatted_rules)
+        r = _rule_format_map[options.format](rules)
```
webknjaz, line 131: Please make the variable name readable

lib/ansiblelint/generate_docs.py
```diff
@@ -41,3 +47,20 @@ def rules_as_rst(rules: RulesCollection) -> str:
         r += f"\n\n.. _{d.id}:\n\n{title}\n{'*' * len(title)}\n\n{d.description}"
 
     return r
+
+
+@render_group()
+def rules_as_rich(rules: RulesCollection) -> Iterable[Table]:
+    """Print documentation for a list of rules, returns empty string."""
+    for d in rules:
```
webknjaz, line 55: What is "d"?

test/TestRulesCollection.py
```diff
@@ -92,3 +94,21 @@ def test_no_duplicate_rule_ids(test_rules_collection):
     real_rules = RulesCollection([os.path.abspath('./lib/ansiblelint/rules')])
     rule_ids = [rule.id for rule in real_rules]
     assert not any(y > 1 for y in collections.Counter(rule_ids).values())
+
+
+def test_rich_rule_listing():
+    rules_path = os.path.abspath('./test/rules')
+    command = [sys.executable, "-m", "ansiblelint", "-r", rules_path, "-f", "rich", "-L"]
```
webknjaz, line 101: Don't we have a fixture helper for this already?

ssbarnea: Not a compatible one. The one we have works only with paths.

We could change it to also cover for this use but i did not want to increase the scope of this change. It is better to do that part as a standalone change.

test/TestRulesCollection.py
```diff
@@ -92,3 +93,17 @@ def test_no_duplicate_rule_ids(test_rules_collection):
...
+
+
+def test_rich_rule_listing():
+    rules_path = os.path.abspath('./test/rules')
+    result = run_ansible_lint("-r", rules_path, "-f", "rich", "-L")
+
+    # The only contract we offer is that the output will contain: id, shortdesc
+    # and description of each rule.
+    for rule in RulesCollection([rules_path]):
+        assert result.returncode == 0
```
webknjaz, line 105: Checking the same value in the loop is pointless. Move it out.

test/TestRulesCollection.py
```diff
@@ -92,3 +93,17 @@ def test_no_duplicate_rule_ids(test_rules_collection):
     real_rules = RulesCollection([os.path.abspath('./lib/ansiblelint/rules')])
     rule_ids = [rule.id for rule in real_rules]
     assert not any(y > 1 for y in collections.Counter(rule_ids).values())
+
+
+def test_rich_rule_listing():
```
webknjaz, line 98:
```suggestion
def test_rich_rule_listing():
    """Test that rich list format output is rendered as a table.

    This check also offers the contract of having rule id, short and long
    descriptions in the console output.
    """
```

webknjaz (changes requested):

test/TestRulesCollection.py
```diff
@@ -17,14 +17,15 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
-
 import collections
```
webknjaz, line 20: revert the separator
```suggestion

import collections
```

webknjaz (approved):

Issue #979, opened 2020-08-21
E106: Role name is not loaded from inside meta/main.yml


ssbarnea:
Galaxy mentions that a role name can be overridden by defining
```
# meta/main.yml
galaxy_info:
  role_name: foo
```

Still, E106 rule looks only for folder names and assumes this is the role name, and this breaks all roles that used old recommendation of naming repositories `ansible-role-{rolename}`.

We need to respect role_name in order to allow role authors to avoid such error.

PR #980, opened 2020-08-21
E106: Avoid false positives with old roles

ssbarnea:
Avoid raising E106 (Role name does not match) if:
- role folder starts with `ansible-role-` prefix
- meta/main.yml contains a compliant role_name

Fixes: #979
Fixes: #966


jsf9k:
Are there plans to create a new release containing this fix soon?  #979 and #966 are really wreaking havoc on quite a few folks' Ansible roles (including [mine](https://github.com/cisagov/?q=ansible-role&type=&language=)).

PR #983, opened 2020-08-21
Make coverage report run only on full test runs

ssbarnea:
We add coverage options but not making them mandatory as we do not want to force pytest users to run coverage when they just want to run a single test with `pytest -k test`.

Settings added to repository pytest.ini become defaults for all executions.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -14,12 +14,6 @@ addopts =
     --doctest-modules
     --junitxml=.test-results/pytest/results.xml
 
-    # `pytest-cov`:
-    --cov=ansiblelint
-    --cov-report term-missing:skip-covered
-    --cov-report xml:.test-results/pytest/cov.xml
-    --no-cov-on-fail
-
     # interpret all the target args as importables:
     --pyargs
 
--- a/tox.ini
+++ b/tox.ini
@@ -25,10 +25,16 @@ deps =
 commands =
   # safety measure to assure we do not accidentaly run tests with broken dependencies
   {envpython} -m pip check
+  # We add coverage options but not making them mandatory as we do not want to force
+  # pytest users to run coverage when they just want to run a single test with `pytest -k test`
   {envpython} -m pytest \
-  --cov "{envsitepackagesdir}/ansiblelint" \
   --junitxml "{toxworkdir}/junit.{envname}.xml" \
-  {posargs:}
+  {posargs:\
+    --cov ansiblelint \
+    --cov "{envsitepackagesdir}/ansiblelint" \
+    --cov-report term-missing:skip-covered \
+    --cov-report xml:.test-results/pytest/cov.xml \
+    --no-cov-on-fail}
 install_command =
   {envpython} -m \
     pip install \
@@ -37,6 +43,7 @@ install_command =
 passenv =
   CURL_CA_BUNDLE  # https proxies, https://github.com/tox-dev/tox/issues/1437
   HOME
+  PYTEST_*  # allows developer to define their own preferences
   REQUESTS_CA_BUNDLE  # https proxies
   SSL_CERT_FILE  # https proxies
 # recreate = True


webknjaz (changes requested):
This doesn't look right. If you want something overrideable, you could add `--no-cov`to to posargs. Spreading the args across multiple places does not seem like a good idea maintenance-wise.

ssbarnea:
> This doesn't look right. If you want something overrideable, you could add `--no-cov`to to posargs. Spreading the args across multiple places does not seem like a good idea maintenance-wise.

This is not possible if you do not have pytext-cov installed and would require calling it with extra arguments. I am really curious how many contributors to the project have found any use in being forced to scroll each time they run a single test with "pytest -k foo".

I know that you have a strong passion for coverage but this does not mean that is ok to enforce your personal preferences on anyone else. Coverage arguments were added by you via https://github.com/ansible/ansible-lint/pull/604 **without any peer review**. I am trying to undo that action.

**I will be waiting feedback from others**, I invite them to try running:

```bash
pytest --collect-only
# this command is supposed to list all tests, but due to current settings
# it also dumps code coverage reports, making the output almost useless

pytest -k test_logger_debug
# again adding extra 60 lines of SPAM after what was supposed to be one line
```

I do have few extra examples of why this f

webknjaz:
I don't think it's reasonable to support 100500 ways of running pytest. The official one is to use tox. If somebody wants to use alternative ways they should figure out what works for them on their own.

odyssey4me:
I'm not sure I understand the contention here. This appears to keep the same default, but provides a way to optionally override the arguments.

greg-hellings:
I agree with making these defaults able to be easily overridden. I just finished running tests on a single file during my dev cycle, and it was significantly annoying to scroll back past all the coverage output. I'm curious why you wouldn't have the defaults in the place of {posargs:}, though? I invoked my test through `tox -e py38-ansible29 -- test/TestMissingFilePermissionsRule.py` rather than looking for the envarg here.

tox.ini
```diff
@@ -44,6 +44,7 @@ install_command =
 passenv =
   CURL_CA_BUNDLE  # https proxies, https://github.com/tox-dev/tox/issues/1437
   HOME
+  PYTEST_*  # allows developer to define his own preferences
```
greg-hellings, line 47:
```suggestion
  PYTEST_*  # allows developer to define their own preferences
```

ssbarnea:
I asked on pytest main irc channel about that issue and it seems that others also found coverage report a nuisance. They mentioned another common approach of having separated tox environment (and CI job) solely coverage. In other words, relatively similar methods to avoid the same issue.

ss

greg-hellings:
I have found success with having each tox environment generate its coverage reports, but only outputting them during a call to "tox -e coverage", which then will combine the reports from all other runs and produce the output. That's independent of pytest, which I only have begun to look into recently. It's especially useful to do that if there are code branches to handle multiple versions of e.g. Python, Operating System, etc.

pytest.ini
```diff
@@ -12,13 +12,6 @@ addopts =
     -ra
     --showlocals
     --doctest-modules
-    --junitxml=.test-results/pytest/results.xml
```
webknjaz, line 15: This is totally unrelated to coverage

tox.ini
```diff
@@ -31,11 +31,11 @@ deps =
...
 commands =
   # safety measure to assure we do not accidentaly run tests with broken dependencies
+  # We add coverage options but not making them mandatory as we do not want to force
+  # pytest users to run coverage when they just want to run a single test with `pytest -k test`
   python -m pip check
   {envpython} -m pytest \
-  --cov "{envsitepackagesdir}/ansiblelint" \
-  --junitxml "{toxworkdir}/junit.{envname}.xml" \
-  {posargs:}
+  {posargs:--cov "{envsitepackagesdir}/ansiblelint" --junitxml "{toxworkdir}/junit.{envname}.xml" --cov-report term-missing:skip-covered --cov-report xml:.test-results/pytest/cov.xml --no-cov-on-fail}
```
webknjaz, line 38: Do tox substitutions work across lines? Multi-line would read better

tox.ini
```diff
@@ -44,6 +44,7 @@ install_command =
 passenv =
   CURL_CA_BUNDLE  # https proxies, https://github.com/tox-dev/tox/issues/1437
   HOME
+  PYTEST_*  # allows developer to define their own preferences
```
webknjaz, line 47: This one is controversial. Using posargs already allows to override the args, what else do you need?

tox.ini
webknjaz: Also, for CI runs I'd rather not set no-cov-on-fail...

webknjaz: Also, for CI runs I'd rather not set no-cov-on-fail...

webknjaz, line 38: Oh, you changed the xml location, I'd like to keep it in the old one, the whole dir is gitignored

webknjaz: Let's leave junit changes out and discuss separately, I guess.

ssbarnea: With current Github Actions the impact is not very big because sadly github displays output with top instead of tail (opposed to most other CI systems which display the tail and allow you to scroll down, more like a terminal approach).

Please note that removing `--no-cov-on-fail` option would be a huge problem for local development, because it would force any developer running tox to scroll few screens to see the test failed.

IMHO, there is zero value in producing a coverage report when the outcome if fail. Nobody cares about coverage when the outcome is already a failure. Also that report would clearly not be trustable either because it would not report a real coverage.

As recommended by other pytest users, many do have a different `tox -e coverage` environment that does this. CI would use it, and if we do that it could be run all the time, even consolidating data from multiple environments.

ssbarnea: I removed that as that is outside the scope of this change.

tox.ini
ssbarnea: Is not controversial at all, it allows developer to configure set of persistent personal preferences for running pytest, and avoid having to pass extra options each time you call pytest.

For example, my local ones look like ``export PYTEST_ADDOPTS='-s -n0 --ff --verbosity=0'``

webknjaz: This sounds like it'll allow people to accidentally break our setup and complain about it being broken...

tox.ini
```diff
@@ -31,11 +31,16 @@ deps =
...
   # safety measure to assure we do not accidentaly run tests with broken dependencies
+  # We add coverage options but not making them mandatory as we do not want to force
+  # pytest users to run coverage when they just want to run a single test with `pytest -k test`
   python -m pip check
   {envpython} -m pytest \
-  --cov "{envsitepackagesdir}/ansiblelint" \
   --junitxml "{toxworkdir}/junit.{envname}.xml" \
-  {posargs:}
+  {posargs:\
+    --cov "{envsitepackagesdir}/ansiblelint" \
```
webknjaz, line 40: Don't forget to keep the value from the config:
```suggestion
    --cov ansiblelint \
    --cov "{envsitepackagesdir}/ansiblelint" \
```

pytest.ini
```diff
@@ -13,12 +13,6 @@ addopts =
     --showlocals
     --doctest-modules
     --junitxml=.test-results/pytest/results.xml
-
-    # `pytest-cov`:
-    --cov=ansiblelint
-    --cov-report term-missing:skip-covered
-    --cov-report xml:.test-results/pytest/cov.xml
-    --no-cov-on-fail
```
webknjaz, line 21: Have you considered keeping these here with `--no-cov` and just re-enabling in `tox.ini` (by setting `--cov`)? This would allow keeping more config settings consolidated.

ssbarnea: I did but it would be a mistake. It will fail to run if user does not have coverage plugin installed. Sometimes I do not have it and TBH, I do not see a need to add it as a runtime requirement for cases that do not need it. That is another reason why adding any plugin specific options in defaults is dangerous, as it prevents running pytest without them: aka xdist options, some kind of problem.

Our tests should be to run in absence of all/most pytest plugins, as these are supposed to add extra features, but still nicely degrade in their absence. xdist and cov fit this category.

webknjaz: Well, you know that I'm against supporting this many ways to run the tests, this does not sound convincing.

tox.ini
ssbarnea: Should we stop selling cars because some people may not know how to drive them? If you are really afraid of this, you can make a PR that adds a warning message when tests failed and PYTEST_ADDOPTS was defined. Still, I doubt that this is a real risk. There are lots of other stuff that can produce failures and unless we see a clear pattern (multiple bugs reported), I doubt we should invest in them.

PR #984, opened 2020-08-21
Added type information to tests

ssbarnea:
Partial-Fix: #893

### DIFF
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -190,7 +190,7 @@ class RulesCollection(object):
             self.extend(load_plugins(rulesdir))
         self.rules = sorted(self.rules, key=lambda r: r.id)
 
-    def register(self, obj):
+    def register(self, obj: AnsibleLintRule):
         self.rules.append(obj)
 
     def __iter__(self):
--- a/lib/ansiblelint/runner.py
+++ b/lib/ansiblelint/runner.py
@@ -1,7 +1,7 @@
 """Runner implementation."""
 import logging
 import os
-from typing import TYPE_CHECKING, Any, FrozenSet, Generator, List, Set
+from typing import TYPE_CHECKING, Any, FrozenSet, Generator, List, Optional, Set
 
 import ansiblelint.file_utils
 import ansiblelint.skip_utils
@@ -23,9 +23,9 @@ class Runner(object):
             self,
             rules: "RulesCollection",
             playbook: str,
-            tags: FrozenSet[Any],
-            skip_list: FrozenSet[Any],
-            exclude_paths: List[str],
+            tags: FrozenSet[Any] = frozenset(),
+            skip_list: Optional[FrozenSet[Any]] = frozenset(),
+            exclude_paths: List[str] = [],
             verbosity: int = 0,
             checked_files: Set[str] = None) -> None:
         """Initialize a Runner instance."""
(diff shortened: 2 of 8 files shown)


lib/ansiblelint/runner.py
```diff
@@ -23,9 +23,9 @@ def __init__(
             self,
             rules: "RulesCollection",
             playbook: str,
-            tags: FrozenSet[Any],
-            skip_list: FrozenSet[Any],
-            exclude_paths: List[str],
+            tags: FrozenSet[Any] = frozenset(),
+            skip_list: Optional[FrozenSet[Any]] = frozenset(),
+            exclude_paths: List[str] = [],
```
webknjaz, line 28: Lists are mutable. You shouldn't use them as method defaults.

lib/ansiblelint/rules/__init__.py
```diff
@@ -190,7 +190,7 @@ def __init__(self, rulesdirs=None) -> None:
             self.extend(load_plugins(rulesdir))
         self.rules = sorted(self.rules, key=lambda r: r.id)
 
-    def register(self, obj):
+    def register(self, obj: AnsibleLintRule):
```
webknjaz, line 193: How is this related to tests?

webknjaz (changes requested):

lib/ansiblelint/runner.py
ssbarnea: Using an empty list is the only way to make the parameter optional without changing the API. I do not see any problem with having a list passed. 

If you want to change exclude_paths to be something else I invite you to make PR that does this, outside of this change.

lib/ansiblelint/rules/__init__.py
ssbarnea: Yes it does. I would really recommend you to try adding some typing information yourself on ansible-lint, because that is something you will realise right away: adding type to one method does create a long chain of dependencies.

Here is what happens if I remove it:
```
test/TestWithSkipTagId.py:10: error: Call to untyped function "register" in typed context
test/TestMetaMainHasInfo.py:61: error: Call to untyped function "register" in typed context
test/TestUseHandlerRatherThanWhenChanged.py:77: error: Call to untyped function "register" in typed context
test/TestComparisonToEmptyString.py:28: error: Call to untyped function "register" in typed context
test/TestSudoRule.py:48: error: Call to untyped function "register" in typed context
test/TestLineTooLong.py:17: error: Call to untyped function "register" in typed context
test/TestVariableHasSpaces.py:47: error: Call to untyped function "register" in typed context
test/TestMetaChangeFromDefault.py:19: error: Call to untyped function "register" in typed context
test/TestMetaVideoLinks.py:23: error: Call to untyped function "register" in typed context
test/TestTaskNoLocalAction.py:17: error: Call to untyped function "register" in typed context
test/TestDeprecatedModule.py:19: error: Call to untyped function "register" in typed context
test/TestShellWithoutPipefail.py:73: error: Call to untyped function "register" in typed context
test/TestEnvVarsInCommand.py:79: error: Call to untyped function "register" in typed context
test/TestRoleRelativePath.py:41: error: Call to untyped function "register" in typed context
test/TestMetaTagValid.py:17: error: Call to untyped function "register" in typed context
test/TestComparisonToLiteralBool.py:46: error: Call to untyped function "register" in typed context
Found 16 errors in 16 files (checked 117 source files)
```

PR #989, opened 2020-08-24
Add ability to only warn about some rules

ssbarnea:
Adds a `warn_list` option were user can add rules that should not
affect the outcome of the linting, while they will still be identified.

This softer option makes linter easier to adopt while still reminding
users about issues they should seek to address.

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -79,19 +79,31 @@ def choose_formatter_factory(
     return r
 
 
-def hint_about_skips(matches: List["MatchError"]):
-    """Display information about how to skip found rules."""
+def report_outcome(matches: List["MatchError"], options) -> int:
+    """Display information about how to skip found rules.
+
+    Returns exit code, 2 if erros were found, 0 when only warnings were found.
+    """
+    failure = False
     msg = """\
 You can skip specific rules by adding them to the skip_list section of your configuration file:
 ```yaml
 # .ansible-lint
-skip_list:
+warn_list:  # or 'skip_list' to silence them completely
 """
     matched_rules = {match.rule.id: match.rule.shortdesc for match in matches}
     for id in sorted(matched_rules.keys()):
-        msg += f"  - '{id}'  # {matched_rules[id]}'\n"
+        if id not in options.warn_list:
+            msg += f"  - '{id}'  # {matched_rules[id]}'\n"
+            failure = True
     msg += "```"
-    console.print(Markdown(msg))
+
+    if failure:
+        if not options.quiet and not options.parseable:
+            console.print(Markdown(msg))
+        return 2
+    else:
+        return 0
 
 
 def main() -> int:
@@ -154,9 +166,7 @@ def main() -> int:
             print(formatter.format(match))
 
     if matches:
-        if not options.quiet and not options.parseable:
-            hint_about_skips(matches)
-        return 2
+        return report_outcome(matches, options=options)
     else:
         return 0
 
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -134,6 +134,8 @@ def get_cli_parser() -> argparse.ArgumentParser:
     parser.add_argument('-x', dest='skip_list', default=[], action='append',
                         help="only check rules whose id/tags do not "
                         "match these values")
+    parser.add_argument('-w', dest='warn_list', default=[], nargs="+",
+                        help="only warn about these rules")
     parser.add_argument('--nocolor', dest='colored',
                         default=hasattr(sys.stdout, 'isatty') and sys.stdout.isatty(),
                         action='store_false',
@@ -173,6 +175,7 @@ def merge_config(file_config, cli_config) -> NamedTuple:
         'rulesdir',
         'skip_list',
         'tags',
+        'warn_list',
     )
 
     if not file_config:


nre-ableton:
This looks like a great way to solve some of the issues raised in the 4.3.0 update; thanks so much for this patch!

lib/ansiblelint/__main__.py
```diff
@@ -79,19 +79,30 @@ def choose_formatter_factory(
...
+
+    Returns exit code, 2 if erros were found, 0 when only warnings were found.
+    """
+    failure = False
     msg = """\
 You can skip specific rules by adding them to the skip_list section of your configuration file:
 ```yaml
 # .ansible-lint
-skip_list:
+warn_list:  # or 'skip_list' to silence them completly
```
rafaelfolco, line 92:
```suggestion
warn_list:  # or 'skip_list' to silence them completely
```

lib/ansiblelint/cli.py
```diff
@@ -160,7 +162,7 @@ def get_cli_parser() -> argparse.ArgumentParser:
     return parser
 
 
-def merge_config(file_config, cli_config) -> NamedTuple:
+def merge_config(file_config, cli_config) -> NamedTuple:  # noqa: C901
```
webknjaz, line 165: I don't have time to do the full review right now but adding complexity is unacceptable.

lib/ansiblelint/__main__.py
```diff
@@ -154,8 +165,7 @@ def main() -> int:
             print(formatter.format(match))
 
     if matches:
-        hint_about_skips(matches)
-        return 2
+        return report_outcome(matches, options=options)
```
jbackman, line 168: https://github.com/ansible/ansible-lint/pull/992 will gate this so it only shows up when NOT quiet, paresable, or parsable_severity are True. Since again this is adding verbose output, I would expect it not to be enabled for those modes

lib/ansiblelint/cli.py
```diff
@@ -194,6 +196,9 @@ def merge_config(file_config, cli_config) -> NamedTuple:
     if 'skip_list' in file_config:
         cli_config.skip_list = cli_config.skip_list + file_config['skip_list']
 
+    if 'warn_list' in file_config:
+        cli_config.warn_list = cli_config.skip_list + file_config['warn_list']
```
jbackman, line 200: I think this is incorrect - cli_config.skip_list is not what you are working with - I think it should be cli_config.warn_list

lib/ansiblelint/cli.py
ssbarnea: That is not real complexity an you know it well, clearly not a reason to block this change. If you know how to avoid scalability issues with merge_config, you are welcomed to refactor it to avoid triggering C901 when newer option are added.

lib/ansiblelint/cli.py
ssbarnea: Indeed. Thanks, that was a genuine bug.

lib/ansiblelint/cli.py
greg-hellings: There is a lot of repetition here. It's not complex, but it is redundant. Maybe the complexity warning could be squashed with a refactor. Something along the lines of (untested)

```python
for field in ('quiet', 'parseable', 'parseable_severity' ... ):
    if field in file_config:
        setattr(cli_config, field, getattr(cli_config, field) or file_config[field]))
# Repeat again for addition/extend
```

ssbarnea: @greg-hellings Yes, it may work but it should be a follow-up and C901 should not be used as a ransom for forcing someone to refactor merge_config. This change itself does not add that kind of complexity we want to avoid with C901.

greg-hellings: @ssbarnea I agree. Spurious complexity should not be a reason to hold up positive changes. But refactoring should be tossed on the backlog of things to tackle. I've added an issue [here](https://github.com/ansible/ansible-lint/issues/998) for that.

lib/ansiblelint/__main__.py
```diff
@@ -79,19 +79,31 @@ def choose_formatter_factory(
...
+
+    Returns exit code, 2 if erros were found, 0 when only warnings were found.
+    """
+    failure = False
     msg = """\
 You can skip specific rules by adding them to the skip_list section of your configuration file:
 ```yaml
 # .ansible-lint
-skip_list:
+warn_list:  # or 'skip_list' to silence them completly
```
greg-hellings, line 92: You resolved this, but the typo is still there.

lib/ansiblelint/__main__.py
greg-hellings: It looks like the report_outcome method does the guarding around quiet/parseable itself.

lib/ansiblelint/cli.py
awcrosby: While the addition of this branch is not complex in itself, it does underscore this method should be refactored as it grows... I am ok with having that on the backlog as added ^.

greg-hellings: @awcrosby #999 was already submitted by @ssbarnea to address the complexity refactor.

lib/ansiblelint/__main__.py
ssbarnea: done now

PR #990, opened 2020-08-24
lint file type modules explicit use of mode:

chrismeyersfsu:
* Recent Ansible CVE required Ansible to change the default behavior of
file type modes such that if mode is node specified explicitly, newly
created files would default to mode '0600'. Before this CVE fix the
default mode was '0644`. This can cause problems.


greg-hellings:
How is this better than the existing rule id 208? It looks like maybe you just need to add some modules into the list for that one?

ssbarnea:
Clearly you missed to observe that https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/MissingFilePermissionsRule.py is already doing this.

Closing as duplicate but I hope to see more contributions from you.

PR #992, opened 2020-08-24
Avoid non parsable output on stdout

jbackman:
hint_about_skips breaks behavior of  quiet, parseable, and parsable_severity by adding non complaint lines. This quick fix will not add the hint_about_skips when one of those flags is specified

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -153,7 +153,7 @@ def main() -> int:
         for match in matches:
             print(formatter.format(match))
 
-    if matches:
+    if matches and not (options.quiet or options.parseable or options.parseable_severity):
         hint_about_skips(matches)
         return 2
     else:
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -155,6 +155,8 @@ def main() -> int:
 
     if matches and not (options.quiet or options.parseable or options.parseable_severity):
         hint_about_skips(matches)
+    
+    if matches:
         return 2
     else:
         return 0
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -153,10 +153,9 @@ def main() -> int:
         for match in matches:
             print(formatter.format(match))
 
-    if matches and not (options.quiet or options.parseable or options.parseable_severity):
-        hint_about_skips(matches)
-
-    if matches:
+    if matches :
+        if not options.quiet and not options.parseable:
+            hint_about_skips(matches)
         return 2
     else:
         return 0
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -153,7 +153,7 @@ def main() -> int:
         for match in matches:
             print(formatter.format(match))
 
-    if matches :
+    if matches:
         if not options.quiet and not options.parseable:
             hint_about_skips(matches)
         return 2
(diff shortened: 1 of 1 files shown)


jbackman:
https://github.com/ansible/ansible-lint/pull/971 introduced this feature but didn't take into consideration the modes where specific types of output are expected

ssbarnea:
There is better way to address this, just changing the report to go to stderr instead of stdout. See https://rich.readthedocs.io/en/latest/console.html

jbackman:
I disagree - This is not actually an error, it is a more verbose output which is useful, but should be suppressed if it is not desired. The three modes I specify are to suppress output that is not desired

ssbarnea:
> I disagree - This is not actually an error, it is a more verbose output which is useful, but should be suppressed if it is not desired. The three modes I specify are to suppress output that is not desired

Not sure what to say. I seen quiet and parsable used in practice and if we do it like this the user will likely never get to see the hits.

I am curious what @webknjaz or @awcrosby think about it. I will let them decide on how to fix it.

lib/ansiblelint/__main__.py
```diff
@@ -153,7 +153,7 @@ def main() -> int:
         for match in matches:
             print(formatter.format(match))
 
-    if matches:
+    if matches and not (options.quiet or options.parseable or options.parseable_severity):
```
webknjaz, line 156: I agree with `-q` but not the others.
Also, @ssbarnea is correct about the need to redirect the hint output to stderr.

There's other problems with this too:
* Adding the flag check to this if-clause changes the return code behavior effectively breaking those flags and making them trick ansible-lint into returning the successful outcome regardless of the violations present
* This change must include regression tests, both for the hint absence with `-q` and checking that adding it does not affect the return code

webknjaz (changes requested):
This needs a regression

lib/ansiblelint/__main__.py
jbackman: I personally use parseable and parsable_severity as inputs into a larger testing framework. I'm sure others do as well. The adding of lines that do not follow the established pattern requires additional parsing that is unnecessary. STDERR is generally reserved for error messages of which this is not. It is an informational message that although provides value, is not necessary.

As for your comment on the return codes - I agree. I would propose reverting to the original if statement and creating a new if statement for this function gated by the flags.

ssbarnea: Alter the code and use a nested conditional approach similar to the warning feature PR currenly open:

```
if matches:
   if not options.quiet and not options.parseable:
       ...
```

That is much easier to read and works file. There is no need to look at parseable_severity -- we do not care about its value. The only case where we want to skip printing is when we are in parsable mode or quiet.

I know I proposed stderr, but that is tricky as it needs lots of testing, especially around ANSI support on stderr, which can vary.

Stderr is for outof the band messaging, logging most of the time, not only error, also debug, and "info". This message fits well in info category. Still, I cannot say for sure that either stdour or stderr are the ultimate correct stream to use for this message. Is a bit of grey area. So, when in doubt I am inclined to avoid making a change on this.

ssbarnea (approved):

ssbarnea:
Added https://github.com/ansible/ansible-lint/issues/997 for implementing a regression test.

lib/ansiblelint/__main__.py
webknjaz: @jbackman stderr does not exist solely for error messages. It should be used for any (often human-readable) diagnostic output so that stdout is clean and can be piped into other programs for parsing.

Issue #998, opened 2020-08-25
Reduce redundancy in merge_config method


greg-hellings:
##### Summary
Refactor cli.py:merge_config to remove redundancy. See [this comment](https://github.com/ansible/ansible-lint/pull/989#discussion_r476516014) for more context.

##### Issue Type

- Feature Idea

##### Additional Information
This will reduce the verbosity of the config merge code and make it easier to account for additional config values in the future.

It will also eliminate the need for skipping checks on the method.

PR #999, opened 2020-08-25
Simplify merge_config

ssbarnea:
Makes config_merge implementation easier to maintain and avoid C901 when new options will be added.

Fixes: #998

### DIFF
--- a/lib/ansiblelint/cli.py
+++ b/lib/ansiblelint/cli.py
@@ -161,41 +161,33 @@ def get_cli_parser() -> argparse.ArgumentParser:
 
 
 def merge_config(file_config, cli_config) -> NamedTuple:
+    bools = (
+        'display_relative_path',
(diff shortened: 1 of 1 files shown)


lib/ansiblelint/cli.py
```diff
@@ -161,42 +161,37 @@ def get_cli_parser() -> argparse.ArgumentParser:
 
 
 def merge_config(file_config, cli_config) -> NamedTuple:
+    bools = (
+        'display_relative_path',
+        'parseable',
+        'quiet',
+        'use_default_rules',
+    )
```
greg-hellings, line 169: Any reason to not include 'parseable_severity' here?

ssbarnea: Only lack of attention. Thanks fixed.

PR #1003, opened 2020-08-26
Refactor run_ansible_lint test util

ssbarnea:
Changes implementation of run_ansible_lint testing method in order
to make it easier to reuse, multiple arguments or commands.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -59,26 +59,24 @@ class RunFromText(object):
 
 
 def run_ansible_lint(
-        cwd: os.PathLike = None,
-        role_path: os.PathLike = None,
+        *argv: str,
+        cwd: str = None,
         bin: str = None,
-        env: Dict[str, str] = None):
+        env: Dict[str, str] = None) -> subprocess.CompletedProcess:
     """Run ansible-lint on a given path and returns its output."""
-    if bin:
-        command = [bin, "-v"]
+    if not bin:
+        bin = sys.executable
+        args = [sys.executable, "-m", "ansiblelint", *argv]
     else:
-        command = [sys.executable, "-m", "ansiblelint", "-v"]
-    if role_path:
-        command.append(str(role_path))
+        args = [bin, *argv]
 
-    result, err = subprocess.Popen(
-        command,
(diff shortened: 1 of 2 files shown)


test/__init__.py
```diff
@@ -59,26 +59,25 @@ def run_role_meta_main(self, meta_main_text):
...
+    return subprocess.Popen(
         command,
         cwd=cwd,
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         shell=False,  # needed when command is a list
         env=env,
+        universal_newlines=True
     ).communicate()
```
webknjaz, line 83: Since you're already refactoring this...
The docs recommend using run instead of Popen starting Python 3.5: https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module.
```suggestion
    proc_result = subprocess.run(
        command,
        cwd=cwd,
        stdin=subprocess.PIPE,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        shell=False,  # needed when command is a list
        env=env,
        universal_newlines=True
    )
    return proc_result.stdout, proc_result.stderr
```
(Although, I'd probably return the whole `subprocess.CompletedProcess` instance)

test/__init__.py
```diff
@@ -59,26 +59,25 @@ def run_role_meta_main(self, meta_main_text):
...
-        role_path: os.PathLike = None,
         bin: str = None,
-        env: Dict[str, str] = None):
+        env: Dict[str, str] = None) -> Tuple[str, str]:
     """Run ansible-lint on a given path and returns its output."""
     if bin:
-        command = [bin, "-v"]
+        command = [bin]
     else:
-        command = [sys.executable, "-m", "ansiblelint", "-v"]
```
webknjaz, line 70: I think removing `-v` will affect all the tests relying on verbosity being set...

ssbarnea: It does not affect because we did not check stderr anyway. Lucky this time.

webknjaz (approved):

PR #1009, opened 2020-08-27
Use embedded rule testing for MetaTagValidRule

webknjaz:
Partial-Fix: #947


lib/ansiblelint/__init__.py
```diff
@@ -19,8 +19,27 @@
...
 
-from ansiblelint.rules import AnsibleLintRule
-from ansiblelint.version import __version__
+import builtins
+from typing import Any
+
+# NOTE: Importing builtins as _builtins does not allow to redefine the
+# NOTE: type required to trick mypy into thinking that we can set the
+# NOTE: "pytest" attribute:
+_builtins: Any = builtins
```
ssbarnea, line 28: Sorry but adding this bit in __init__ of the tool scares the ... out of me, making me want to really avoid it. Looks like a too low level hack that can explode in our face.

PR #1010, opened 2020-08-27
Move test utils into ansiblelint.testing

ssbarnea:
Moves our few test utils inside a library module in order to make them available to other consumers, like rules.


lib/ansiblelint/test/__init__.py
```diff
@@ -0,0 +1,85 @@
+"""Test utils for ansible-lint."""
```
webknjaz, line 1: Naming the package "test" suggests that it contains tests. I'd like it to be "testing" so that it feels like test helpers/framework instead.

test/__init__.py
```diff
@@ -1,82 +1 @@
...
-        args,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        shell=False,  # needed when command is a list
-        check=False,
-        cwd=cwd,
-        env=env,
-        universal_newlines=True
-    )
+"""Use ansiblelint.test instead for test reusables."""
```
webknjaz, line 1: Typo ^

lib/ansiblelint/conftest.py
```diff
@@ -0,0 +1,33 @@
+"""PyTest Fixtures."""
```
webknjaz, line 1: Out of scope

webknjaz (approved):

PR #1017, opened 2020-08-28
Refactor conftest.py

ssbarnea:
- move conftest.py to the root of the repository in order to make
  fixture accesible to tests from inside the library too.
- adds missing docstrings
- remove editor config lefover
- provide better failure message when pytest is run without the
  library being installed
- improves default_rules_collection fixture to make it use
  ansiblelint.constants.DEFAULT_RULESDIR

This is in sync with documented use from:
https://docs.pytest.org/en/stable/writing_plugins.html#conftest-py-plugins


test/conftest.py
```diff
@@ -1,30 +0,0 @@
...
-@pytest.fixture
-def rule_runner(request):
-    """Return runner for a specific rule class."""
-    rule_class = request.param
-    collection = RulesCollection()
-    collection.register(rule_class())
-    return RunFromText(collection)
-
-
-# vim: et:sw=4:syntax=python:ts=4:
```
webknjaz, line 30: Please keep this.

conftest.py
```diff
@@ -0,0 +1,33 @@
...
+import os
+
+import pytest
+
+try:
+    from ansiblelint.constants import DEFAULT_RULESDIR
+    from ansiblelint.rules import RulesCollection
+    from ansiblelint.testing import RunFromText
+except ImportError:
+    pytest.exit("You need to install ansiblelint to be able to load test fixtures.")
```
webknjaz, line 11: I don't think this is necessary.

ssbarnea: It is highly desirable. Try to just run "pytest" in repository folder without it and it will fail in an ugly way without it. I prefer to give a hint to the developer instead of risking more tickets, especially because installing the library before running tests is not something most project require, many of them work without that.

webknjaz: Supporting pytest outside of tox is wrong.

test/conftest.py
ssbarnea: Why to keep this? What makes this file unique to have it an other not? It is noise and we do not want noise.

webknjaz (approved):

PR #1027, opened 2020-09-02
Replace a tmp RST document hack with a directive

webknjaz:
The previous implementation used a hack of temporarily generating an RST document in the docs tree that I looked up elsewhere when I didn't know better. This approach makes it unobvious how the document even appears in docs, which is bad.

Now that I know how to make nodes from RST, I've changed it to a directive that expands as a table with the default rules. The intermediate contents don't even hit the disk unnecessarily!

In docs, it renders the same. In source, we now have actual non-gitignored `default_rules.rst` that contains the new directive.


ssbarnea:
Look nice but would make debugging of rule listing generation a serious PITA as we loose the ability to look at intermediary file.

webknjaz:
FWIW the proof is also viewable on the preview page: https://ansible-lint--1027.org.readthedocs.build/en/1027/default_rules.html.

webknjaz:
> Look nice but would make debugging of rule listing generation a serious PITA as we loose the ability to look at intermediary file.

Not really, we used to autoremove it right away. It's not worse than trying to catch a moment when it's still there.

ssbarnea (approved):

webknjaz:
Plus, you can still get the RST output from CLI anyway.

Issue #1026, opened 2020-09-02
Add rule warning about use of run_once with strategy: free


cboylan:
##### Summary

It would be useful for ansible-lint to have a rule that warns against the use of `run_once` with `strategy: free`. Ansible ignores `run_once` with the free strategy which means your tasks are run many times, once for each valid inventory host. Ansible itself does warn against this as well but doesn't fail directly. Being able to catch this early would be great.

##### Issue Type

- Feature Idea

ssbarnea:
ansible 2.10 does produce runtime warning about use of `run_once` when used with `strategy: free` but it does not produce any warnings when doing syntax check or when running without free strategy.

This behavior makes the `run_once` a potential production bomb, as users may change strategy when deploying in production, in some cases when they use tower/awx this can also be changed by the operator.

I am inclined to consider `run_once` an unpredictable option to use and always warn about its presences, regardless the strategy (especially as it cannot reliably be inferred during linting).  I would suggest manually whitelisting each usage, unless someone comes with a better recommendation.

ansiblejunky:
Just for info, providing what ansible 2.9.x shows as far as the warning message:

playbook:
```yaml
---
- hosts: all
  strategy: free
  gather_facts: false

  tasks:

  - debug:
      msg: "Test"
    run_once: True
```

output from `ansible-playbook -i localhost, test.yml`:
```
PLAY [all] ***********************************************************************************************************************
[WARNING]: Using run_once with the free strategy is not currently supported. This task will still be executed for every host in
the inventory list.

TASK [debug] *********************************************************************************************************************
ok: [localhost] => {
    "msg": "Test"
}

PLAY RECAP ***********************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

I would be ok with a rule that gives a warning and leaving it up to the individual to make it an error or leave it as warning.

This is another rule where it would be beneficial for lint rules to know a bit more about the context (in this case if the task is within a Play where strategy=free). However, that's a separate enhancement.

bluikko:
> use when: inventory_hostname == ansible_play_hosts_all[0] construct

If the rule proposed in the OP will be added, I for one would not want see the above construct as a remedy recommendation. It may work in some cases but in many cases (especially with the `free` strategy, where some similar construct is the only option) it is not equivalent at all and is a bad advise.

PR #1028, opened 2020-09-02
E107: Added rule for run_once

ssbarnea:
Implements a new rule that warns about use of `run_once` as it can have unpredictable behaviors when used with free strategy.

Fixes: #1026

### DIFF
--- a/lib/ansiblelint/rules/RunOnceRule.py
+++ b/lib/ansiblelint/rules/RunOnceRule.py
@@ -1,0 +1,118 @@
+
+# Copyright (c) 2020 Sorin Sbarnea <sorin.sbarnea@gmail.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+import sys
+
+from ansiblelint.rules import AnsibleLintRule
+
+
+def _nested_search(term, data):
+    return ((term in data) or
+            reduce((lambda x, y: x or _nested_search(term, y)), _get_subtasks(data), False))
+
+
+def _become_user_without_become(becomeuserabove, data):
+    if 'become' in data:
+        # If become is in lineage of tree then correct
+        return False
+    if ('become_user' in data and _nested_search('become', data)):
+        # If 'become_user' on tree and become somewhere below
+        # we must check for a case of a second 'become_user' without a
+        # 'become' in its lineage
+        subtasks = _get_subtasks(data)
+        return reduce((lambda x, y: x or _become_user_without_become(False, y)), subtasks, False)
+    if _nested_search('become_user', data):
+        # Keep searching down if 'become_user' exists in the tree below current task
+        subtasks = _get_subtasks(data)
+        return (len(subtasks) == 0 or
+                reduce((lambda x, y: x or
+                        _become_user_without_become(
+                            becomeuserabove or 'become_user' in data, y)), subtasks, False))
+    # If at bottom of tree, flag up if 'become_user' existed in the lineage of the tree and
+    # 'become' was not. This is an error if any lineage has a 'become_user' but no become
+    return becomeuserabove
+
+
+class RunOnceRule(AnsibleLintRule):
+    """RunOnceRule Class."""
+
+    id = "107"
+    shortdesc = "Unpredictable run_once"
+    description = (
+        "Use of run_once does not work with free execution strategy. "
+        "Avoid it, or use noqa comments to ignore it.")
+    severity = 'MEDIUM'
+    tags = ['unpredictability', 'experimental']
+    version_added = 'v4.4.0'
+
+    def matchtask(self, file, task):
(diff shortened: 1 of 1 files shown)


lib/ansiblelint/rules/RunOnceRule.py
```diff
@@ -0,0 +1,67 @@
...
+if "pytest" in sys.modules:
+
+    import pytest
+
+    @pytest.mark.parametrize('rule_runner', (RunOnceRule, ), indirect=['rule_runner'])
+    def test_107(rule_runner):
+        """Test rule matches."""
+        results = rule_runner.run_playbook(EXAMPLE_PLAYBOOK)
+        assert len(results) == 1
+        assert RunOnceRule == results[0].rule.__class__
```
webknjaz, line 66: Did you mean
```suggestion
        assert RunOnceRule is type(results[0].rule)
```

lib/ansiblelint/rules/RunOnceRule.py
```diff
@@ -0,0 +1,67 @@
...
+      command: touch /tmp/foo
+      run_once: true
+
+    - name: task that does not match
+      command: touch /tmp/foo
+      run_once: true  # noqa 107
+"""
+
+
+if "pytest" in sys.modules:
```
greg-hellings, line 57: This is... different. Do we want tests to live directly in the rule files, now?

webknjaz: There's actually two other PRs attempting to make it possible/easy to couple tests with the rule modules. I still feel weird about it too...

pytest.ini
```diff
@@ -54,5 +54,7 @@ python_files =
     # Ref: https://docs.pytest.org/en/latest/reference.html#confval-python_files
     # Needed to discover legacy nose test modules:
     Test*.py
-testpaths = test/
+    # Needed to discover embedded Rule tests
+    *Rule.py
+testpaths = test/ lib/ansiblelint/rules/
```
webknjaz, line 59: This is why I used pyargs in #1009:
https://github.com/ansible/ansible-lint/pull/1028/checks?check_run_id=1062724401#step:11:582

lib/ansiblelint/rules/RunOnceRule.py
ssbarnea: That is the goal, in fact the older PR also includes some documentation about how to write rules. Still we can wait for that to settle.

The main advantage is that I can use the example used in tests to create a rich documentation page for the rule.

My proposal is to experiment this approach with a limited set of rules and see how it scales. If we find deal-breakers, I will take care of moving the tests back.

pytest.ini
ssbarnea: I think that flaky failure (happens only on macos, ~1/4) is unrelated to how we pass the test args. Are you sure it does not happen with pyargs?

I asked about pyargs vs testpaths but nobody was able to provide a comparison, yet.

lib/ansiblelint/rules/RunOnceRule.py
greg-hellings: I like the idea. It can make it clear to me as a user what exactly the rule is testing for. I just hadn't looked into those other PRs.

I'm wondering if there is a more robust way for us to detect that we are running under our own tests? I'm just imagining a wild scenario where someone invokes ansible-lint while pytest is active and imported and somehow ends up running the test suite while trying to lint their own code. Is that safe from happening, or is there a way for us to make it safe? A custom environment variable, or command line argument, or something like that?

I just see this and immediately think it's far too easy for someone to have pytest autoloaded somehow and whoops, now they've got tests they didn't intend running.

ssbarnea: I think we do not need to worry about this because we do not support use of the linter other than from the command line. The presence of pylint in sys.modules has serious implications as loading pytest module does a lot of stuff under the hood.

Also keep in mind that under that conditional branch we only define test classes or method, not really executing stuff. So, even if you have pytest imported and you import the rule, it will not execute the tests.

gundalow:
Need review from @ansible-core-architects before merging

lib/ansiblelint/rules/RunOnceRule.py
```diff
@@ -0,0 +1,67 @@
...
+    shortdesc = "Unpredictable run_once"
+    description = (
+        "Use of run_once does not work with free execution strategy. "
+        "Avoid it, or use noqa comments to ignore it.")
+    severity = 'MEDIUM'
+    tags = ['deprecated']
+    version_added = 'v4.4.0'
+
+    def matchtask(self, file, task):
+        return 'run_once' in task
```
sivel, line 39: Where does this rule limit to only plays that are running with `free`?

Generally speaking I'm -1 on this rule, unless it can be explicitly targeted to the `free` strategy.  It's a very commonly used, and useful keyword.  

Blanket discouraging its use is not in alignment with Ansible. ansible-lint should not be instructing users to not use functionality of Ansible, unless it's been deprecated within Ansible itself.  If ansible-lint cannot properly apply context, and someone feels like this should definitely be included, then it should not be a rule that is enabled by default, and a rule that concerned authors can enable.

nitzmahone: +1 for "only if it's off by default" in its current form, since there's not really a viable alternative I can think of to `run_once` in eg, role contexts running with >1 host. Even in playbook contexts, avoiding `run_once` would generally require multiple plays, one with an explicit host, which has other knock-on effects in the face of failed hosts that `run_once` already handles more gracefully. `run_once` generally works quite well under linear strategies (eg propagating the registered result to all hosts), and we're already issuing a runtime warning under `strategy: free`. A rule might be able to detect that set of conditions in the simplest cases, but for roles, includes, etc, there's no way to know for sure until runtime. So seems like "off by default" is the safest thing here, absent a better alternative to `run_once`.

ssbarnea: While I think it can partially avoided when strategy is mentioned explicitly, I doubt we can have any guarantee regarding now code run in production. Imagine someone uses it inside a role, you do not have any control on how users are consuming this role. 

That us where static analyser, the linter, comes in, warning the role author that use of run_one does not always work the way he would imagine. If he accepts that, he can document this aspect and bypass the rule violation with a simple comment.

If we do not due this on all/most rules, we avoid to warn the author before is too late.

Still, I can see one genuine case where we can avoid triggering the rule when strategy is mentioned at current playbook and explicitly using liner. In that particular case we should to trigger it.

We may also need to rephrase the message and make it more descriptive, explaining user which options he has. Such a rule does not mean that code is bad, it only needs confirmation that the risk was already considered.

ansiblejunky: Where does the code check for the current strategy?

Also, `run_once` does work with `free` strategy. See documentation [here](https://docs.ansible.com/ansible/latest/user_guide/playbooks_delegation.html#run-once) and a note from that page states

_When used together with “serial”, tasks marked as “run_once” will be run on one host in each serial batch. If it’s crucial that the task is run only once regardless of “serial” mode, use when: inventory_hostname == ansible_play_hosts_all[0] construct._

Also not sure why its tagged with: `tags = ['deprecated']` ? Is something deprecated?

ssbarnea:
Closing it because I do not have time to implement all mentioned request. If someone can do them, I will be happy to review them as I do think this rule could prove useful, even it would just raise warnings by default (due to experimental tag).

Issue #1007, opened 2020-08-26
E208 tests include a task that would raise an Ansible error


moreati:
##### Summary
In the unit tests for E208 'File permissions not mentioned', there is a task would fail to execute if in an Ansible playbook.

https://github.com/ansible/ansible-lint/blob/cc23bae3c9912aa75658bd3aee75648e80d41dba/test/TestMissingFilePermissionsRule.py#L29-L32

Similarly the documentation for E208 suggests using `mode: preserve`, but a task using that with the file module and `state: file`, or `state: preserve` would similarly fail.

I'm unsure at the moment if the Ansible error is a bug in Ansible or working as intended. I'm raising this in case other lost souls make the change motivated on ansible-lint.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
:~/venvs$ ./ansible-2.9.12/bin/ansible --version
ansible 2.9.12
  config file = /home/alex/.ansible.cfg
  configured module search path = ['/home/alex/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/alex/venvs/ansible-2.9.12/lib/python3.8/site-packages/ansible
  executable location = ./ansible-2.9.12/bin/ansible
  python version = 3.8.2 (default, Jul 16 2020, 14:00:26) [GCC 9.3.0]

:~/venvs$ ./ansible-2.9.12/bin/ansible-lint --version
ansible-lint 4.3.3.dev1+gcc23bae

```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
Ubuntu 20.04 x86_64

##### STEPS TO REPRODUCE

```console (paste below)
:~/venvs$ cat playbook.yml 
- name: Test E208
  hosts: localhost
  tasks:
    - name: File module without mode
      file:
        path: canary
        state: file
        mode: preserve

:~/venvs$ touch canary

:~/venvs$ ./ansible-2.9.12/bin/ansible-playbook playbook.yml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost
does not match 'all'

PLAY [Test E208] ****************************************************************************************

TASK [Gathering Facts] **********************************************************************************
ok: [localhost]

TASK [File module without mode] *************************************************************************
[...]

PR #1030, opened 2020-09-03
E208: Corrected file mode checking

ssbarnea:
Takes care of incorrect Ansible documentation which misleads users about support for `mode: preserve` on modules which do not accept it.

Fixes: #1007

### DIFF
--- a/lib/ansiblelint/rules/MissingFilePermissionsRule.py
+++ b/lib/ansiblelint/rules/MissingFilePermissionsRule.py
@@ -19,44 +19,59 @@
 # THE SOFTWARE.
 from ansiblelint.rules import AnsibleLintRule
 
+# Despite documentation mentioning 'preserve' only these modules support it:
+_modules_with_preserve = (
+    'copy',
+    'template',
+)
+
 
 class MissingFilePermissionsRule(AnsibleLintRule):
     id = "208"
-    shortdesc = 'File permissions not mentioned'
+    shortdesc = 'File permissions unset or incorrect'
     description = (
-        "Missing mode parameter can cause unexpected file permissions based "
-        "on version of Ansible being used. Be explicit, or if you still "
-        "want the default behavior you can use ``mode: preserve`` to avoid "
-        "hitting this rule. See "
-        "https://github.com/ansible/ansible/issues/71200"
+        "Missing or unsupported mode parameter can cause unexpected file "
+        "permissions based "
+        "on version of Ansible being used. Be explicit, like ``mode: 0644`` to "
+        "avoid hitting this rule. Special ``preserve`` value is accepted "
+        f"only by {', '.join(_modules_with_preserve)} modules. "
+        "See https://github.com/ansible/ansible/issues/71200"
     )
     severity = 'VERY_HIGH'
     tags = ['unpredictability', 'experimental']
     version_added = 'v4.3.0'
 
-    _modules = (
-        'assemble',
+    _modules = {
         'archive',
-        'copy',
+        'assemble',
+        'copy',  # supports preserve
         'file',
-        'template',
+        'replace',  # implicit preserve behavior but mode: preserve is invalid
+        'template',  # supports preserve
         'unarchive',
-    )
+    }
 
-    _modules_with_create = (
-        'blockinfile',
-        'ini_file',
-        'lineinfile'
-    )
+    _modules_with_create = {
+        'blockinfile': False,
+        'htpasswd': True,
+        'ini_file': True,
+        'lineinfile': False,
+    }
 
     def matchtask(self, file, task):
-        if task["action"]["__ansible_module__"] not in self._modules and \
-                task["action"]["__ansible_module__"] not in self._modules_with_create:
+        module = task["action"]["__ansible_module__"]
+        mode = task['action'].get('mode', None)
+
+        if module not in self._modules and \
+                module not in self._modules_with_create:
             return False
 
-        if task["action"]["__ansible_module__"] in self._modules_with_create and \
-                not task["action"].get("create", False):
-            return False
+        if mode == 'preserve' and module not in _modules_with_preserve:
+            return True
+
+        if module in self._modules_with_create:
+            create = task["action"].get("create", self._modules_with_create[module])
+            return create and mode is None
 
         # A file that doesn't exist cannot have a mode
         if task['action'].get('state', None) == "absent":
@@ -67,9 +82,14 @@ class MissingFilePermissionsRule(AnsibleLintRule):
             return False
 
         # The file module does not create anything when state==file (default)
-        if task["action"]["__ansible_module__"] == "file" and \
+        if module == "file" and \
                 task['action'].get('state', 'file') == 'file':
             return False
 
-        mode = task['action'].get('mode', None)
+        # replace module is the only one that has a valid default preserve
+        # behavior, but we want to trigger rule if user used incorrect
+        # documentation and put 'preserve', which is not supported.
+        if module == 'replace' and mode is None:
+            return False
+
         return mode is None
--- a/test/TestMissingFilePermissionsRule.py
+++ b/test/TestMissingFilePermissionsRule.py
@@ -26,10 +26,6 @@ SUCCESS_TASKS = '''
 ---
 - hosts: hosts
   tasks:
-    - name: permissions not missing and string
-      file:
-        path: foo
-        mode: preserve
     - name: permissions not missing and numeric
       file:
         path: foo
@@ -49,13 +45,21 @@ SUCCESS_TASKS = '''
     - name: file edit when create is false
       lineinfile:
         path: foo
+        create: false
         line: some content here
+    - name: replace should not require mode
+      replace:
+        path: foo
 '''
 
 FAIL_TASKS = '''
 ---
 - hosts: hosts
   tasks:
+    - name: file does not allow preserve value for mode
+      file:
+        path: foo
+        mode: preserve
     - name: permissions missing and might create file
       file:
         path: foo
@@ -64,10 +68,24 @@ FAIL_TASKS = '''
       file:
         path: foo
         state: directory
-    - name: permissions needed if create is possible
+    - name: permissions needed if create is used
       ini_file:
         path: foo
         create: true
+    - name: lineinfile when create is true
+      lineinfile:
+        path: foo
+        create: true
+        line: some content here
+    - name: replace does not allow preserve mode
+      replace:
+        path: foo
+        mode: preserve
+    - name: ini_file does not accept preserve mode
+      ini_file:
+        path: foo
+        create: true
+        mode: preserve
 '''
 
 
@@ -82,4 +100,11 @@ def test_success(rule_runner):
 def test_fail(rule_runner):
     """Validate that missing mode triggers the rule."""
     results = rule_runner.run_playbook(FAIL_TASKS)
-    assert len(results) == 3
+    assert len(results) == 7
+    assert results[0].linenumber == 5
+    assert results[1].linenumber == 9
+    assert results[2].linenumber == 13
+    assert results[3].linenumber == 17
+    assert results[4].linenumber == 21
+    assert results[5].linenumber == 26
+    assert results[6].linenumber == 30
--- a/test/nomatchestest.yml
+++ b/test/nomatchestest.yml
@@ -6,4 +6,4 @@
       action: debug msg="Hello!"
 
     - name: this should be fine too
-      action: file state=touch dest=./wherever mode=preserve
+      action: file state=touch dest=./wherever mode=0600
--- a/test/unicode.yml
+++ b/test/unicode.yml
@@ -6,4 +6,4 @@
 
   tasks:
   - name: bonjour, ça va?
-    file: state=touch dest=/tmp/naïve.yml mode=preserve
+    file: state=touch dest=/tmp/naïve.yml mode=0600


sivel:
We've talked about this rule a little bit and aren't sure this is beneficial.  The original change that necessitated this rule, has been reverted, and we have no intention on adding a change back that affects the result like the previous change did.

I question what this rule is attempting to achieve at this point.

If you think it should be kept, I'm open to hearing the reasoning to understand your POV.  However, I don't think we should really worry with telling people about `preserve`.

Generally speaking, I feel like this rule isn't necessarily "best practice". The problem with ansible-lint, compared to something like pycodestyle for example, is that it's hard to understand intent and need.  While it may be a good idea to use `mode` in some situations, it's not something that is required everywhere.  So making it a rule that people have to disregard in many situations seems like busy work for no benefit.

We've discussed that while we could have many pedantic rules, many should probably be disabled by default.  And maybe there could be a CLI flag called `--strict` or `--pedantic` that enables them all.

nitzmahone:
+1 to @sivel 's comments. A pattern I've liked in other linters is user-definable rule set files rather than (or in addition to) skiplists. The tool defaults to one of the shipped rule sets, but that default is configurable, and users can trivially supply their own rulesets as well (or copy one of the builtins and augment). Something like that would accommodate the "off by default" behavior that we'd like to see for the previous new rule- it's not part of the default rule set, but could be in a `pedantic` or dynamic `all` ruleset (and it's there for users to include in their own rulesets should they choose).

ssbarnea:
Regardless the fact that the CVE related mode change was reverted the subject of missing `mode:` is as valid as possible, as Ansible behavior is still not documented when the mode is not mentioned.

The documentation about "mode" which is also false, advertising unsupported values like "preserve". Only 3/11 modules supporting preserve in reality.

I seen a patch being merged to remove the mention of preserve but this was not backported to all previous versions of ansible, which people may still use. Some of them may even have invalid values in their code because the code was not effectively tested.

IMHO, I would be a serious mistake to ignore the `mode` range of issues.

Even if documentation would be updated to mention clearly what is the file mode behavior for all 11 modules when mode is not mentioned, we still need this rule to underline the wrong usage of "preserve" on those 8 modules that do not allow it. Such a documentation mistake that was online for more than an year does not go away over night. I still see preserve being advertised on both 2.8 and 2.9.

sivel:
Sure, then I suppose at that level, the rule becomes only a way to make sure people are _not_ using `preserve` on a subset of modules, not giving them instruction that they _need_ to use `mode`.

ssbarnea:
> Sure, then I suppose at that level, the rule becomes only a way to make sure people are _not_ using `preserve` on a subset of modules, not giving them instruction that they _need_ to use `mode`.

I am more than happy to repurpose the rule to make it only spot incorrect use of preserve but can we first get the mode behavior documented for cases when it is not mentioned?

At this moment the documentation does not document this:

![https://sbarnea.com/ss/Screen-Shot-2020-09-03-18-59-38.44.png][]

Having examples on how to use it is good, but not mentioning what happens when you don't is clearly an issue, a potential security issue because created files may contain credentials.

The expectation of the most people I asked was the umode would be used, but apparently that is not the case.

sivel:
This was just merged https://github.com/ansible/ansible/pull/71486

ssbarnea:
I am aware about the partial removal of `preseve` from documentation, but that is only half of the story. We still do not have documentation about which file modes ansible will use for creating these files/directories.

I asked @acozine for help here, with bit of luck we may sort this issue.

ssbarnea:
Regarding the proposal of avoiding new rules or specific groups of rules, I raised https://github.com/ansible/ansible-lint/issues/1031 proposal. Please review it.

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -19,45 +19,58 @@
 # THE SOFTWARE.
 from ansiblelint.rules import AnsibleLintRule
 
+# Despite documentation mentioning 'preserve' only these modules support it:
+_modules_with_preserve = (
+    'copy',
+    'ini_file',
```
Sacrome, line 25: Looking at the code of `ini_file` I don't see any references to the argument `mode: preserve`. 
I suspect that only the modules `copy` and `template` have this functionality.

Sacrome: Comment by @sivel confirms my claim: https://github.com/ansible/ansible/issues/56839#issuecomment-680082091
> @geerlingguy that is correct. Per my comment above, `preserve` only applies to `copy` and by extension `template`. Someone incorrectly moved some docs into the shared fragment which was wrong, no one has submitted a fix for the docs issue as of yet.

Sacrome (changes requested):

lib/ansiblelint/rules/MissingFilePermissionsRule.py
ssbarnea: Thanks for spotting the bug. I fixed it and also extended the tests. Please take another look, hopefully we will do it right this time.

Sacrome (approved):

Sacrome:
@ssbarnea Thanks for updating the linter! 
I've already approved your PR, but left a small comment, please take a look at it and see if you agree 😄

odyssey4me:
re: ``preserve``
It seems to me that if a module doesn't accept/use a parameter then that's a bug which should be solved in ansible itself - not in a linting tool.

Sacrome:
> re: `preserve`
> It seems to me that if a module doesn't accept/use a parameter then that's a bug which should be solved in ansible itself - not in a linting tool.

The documentation has actually been fixed in Ansible v2.10 and forward, with two PRs open for backporting this to v2.9 and v2.8. The linter however, incorrectly suggests `mode: preserve` on a number of modules. Which is why this PR exists.

Sacrome:
Note that as of Ansible v2.9.13 the behaviour of the `mode` parameter has been changed, reverting the fix for CVE-2020-1736. 

https://github.com/ansible/ansible/blob/stable-2.9/changelogs/CHANGELOG-v2.9.rst#security-fixes

ssbarnea:
> Note that as of Ansible v2.9.13 the behaviour of the `mode` parameter has been changed, reverting the fix for CVE-2020-1736.
> 
> [https://github.com/ansible/ansible/blob/stable-2.9/changelogs/CHANGELOG-v2.9.rst#security-fixes](https://github.com/ansible/ansible/blob/stable-2.9/changelogs/CHANGELOG-v2.9.rst?rgh-link-date=2020-09-07T12%3A07%3A20Z#security-fixes)

E208 has almost nothing to do with the mentioned CVE. It has to do with the fact that with any version of Ansible, if you do not specify a value for "mode", you will end-up having unpredictable file permissions. The final result does not only depend on the version of Ansible but also on the operating system and the connection method. That is issue that this rule is trying to warn about and this is why is tagged with `unpredictability`. Arguably we could also use `security` as not being explicit about it, is also a potential security issue.

Please note that rule was tagged as experimental, this means it will no longer produce errors by default (they will only be warnings).

ssbarnea:
> re: `preserve`
> It seems to me that if a module doesn't accept/use a parameter then that's a bug which should be solved in ansible itself - not in a linting tool.

Ansible itself is moving at a very slow pace and they struggle even to fix documentation errors. For example, even if we all discovered that preserve was documented on all 11 modules when only 3 were supporting it week ago, preserve was removed only from devel. Guess what, >95% of ansible users are reading documentation of 2.7, 2.8 and 2.9. For good reasons they read the same version of documentation as the version of the ansible they use in production.

As of today there are not even opened PR to backport the documentation fixes. These being said, that is very different from the linter, which can warn about both unsafe code. The linter is also a primary input tool that enables people to upgrade their roles/playbooks to newer versions of Ansible.

Sacrome:
> > re: `preserve`
> > It seems to me that if a module doesn't accept/use a parameter then that's a bug which should be solved in ansible itself - not in a linting tool.
> 
> Ansible itself is moving at a very slow pace and they struggle even to fix documentation errors. For example, even if we all discovered that preserve was documented on all 11 modules when only 3 were supporting it week ago, preserve was removed only from devel. Guess what, >95% of ansible users are reading documentation of 2.7, 2.8 and 2.9. For good reasons they read the same version of documentation as the version of the ansible they use in production.
> 
> As of today there are not even opened PR to backport the documentation fixes. These being said, that is very different from the linter, which can warn about both unsafe code. The linter is also a primary input tool that enables people to upgrade their roles/playbooks to newer versions of Ansible.

Just to add to that;
I have actually opened up PR's last weekend for backporting the documentation to 2.8 and 2.9: 
https://github.com/ansible/ansible/pull/71632
https://github.com/ansible/ansible/pull/71633

But indeed, I'm still waiting for them to be reviewed and merged.

Sacrome:
> > re: `preserve`
> > It seems to me that if a module doesn't accept/use a parameter then that's a bug which should be solved in ansible itself - not in a linting tool.
> 
> Ansible itself is moving at a very slow pace and they struggle even to fix documentation errors. For example, even if we all discovered that preserve was documented on all 11 modules when only 3 were supporting it week ago, preserve was removed only from devel. Guess what, >95% of ansible users are reading documentation of 2.7, 2.8 and 2.9. For good reasons they read the same version of documentation as the version of the ansible they use in production.
> 
> As of today there are not even opened PR to backport the documentation fixes. These being said, that is very different from the linter, which can warn about both unsafe code. The linter is also a primary input tool that enables people to upgrade their roles/playbooks to newer versions of Ansible.

Alright, thanks for looking into that. I found the changelog to be very sparse and wasn't aware what it was exactly - as no ticket or PR was mentioned. So I felt like mentioning it here might've been useful 😉

Sacrome:
Alright, thanks for looking into that. I found the changelog to be very sparse and wasn't aware what it was exactly - as no ticket or PR was mentioned. So I felt like mentioning it here might've been useful 😉

> > Note that as of Ansible v2.9.13 the behaviour of the `mode` parameter has been changed, reverting the fix for CVE-2020-1736.
> > [https://github.com/ansible/ansible/blob/stable-2.9/changelogs/CHANGELOG-v2.9.rst#security-fixes](https://github.com/ansible/ansible/blob/stable-2.9/changelogs/CHANGELOG-v2.9.rst?rgh-link-date=2020-09-07T12%3A07%3A20Z#security-fixes)
> 
> E208 has almost nothing to do with the mentioned CVE. It has to do with the fact that with any version of Ansible, if you do not specify a value for "mode", you will end-up having unpredictable file permissions. The final result does not only depend on the version of Ansible but also on the operating system and the connection method. That is issue that this rule is trying to warn about and this is why is tagged with `unpredictability`. Arguably we could also use `security` as not being explicit about it, is also a potential security issue.
> 
> Please note that rule was tagged as experimental, this means it will no longer produce errors by default (they will only be warnings).

Alright, thanks for looking into that. I found the changelog to be very sparse and wasn't aware what it was exactly - as no ticket or PR was mentioned. So I felt like mentioning it here might've been useful 😉

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -19,57 +19,77 @@
...
+_modules_with_preserve = (
+    'copy',
+    'template',
+)
+
 
 class MissingFilePermissionsRule(AnsibleLintRule):
     id = "208"
-    shortdesc = 'File permissions not mentioned'
+    shortdesc = 'File permissions not mentioned or incorrect'
```
webknjaz, line 31:
```suggestion
    shortdesc = 'File permissions unset or incorrect'
```

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -19,57 +19,77 @@
 # THE SOFTWARE.
 from ansiblelint.rules import AnsibleLintRule
 
+# Despite documentation mentioning 'preserve' only these modules support it:
+_modules_with_preserve = (
```
webknjaz, line 23: It looks like it's semantically better to use set and not a tuple here.

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -19,57 +19,77 @@
...
     id = "208"
-    shortdesc = 'File permissions not mentioned'
+    shortdesc = 'File permissions not mentioned or incorrect'
     description = (
-        "Missing mode parameter can cause unexpected file permissions based "
-        "on version of Ansible being used. Be explicit, or if you still "
-        "want the default behavior you can use ``mode: preserve`` to avoid "
-        "hitting this rule. See "
-        "https://github.com/ansible/ansible/issues/71200"
+        f"Missing or unsupported mode parameter can cause unexpected file "
```
webknjaz, line 33: f-prefix is only necessary on lines with interpolation
```suggestion
        "Missing or unsupported mode parameter can cause unexpected file "
```

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -19,57 +19,77 @@
...
-        "Missing mode parameter can cause unexpected file permissions based "
-        "on version of Ansible being used. Be explicit, or if you still "
-        "want the default behavior you can use ``mode: preserve`` to avoid "
-        "hitting this rule. See "
-        "https://github.com/ansible/ansible/issues/71200"
+        "Missing or unsupported mode parameter can cause unexpected file "
+        f"permissions based "
+        f"on version of Ansible being used. Be explicit, like ``mode: 0644`` to "
+        f"avoid hitting this rule. Special ``preserve`` value is accepted "
+        f"only by {', '.join(_modules_with_preserve)} modules. "
```
webknjaz, line 37: Only thing line needs an f prefix

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -19,57 +19,77 @@
...
+        if module in self._modules_with_create:
+            create = task["action"].get("create", self._modules_with_create[module])
+            return create and mode is None
+
         # A file that doesn't exist cannot have a mode
         if task['action'].get('state', None) == "absent":
             return False
 
-        # A symlink always has mode 0o777
+        # A symlink always has mode 0777
```
webknjaz, line 80: This change is misleading. Before it, it was clearly octal, now it's not

lib/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -19,57 +19,77 @@
...
-    _modules_with_create = (
-        'blockinfile',
-        'ini_file',
-        'lineinfile'
-    )
+    _modules_with_create = {
+        'blockinfile': False,
+        'ini_file': True,
+        'lineinfile': False,
+        'htpasswd': True,
```
rafaelfolco, line 58: ok htpasswd apprently supports preserve from 2.6: https://docs.ansible.com/ansible/latest/modules/htpasswd_module.html#parameter-mode

don't you keep these lists in alphabetical order ?

rafaelfolco (approved):
lgtm

lib/ansiblelint/rules/MissingFilePermissionsRule.py
Sacrome: The documentation of Ansible 2.9 is incorrect. The `htpasswd` module does not support `preserve`. 
I have verified this myself and created PRs for fixing the docs. You can read more on this problem here: https://github.com/ansible/ansible/issues/56839

rafaelfolco (approved):
lgtm

gundalow:
What's needed to progress this?

sivel (approved):

gundalow (approved):

bbaassssiiee:
I wonder what the correct permission would be for `unarchive:`? Is it the top-level directory? I always thought if `mode:` was undefined that it was governed by the umask of the use connecting.

ssbarnea:
Unarchive is quite tricky because based on archive type it may or may not inherit permissions from it. It is up to the user to make an informed decision, case by case.

bbaassssiiee:
For unarchive of tarfile a proper implementation of "mode: preserve" would be quite useful...

PR #1033, opened 2020-09-04
Use NamedTemporaryFile for run_playbook

ssbarnea:
Avoid lower level functions and creation of temporary directory when we just want to test a simple playbook.

### DIFF
--- a/lib/ansiblelint/testing/__init__.py
+++ b/lib/ansiblelint/testing/__init__.py
@@ -31,11 +31,10 @@ class RunFromText(object):
 
     def run_playbook(self, playbook_text):
         """Lints received text as a playbook."""
-        play_root = tempfile.mkdtemp()
-        with open(os.path.join(play_root, 'playbook.yml'), 'w') as fp:
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".yml", prefix="playbook") as fp:
             fp.write(playbook_text)
-        results = self._call_runner(fp.name)
-        shutil.rmtree(play_root)
+            fp.flush()
+            results = self._call_runner(fp.name)
         return results
 
     def run_role_tasks_main(self, tasks_main_text):


lib/ansiblelint/testing/__init__.py
```diff
@@ -31,12 +31,11 @@ def _call_runner(self, path) -> List["MatchError"]:
...
-        play_root = tempfile.mkdtemp()
-        with open(os.path.join(play_root, 'playbook.yml'), 'w') as fp:
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".yml", prefix="playbook") as fp:
             fp.write(playbook_text)
-        results = self._call_runner(fp.name)
-        shutil.rmtree(play_root)
-        return results
+            fp.flush()
+            results = self._call_runner(fp.name)
+            return results
```
webknjaz, line 38: Only `self._call_runner()` needs to be inside of CM, let's return after the tmp dir has been cleaned up:
```suggestion
        return results
```

webknjaz, line 38: Alternatively:
```suggestion
            return self._call_runner(fp.name)
```

lib/ansiblelint/testing/__init__.py
```diff
@@ -31,11 +31,10 @@ def _call_runner(self, path) -> List["MatchError"]:
 
     def run_playbook(self, playbook_text):
         """Lints received text as a playbook."""
-        play_root = tempfile.mkdtemp()
-        with open(os.path.join(play_root, 'playbook.yml'), 'w') as fp:
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".yml", prefix="playbook") as fp:
             fp.write(playbook_text)
-        results = self._call_runner(fp.name)
-        shutil.rmtree(play_root)
+            fp.flush()
```
webknjaz, line 36: this may be useful too
```suggestion
            fp.flush()  # guarrantee it's on disk before linting
```

webknjaz (approved):

Issue #1031, opened 2020-09-03
proposal: require experimental tag for new rules


ssbarnea:
##### Summary

1. Newly introduced rules should always have `experimental` tag added to them.
2. The experimental tag is to be removed when a major version is released
3. A major release would never be made if not more than two weeks passed since all experimental new rules were released, giving a window of at least two weeks between a new rule becomes not experimental.

##### Issue Type

- Feature Idea

##### Additional Information

By combining the 3 proposed changes a user can express its preferences like below:

```yaml
# .ansiblelint
skip_list:
  - experimental  # experimental rule violations are skipped
```

```yaml
# .ansiblelint
warn_list:
  - experimental  # experimental rule violations are displayed but do not affect the result
```

It should be mentioned that the default behavior still remains to have all rules enabled. This is an important aspect to remember because it assures that new users will be able to identify problems and is also compatible with use of custom rules. could 

If we implement this we also enable users to also disable any set of tags, for example they could easily add `galaxy` tag to ignore all current and future rules that are related to galaxy.

geerlingguy:
I would be in favor of this but would like to make sure the default behavior is equivalent to:

```
warn_list:
  - experimental
```

That way while a rule is still 'baking', I would start seeing the warnings in output, but they wouldn't immediately kill all 100+ of my CI jobs that run either daily or weekly.

ssbarnea:
I am concerned that a default warning may confuse users but I lets see what other thing about your proposal. If we decide to do it like this it would mean that as soon you put any value in warn_list in the config, the default is overridden. So if someone wants a warn as errors behavior, they would have to write `warn_list: []`

Issue #997, opened 2020-08-25
Create test that prevent non parsable stdout content


ssbarnea:
##### Summary
We need a new test (or update on existing test) that would assure that when running with parsable format, we do get only parsable format on stdout (regardless if is success or failure exit code).

##### Additional Information

This would prevent regressions that required a fix like the one in https://github.com/ansible/ansible-lint/pull/992/files 

If done well it should catch a random print() that could make the output, not parsable.

ssbarnea:
@jbackman If you can do this it would be really nice.

PR #1034, opened 2020-09-04
Improve violation avoidance hints

ssbarnea:
- Assure hints are sent on stderr in order to avoid messing parseable output
- Hints are not printed on quiet mode
- If an experimental rule violation is found, it includes example on how to disable experimental rules
- Fixes wrong single quote at the end of the rule comment

sample-output

Related: #1031


webknjaz (approved):

webknjaz:
@ssbarnea approving but I'd prefer to have actual tests per https://github.com/ansible/ansible-lint/issues/997.

PR #1037, opened 2020-09-04
Ignore experimental violations for final result

ssbarnea:
Implement changes documented in #1031:
- include `experimental` tag in `warn_list` unless user specify other value inside the config file.
- assures final report (hint) is displayed whenever we have any violations, regardless the result code.
- tag E208 as experimental

Fixes: #1031

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -87,7 +87,7 @@ def choose_formatter_factory(
 def report_outcome(matches: List["MatchError"], options) -> int:
     """Display information about how to skip found rules.
 
-    Returns exit code, 2 if erros were found, 0 when only warnings were found.
+    Returns exit code, 2 if errors were found, 0 when only warnings were found.
     """
     failure = False
     msg = """\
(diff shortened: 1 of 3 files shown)


geerlingguy (approved):

lib/ansiblelint/__main__.py
```diff
@@ -97,20 +98,21 @@ def report_outcome(matches: List["MatchError"], options) -> int:
 warn_list:  # or 'skip_list' to silence them completely
 """
 
-    matched_rules = {match.rule.id: match.rule.shortdesc for match in matches}
-    for id in sorted(matched_rules.keys()):
-        if id not in options.warn_list:
-            msg += f"  - '{id}'  # {matched_rules[id]}\n"
+    matched_rules = OrderedDict({match.rule.id: match.rule for match in matches})
```
webknjaz, line 101: There's no point in making an unordered dict unless you keep it ordered originally.

webknjaz: Also, intermediate in-memory dict is unnecessary

webknjaz:
```suggestion
    matched_rules = OrderedDict((match.rule.id, match.rule) for match in matches)
```

lib/ansiblelint/__main__.py
```diff
@@ -97,20 +98,21 @@ def report_outcome(matches: List["MatchError"], options) -> int:
...
+    for id, rule in matched_rules.items():
+        if set([id, *rule.tags]).isdisjoint(options.warn_list):
```
webknjaz, line 103: No need to make a throw-away in-memory list here:
```suggestion
        if {id, *rule.tags}.isdisjoint(options.warn_list):
```

lib/ansiblelint/__main__.py
webknjaz: P.S. Starting Python 3.6 dicts are ordered in CPython (de-facto) and with 3.7+ they are ordered de-jure. So there's probably no point in importing `OrderedDict`.

webknjaz (approved):

PR #1038, opened 2020-09-04
Fix improper syntax highlighting in README

andrewimeson:
The syntax highlighting looks odd when set to Bash since it's output and not actual shell code. This makes it plain instead.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,43 +78,37 @@ The following is the output from ``ansible-lint --help``, providing an overview 
 
 .. code-block::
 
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

README.rst
```diff
@@ -155,7 +149,7 @@ It's important to note that ``ansible-lint`` accepts a list of Ansible playbook
 
 The following lints the role ``geerlingguy.apache``:
 
-.. code-block:: bash
```
webknjaz, line 158: This should've been `shell-session` instead.

PR #1039, opened 2020-09-05
Enable retries on IOErrors

ssbarnea:
As we observed random IOErrors while running tests on MacOS, it is better to include a minimal set of retries, so we can resume IO flakiness.

### DIFF
--- a/test/TestIncludeMissFileWithRole.py
+++ b/test/TestIncludeMissFileWithRole.py
@@ -93,7 +93,7 @@ def test_cases_warning_message(runner, caplog):
         if "Couldn't open" in str(record):
             noexist_message_count += 1
 
-    assert noexist_message_count == 1
+    assert noexist_message_count == 3  # 3 retries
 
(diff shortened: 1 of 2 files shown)


test/TestIncludeMissFileWithRole.py
```diff
@@ -93,7 +93,7 @@ def test_cases_warning_message(runner, caplog):
         if "Couldn't open" in str(record):
             noexist_message_count += 1
 
-    assert noexist_message_count == 1
+    assert noexist_message_count == 3  # 3 retries
```
webknjaz, line 96: Why would it be exactly 3?

ssbarnea: Because this is what we added as maximum retries. That test is testing that an invalid path does trigger the warning. It fails and needs to fail with 3 warnings not one like previously.

webknjaz (approved):

Issue #1001, opened 2020-08-26
E106 is raised if subdirectories are present in the "tasks" folder


nre-ableton:
##### Summary

Sorry to report this bug after 4.3.2 was released; I didn't have time to test ansible-lint from `master` before the release was cut. Anyways, the fix made in https://github.com/ansible/ansible-lint/pull/980 does not work if a role has subdirectories in the `tasks` folder, for example:

```
ansible/roles/foo.bar/tasks/debian/install.yml:1: [E106] Role name foo.bar does not match ``^[a-z][a-z0-9_]+$`` pattern
```

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details
ansible-lint --version: 4.3.2

- ansible installation method: pip
- ansible-lint installation method: pip

ssbarnea:
Ohh joy. Please make a PR and I can help making a new release. Include a test to avoid regression.

nre-ableton:
@ssbarnea I'm not terribly familiar with the codebase, but I'll give it a shot!

nre-ableton:
Ok, I finally got around to looking at this issue. Fix is here: https://github.com/ansible/ansible-lint/pull/1044

PR #1044, opened 2020-09-09
E106: Don't use relative paths for meta/main.yml

nre-ableton:
This change calculates the role's root directory by finding the
"tasks" directory first, rather than using the parent of the parent of
the current file. The problem with using the parent/parent scheme is
that roles which organize their tasks into subdirectories would cause
ansible-lint to erroneously raise E106 for an invalid role name.

---

Fixes https://github.com/ansible/ansible-lint/issues/1001

### DIFF
--- a/lib/ansiblelint/rules/RoleNames.py
+++ b/lib/ansiblelint/rules/RoleNames.py
@@ -54,8 +54,9 @@ class RoleNames(AnsibleLintRule):
     def match(self, file, text):
         path = file['path'].split("/")
         if "tasks" in path:
-            meta = Path(file['path']).parent.parent / "meta" / "main.yml"
             role_name = _remove_prefix(path[path.index("tasks") - 1], "ansible-role-")
+            role_root = path[:path.index("tasks")]
+            meta = Path("/".join(role_root)) / "meta" / "main.yml"
 
             if meta.is_file():
                 meta_data = parse_yaml_from_file(str(meta))
--- a/test/TestCliRolePaths.py
+++ b/test/TestCliRolePaths.py
@@ -95,7 +95,6 @@ class TestCliRolePaths(unittest.TestCase):
         assert len(result.stderr) == 0
         assert result.returncode == 0
 
-    @pytest.mark.xfail
     def test_run_role_name_from_meta(self):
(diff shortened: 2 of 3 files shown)


ssbarnea (changes requested):
AFAIK, roles can exist without tasks but they cannot exist without meta. Also, meta is always at predictable location, regardless how the tasks are organized.

If we are fixing an existing bug, we clearly need a test that prevents a future regression, a test that would reproduce the current bug.

It would even be better if you can add a test that reproduces the current bug and add it using @pytest.mark.xfail before making this fix.

I am saying this because I want to avoid future regressions, especially as this area was touched multiple times, and recently too.

nre-ableton:
Adding a task in a subdirectory of an existing role (such as `valid-due-to-meta`) reproduces the behavior. I tried to explain this in the commit message, but I think that your suggestion is much better. As such, I split the original commit to added the subdirectory and `@pytest.mark.xfail` in a separate commit with explanation there.

ssbarnea:
> Adding a task in a subdirectory of an existing role (such as `valid-due-to-meta`) will result in a failing test and reproduce the behavior. I tried to explain this in the commit message, but I think that your suggestion is much better. As such, I split the original commit to added the subdirectory and `@pytest.mark.xfail` in a separate commit with explanation there.

In that case maybe just add a comment on the empty task file explaining why the file is there, so we do not remove it by mistake?

nre-ableton:
> > Adding a task in a subdirectory of an existing role (such as `valid-due-to-meta`) will result in a failing test and reproduce the behavior. I tried to explain this in the commit message, but I think that your suggestion is much better. As such, I split the original commit to added the subdirectory and `@pytest.mark.xfail` in a separate commit with explanation there.
> 
> In that case maybe just add a comment on the empty task file explaining why the file is there, so we do not remove it by mistake?

Great suggestion, done! Also, I removed the `@pytest.mark.xfail` comment in the subsequent commit that fixes the problem, I forgot to do this before. :sweat_smile:

ssbarnea (approved):

nre-ableton:
@ssbarnea Thanks for the quick review and merge! :clap:

PR #1045, opened 2020-09-09
Adds a one line stats line

ssbarnea:
Adds a one line starts at the end with information about number of found errors, warnings and files.

This adds one additional line with "Finished ..." on stderr that can be seen below: 
sample

stderrr is used instead of stdout in order to avoid affecting result that could be machine parseable. That also applies to any logging.


lib/ansiblelint/__main__.py
```diff
@@ -84,24 +84,30 @@ def choose_formatter_factory(
...
+        if {match.rule.id, *match.rule.tags}.isdisjoint(options.warn_list):
+            failures += 1
+        else:
+            warnings += 1
+
     for id in sorted(matched_rules.keys()):
         if {id, *matched_rules[id].tags}.isdisjoint(options.warn_list):
             msg += f"  - '{id}'  # {matched_rules[id].shortdesc}\n"
-            failure = True
     for match in matches:
```
rafaelfolco, line 111: do we really need to iterate over matches again?

rafaelfolco (approved):
lgtm

lib/ansiblelint/__main__.py
ssbarnea: That is not the right place to look for performance improvements. It is written as is for code readability, which is more important in this case.  Even in the worst case possible of having >1000 violations found, the cost is likely measures in few milliseconds.

Looking for performance is good, there are places where this really matters (see recent fix around parsing files multiple times, which made a 10x speed improvement).

greg-hellings (approved):

ssbarnea:
Already merged.

Issue #675, opened 2020-02-10
test_module_deprecated test fails on newer ansible


webknjaz:
Because of some changes in the Ansible Core repo, tests started failing 4 days ago. This needs to be investigated and fixed. PRs are welcome. If anybody wants to take this task, feel free to drop a comment here and then send a PR.

# Issue Type
- Bug report

# Ansible and Ansible Lint details

This is caused by a hardcoded list of deprecations from inside the linter at https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/DeprecatedModuleRule.py#L19-L29

This list does change with each version of ansible and some of these are now total failures. 

We identified two potential new ways of interacting with ansible in a portable way:

```
ansible-doc -l -j |grep docker   
# ~3s to run, and returns only a list of modules, and deprecations still not visible
{                                                                                                                  	"docker_compose": "Manage multi-container Docker applications with Docker Compose",
    "docker_config": "Manage docker configs",
    "docker_container": "manage docker containers",
}
```

```
ansible-doc --metadata-dump
# expensive (10s) to run and produces a dictionary of modules like
# biggest problem: it does not include deprecated modules or info about deprecation!!!
{
    "module_name": {
        "description": "...",
        "name": "...",
        "namespace": "...",
        "version_added": "..."
    }
}
```

Apparently ansible displays the deprecation warnings (module renames), only on STDERR when it encounteres an outdated name. Even worse is the fact that the stderr output is split on multiple lines due to https://github.com/ansible/ansible/issues/71461 and there is no option to collect these warnings. Apparently they were not implemented using python warnings support, and thus is not possible to collect them from outside.

> The biggest challenge here is to find the list of deprecations and do it in a way that works for 2.8. 2.9 and 2.10

# Desired Behaviour

List of deprecated modules should be dynamically retried from ansible itself and not being a static list.

# Actual Behaviour (Bug report only)

One test started failing in the CI when running against ansible/ansible@devel.
`test/TestDeprecatedModule.py::TestDeprecatedModuleRule::test_module_deprecated`: https://travis-ci.com/ansible/ansible-lint/jobs/285300600#L391

webknjaz:
I've xfailed it. Whoever is going to implement the fix, must start `git revert e766ead` so that the actual test result will show up in the test results.

Ref: https://pganssle-talks.github.io/xfail-lightning.

greg-hellings:
[...]

PR #1046, opened 2020-09-09
Tell Ansible to skip module parsing

### LINKED ISSUE #538: Collection modules inside tasks cause exceptions

greg-hellings:
To make way for modules called from a collection, ask Ansible to hold
off on parsing the module names. We cannot validate and guarantee,
successfully, that such modules are valid or not as part of a static
linting tool

Fixes #538

### DIFF
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -32,7 +32,7 @@ from pathlib import Path
 from typing import Callable, ItemsView, List, Optional, Tuple
 
 import yaml
-from ansible import constants
+from ansible import cli, constants
 from ansible.errors import AnsibleError, AnsibleParserError
 from ansible.parsing.dataloader import DataLoader
 from ansible.parsing.mod_args import ModuleArgsParser
@@ -416,8 +416,14 @@ def normalize_task_v2(task: dict) -> dict:  # noqa: C901
 
     sanitized_task = _sanitize_task(task)
     mod_arg_parser = ModuleArgsParser(sanitized_task)
+    ansible_version = cli.CLI.version_info()
     try:
-        action, arguments, result['delegate_to'] = mod_arg_parser.parse()
+        if ansible_version['major'] > 2 or \
+           (ansible_version['major'] == 2 and ansible_version['minor'] >= 9):
+            action, arguments, result['delegate_to'] = \
+                mod_arg_parser.parse(skip_action_validation=True)
+        else:
+            action, arguments, result['delegate_to'] = mod_arg_parser.parse()
     except AnsibleParserError as e:
         try:
             task_info = "%s:%s" % (task[FILENAME_KEY], task[LINE_NUMBER_KEY])


greg-hellings:
The failures seem to be limited to the known bad issue #675

rafaelfolco (approved):
lgtm

ssbarnea:
We are days away from `ansible==2.10` being released and on the same date we will make minimal version required 2.9 and avoid that conditional. I am inclined to wait for that to happen first.

greg-hellings:
Waiting seems reasonable. I've been puzzling over the failures. Technically this change does remove the need for the XFail referenced in #675, but I don't know if it addresses it in a way that we want to.

Should I go ahead and pull the XFail attribute on that test, or do we want to try and solve this a different way. The test is now behaving properly, but it doesn't fully address #675, because we're not dynamically pulling the list of deprecated modules with this. I don't touch the deprecated bit. I just tell Ansible not to dig too deeply into the module name.

Personally I'm inclined to pull the XFail, because that is now fixed. But we won't say that this fixes #675, as that is an entirely different problem that still requires the implementation of dynamic deprecation lists in Ansible.

caleb15:
Ansible 2.10 seems to be released:

https://docs.ansible.com/ansible/devel/roadmap/ROADMAP_2_10.html

```
Collecting ansible-lint==4.2.0
  Downloading ansible_lint-4.2.0-py2.py3-none-any.whl (57 kB)
     |████████████████████████████████| 57 kB 10.8 MB/s eta 0:00:01
Collecting ansible>=2.7
  Downloading ansible-2.10.0.tar.gz (25.5 MB)
     |████████████████████████████████| 25.5 MB 12.8 MB/s eta 0:00:0101
```

```
#!/bin/bash -eo pipefail
ansible-lint *.yml -x idempotency,201,302,303,501,602
Couldn't parse task at /home/circleci/project/roles/os/tasks/deploy.yml:3 (conflicting action statements: copy, __line__

The error appears to be in '<unicode string>': line 3, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

(could not open file to display line))
AnsibleMapping([ ('name', 'Copy clean_tmp cron job file'),
                 ( 'copy',
                   AnsibleMapping([ ('src', 'clean_tmp'),
                                    ('dest', '/etc/cron.d/clean_tmp'),
                                    ('__line__', 5),
                                    ( '__file__',
                                      '/home/circleci/project/roles/os/tasks/deploy.yml')])),
                 ('skipped_rules', [])])

Exited with code exit status 1
CircleCI received exit code 1
```

americanwookie:
I'm using Ansible 2.10.3 with Ansible-lint 4.3.7, and I've installed a collection using `ansible-galaxy collection install`, but ansible-lint is unable to find it. I found the PR fixes this for me. Is this PR still a good solution for this problem?

lib/ansiblelint/utils.py
```diff
@@ -416,8 +416,14 @@ def normalize_task_v2(task: dict) -> dict:  # noqa: C901
 
     sanitized_task = _sanitize_task(task)
     mod_arg_parser = ModuleArgsParser(sanitized_task)
+    ansible_version = cli.CLI.version_info()
```
ssbarnea, line 419: This is no longer needed because we already require 2.9 or newer and we check for it.

I still think that skip_action_validation should be something that user can enable or disable using config file. Maybe disable them with.:
```
features:
  action-validation: false  # default true
```

ssbarnea (changes requested):

ssbarnea:
Closing as I started to work on it via https://github.com/ansible-community/ansible-lint/pull/1134 (will keep the credits).

Issue #468, opened 2019-01-18
Empty task "-" crashes ansible-lint


qinqon:
# Issue Type
- Bug report

# Ansible and Ansible Lint details

```
ansible --version
ansible 2.7.5
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/home/ellorent/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ellorent/.local/lib/python2.7/site-packages/ansible
  executable location = /home/ellorent/.local/bin/ansible
  python version = 2.7.15 (default, Oct 15 2018, 15:24:06) [GCC 8.1.1 20180712 (Red Hat 8.1.1-5)]

ansible-lint --version
ansible-lint 3.4.23
```

- ansible installation method: one of source, pip, OS package
  pip install --user ansible
- ansible-lint installation method: one of source, pip, OS package
  pre-commit call with git commit -a --amend

# Desired Behaviour

If you have a list of tasks with an empty want if have to report you as failure
- name: foobar
  debug:
    msg: Hello world
- 

# Actual Behaviour (Bug report only)

Instead of failing with a proper message it crashes
Traceback (most recent call last):
  File "/home/ellorent/.cache/pre-commit/repoHkh3h8/py_env-python2/bin/ansible-lint", line 11, in <module>
    sys.exit(main())
  File "/home/ellorent/.cache/pre-commit/repoHkh3h8/py_env-python2/lib/python2.7/site-packages/ansiblelint/__main__.py", line 176, in main
    matches.extend(runner.run())
  File "/home/ellorent/.cache/pre-commit/repoHkh3h8/py_env-python2/lib/python2.7/site-packages/ansiblelint/__init__.py", line 229, in run
    for child in ansiblelint.utils.find_children(arg, self.playbook_dir):
  File "/home/ellorent/.cache/pre-commit/repoHkh3h8/py_env-python2/lib/python2.7/site-packages/ansiblelint/utils.py", line 169, in find_children
    items = _playbook_items(playbook_ds)
  File "/home/ellorent/.cache/pre-commit/repoHkh3h8/py_env-python2/lib/python2.7/site-packages/ansiblelint/utils.py", line 154, in _playbook_items
    return [item for play in pb_data for item in play.items()]
AttributeError: 'NoneType' object has no attribute 'items'

awcrosby:
[...]

PR #1048, opened 2020-09-10
Avoid crash with None tasks

ssbarnea:
As ansible allows tasks to be None, we should also avoid crashing when
encountering them.

Includes regression prevention by assuring we can parse files
with empty tasks.

Fixes: #468

### DIFF
--- a/examples/example.yml
+++ b/examples/example.yml
@@ -47,3 +47,6 @@
     - name: always run
       debug: msg="always_run is deprecated"
       always_run: true
+
+    # empty task is currently accepted by ansible as valid code:
+    -
--- a/lib/ansiblelint/rules/BecomeUserWithoutBecomeRule.py
+++ b/lib/ansiblelint/rules/BecomeUserWithoutBecomeRule.py
@@ -34,14 +34,15 @@ def _get_subtasks(data):
             'always',
             'rescue']
     for name in block_names:
-        if name in data:
+        if data and name in data:
             result += (data[name] or [])
     return result
 
 
 def _nested_search(term, data):
-    return ((term in data) or
-            reduce((lambda x, y: x or _nested_search(term, y)), _get_subtasks(data), False))
+    if data and term in data:
+        return True
+    return reduce((lambda x, y: x or _nested_search(term, y)), _get_subtasks(data), False)
 
 
 def _become_user_without_become(becomeuserabove, data):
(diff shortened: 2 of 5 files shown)


examples/example.yml
```diff
@@ -47,3 +47,6 @@
     - name: always run
       debug: msg="always_run is deprecated"
       always_run: true
+
+    # empty task is currently accepted by ansible as valid code:
+    -
```
rafaelfolco, line 52:
```suggestion
    - name: empty
```
This test is also covered, correct me otherwise.

rafaelfolco (changes requested):
empty name with no action would be a good test case, isn't it?

examples/example.yml
ssbarnea: Putting a name there would render the test useless because it will produce a dictionary with a key name in it. We need a real `null` there for testing, not even an empty dictionary would test it correctly.

greg-hellings (approved):
The `if foo in (bar or []):` is weird syntax to my eyes. Took me a minute to get what it's doing, I've just never seen it used like this in Python before. Once I got my eyes wrapped around it, this looks good.

ssbarnea:
> The `if foo in (bar or []):` is weird syntax to my eyes. Took me a minute to get what it's doing, I've just never seen it used like this in Python before. Once I got my eyes wrapped around it, this looks good.

Yes is weird, probably `if bar and foo in bar` would look better? Even so they both serve the same purpose.

I guess you need a better python knowledge to know that "False or []" returns `[]` and not False (but is evaluated as False if needed).

greg-hellings:
> > The `if foo in (bar or []):` is weird syntax to my eyes. Took me a minute to get what it's doing, I've just never seen it used like this in Python before. Once I got my eyes wrapped around it, this looks good.
> 
> Yes is weird, probably `if bar and foo in bar` would look better? Even so they both serve the same purpose.
> 
> I guess you need a better python knowledge to know that "False or []" returns `[]` and not False (but is evaluated as False if needed).

Yes, I've seen it used in assignment and short-hand execution but hadn't seen it used in a conditional before. I agree that this form is the more commonly encountered conditional guard from my experience.

greg-hellings (approved):
lgtm

albinvass (approved):
lgtm

Issue #1054, opened 2020-09-14
Parseable output including severity and pep8 not working


JiriSko:
##### Summary
Parseable-severity mode together with parseable output in the format of pep8 doesn't show severity in output.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details
```console
bash-4.2# ansible --version
ansible 2.9.13
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /var/task/python/lib/python3.8/site-packages/ansible
  executable location = /var/task/python/bin/ansible
  python version = 3.8.5 (default, Sep 14 2020, 06:51:44) [GCC 7.3.1 20180712 (Red Hat 7.3.1-6)]
bash-4.2# ansible-lint --version
ansible-lint 4.3.4
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
Custom docker image based on `amazonlinux:2.0.20190508` with installed python 3.8.5

##### STEPS TO REPRODUCE
```console
ansible-lint -p --parseable-severity roles/test.yml
```

##### Desired Behaviour
```
roles/test.yml:5: [E301] [HIGH] Commands should not change things if nothing needs doing
roles/test.yml:5: [E303] [HIGH] service used in place of service module
roles/test.yml:8: [E301] [HIGH] Commands should not change things if nothing needs doing
roles/test.yml:8: [E303] [HIGH] service used in place of service module
roles/test.yml:13: [E301] [HIGH] Commands should not change things if nothing needs doing
roles/test.yml:13: [E303] [HIGH] service used in place of service module
roles/test.yml:13: [E502] [MEDIUM] All tasks should be named
```

This behavior is captured from ansible-lint v4.2.0.

##### Actual Behaviour
```
roles/test.yml:5: [E301] Commands should not change things if nothing needs doing
roles/test.yml:5: [E303] service used in place of service module
roles/test.yml:8: [E301] Commands should not change things if nothing needs doing
roles/test.yml:8: [E303] service used in place of service module
roles/test.yml:13: [E301] Commands should not change things if nothing needs doing
roles/test.yml:13: [E303] service used in place of service module
roles/test.yml:13: [E502] All tasks should be named
```

`roles/test.yml` contents:
```yml
---
- hosts: bobbins

  tasks:
  - name: a bad play
    action: command service blah restart

  - name: Start Nginx
    command: service nginx start
[...]

PR #1055, opened 2020-09-14
Show severity in Github Workflow command message

JiriSko:
Fixes: #1054

### DIFF
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -107,9 +107,9 @@ class AnnotationsFormatter(BaseFormatter):
     in a form of GitHub Checks API annotations that appear rendered in pull-
     request files view.
 
-    ::debug file={name},line={line},col={col}::{message}
-    ::warning file={name},line={line},col={col}::{message}
-    ::error file={name},line={line},col={col}::{message}
+    ::debug file={name},line={line},col={col},severity={severity}::{message}
+    ::warning file={name},line={line},col={col},severity={severity}::{message}
+    ::error file={name},line={line},col={col},severity={severity}::{message}
 
     Supported levels: debug, warning, error
     """
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -123,10 +123,11 @@ class AnnotationsFormatter(BaseFormatter):
         file_path = self._format_path(match.filename or "")
         line_num = match.linenumber
         rule_id = match.rule.id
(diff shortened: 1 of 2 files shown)


lib/ansiblelint/formatters/__init__.py
```diff
@@ -123,10 +123,11 @@ def format(self, match: "MatchError", colored: bool = False) -> str:
         file_path = self._format_path(match.filename or "")
         line_num = match.linenumber
         rule_id = match.rule.id
+        severity = match.rule.severity
         violation_details = match.message
         return (
             f"::{level} file={file_path},line={line_num}"
-            f"::[E{rule_id}] {violation_details}"
+            f"::[E{rule_id}] [{severity}] {violation_details}"
```
ssbarnea, line 130: Based on how the format looks, what you want is to add `,severity={}` on the first line and avoid polluting the message itself.

rafaelfolco (commented):
This worked for me only when I specify
ansible-lint --parseable-severity roles/test.yml 

When I use -p (pep8 format), the severity info is gone.
Is this expected ?

ssbarnea (changes requested):
Please convert it to draft until you pass all tests.

ssbarnea:
Code looks ok now but before merging it, do you have a proof it does work? My guess is that github should just ignore the message but it could also prevent it from rendering them. 

Can you please made a manual test of producing such an annotation and confirm that it works?

The second question would, what if github does not display severity at all in the UI. How is this feature going to help the user?

I am quite curious about you use-case around priority as I never seen it more than something put inside the rule documentation.

ssbarnea:
@JiriSko If you cannot produce an example that proves that github actions would not choke when receiving an additional parameter and that it is possible to view the severity on the annotations produced I will have to close this PR.

Even a screenshot is enough. You should be able to test this with a simple print that produces an annotation that is using this format.

The only reason I am asking this is because I want to prevent a potential break in current code by merging this feature without confirmation that it works.

JiriSko:
Sorry it took so long. Here's test with old and new annotation format:
https://github.com/codeacio/ansible-lint/actions/runs/289711321

ssbarnea:
@jimi-c No worries. Now we know that github accepts extra data in severity. Still, my impression is that the severity will not be visible to the user. How is user going to make use of it?

JiriSko:
It's part of my CI to categorize issues to errors and warnings. Isn't visible directly in Github. I'm using it in previous version of ansible-lint.

ssbarnea (approved):

PR #1065, opened 2020-10-02
Add progressive mode

ssbarnea:
Adds a new option --progressive that ease adoption of the linter as it will only report failure if it detects an increase number of violations since previous commit.

An example of the output when run in progressive mode can be seen below:
```
WARNING  Total violation(s) reduced from 165 to 0 since previous commit. Will mark result as success
WARNING  Marked 99 previously known violation(s) as ignored due to progressive mode.
WARNING  Listing 99 violation(s) marked as ignored, likely already known
```

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -24,13 +24,16 @@ import errno
 import logging
 import os
 import pathlib
+import subprocess
 import sys
-from typing import TYPE_CHECKING, List, Set, Type
+from contextlib import contextmanager
+from typing import TYPE_CHECKING, Any, List, Set, Type, Union
 
 from rich.markdown import Markdown
 
 from ansiblelint import cli, formatters
 from ansiblelint.color import console, console_stderr
+from ansiblelint.file_utils import cwd
 from ansiblelint.generate_docs import rules_as_rich, rules_as_rst
 from ansiblelint.rules import RulesCollection
 from ansiblelint.runner import Runner
@@ -96,8 +99,9 @@ You can skip specific rules or tags by adding them to your configuration file:
 # .ansible-lint
 warn_list:  # or 'skip_list' to silence them completely
 """
+    matches_unignored = [match for match in matches if not match.ignored]
 
-    matched_rules = {match.rule.id: match.rule for match in matches}
+    matched_rules = {match.rule.id: match.rule for match in matches_unignored}
     for id in sorted(matched_rules.keys()):
         if {id, *matched_rules[id].tags}.isdisjoint(options.warn_list):
             msg += f"  - '{id}'  # {matched_rules[id].shortdesc}\n"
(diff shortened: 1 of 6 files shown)


lib/ansiblelint/__main__.py
```diff
@@ -151,23 +154,28 @@ def main() -> int:
...
-        matches.extend(runner.run())
+    matches = _get_matches(rules, options)
 
     # Assure we do not print duplicates and the order is consistent
     matches = sorted(set(matches))
 
+    mark_as_success = False
+    if matches and options.progressive:
+        _logger.info(
+            "Matches found, running again on previos revision in order to detect regressions")
```
MarkusTeufelberger, line 165:
```suggestion
            "Matches found, running again on previous revision in order to detect regressions")
```

lib/ansiblelint/__main__.py
```diff
@@ -151,23 +154,28 @@ def main() -> int:
...
+        with _previous_revision():
+            old_matches = _get_matches(rules, options)
+            # remove old matches from current list
+            if len(old_matches) > len(matches):
+                _logger.warning(
+                    "Total violation(s) reducted from %s to %s since previous "
```
MarkusTeufelberger, line 171:
```suggestion
                    "Total violation(s) reduced from %s to %s since previous "
```

lib/ansiblelint/__main__.py
```diff
@@ -177,12 +185,46 @@ def main() -> int:
...
+        runner = Runner(rules, playbook, options.tags,
+                        options.skip_list, options.exclude_paths,
+                        options.verbosity, checked_files)
+        matches.extend(runner.run())
+    return matches
+
+
+@contextmanager
+def _previous_revision():
+    """Create or updates a temporary workdir containing the previous revious."""
```
MarkusTeufelberger, line 214:
```suggestion
    """Create or updatesa temporary workdir containing the previous revision."""
```

MarkusTeufelberger:
If I have 5 errors of type A, fix them in one commit but create 4 new errors of type B in that fix, will this fail (since I introduced 4 new errors) or pass (since only 4 errors are now in the repository, even though they are new)?

ssbarnea:
> If I have 5 errors of type A, fix them in one commit but create 4 new errors of type B in that fix, will this fail (since I introduced 4 new errors) or pass (since only 4 errors are now in the repository, even though they are new)?

I rewrote a lot of the implementation and now it will always fail if newer violations are introduced. Take a look and more important, test locally.

geerlingguy:
This is an interesting feature, for sure—though the first run of Ansible Lint on an existing repo would fail, correct? Or I guess if it runs against a commit that had no new violations compared to the previous commit, then it wouldn't fail.

It would be useful for those who are adopting Ansible Lint on a repo that has a lot of history and needs a lot of cleanup.

lib/ansiblelint/cli.py
```diff
@@ -111,6 +111,12 @@ def get_cli_parser() -> argparse.ArgumentParser:
                         default=False,
                         action='store_true',
                         help="parseable output including severity of rule")
+    parser.add_argument('--progressive', dest='progressive',
+                        default=False,
+                        action='store_true',
+                        help="Return success if it detects a reduction in number"
+                        " of violations compared with previous git commit. This "
+                        "feature works only on git repository clones.")
```
albinvass, line 119: > This feature only works in git repositories

albinvass (approved):

Issue #1066, opened 2020-10-02
improve integration into pre-commit


SimonHeimberg:
##### Summary

Currently the hook of pre-commit is run *on every commit* for *all files*. This two points can be improved.

##### Issue Type

- Feature Idea

##### Additional Information

improve ...

###### on every run
Since fixing #611 the hook is called always, also when no yaml file is modified. This can simply be improved by removing `always: true` from .pre-commit-hooks.yaml and adding `files: files: \.(yaml|yml)$` (or even better `type: yaml`). Because of [`pass_filenames: false`](https://github.com/ansible/ansible-lint/blob/e133bdc/.pre-commit-hooks.yaml#L13), the file names are not passed, but the hook is only triggered when there are yaml files.

###### all files
Since #615 detecting of roles and playbooks is supported. This could be extended so that roles and playbooks can be detected when given from command line. Maybe this should be enabled by an argument, because humans should normally only pass playbooks or rule directories. But the pre-commit could use this to not always test all playbooks of the repo.

SimonHeimberg:
pr fixing *on every run* (1st point), see above

ssbarnea:
Current behavior is by design mainly because this linters is not limited to yaml files. There are changes that can break ansible (and the linter) that do touch non yaml files. Also a change made to one file can have impact on another one making the file by file linting of less use.

That is one of the reasons why linter is recommended to be run on entire repository and not on specific files, allowing it to detect things like roles, which are not even files.

If we would run on files we would only be able to catch around 2/3 of current issues, also file type detection would be crippled. 

Unless someone with good knowledge of the linter is able to convince me it is doable, this will not change.

Hint: if you start passing it yaml files, it will fail when encountering non ansible files, so it will not work.

SimonHeimberg:
I see that only triggering on yaml files is bad. I use an old version of ansible-lint because of old python, where the pre-commit-hook is wrong. So I never detected an error in a template file or a similar thing.

I still suggest to remove `always_run: true`, but to not add any `type: ...` or `file: ...`.
This has the advantage I can use file-related arguments myself (like `exclude: README.md`, `exclude_type: [markdown, shell]` or file: `^ansible/`). So the user could limit when the linter is triggered.

(Passing a non ansible yaml file would not fail when the "playbooks and roles" detection were enhanced.)

ssbarnea:
Ignoring unknown files behavior is going away, in future version it will return as error, just to avoid accidents.

[...]

PR #1067, opened 2020-10-02
only run from pre-commit when a yaml file changed

SimonHeimberg:
Files names are still not passed to the script (because of pass_filenames: false),
but pre-commit does not call ansible-lint when no matching file has changed.

improves #1066


ssbarnea:
Current behavior is by design, I will put more details in ticket.

PR #1077, opened 2020-10-19
Reduce ansible dependency a little bit

felixfontein:
~~This is causing so much pain, let's get rid of it. Obviously we still need the `ansible` Python lib, but the user can provide that also by installing `ansible-base` instead of Ansible.~~

This has mostly been done by @ssbarnea already, I reduced this PR to the part that still makes sense.

### DIFF
--- a/lib/ansiblelint/rules/CommandsInsteadOfArgumentsRule.py
+++ b/lib/ansiblelint/rules/CommandsInsteadOfArgumentsRule.py
@@ -21,19 +21,7 @@
 import os
 
 from ansiblelint.rules import AnsibleLintRule
-from ansiblelint.utils import get_first_cmd_arg
-
-try:
-    from ansible.module_utils.parsing.convert_bool import boolean
-except ImportError:
-    try:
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -48,7 +48,9 @@ Using Pip
```
marleyjustin, line 48: test-requirements.txt

marleyjustin (approved):

marleyjustin (approved):
🤘

README.rst
```diff
@@ -48,7 +48,9 @@ Using Pip
 
 .. code-block:: bash
 
-    pip install ansible-lint
+    pip install ansible-lint                # Assuming you already installed ansible or ansible-base
+    pip install ansible-lint[ansible]       # Installs Ansible
```
ssbarnea, line 52: Add double quotes as zsh may choke on use of brackets in absence of quotes.

ssbarnea (approved):
Thanks.

felixfontein:
@ssbarnea thanks for reviewing and merging!

PR #1087, opened 2020-11-02
Avoid reporting main role task files as unknown

ssbarnea:
Currently main.y[a]ml files are reported initially as unknown files while the role is detected correctly. Later they are linted but we should avoid printing confusing messages.


albinvass (approved):
lgtm

PR #1088, opened 2020-11-02
Refactor ANSI implementation

ssbarnea:
* Replace internal colorize() method with rich own one. This makes the our code much cleaner and easier to maintain.
* Disabling emoji support is needed because otherwise `file:100:` would become an `file`:100: emoji.
* Now "rich" is imported only from inside color module, making easier to refactor
* Assured that command line overrides for coloring do reconfigure consoles
* Make YAML output easier to read and consistent.

Example after:

### DIFF
--- a/lib/ansiblelint/formatters/__init__.py
+++ b/lib/ansiblelint/formatters/__init__.py
@@ -3,8 +3,6 @@ import os
 from pathlib import Path
 from typing import TYPE_CHECKING, Generic, TypeVar, Union
 
-from ansiblelint.color import Color, colorize
-
 if TYPE_CHECKING:
     from ansiblelint.errors import MatchError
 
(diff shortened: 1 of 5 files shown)


lib/ansiblelint/formatters/__init__.py
```diff
@@ -143,25 +116,13 @@ def _severity_to_level(severity: str) -> str:
...
-        return formatstr.format(
-            filename,
-            linenumber,
-            rule_id,
-            severity,
-            message,
-        )
+        return (
+            f"[filename]{filename}[/]:{linenumber}: [[error_code]{rule_id}[/]] "
+            f"[[error_code]{severity}[/]] [dim]{message}[/]")
```
albinvass, line 128: Does this mean the output is always colorized now?

lib/ansiblelint/formatters/__init__.py
```diff
@@ -143,25 +116,13 @@ def _severity_to_level(severity: str) -> str:
 class ParseableSeverityFormatter(BaseFormatter):
 
     def format(self, match: "MatchError", colored: bool = False) -> str:
```
albinvass, line 118:
```suggestion
    def format(self, match: "MatchError") -> str:
```

Looks like the colored argument is not used.

lib/ansiblelint/formatters/__init__.py
```diff
@@ -51,52 +49,27 @@ def format(self, match: "MatchError", colored: bool = False) -> str:
...
-            return formatstr.format(match.rule.id, self._format_path(match.filename or ""),
-                                    match.linenumber)
+        return (
+            f"[error_code]{match.rule.id}[/] "
+            f"[filename]{self._format_path(match.filename or '')}[/]:{match.linenumber}")
 
 
 class ParseableFormatter(BaseFormatter):
 
     def format(self, match: "MatchError", colored: bool = False) -> str:
```
albinvass, line 69:
```suggestion
    def format(self, match: "MatchError") -> str:
```

Unused argument

lib/ansiblelint/formatters/__init__.py
```diff
@@ -51,52 +49,27 @@ def format(self, match: "MatchError", colored: bool = False) -> str:
...
-                                    match.details)
+        return (
+            f"[error_code]{_id}[/] [error_title]{match.message}[/]\n"
+            f"[filename]{self._format_path(match.filename or '')}[/]:{match.linenumber}\n"
+            f"[dim]{match.details}[/]\n")
 
 
 class QuietFormatter(BaseFormatter):
 
     def format(self, match: "MatchError", colored: bool = False) -> str:
```
albinvass, line 61:
```suggestion
    def format(self, match: "MatchError") -> str:
```
Unused argument

lib/ansiblelint/formatters/__init__.py
```diff
@@ -51,52 +49,27 @@ def format(self, match: "MatchError", colored: bool = False) -> str:
 class Formatter(BaseFormatter):
 
     def format(self, match: "MatchError", colored: bool = False) -> str:
```
albinvass, line 51:
```suggestion
    def format(self, match: "MatchError") -> str:
```

Unused argument

albinvass (changes requested):
It looks there's no support to disable colored output but I could be missing something

lib/ansiblelint/formatters/__init__.py
ssbarnea: Nope. Rich does auto fallback to plain when terminal is not available. This means that we are not supposed to have conditionals inside the code regarding if ansi foo else bar.

Issue #1064, opened 2020-10-01
Rule 208 not possible with unarchive or directory with recurse


nerrehmit:
##### Summary
How can one set a mode on the "unarchive" module and on the file module with state:directory and recurse: yes so that rule 208 is ok?

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible 2.9.13

ansible-lint --version
ansible-lint 4.3.5
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
Linux

##### STEPS TO REPRODUCE

unarchive example from the ansible documentation

file module example with state: directory and recurse: yes from documentation

**command**
```
ansible-lint example/test.yml
```
**example playbook**
```
---
- hosts: somehosts

  tasks:
    - name: Extract foo.tgz into /var/lib/foo
      unarchive:
        src: foo.tgz
        dest: /var/lib/foo

    - name: Recursively change ownership of a directory
      file:
        path: /etc/foo
        state: directory
        recurse: yes
        owner: foo
        group: foo
```

##### Desired Behaviour

ansible-lint reports no problems.

##### Actual Behaviour

ansible-lint rule 208 requires a mode for the ansible module `unarchive` https://docs.ansible.com/ansible/latest/collections/ansible/builtin/unarchive_module.html
However the archive might contain files with different permissions (eg. an executable and a README, License, etc.) and therefore there is no single mode that's fitting for all unarchived contents.

ansible-lint rule 208 also requires that a mode is set for the `file` module https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html 
This is problematic in case I apply the module towards a directory with recurse: yes (eg. to change the owner recursively) as directories mainly have `0755` and the containing files are not necessarily executable.

```
[208] File permissions unset or incorrect
example/test.yml:5
Task/Handler: Extract foo.tgz into /var/lib/foo

[208] File permissions unset or incorrect
example/test.yml:10
Task/Handler: Recursively change ownership of a directory
```

abdennour:
I've same issue with `unarchive` module:
If the rule 208 instructs to set the mode, i will lose default mode of each file.
For now, i made it as permissive quality gate  :

```.ansible-lint
# .ansible-lint file

warn_list:
  - experimental
```

nerrehmit:
Thank you for the pull request to fix the issue with the unarchive module.

[...]

PR #1089, opened 2020-11-03
Avoid 208 rule with unarchive module

ssbarnea:
As archives can also contain permissions (like tar ones), we should avoid triggering 208 rule when unarchive module is used.

Fixes: #1064


albinvass (approved):
lgtm, unarchive shouldn't be forced to set mode.

PR #1090, opened 2020-11-03
Avoid rule 208 with file module and recursive

nerrehmit:
If the file module is applied recursively on a directory, for example to change
ownership of files mode should not be enforced.
The same mode is seldom fitting for both the directory and the files contained within.


nerrehmit:
This fixes the second reported problem in #1064

ssbarnea (approved):

PR #1097, opened 2020-11-13
Bump linters

ssbarnea:
- bump linters
- temporary disable new pylint rule raise-missing-from
- needed as current master fails on CI


albinvass (approved):
lgtm

PR #1104, opened 2020-11-24
Fix rotten github actions

ssbarnea:
Disable broken caching on github-actions: as Github finally disabled
set-env due to security issues we are forced to disable our caching
implementation until we find a solution that works.

Switches to native py39 testing as we no longer need deadsnakes,
which are not dead for good due to the same security fix.

Related: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -42,24 +42,24 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: set PY_SHA256
-      run: echo "::set-env name=PY_SHA256::$(python -VV | sha256sum | cut -d' ' -f1)"
-    - name: Pre-commit cache
-      uses: actions/cache@v1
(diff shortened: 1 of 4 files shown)


.github/workflows/tox.yml
```diff
@@ -42,24 +42,24 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: set PY_SHA256
-      run: echo "::set-env name=PY_SHA256::$(python -VV | sha256sum | cut -d' ' -f1)"
```
webknjaz, line 46: You can set output instead and it'll work.

webknjaz (commented):
Also env example: https://github.com/pypa/setuptools/pull/2461/files#diff-207f09dfa0bc604a684631a1c78be7d9aa00f9a6c778eec30afa287990be3da1R110

PR #1115, opened 2020-12-11
Require ansible>=2.9 using extras to avoid direct dependency

ssbarnea:
This change enables the linter to be used with the thin version of Ansible as not everyone may want to install all community collections from the ACD. 

This change should be marked as major and assume a major version bump, as it also marks removal of support for Ansible 2.8 and because it may require users to manually specify which version of ansible they want to install, if they did not mention it already as a separated requirement.

This detects absence of Ansible at runtime (or version mismatch) and fail with a clear message that tells user how to correctly install it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,10 @@ setup_requires =
 
 # These are required in actual runtime:
 install_requires =
-  ansible >= 2.8
+  # ansible-lint does not list ansible as direct dependency in order to
+  # allow user to choose his target version. Be sure that you mention one
+  # of the extras or install a specific version of ansible yourself.
+  packaging
   pyyaml
   rich
   ruamel.yaml >= 0.15.34,<1; python_version < "3.7"
@@ -78,6 +81,20 @@ install_requires =
 [options.entry_points]
 console_scripts =
   ansible-lint = ansiblelint.__main__:main
+
+[options.extras_require]
+# Anyone wanting the full ansible (ACD) should mention it as a standalone
+# dependency as our extras bring the slim versions.
+
+# core will point to ansible-core>=2.11 as soon that is released
+community =
+  ansible>=2.10
+# this will move to latest stable ansible as soon that is released
+core =
+  ansible-base>=2.10
+# will install ansible from devel branch, may break at any moment.
+devel =
+  ansible-core @ git+https://github.com/ansible/ansible.git
 
 [options.packages.find]
 where = lib
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -27,7 +27,7 @@ jobs:
         python-version:
         - 3.8
         os:
-        - ubuntu-latest
+        - ubuntu-20.04
         env:
         - TOXENV: lint
         - TOXENV: docs
@@ -41,7 +41,7 @@ jobs:
       with:
         fetch-depth: 0  # needed by setuptools-scm
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
     # - name: set PY_SHA256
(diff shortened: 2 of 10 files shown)


webknjaz:
@ssbarnea 4.4 is minor, 5.0 is major

ssbarnea:
> @ssbarnea 4.4 is minor, 5.0 is major

Better preach this to Ansible team ;) -- For the moment I was hoping to be able to get versioning in sync with Ansible but I guess I will not see Ansible 4.4 being released before I retire.

Your question is still valid and makes me think even more about adopting the year as the major number, as it seams to become very popular, and an easy indicator of the age of any piece of software (pip 20., docker 20...).

setup.cfg
```diff
@@ -79,5 +81,19 @@ install_requires =
...
+# Anyone wanting the full ansible (ACD) should mention it as a standalone
+# dependency as our extras bring the slim versions.
+
+# current will become to ansible-core>=2.11 once is released
+current =
+  ansible-base>=2.10
+devel =
+  ansible-core @ git+https://github.com/ansible/ansible.git
+ansible210 =
+  ansible-base>=2.10
```
felixfontein, line 94: I really don't think `ansible210` should refer to ansible-*base*. It should refer to Ansible, or it should be called `ansiblebase210`.

felixfontein:
> but I guess I will not see Ansible 4.4 being released before I retire.

Ansible 4.4.0 will be released somewhen next summer (https://hackmd.io/y7BBcweNR3aRVLuMbKkDxw#Proposed-Extremely-Tentative-Schedule-for-Ansible-400).

felixfontein:
Something that seems to be missing: a test which checks whether `check_ansible_presence` actually does as advertised if no form of Ansible is installed.

setup.cfg
ssbarnea: I reworked the extras we defined to avoid generating too many of them. Latest version does expose only 3 extras and users that do not want to use them are welcomed to BYOD (dependency). I also updated the install readme in order to clarify installation options.

ssbarnea:
> Something that seems to be missing: a test which checks whether `check_ansible_presence` actually does as advertised if no form of Ansible is installed.

Believe me that I already considered it but mocking absence of python modules is not easy and what is worse is that mocking it does not reproduce real life scenarios. IMHO that would nullify the benefits of testing. Still, I would be very happy to see a follow-up that addresses it, before we make a release.

ssbarnea:
Unless someone spots a serious problem with the current PR, I will mere it in an hour. I know it may not be perfect and we are likely to do other follow-ups but it **brings the CI to green status** by addressing a long list of problems that broke during the last months.

Having a broken test suite and/or CI prevents others from contributing useful patches, especially when facing multiple failure points.

Please add a comment here mentioning me and I will try to address them.

PS. I will take extra care for the next release in order assure we get enough feedback before tagging it.

tadeboro (approved):

felixfontein:
> Believe me that I already considered it but mocking absence of python modules is not easy and what is worse is that mocking it does not reproduce real life scenarios.

How about creating a new venv without `--system-site-packages`, installing ansible-lint in there without any extras, and try to run it and see what happens?

I don't really know the test framework you're using, so no idea how hard that is to implement. Doing this as a `runme.sh` like test (as the ones in ansible-test) should be pretty easy.

ssbarnea:
We are using pytest and tox to orchestrate the testing, so 100% of our testing happens only inside venvs. Creating a new venv during testing may be possible but would need to include cleanup and will likely double the test runtime, as most of the time is spent installing dependencies.

Another option would be to add new environments to tox matrix, but this means more CI jobs.

We are not using system-site-packages anyway, so that is not important.

Probably the least intrusive way to test for this particular case is to add something to the "linters" jobs, either an extra command in tox.ini that tests for this or a custom in-repo pre-commit hook that tests for this.

A 3rd option would be to collapse build-dists and metadata-validation into a generic packaging env, which will build package, test metadata, install packages, test that it fails correctly due to missing ansible (already missing in that env).

I am inclined to like more this last option as it also simplifies how we run testing and it makes sense as all these checks are part of the packaging/distribution.

PR #1120, opened 2020-12-13
WIP: Change AnsibleLintRule.match signature

ssbarnea:
Adds type checking to match and enforces it to return a specific format instead of the previous very-loose return
type.

`List[MatchError]` return type was picked because it presented few particular benefits:
- still evaluates as bool correctly, empty list means no match(es)
- easy to process (extend list, no conditionals)
- already used as return type by 22/30 internal rules
- would allow us to return multiple matches with different message  each

This change will allow us to improve its use and avoid weird bugs cause by unexpected objects. This change counts as **major** because is likely to affect users that have their own custom rules.
Updating the code to match the new required return type should be quite easy.

In order to avoid runtime exceptions when running with old custom rules that were not fixed, it will raise a specific new `E901: RuntimeErrorRule (901)` when it detects unexpected return value. This will allow user to either fix the broken rule or temporary ignore 901 errors until the custom rule is fixed.

Another change introduced by this change is introduction of TargetFile, class which allows us to group files with their types. This is very important feature because we need to easily distinguish between various file types (tasks, playbooks, galaxy, deps, ....) and to extend them without having to change method signatures in a huge number
of places.

Partial: #893

### DIFF
--- a/lib/ansiblelint/rules/RoleNames.py
+++ b/lib/ansiblelint/rules/RoleNames.py
@@ -21,15 +21,19 @@
 
 import re
 from pathlib import Path
-from typing import List
+from typing import TYPE_CHECKING, List, Optional
 
+from ansiblelint.errors import MatchError
 from ansiblelint.rules import AnsibleLintRule
 from ansiblelint.utils import parse_yaml_from_file
+
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import TargetFile
 
 ROLE_NAME_REGEX = '^[a-z][a-z0-9_]+$'
 
 
-def _remove_prefix(text, prefix):
+def _remove_prefix(text: str, prefix: str) -> str:
     return re.sub(r'^{0}'.format(re.escape(prefix)), '', text)
 
 
@@ -51,7 +55,9 @@ class RoleNames(AnsibleLintRule):
 
     ROLE_NAME_REGEXP = re.compile(ROLE_NAME_REGEX)
 
-    def match(self, file, text):
+    def match(self, file: "TargetFile", line: str, line_no: Optional[int]) -> List["MatchError"]:
+        if not file:
+            return []
         path = file['path'].split("/")
         if "tasks" in path:
             role_name = _remove_prefix(path[path.index("tasks") - 1], "ansible-role-")
@@ -67,8 +73,11 @@ class RoleNames(AnsibleLintRule):
                         pass
 
             if role_name in self.done:
-                return False
+                return []
             self.done.append(role_name)
             if not re.match(self.ROLE_NAME_REGEXP, role_name):
-                return self.shortdesc.format(role_name)
-        return False
+                return [MatchError(
+                    message=self.shortdesc.format(role_name),
+                    filename=file,
+                    rule=self.__class__)]
+        return []
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -5,12 +5,14 @@ import logging
 import os
 import re
 from collections import defaultdict
+from collections.abc import Sequence
 from importlib.abc import Loader
 from time import sleep
-from typing import List
+from typing import List, Optional
 
 import ansiblelint.utils
 from ansiblelint.errors import BaseRule, MatchError, RuntimeErrorRule
+from ansiblelint.file_utils import TargetFile
 from ansiblelint.skip_utils import append_skipped_rules, get_rule_skips_from_line
 
 _logger = logging.getLogger(__name__)
@@ -21,6 +23,9 @@ class AnsibleLintRule(BaseRule):
     def __repr__(self) -> str:
         """Return a AnsibleLintRule instance representation."""
         return self.id + ": " + self.shortdesc
+
+    def match(self, file: "TargetFile", line: str, line_no: Optional[int]) -> List[MatchError]:
+        return []
 
     @staticmethod
     def unjinja(text):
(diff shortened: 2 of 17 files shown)


lib/ansiblelint/rules/RoleNames.py
```diff
@@ -21,15 +21,19 @@
...
 
+if TYPE_CHECKING:
+    from ansiblelint.file_utils import TargetFile
+
 ROLE_NAME_REGEX = '^[a-z][a-z0-9_]+$'
 
 
-def _remove_prefix(text, prefix):
+def _remove_prefix(text: str, prefix: str) -> str:
     return re.sub(r'^{0}'.format(re.escape(prefix)), '', text)
```
felixfontein, line 37: This looks extremely inefficient. Why not simply do:
```suggestion
    return text[len(prefix):] if text.startswith(prefix) else text
```

lib/ansiblelint/rules/__init__.py
```diff
@@ -71,17 +77,15 @@ def matchlines(self, file, text) -> List[MatchError]:
...
-                message=message,
-                linenumber=prev_line_no + 1,
-                details=line,
-                filename=file['path'])
-            matches.append(m)
+            if not isinstance(result, Sequence) or isinstance(result, str):
+                result = [
+                    MatchError(
+                        message=(
+                            f"Ignored deprecated match() return value: {result}, see "
```
felixfontein, line 84:
```suggestion
                            f"Ignored deprecated match() return value: {result!r}, see "
```

lib/ansiblelint/rules/__init__.py
```diff
@@ -174,6 +178,17 @@ def matchyaml(self, file: str, text: str) -> List[MatchError]:
...
 
+class RuntimeErrorRule(AnsibleLintRule):
+    """Used to identify errors."""
+
+    id = '902'
+    shortdesc = 'Unexpected internal error.'
+    description = 'This error can be caused by outdated custom rules.'
+    severity = 'VERY_HIGH'
+    tags = ['core']
+    version_added = 'v4.4.0'
```
felixfontein, line 189: Isn't this a breaking change and should end up in a new major release?

lib/ansiblelint/rules/RoleNames.py
ssbarnea: Outside the scope of this change, but we can it separated. In fact the only reason I touched it was because adding type hints to the other methods required type of it too.

lib/ansiblelint/rules/__init__.py
ssbarnea: Done.

tadeboro (changes requested):
I went over the PR and the change of the `match()` return type to the list of errors seems a bit unnecessary. The benefit of being able to return multiple matches does not sound convincing enough to justify the change.

I went through existing rules and check what they return back. I found 9 rules that define the ` match()` method and they all return back either a boolean value, a match object, or `None`. Additionally, none of the rules define a custom message. This makes sense because each class is a rule with a single error message for a line. You need multiple rules to produce multiple errors on a line.

I would suggest that we define the return value as `bool` (if we do not care about breaking backward-compatibility) or as `Optional[str]`  if we want to retain the current behavior of being able to override the default rule message with a custom one. This would allow us to keep the `match` implementation simple and concise.

lib/ansiblelint/rules/__init__.py
```diff
@@ -49,14 +53,14 @@ def create_matcherror(
             details: str = "",
             filename: str = None) -> MatchError:
         return MatchError(
-            message=message,
+            message=message or self.shortdesc,
```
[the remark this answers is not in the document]
ssbarnea: It is normal that you did not see any rule using custom message as it was not possible but I can give you a practical example were this is needed.

Remember about galaxy.yml file and the platforms that you can write there, some values may be invalid. That means that a single rule may want to generate something like "Platform CentOS was not found among the list recognized by galaxy server". If you list multiple ones, you may get multiple errors.

In fact the invalid role-name does have a variant error message that includes the invalid role name, but is not really affected by this limitation because we will never have a single role generating two invalid name errors.

Another example that used in the past was a hypothetical spell checker which could identify two different bad words on the same file and line.

The same kind of issue could appear on a line that looks like `foo: [invalid-value-a, invalid-value-b]`, where same rule may identify two separated issues: `invalid-value-a is not defined` and `invalid-value-b is not defined`. Keep in mind that some lines are quite long (aka shell/command can be entire scripts).

As these changes will only go into 5.0 version we can afford to break things in order to correct historical mistakes. Still, I plan to also minimize breaking changes and make them only were there are benefits.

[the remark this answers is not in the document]
tadeboro: > Remember about galaxy.yml file and the platforms that you can write there, some values may be invalid. That means that a single rule may want to generate something like "Platform CentOS was not found among the list recognized by galaxy server". If you list multiple ones, you may get multiple errors.

I am pretty sure this is not a kind of error ` match` is expected to find since it is expected to operate on a single line, not a structure of some sort. I am not familiar enough with the codebase to be sure, but I would expect this to be handled by some other method that operates on more than a single line.

> Another example that used in the past was a hypothetical spell checker which could identify two different bad words on the same file and line.
>
> The same kind of issue could appear on a line that looks like `foo: [invalid-value-a, invalid-value-b]`, where same rule may identify two separated issues: `invalid-value-a is not defined` and `invalid-value-b is not defined`. Keep in mind that some lines are quite long (aka shell/command can be entire scripts).

These are all valid scenarios. So I guess we do need to be able to return multiple error messages. But maybe we can just return back a list of strings? Forcing each `match` method to mess around with line numbers feels like a bad design choice since the caller that iterates over the file already has this information. If we return back a list of strings, we probably get the best of both worlds: the match method can be oblivious to the line number it operates on while still being able to return multiple messages, and the match error construction stays centralized in one place that already has enough content to build it.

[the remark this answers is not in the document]
Highaboveus2: Github highaboveus2

Highaboveus2 (commented):
Wish it could be refined  a little

ssbarnea:
Closing as I went for a different approach that minimized the changes to the function signature.

PR #1123, opened 2020-12-14
Add RuntimeErrorRule internal rule

ssbarnea:
This change should add a RuntimeErrorRule which would be the implicit
rule used when creating MatchErrors without mentioning a specific rule.

Current implementation of MatchError has a default rule=None but we
want to drop it so we would no longer have a rule as Optional, as
this can easily cause runtime bug (like the example seen where
we fed it an object instead of a class inside create_matcherror()

### DIFF
--- a/lib/ansiblelint/errors.py
+++ b/lib/ansiblelint/errors.py
@@ -1,6 +1,8 @@
 """Exceptions and error representations."""
 import functools
+from typing import Type
 
+from ansiblelint._internal.rules import BaseRule, RuntimeErrorRule
 from ansiblelint.file_utils import normpath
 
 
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/errors.py
```diff
@@ -26,11 +68,12 @@ def __init__(
...
             details: str = "",
             filename=None,
-            rule=None) -> None:
+            rule: Type[BaseRule] = RuntimeErrorRule
+            ) -> None:
         """Initialize a MatchError instance."""
         super().__init__(message)
 
-        if not (message or rule):
+        if rule == RuntimeErrorRule and not message:
```
webknjaz, line 76: I think it's better to use `is`

PR #1126, opened 2020-12-15
Change deprecated tag to deprecations

### LINKED ISSUE #1106: 'Deprecated rules' should be renamed 'Deprecation rules' to avoid confusion
##### SUMMARY

The documentation refers to the 1xx rules as "Deprecated rules" but this can be misinterpreted as "rules that have been deprecated". To avoid confusion, they should be called "Deprecation rules" instead or something similar.

netzvieh:
To avoid possible confusion that rules with the deprecated tag are
themselves deprecated, we rename the tag to deprecations.

Fixes #1106

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -486,7 +486,7 @@ The following shows the available tags in an example set of rules, and the rules
     behaviour ['[503]']
     bug ['[304]']
     command-shell ['[305]', '[302]', '[304]', '[306]', '[301]', '[303]']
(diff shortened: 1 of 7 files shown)


netzvieh:
I'm not sure if we want to change some of the `shortdesc` fields aswell, "Deprecated always_run" could probably be "always_run is deprecated" etc.

ssbarnea (approved):

ssbarnea (approved):

PR #1127, opened 2020-12-16
Remove old ANSIBLE00.. tags

netzvieh:
The numeric ids should be used instead.


netzvieh:
I'd be open to implement some kind of deprecation if directly removing them is to abrupt. We could show a warning when using these for some releases and at some later point remove them completely.

Also needs a rebase if #1126 is merged. But I thought building both branches from master would be saner.

ssbarnea (approved):

PR #1133, opened 2020-12-17
Move custom rule docs to standalone page

ssbarnea:
This change reduces the size of root README.rst and will allow us to refactor the examples used there and test them.


docs/custom-rules.rst
```diff
@@ -0,0 +1,92 @@
...
+            # If the task include another task or make the playbook fail
+            # Don't force to have a tag
+            if not set(task.keys()).isdisjoint(['include','fail']):
+                return False
+
+            # Task should have tags
+            if not task.has_key('tags'):
+                  return True
+
+        return False
```
greg-hellings, line 61: Check the indentation on this line
```suggestion
            return False
```

docs/custom-rules.rst
```diff
@@ -0,0 +1,92 @@
...
+In ansible-lint 2.0.0, *task['action']['args']* was renamed *task['action']['module_arguments']* to avoid a clash when a module actually takes args as a parameter key (e.g. ec2_tag)
+
+In ansible-lint 3.0.0 *task['action']['module']* was renamed *task['action']['__ansible_module__']* to avoid a clash when a module take module as an argument. As a precaution, *task['action']['module_arguments']* was renamed *task['action']['__ansible_arguments__']*.
+
+Packaging Custom Rules
+----------------------
+
+Ansible-lint provides a sub directory named *custom* in its built-in rules,
+``/usr/lib/python3.8/site-packages/ansiblelint/rules/custom/`` for example, to
+install custom rules since v4.3.1. The custom rules which are packaged as an
```
greg-hellings, line 74:
```suggestion
install custom rules since v4.3.1. The custom rules which are packaged as a
```

Odd bit of grammar, but "a" is used with "usual" because the first sound is not a vowel when pronounced.

docs/custom-rules.rst
```diff
@@ -0,0 +1,92 @@
...
+usual python package installed into this directory will be loaded and enabled
+automatically by ansible-lint.
+
+To make custom rules loaded automatically, you need the followings:
+
+- Packaging your custom rules as an usual python package named some descriptive ones like ``ansible_lint_custom_rules_foo``.
```
greg-hellings, line 80:
```suggestion
- Packaging your custom rules as a usual python package named some descriptive ones like ``ansible_lint_custom_rules_foo``.
```

greg-hellings (changes requested):
Other than those tiny nitpicks, looks good

docs/custom-rules.rst
webknjaz: Good catch! It was broken in the original snippet too.

docs/custom-rules.rst
```diff
@@ -0,0 +1,92 @@
...
+* Description: Behaviour the rule is looking for
+* Tags: one or more tags that may be used to include or exclude the rule
+* At least one of the following methods:
+
+  * ``match`` that takes a line and returns None or False, if the line doesn't match the test, and True or a custom message, when it does. (This allows one rule to test multiple behaviours - see e.g. the *CommandsInsteadOfModulesRule*.)
+  * ``matchtask`` that operates on a single task or handler, such that tasks get standardized to always contain a *module* key and *module_arguments* key. Other common task modifiers, such as *when*, *with_items*, etc., are also available as keys, if present in the task.
+
+An example rule using ``match`` is:
+
+.. code-block:: python
```
webknjaz, line 23: Moving snippets away from the README means that you no longer need to keep the source in RST files. Instead of relying on humans to detect typos, it's easier to rely on actual linters. So consider moving all of the examples to files and just including them here + start linting them.

Here are some examples of how I did this in the past:
* https://github.com/pypa/packaging.python.org/blob/8c1f843/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst#L76-L79
* https://github.com/ansible/pylibssh/blob/55b241a/docs/user_guide.rst#L54-L57

P.S. Also consider emphasizing the most important lines in snippets.

docs/custom-rules.rst
```diff
@@ -0,0 +1,92 @@
...
+---------------------
+
+Rules are described using a class file per rule. Default rules are named *DeprecatedVariableRule.py*, etc.
+
+Each rule definition should have the following:
+
+* ID: A unique identifier
+* Short description: Brief description of the rule
+* Description: Behaviour the rule is looking for
+* Tags: one or more tags that may be used to include or exclude the rule
```
webknjaz, line 15: This is inconsistent: some lines have a capital after a colon, others don't.

ssbarnea:
@webknjaz @greg-hellings Please note that I do not want to change the content of the docs at the same time as moving it. I will implement the mentioned changes but in follow-up after the move. That is because it would much easier to review. Usually not a good idea to combine moves with changes in content.

docs/custom-rules.rst
webknjaz: Not at all odd but people often don't make the distinction between letters and sounds. If you think about it's pretty obvious that "u" is made up of something like `[ju:]`.

docs/rules.rst
```diff
@@ -4,10 +4,102 @@
...
+It's also possible to use the default rules, plus custom rules. This can be done by passing the ``-R`` to indicate that the default rules are to be used, along with one or more ``-r`` options.
+
+Using Tags to Include Rules
+```````````````````````````
+
+Each rule has an associated set of one or more tags. To view the list of tags for each available rule, use the ``-T`` option.
+
+The following shows the available tags in an example set of rules, and the rules associated with each tag:
+
+.. code-block:: bash
```
webknjaz, line 21: "bash" is the wrong lexer choice for this. The correct one is `console` or `shell-session`. Otherwise, it'll attempt to highlight both the prompt and the command output as bash which would look funny.

Ref: https://pygments.org/languages/#interactive-terminal-shell-sessions.

webknjaz:
@ssbarnea noted. I expect that you'll just revisit the comments made here in a separate PR. I also suggest fixing that `return` indentation in README before proceeding with this change.

PR #1134, opened 2020-12-17
Allow skipping action validation

### LINKED ISSUE #538: Collection modules inside tasks cause exceptions

ssbarnea:
To make way for modules called from a collection, ask Ansible to hold
off on parsing the module names. We cannot validate and guarantee,
successfully, that such modules are valid or not as part of a static
linting tool

Fixes #538

### DIFF
--- a/lib/ansiblelint/__main__.py
+++ b/lib/ansiblelint/__main__.py
@@ -19,7 +19,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 """Command line implementation."""
-
 import errno
 import logging
 import os
@@ -32,7 +31,8 @@ from typing import TYPE_CHECKING, Any, List, Set, Type, Union
 from rich.markdown import Markdown
 from rich.syntax import Syntax
 
-from ansiblelint import cli, formatters
+from ansiblelint import formatters
+from ansiblelint.cli import configure
 from ansiblelint.color import console, console_stderr
 from ansiblelint.file_utils import cwd
 from ansiblelint.generate_docs import rules_as_rich, rules_as_rst
@@ -122,11 +122,17 @@ warn_list:  # or 'skip_list' to silence them completely
         return 0
 
 
-def main() -> int:
+def main(args=None) -> int:
     """Linter CLI entry point."""
+    # We want to persist loaded config inside the module
+
+    if args is None:
+        args = sys.argv[1:]
+
     cwd = pathlib.Path.cwd()
 
-    options = cli.get_config(sys.argv[1:])
+    # this will also set options
+    options = configure(args)
 
     initialize_logger(options.verbosity)
     _logger.debug("Options: %s", options)
--- a/lib/ansiblelint/utils.py
+++ b/lib/ansiblelint/utils.py
@@ -57,6 +57,7 @@ from yaml.composer import Composer
 from yaml.representer import RepresenterError
 
 from ansiblelint._internal.rules import AnsibleParserErrorRule
+from ansiblelint.cli import options
 from ansiblelint.constants import CUSTOM_RULESDIR_ENVVAR, DEFAULT_RULESDIR, FileType
 from ansiblelint.errors import MatchError
 from ansiblelint.file_utils import normpath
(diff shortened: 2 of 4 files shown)


ssbarnea:
AFAIK, there is another missing bit: that feature is not documented and currently listing tags via `-T` does not print any descriptions about tags.

Before merging this I will also have to refactor the way we print the tags in order to allow doc strings about them.

lib/ansiblelint/__main__.py
```diff
@@ -125,8 +126,6 @@ def main() -> int:
     """Linter CLI entry point."""
     cwd = pathlib.Path.cwd()
 
-    options = cli.get_config(sys.argv[1:])
```
webknjaz, line 128: Removing this doesn't look right. Look for something that may be mutating this list (it's not immutable).

ssbarnea: @webknjaz we are not removing it, it is happening inside the `config` module. The only issue that we have is that  now we need options to be loaded inside `utils` module, and this affected test suites where we used to test different by just importing the methods we wanted to test. This now triggers config loading as side effect and fails because the command line arguments are not the same.

We could likely avoid it by mocking the import or mocking the cli arguments.

A longer term solution would be to move stuff outside the utils module which used to be the kitchensink.

webknjaz: That doesn't look right either. The typical idiom is to access `sys.argv` on the top level (that would be this module). The main function is supposed to accept it as an argument and the outer func would access `sys.argv` and do return `main(sys.argv)`. That outer func would also have to be specified as a console script entrypoint.

This way, anybody can cleanly call `main()` with any args they want from tests w/o having to worry about the state of the global `sys.argv`.

lib/ansiblelint/_internal/config.py
```diff
@@ -0,0 +1,9 @@
+"""Loads config options."""
+import sys
+
+from ansiblelint import cli
+
+if any(i in ['pytest', 'sphinx'] for i in sys.modules):
+    options = cli.get_config([])
+else:
+    options = cli.get_config(sys.argv[1:])
```
webknjaz, line 9: This is a bad approach because it heavily relies on the global state of `sys.argv` and `sys.modules`. I don't see how this code can be treated as predictable.

webknjaz:
@ssbarnea here's how the work with `sys.argv` is normally organized: https://github.com/ansible-community/ansible-lint/pull/1146

lib/ansiblelint/__main__.py
ssbarnea: Please check again, I rewrote it.

lib/ansiblelint/__main__.py
```diff
@@ -122,11 +123,15 @@ def report_outcome(matches: List["MatchError"], options) -> int:
...
-def main() -> int:
+def main(args=None) -> int:
     """Linter CLI entry point."""
+    if args is None:
+        args = sys.argv[1:]
+
     cwd = pathlib.Path.cwd()
 
-    options = cli.get_config(sys.argv[1:])
+    constants.options = get_config(args)
```
webknjaz, line 133: I don't think you can claim something to be constant if you literally rewrite its value spaghetti-style. Global vars that can be randomly modified from many places in the control flow are something to avoid at any cost.

lib/ansiblelint/utils.py
```diff
@@ -57,7 +57,7 @@
 from yaml.representer import RepresenterError
 
 from ansiblelint._internal.rules import AnsibleParserErrorRule
-from ansiblelint.constants import CUSTOM_RULESDIR_ENVVAR, DEFAULT_RULESDIR, FileType
+from ansiblelint.constants import CUSTOM_RULESDIR_ENVVAR, DEFAULT_RULESDIR, FileType, options
```
webknjaz, line 60: This is going to be problematic depending on whether `options` import happens before or after calling `main()` :exploding_head:

lib/ansiblelint/utils.py
```diff
@@ -435,14 +435,16 @@ def _sanitize_task(task: dict) -> dict:
 
 # FIXME: drop noqa once this function is made simpler
 # Ref: https://github.com/ansible-community/ansible-lint/issues/744
-def normalize_task_v2(task: dict) -> dict:  # noqa: C901
+def normalize_task_v2(task: dict, options=options) -> dict:  # noqa: C901
```
webknjaz, line 438: Using mutable objects as defaults is a known and dangerous antipattern that can lead to hard to follow behaviors.

ssbarnea:
got too tired, closing for now.

PR #1136, opened 2020-12-17
Improved tags listing

ssbarnea:
From now tags are listed with descriptions and easier to read output.

Includes combining `oddity` tag into `unpredictability` as both were mostly synonyms.

### DIFF
--- a/lib/ansiblelint/rules/__init__.py
+++ b/lib/ansiblelint/rules/__init__.py
@@ -245,11 +245,37 @@ class RulesCollection(object):
                           for rule in sorted(self.rules, key=lambda x: x.id)])
 
     def listtags(self) -> str:
+        tag_desc = {
+            "behaviour": "Indicates a bad practice or behavior",
+            "bug": "Likely wrong usage pattern",
+            "command-shell": "Specific to use of command and shell modules",
+            "core": "Related to internal implementation of the linter",
+            "deprecations": "Indicate use of features that are removed from Ansible",
+            "experimental": "Newly introduced rules, by default triggering only warnings",
+            "formatting": "Related to code-style",
+            "idempotency":
+                "Possible indication that consequent runs would produce different results",
+            "idiom": "Anti-pattern detected, likely to cause undesired behavior",
+            "metadata": "Invalid metadata, likely related to galaxy, collections or roles",
+            "module": "Incorrect module usage",
+            "readability": "Reduce code readability",
+            "repeatability": "Action that may produce different result between runs",
+            "resources": "Unoptimal feature use",
+            "safety": "Increase security risk",
+            "task": "Rules specific to tasks",
+            "unpredictability": "This will produce unexpected behavior when run",
+        }
+
         tags = defaultdict(list)
         for rule in self.rules:
             for tag in rule.tags:
-                tags[tag].append("[{0}]".format(rule.id))
-        results = []
+                tags[tag].append(rule.id)
+        result = "# List of tags and how they are used\n"
         for tag in sorted(tags):
-            results.append("{0} {1}".format(tag, tags[tag]))
-        return "\n".join(results)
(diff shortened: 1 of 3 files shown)


lib/ansiblelint/rules/__init__.py
```diff
@@ -245,11 +245,37 @@ def __repr__(self) -> str:
...
 
     def listtags(self) -> str:
+        tag_desc = {
+            "behaviour": "Indicates a bad practice or behavior",
+            "bug": "Likely wrong usage pattern",
+            "command-shell": "Specific to use of command and shell modules",
+            "core": "Related to internal implementation of the linter",
+            "deprecations": "Indicate use of features that are removed from Ansible",
+            "experimental": "Newly introduced rules, by default triggering only warnings",
+            "formatting": "Related to code-style",
```
ssbarnea, line 255: I am also considering combining exiting "formatting" and "readability" tags into a "style", as I seen them to cover the same area, both aspects being part of code style. Use thump up/down for that.

greg-hellings (approved):

lib/ansiblelint/rules/__init__.py
```diff
@@ -245,11 +245,37 @@ def __repr__(self) -> str:
                           for rule in sorted(self.rules, key=lambda x: x.id)])
 
     def listtags(self) -> str:
+        tag_desc = {
+            "behaviour": "Indicates a bad practice or behavior",
+            "bug": "Likely wrong usage pattern",
+            "command-shell": "Specific to use of command and shell modules",
+            "core": "Related to internal implementation of the linter",
+            "deprecations": "Indicate use of features that are removed from Ansible",
```
samccann, line 253: coming in late w/ some comments - deprecation is supposed to mean features that will be removed. Consider changing to  'features that will be removed from Ansible'.

lib/ansiblelint/rules/__init__.py
```diff
@@ -245,11 +245,37 @@ def __repr__(self) -> str:
...
+            "experimental": "Newly introduced rules, by default triggering only warnings",
```
samccann, line 254: "by default triggers as only warnings."

lib/ansiblelint/rules/__init__.py
```diff
@@ -245,11 +245,37 @@ def __repr__(self) -> str:
...
+            "formatting": "Related to code-style",
+            "idempotency":
+                "Possible indication that consequent runs would produce different results",
+            "idiom": "Anti-pattern detected, likely to cause undesired behavior",
```
samccann, line 258: I'm not sure what an anti-pattern would be... but if your users know, then ignore my comment :-)

lib/ansiblelint/rules/__init__.py
```diff
@@ -245,11 +245,37 @@ def __repr__(self) -> str:
...
+            "metadata": "Invalid metadata, likely related to galaxy, collections or roles",
+            "module": "Incorrect module usage",
+            "readability": "Reduce code readability",
+            "repeatability": "Action that may produce different result between runs",
```
samccann, line 262: how is repeatability different from idempotency problems?

lib/ansiblelint/rules/__init__.py
```diff
@@ -245,11 +245,37 @@ def __repr__(self) -> str:
...
+            "resources": "Unoptimal feature use",
```
samccann, line 263: "suboptimal.."

lib/ansiblelint/rules/__init__.py
```diff
@@ -245,11 +245,37 @@ def __repr__(self) -> str:
...
+            "safety": "Increase security risk",
```
samccann, line 264: Any reason not to call this 'unsecure' or 'security' to make it obvious what the problem could be?

samccann (commented):
coming in late with some suggestions and questions

PR #1152, opened 2020-12-24
Allow linting of roles in the current directory

netzvieh:
Make the following return the same results:

/tmp/playbooks$ ansible-lint roles/foo
/tmp/playbooks/roles$ ansible-lint foo
/tmp/playbooks/roles/foo$ ansible-lint .

Before the second command was not picking up the 'foo' role in the directory
ansible-lint was run from, if a role with the same name exists in 
$ANSIBLE_ROLES_PATH.


ssbarnea (changes requested):
It needs a test that prevents a regression, especially as more refactoring are planned.

ssbarnea (approved):

PR #1185, opened 2021-01-01
Add ability to report column position

ssbarnea:
Enable MatchError to store column number and enable formatters from printing it when available.


tadeboro:
Changes in this PR broke the total ordering of the match errors because it did not add the newly added `column` data field to the hash key.

PR #1200, opened 2021-01-04
E304: environment variable with shell is allowed

senyoltw:
The default rule 304 description is  “Environment variables should be passed to ``shell`` or ``command`` through environment argument“.  but in the code only ``command``, so need to fix it.
Reference: Pull Request https://github.com/ansible-community/ansible-lint/pull/477

### DIFF
--- a/lib/ansiblelint/rules/EnvVarsInCommandRule.py
+++ b/lib/ansiblelint/rules/EnvVarsInCommandRule.py
@@ -26,7 +26,7 @@ class EnvVarsInCommandRule(AnsibleLintRule):
     id = '304'
     shortdesc = "Environment variables don't work as part of command"
     description = (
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
There is some confusion as both modules accept environment argument but on command you cannot put in-line vars.

To avoid confusing users even more, lets rephrase both shortdesc and description to make this clear. This rule is only about command module. I am thinking about something like: 

Command module does not accept setting environmnet variables inline. 

Use environment argument to set variables or use shell module which accepts both.

senyoltw:
Thanks for the review. I fixed it.

ssbarnea (approved):

tadeboro (approved):

Issue #1188, opened 2021-01-01
ansible-lint's license (MIT) is not compatible with Ansible's license (GPLv3)


tadeboro:
##### Summary

License file in the ansible-lint's repository and some of the source files contain an MIT license. But since ansible-lint imports GPLv3-licensed parts from the ansible package, it should itself be licensed under GPLv3 as a whole.

Internal-Reference: LOS0549210

ssbarnea:
That is a known issue that needs clarification for sure. The code inside repository itself can be MIT, only the runtime code (.pyc?) would be affected by GPL. Keep in mind that GPL enforcement applies only when you distribute code that combines GPL parts, at this moment, **neither the the source archive, wheel or source code itself contains any GPL code.** 

If in the future we drop the Ansible imports this confusion would also be sorted.

On the other hand, we should update the license of https://github.com/ansible-community/toolset to make it GPL to match the built container license.

I am personally against changing the original license of the code from MIT to GPL as this would be a one way, making impossible to go back.

tadeboro:
WARNING: I am not a lawyer.

The code can probably be licensed under MIT, but as soon as the package is created and distributed o PyPI, that package must be licensed under GPLv3 (see bullet c in section 5 of https://www.gnu.org/licenses/gpl-3.0.html).

tadeboro:
Also, I did look at what it would take to get rid of Ansible dependency, and I am not if we can remove it. Things around the collection loader are especially troublesome.

ssbarnea:
I reopened it a the subject is not yet fully addressed. See https://github.com/ansible-community/ansible-lint/discussions/1832 for details.

I will keep this open and assigned to me.

ssbarnea:
Fixed by #1882

PR #1213, opened 2021-01-06
Clarify licensing

ssbarnea:
Fixes: #1188

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,6 +32,7 @@ classifiers =
   Operating System :: OS Independent
 
   License :: OSI Approved :: MIT License
+  License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)
 
   Programming Language :: Python
   Programming Language :: Python :: 3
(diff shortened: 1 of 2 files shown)


gundalow (changes requested):
@ssbarnea Thank you for updating this.

I'm just putting -1 till I get someone else to review this.

gundalow:
@tadeboro Thank you for raising this (via https://github.com/ansible-community/ansible-lint/issues/1188) how does this working sound?

ssbarnea:
@gundalow While this explanation does not have any legal value, it would worth contacting our legal team for feedback on licensing.

I am also quite curious to find about other Python packages that faced the same issue. I really doubt that **dynamic languages** that do not effectively embed any code from the GPL dependency are affected. In the end the string of bits, as source or package does not contain any GPL code and producing it (archiving) does not make use of any GPL code, so no lawyer would ever be able to win.

On the other hand, the reality is that you cannot make practical use of the project without also installing the GPL bits. 

AFAIK, that counts as weak-dependency  but it worth being documented in order to avoid confusions.

acozine (commented):
I've tried to make the license statement easier to read and understand. Have I retained the correct meaning? 

Once we've agreed on the wording, it would be nice to add links to the licenses in the text.

tadeboro:
The additions do solve the main issue I had: knowing what license applies to contributions. But because IANAL, I have no idea if this is enough to satisfy the GPLv3 requirements.

abadger:
This is the clearest wording I've ever seen concerning this issue.  Well done ssbarnea and acozine!  

This looks like it agrees with the legal advice the lawyers have always given me about similar licensing situations.  +1 from me on that count.

setup.cfg
```diff
@@ -88,16 +89,16 @@ console_scripts =
 
 # core will point to ansible-core>=2.11 as soon that is released
 community =
-  ansible>=2.10
+  ansible>=2.10  # GPLv3+
```
gundalow, line 92: Tracking like this is a great idea, thank you

gundalow (approved):
Given @abadger's comment I'm changing my -1 to +1
Thank you all

PR #1221, opened 2021-01-08
Refactor logging

ssbarnea:
- Makes use of enrich log formatter in order to add coloring to log
  messages
- Use a single logger instance across the entire application

Example of logging output after this change:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,6 +71,7 @@ repos:
     - Sphinx>=3.1.2
     - enrich
     - yamllint
+    - enrich>=1.2.6
 - repo: https://github.com/pre-commit/mirrors-pylint
   rev: v2.6.0
   hooks:
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -30,6 +30,7 @@ from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, List, Set, Type, Union
 
 from enrich.console import should_do_markup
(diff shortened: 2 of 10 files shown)


tadeboro:
Is there a reason for removing the per-module loggers and using one global logger? This feels strange and contradicts the "official" usage patterns from the logging manual.

ssbarnea:
> Is there a reason for removing the per-module loggers and using one global logger? This feels strange and contradicts the "official" usage patterns from the logging manual.

I think that the guidelines you mentioned are for **python libraries**, and from practical point of views ansible-lint is a **python application**, as we do not expect anyone to consume it other than from the command line.

I can look into using per module loggers but I am not sure that is needed.

Note: I will not merge this very soon as I want to get feedback. Even if we agree on all changes I will likely split it in several smaller ones, one per item touched.

tadeboro:
I do not distinguish between libraries and applications since the only difference is the latter one contains an additional module or two for the CLI stuff and an entry point in `setup.cfg`. And as far as I can remember, python docs make no distinction also (but it has been a while since I looked at the docs for the last time).

But even if we want to use a single logger in ansible-lint, maybe we should at least name it `ansiblelint` (after the top-level module name) instead of the current `ansiblelint.logger`?

tadeboro:
And the main reason for suggesting to use per-module loggers is the fact that code already does that, so we can save some work by not touching those parts.

ssbarnea:
Closing because I decided to perform other refactorings first, ones that would ease logging bits.

PR #1231, opened 2021-01-14
Auto install requirements into temp directory

ssbarnea:
If `requirements.yml` is found, install **roles** and **collections** from it into a temporary folder in order to avoid runtime errors when ansible would not be able to find roles or modules.

### DIFF
--- a/lib/ansiblelint/_prerun.py
+++ b/lib/ansiblelint/_prerun.py
@@ -1,4 +1,5 @@
 import os
+import subprocess
 import sys
 
 from packaging import version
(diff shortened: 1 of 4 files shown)


lib/ansiblelint/_prerun.py
```diff
@@ -39,6 +40,56 @@ def prepare_environment() -> None:
...
+            universal_newlines=True,
+            check=False,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        if run.returncode != 0:
+            sys.exit(run.returncode)
+
+        os.environ['ANSIBLE_ROLES_PATH'] = f".cache/roles:{os.environ['ANSIBLE_ROLES_PATH']}"
+        if 'ANSIBLE_COLLECTIONS_PATHS' in os.environ:
```
greg-hellings, line 87: Why is line 87 guarded, but line 86 is not? Is there a guarantee somewhere that ANSIBLE_ROLES_PATH exists?

ssbarnea: Fixed.

PR #1242, opened 2021-01-18
Use glob to generate file list from args 

xoxys:
Addresses https://github.com/ansible-community/ansible-lint/issues/1241

### DIFF
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -752,81 +752,86 @@ def get_lintables(
 
     # passing args bypass auto-detection mode
     if args:
+        matches = []
         for arg in args:
             if os.path.isdir(arg):
-                lintables.append(Lintable(arg, kind="role"))
+                matches.append(Path(arg).resolve())
+                for p in Path(arg).rglob(".*"):
+                    if p.suffix in [".yaml", ".yml"]:
+                        matches.append(p)
             elif os.path.isfile(arg):
-                lintables.append(Lintable(arg, kind="playbook"))
+                matches.append(arg)
             else:
                 _logger.warning("Unable to access %s", arg)
-    else:
 
+        files = OrderedDict.fromkeys(sorted(matches))
+    else:
         files = get_yaml_files(options)
 
-        playbooks: List[str] = []
-        role_dirs: List[str] = []
-        role_internals = {
-            'defaults',
-            'files',
-            'handlers',
-            'meta',
-            'tasks',
-            'templates',
-            'vars',
-        }
+    playbooks: List[str] = []
+    role_dirs: List[str] = []
+    role_internals = {
+        'defaults',
+        'files',
+        'handlers',
+        'meta',
+        'tasks',
+        'templates',
+        'vars',
+    }
 
-        # detect role in repository root:
-        if 'tasks/main.yml' in files or 'tasks/main.yaml' in files:
-            role_dirs.append('.')
+    # detect role in repository root:
+    if 'tasks/main.yml' in files or 'tasks/main.yaml' in files:
+        role_dirs.append('.')
 
-        for p in map(Path, files):
+    for p in map(Path, files):
 
-            try:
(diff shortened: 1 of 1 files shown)


src/ansiblelint/utils.py
```diff
@@ -752,81 +752,88 @@ def get_lintables(
 
     # passing args bypass auto-detection mode
     if args:
+        matches = []
         for arg in args:
             if os.path.isdir(arg):
-                lintables.append(Lintable(arg, kind="role"))
+                for p in Path(arg).rglob("*"):
```
ssbarnea, line 758: The fact that we treat every folder as a role is by design, not accident. I am afraid we cannot start doing glob because it would conflict with role detection.

Example: it is invalid to have a playbookbook at the root of the role. What if user puts one there?  it means that we would not be able to decide if that folder hosts playbooks or is just a role.

Have a look at the auto-detection logic which happens when you do not pass any arguments. Likely this will produce the result you want. If it does not, look into improving it.

I do already have some plans to improve the detection.

xoxys: Ok. But how is it intended to work if someone is running `ansible-lint my_role`? E.g. the molecule playbooks are not checked right now. For a git repo git ls-files is already used to get all YAML files, so I though using a glob to achieve the same for a passed folder should be the same result.

xoxys: `get_yaml_files` is producing an OrderedDict of files to process, the idea was to produce the same OrderedDict for folders passed as args so both cases can be handled in the same way.

ssbarnea: When I initially implemented the auto-detection I did want to do the same and allow folders to be considered the same way.

After spending many days trying to make it work with very weird side-effects, I renounced and only did it for no-args case.

The `git ls-files` approach address more than you may think under the hood:
* allows you to call the linter from any subdirectory of your project, with same behavior
* makes use of gitignore, so you do not endup having to maintain a mirror between .gitignore and ansiblelint own config.
* allows us to have a reference root project directory, something essential as lots of things are directly dependent on it. For example requirements.yml can exist only at root level

You should also note that the only code layout official supported is the collection layout, anything not following it is dangerous and likely to cause weird bugs, sometimes even missing to identify files.

Once we improve the detection and add some testing, we ware likely to even make failure to identify a file type an error, forcing user to make a decision to either ignore or move the file. Still, this is months away as we do not want to alienate users, likely something for 5.1 or 6.0 releases.

The blocker on this is lack of tests for determine filetypes. We need to build a really big list of filenames and what they are supposed to be identified as.

xoxys: Ok

xoxys: Let's close this PR for now as it's going in a completely wrong direction?

xoxys: What I can't get out of my mind is:

> You should also note that the only code layout official supported is the collection layout

What about the legacy standalone roles layout? AFAIK it is still valid to use simple Ansible roles instead of collections, isn't it? Are there any plans from Ansible itself to completely remove standalone roles and just support collections?

ssbarnea:
@xoxys I will close this as its current form is not what we need but I hope it will not deter you from making other pull-requests.

In fact I do think that we will be able to add this feature in the future, but only after we assure that we have a very reliable file-type detection logic in place.

Recently I was able to add the initial set of tests for this, you can see them at https://github.com/ansible-community/ansible-lint/blob/master/test/TestUtils.py#L276-L293 

I really need help testing the new code on various repositories and identify real-world patterns that we need to identify correctly. Once we have a longer list of  examples added to that test, we would be able to change the behavior and allow paths passed to the cli to be considered as globs instead of roles, mainly your original intent.

I am still aware of lots of identification issues as not everyone is following 

Keep in mind: not every usage patterns needs to pass, there are lots of cases where the user made a mistake. Ideally we should be able to raise a specific kind of rule for these (let me take care of this).

Issue #1241, opened 2021-01-18
Inconsistent file list while loading files form args instead of a git repo


xoxys:
##### Summary

Detected files are different if args passed to ansible-lint command.

Example: `ansible-lint my_role` is not linting files in `my_role/molecule/` while `ansible-lint` from inside `my_role` (using git ls-files) does.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details
ansible-lint master branch

##### Desired Behaviour

The File list should be (as much as possible) the same regardless of the used method.

ssbarnea:
This is known and documented, run with --help and you will see that args can only be roles and playbooks, not random folders, at least for now.

xoxys:
To be honest I'm still confused. In my example a role called `my_role` is a folder following the regular Ansible role structure:

```
├── defaults
│   └── main.yml
├── LICENSE
├── meta
│   └── main.yml
├── molecule
│   ├── centos7
│   ├── pytest.ini
│   └── requirements.yml
├── README.md
├── tasks
│   ├── main.yml
│   └── setup.yml
└── templates
    └── services
```

Running `ansible-lint` from inside this folder, everything works as expected. Running it from outside the folder and passing the path to this role as arg (which should be supported if I don't misunderstood your answer) does not produce the same result as e.g. the molecule folder will be ignored completely.

ssbarnea:
No arguments is not the same thing as current directory (".") for ansible-lint. No arguments it means: use git ls-files and try to guess each file type.

When you pass any folder to ansible-lint, you tell it not to look into it but to assume is a role. For files it also means expectation to be a a playbook. This means that you cannot pass a simple vars or tasks file to the linter. These are linted only as side-effects of being loaded or when internal detection (no args) decides they should be one of them.

Few hours ago I made a change to improve detection of molecule playbooks via https://github.com/ansible-community/ansible-lint/pull/1257/files -- give this a try and let me know how it works.

ssbarnea:
I will reopen it because

xoxys:
Difficult at all... I like the current idea.

> If done like this we should expect a more consistent behavior. The downside is that it will work only with git repositories, which could prove to be an issue.

Yes, that might be a problem, but why is it just working with a git repo? You already use `os.walk('.')` as fallback of `git ls-files` right? So the first step should be `determine list of all lintables (ls-files or os.walk)`?

xoxys:
[...]

PR #1302, opened 2021-02-03
Recurse inside given folders

ssbarnea:
When given folders, use the normal detection algorithm to look inside
them, making the outcome in sync with user expectations.

Fixes: #1241


ssbarnea:
@xoxys As this is fixing a bug your raised, I am hoping that you can manually test this patch and confirm that the new behavior is fixing it and not causing more confusions.

xoxys:
Sure I'll give it a try later today.

xoxys:
Yeah seems to work better now

PR #1303, opened 2021-02-03
Add optional rule to check for loop var prefix

ssbarnea:
Based on original custom rule written by @albinvass on
https://opendev.org/zuul/zuul-jobs/blame/branch/master/.rules/ZuulJobsNamespaceLoopVar.py

This new rule becomes active only when loop_var_prefix is defined
inside the config and allows user to use dynamic prefix based
on role folder name.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,6 +1,35 @@
+# .ansible-lint
 exclude_paths:
 - .github/
+# parseable: true
+# quiet: true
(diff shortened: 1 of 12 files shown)


albinvass:
lgtm, I guess if we upgrade ansible-lint in zuul-jobs we can replace the rule with a simple config update.

PR #1306, opened 2021-02-04
Make all rule identifiers text based

ssbarnea:
As we v5 is already a major version, we also take the opportunity to replace all numeric rule ids with text one, like pylint has.

The main reason for this is that the numeric ones were cryptic, while the new ones should be self-documented, at least for most cases.

While this rule changes the output due to new codes being printed it is backwards compatible as old codes are recognized, so users are not forced to update their skip lists when they upgrade. This should ease upgrades.

Checklist
- [x] gather feedback from users regarding proposed new ids
- [x] verify that generated documentation (RTD job output on github) for rules and command line listing via `-L` produce better results than in the past
- [x] review listing of tags (categories) `-T` and assure that we do not have incorrect, redundant or confusing entries
- [x] add aliases for old id to new ones and print warning, to allow people to migrate without forcing them to it right away
- [x] add a safety check to enforce use of lowercase-with-dashes for rule ids, so we assure they remain consistent.

Example output of new rule IDs, do you find one that does not make sense? or that can better rephrased? Keep in mind that these IDs must be under ~30ch, and that they are followed by their shortdesc in most cases.:
```yaml
# ansible-liunt -T
# List of tags and rules they cover
command-shell:  # Specific to use of command and shell modules
  - command-instead-of-module
  - command-instead-of-shell
  - deprecated-command-syntax
  - inline-env-var
  - no-changed-when
  - risky-shell-pipe
core:  # Related to internal implementation of the linter
  - internal-error
  - load-failure
  - parser-error
  - syntax-check
deprecations:  # Indicate use of features that are removed from Ansible
  - deprecated-bare-vars
  - deprecated-command-syntax
  - deprecated-local-action
  - deprecated-module
  - no-jinja-when
  - role-name
experimental:  # Newly introduced rules, by default triggering only warnings
  - risky-file-permisions
formatting:  # Related to code-style
  - no-jinja-nesting
  - no-tabs
  - playbook-extension
  - risky-octal
  - var-spacing
  - yaml
idempotency:  # Possible indication that consequent runs would produce different results
  - git-latest
  - hg-latest
  - no-changed-when
  - package-latest
idiom:  # Anti-pattern detected, likely to cause undesired behavior
  - command-instead-of-module
  - command-instead-of-shell
  - empty-string-compare
  - inline-env-var
  - literal-compare
  - missing-import
  - no-handler
  - no-loop-var-prefix
  - no-relative-paths
  - unamed-task
metadata:  # Invalid metadata, likely related to galaxy, collections or roles
  - meta-incorrect
  - meta-no-info
  - meta-no-tags
  - meta-video-links
unpredictability:
  - partial-become
  - risky-file-permisions
yaml:  # External linter which will also produce its own rule codes.
  - yaml
```

### DIFF
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -26,26 +26,11 @@ for each available rule, use the ``-T`` option.
 The following shows the available tags in an example set of rules, and the
 rules associated with each tag:
 
-.. code-block:: console
 
-    $ ansible-lint -v -T
-
-    behaviour ['[503]']
-    bug ['[304]']
-    command-shell ['[305]', '[302]', '[304]', '[306]', '[301]', '[303]']
-    deprecations ['[105]', '[104]', '[103]', '[101]', '[102]']
-    experimental ['[208]']
-    formatting ['[104]', '[203]', '[201]', '[204]', '[206]', '[205]', '[202]']
-    idempotency ['[301]']
-    idiom ['[601]', '[602]']
-    metadata ['[701]', '[704]', '[703]', '[702]']
-    module ['[404]', '[401]', '[403]', '[402]']
-    oddity ['[501]']
-    readability ['[502]']
-    repeatability ['[401]', '[403]', '[402]']
-    resources ['[302]', '[303]']
-    safety ['[305]']
-    task ['[502]', '[503]', '[504]', '[501]']
+.. command-output:: ansible-lint -v -T
+   :cwd: ..
+   :returncode: 0
+   :nostderr:
 
 To run just the *idempotency* rules, for example, run the following:
 
@@ -56,20 +41,13 @@ To run just the *idempotency* rules, for example, run the following:
 Excluding Rules
 ```````````````
 
-To exclude rules from the available set of rules, use the ``-x SKIP_LIST``
+To exclude rules using their identifiers or tags, use the ``-x SKIP_LIST``
 option. For example, the following runs all of the rules except those with the
-tags *readability* and *safety*:
+tags *formatting* and *metadata*:
 
 .. code-block:: bash
 
-    $ ansible-lint -x readability,safety playbook.yml
-
-It's also possible to skip specific rules by passing the rule ID. For example,
-the following excludes rule *502*:
-
-.. code-block:: bash
-
-    $ ansible-lint -x 502 playbook.yml
+    $ ansible-lint -x formatting,metadata playbook.yml
 
 Ignoring Rules
 ``````````````
@@ -101,8 +79,8 @@ a space-separated list.
 
 .. code-block:: yaml
 
-    - name: this would typically fire GitHasVersionRule 401 and BecomeUserWithoutBecomeRule 501
-      become_user: alice  # noqa 401 501
+    - name: this would typically fire git-latest and partial-become
+      become_user: alice  # noqa git-latest partial-become
       git: src=/path/to/git/repo dest=checkout
 
 If the rule is line-based, ``# noqa [rule_id]`` must be at the end of the
@@ -110,10 +88,10 @@ particular line to be skipped
 
 .. code-block:: yaml
 
-    - name: this would typically fire LineTooLongRule 204 and VariableHasSpacesRule 206
+    - name: this would typically fire LineTooLongRule 204 and var-spacing
       get_url:
         url: http://example.com/really_long_path/really_long_path/really_long_path/really_long_path/really_long_path/really_long_path/file.conf  # noqa 204
-        dest: "{{dest_proj_path}}/foo.conf"  # noqa 206
+        dest: "{{dest_proj_path}}/foo.conf"  # noqa var-spacing
 
 
 It's also a good practice to comment the reasons why a task is being skipped.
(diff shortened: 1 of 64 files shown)


docs/rules.rst
```diff
@@ -56,20 +41,13 @@ To run just the *idempotency* rules, for example, run the following:
 Excluding Rules
 ```````````````
 
-To exclude rules from the available set of rules, use the ``-x SKIP_LIST``
+To exclude rules using their identificators or tags, use the ``-x SKIP_LIST``
```
samccann, line 44:
```suggestion
To exclude rules using their identifiers or tags, use the ``-x SKIP_LIST``
```

src/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -29,7 +29,7 @@
 
 
 class MissingFilePermissionsRule(AnsibleLintRule):
-    id = "208"
+    id = "risky-file-permisions"
```
samccann, line 32:
```suggestion
    id = "risky-file-permissions"
```

src/ansiblelint/constants.py
```diff
@@ -57,3 +57,39 @@ def main():
...
+    '102': 'no-jinja-when',
+    '104': 'deprecated-bare-vars',
+    '105': 'deprecated-module',
+    '106': 'role-name',
+    '202': 'risky-octal',
+    '203': 'no-tabs',
+    '205': 'playbook-extension',
+    '206': 'var-spacing',
+    '207': 'no-jinja-nesting',
+    '208': 'risky-file-permisions',
```
samccann, line 72:
```suggestion
    '208': 'risky-file-permissions',
```

src/ansiblelint/constants.py
```diff
@@ -57,3 +57,39 @@ def main():
...
+    '303': 'command-instead-of-module',
+    '304': 'inline-env-var',
+    '305': 'command-instead-of-shell',
+    '306': 'risky-shell-pipe',
+    '401': 'git-latest',
+    '402': 'hg-latest',
+    '403': 'package-latest',
+    '404': 'no-relative-paths',
+    '501': 'partial-become',
+    '502': 'unamed-task',
```
samccann, line 84:
```suggestion
    '502': 'unnamed-task',
```

src/ansiblelint/rules/TaskHasNameRule.py
```diff
@@ -24,14 +24,14 @@
 
 
 class TaskHasNameRule(AnsibleLintRule):
-    id = '502'
+    id = 'unamed-task'
```
samccann, line 27:
```suggestion
    id = 'unnamed-task'
```

test/TestImportPlaybook.py
```diff
@@ -15,5 +15,5 @@ def test_task_hook_import_playbook(default_rules_collection):
     assert len(results) == 2
     # Assures we detected the issues from imported playbook
     assert 'Commands should not change things' in results_text
-    assert '502' in results_text
+    assert 'unamed-task' in results_text
```
samccann, line 18:
```suggestion
    assert 'unnamed-task' in results_text
```

test/playbook-import/playbook_imported.yml
```diff
@@ -3,7 +3,7 @@
   connection: local
   gather_facts: false
   tasks:
-  - command: echo "no name"  # should generate 502
+  - command: echo "no name"  # should generate unamed-task
```
samccann, line 6:
```suggestion
  - command: echo "no name"  # should generate unnamed-task
```

examples/playbooks/tasks/directory with spaces/main.yml
```diff
@@ -1,3 +1,3 @@
-# this should generate: 502: All tasks should be named
+# this should generate: unamed-task: All tasks should be named
```
samccann, line 1:
```suggestion
# this should generate: unnamed-task: All tasks should be named
```

examples/playbooks/tasks/simpletask.yml
```diff
@@ -1,4 +1,4 @@
 ---
-# 502: All tasks should be named
+# unamed-task: All tasks should be named
```
samccann, line 2:
```suggestion
# unnamed-task: All tasks should be named
```

examples/roles/role_with_task_inclusions/tasks/included_tasks.yml
```diff
@@ -1,2 +1,2 @@
-# this task is missing a name (502)
+# this task is missing a name (unamed-task)
```
samccann, line 1:
```suggestion
# this task is missing a name (unnamed-task)
```

examples/roles/role_with_task_inclusions/tasks/imported_tasks.yml
```diff
@@ -1,2 +1,2 @@
-# this task is missing a name (502)
+# this task is missing a name (unamed-task)
```
samccann, line 1:
```suggestion
# this task is missing a name (unnamed-task)
```

test/TestRulesCollection.py
```diff
@@ -117,3 +118,12 @@ def test_rich_rule_listing():
         assert rule.shortdesc in result.stdout
         # description could wrap inside table, so we do not check full length
         assert rule.description[:30] in result.stdout
+
+
+def test_rules_id_format() -> None:
+    """Asure all our rules have consitent format."""
```
samccann, line 124:
```suggestion
    """Asure all our rules have consistent format."""
```

samccann (commented):
just a few  misspellings...

Issue #1311, opened 2021-02-07
exclude_paths no longer properly works


felixfontein:
##### Summary
This was caused by 4faa359ece85a8365f1f741a89fc36e979ed1e0b.

There seem to be three different bugs:

1. `Runner.is_excluded` no longer works, since it does not convert its input paths to absolute paths, and now it no longer always gets absolute paths.
2. While `get_lintables()` properly processes `exclude_paths`, the excluded files are sometimes added back in by `expand_dirs_in_lintables()`.
3. There must be a third bug, because if I comment out `expand_dirs_in_lintables()`, it still checks these files.

Fixing 1. resolves the issue, but only hides 2. (and might or not might solve 3. - that depends on what 3. actually is).

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details
Happens with `master` branch, and latest alpha release.

##### STEPS TO REPRODUCE
Run `ansible-lint` in https://github.com/felixfontein/ansible-acme/: https://github.com/felixfontein/ansible-acme/runs/1847849117?check_suite_focus=true

`roles/acme_certificate/tasks/dns-ns1-cleanup.yml` and `roles/acme_certificate/tasks/dns-ns1-create.yml` are in `exclude_paths` in `.ansiblelint`, but errors in them are reported.

ssbarnea:
As we have multiple places that can increase the effective list of files producing matches I think it would be better to perform a filtering-out of exclusions just before we report the errors, so we avoid missing errors coming from other places.

Thanks for testing on master, this bug counts as release blocker and I am already working on a fix.

PR #1313, opened 2021-02-07
Fix exclude_paths regression

ssbarnea:
Fixes: #1311


felixfontein:
This still produces the error. I guess the problem is that `is_excluded` is still broken (see 1. in the issue).

felixfontein:
It does work now!

felixfontein (approved):

felixfontein:
@ssbarnea thanks for fixing this!

Issue #1339, opened 2021-02-11
German "umlaut" causes "Unable to determine file type for..."


phihos:
##### Summary

When a filename contains a German umlaut like `äöü` `ansible-lint` crashes with
```
Unable to determine file type for "playbook-with-umlaut-\303\274.yml"
```
Note that the `ü` in this case get escaped to `\303\274`.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version

ansible 2.9.13
  config file = ...
  configured module search path = ['...']
  ansible python module location = ...
  executable location = ...
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]


ansible-lint --version

ansible-lint 5.0.0
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Ubuntu 20.04

##### STEPS TO REPRODUCE

Create a playbook `playbook-with-umlaut-ü.yml`. Run `ansible-lint`.

##### Desired Behaviour

`ansible-lint` should not crash.

##### Actual Behaviour

`ansible-lint`  crashes with:
```
Unable to determine file type for "playbook-with-umlaut-\303\274.yml"
```

(PR is already prepared and will be opened within the next 5 mins)

ssbarnea:
Closed by #1340

PR #1340, opened 2021-02-11
Fix parsing of unicode filenames reported by git ls-files (#1339)

phihos:
I fixed German umlauts in filenames by changing this
```
git ls-files '*.yaml' '*.yml'
```
to this
```
git ls-files -z '*.yaml' '*.yml'
```
The output in not delimited by newlines but by `\0` instead. So I also had to change the output parsing.

A test case is included.

### DIFF
--- a/examples/playbooks/with-umlaut-ä.yml
+++ b/examples/playbooks/with-umlaut-ä.yml
@@ -1,0 +1,5 @@
+---
+- hosts:
+  - localhost
+  roles:
+  - name: node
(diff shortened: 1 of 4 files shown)


phihos:
For the tests to pass with Python 3.6 I had to modify the test function `run_ansible_lint` to merge environment variables from the safe list to the `_env` var even if `env` in not `None`. Else variables like `LANG` are not passed and `subprocess.call(...)` tries to convert the unicode output to acii causing an error.

ssbarnea (approved):

ssbarnea (approved):
Thanks for this PR, it really make me happy to review a pull-request that fixes a bug while explaining what and why. I am hoping to see more from you.

PR #1342, opened 2021-02-12
support extra_vars in syntax check rule

### LINKED ISSUE #1299: AnsibleSyntaxCheckRule doesn't support extra vars
##### Summary
When you have a playbook that depends on extra-vars (e.g the var is undefined in playbook) the lint fails.

For example if you have a playbook that expects `my_foo` variable to be passed at launch with `ansible-playbook -e my_foo=foo myplaybook.yml` the lint would obviously fail, because there's no way to specify that lint should add extra var.

I'm not sure if it's feasible to add another special syntax for specifying extra_vars for each playbook, or maybe it's just better to add `--extra-vars` to ansible lint itself, which would be passed to `ansible-playbook --syntax-check`.

I know, that it might not be the best practice to not provide a default, but it's very convenient that ansible-playbook throws an error :) 

I can submit a PR if there's a need and details are figured out.

##### Issue Type

- Bug Report

```console (paste below)
ansible --version
ansible 2.10.5

ansible-lint --version
ansible-lint 5.0.0a2

```

- ansible-lint installation method: pip from git tag

##### Desired Behaviour

Ansible-lint provides a way to specify extra vars
It'd be great if there was a way to specify them either:

 * per-playbook (using something like ```tags:
  - ansible_lint_extra_vars: { "my_foo": "foo"}
  ```
 * for the whole ansible-lint execution (with new parameter `--extra-vars` for `ansible-lint` itself, which will pass it to the syntax check)

##### Actual Behaviour
<!--- Describe what actually happened. If possible run with extra verbosity (-vvvv) -->

Right now `AnsibleSyntaxCheckRule` invokes `cmd = ['ansible-playbook', '--syntax-check', str(lintable.path)]`, so there's no way to provide extra vars.

skarzi:
Adds ability to define a set of extra vars to be passed to ansible when we call `--syntax-check`, so we can avoid failures caused by them.

Closes: #1299

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -38,5 +38,9 @@ warn_list:
   - git-latest
   - experimetal  # experimental is included in the implicit list
 
-# offline mode disables installation of requirements.yml
+# Offline mode disables installation of requirements.yml
 offline: false
+
+# Define required Ansible's variables to satisfy syntax check
+extra_vars:
+  - foo: bar
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -44,3 +44,7 @@ offline: false
 # Define required Ansible's variables to satisfy syntax check
 extra_vars:
   foo: bar
+  multiline_string_variable: |
+    line1
+    line2
+  complex_variable: ":{;\t$()"
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -43,4 +43,4 @@ offline: false
 
 # Define required Ansible's variables to satisfy syntax check
 extra_vars:
-  - foo: bar
+  foo: bar
--- a/src/ansiblelint/rules/AnsibleSyntaxCheckRule.py
+++ b/src/ansiblelint/rules/AnsibleSyntaxCheckRule.py
@@ -143,8 +143,10 @@ if "pytest" in sys.modules:
 
     def test_extra_vars_passed_to_command(config_options) -> None:
         """Validate `extra-vars` are passed to syntax check command."""
-        config_options.extra_vars = {'foo': 'bar', 'additional_var': 'test'}
-        print(options, config_options, options == config_options)
+        config_options.extra_vars = {
+            'foo': 'bar',
+            'complex_variable': ':{;\t$()',
+        }
         lintable = Lintable('examples/playbooks/extra_vars.yml', kind='playbook')
 
         result = AnsibleSyntaxCheckRule._get_ansible_syntax_check_matches(lintable)
--- a/src/ansiblelint/rules/AnsibleSyntaxCheckRule.py
+++ b/src/ansiblelint/rules/AnsibleSyntaxCheckRule.py
@@ -1,10 +1,12 @@
 """Rule definition for ansible syntax check."""
+import json
 import re
 import subprocess
 import sys
 from typing import List
 
 from ansiblelint._internal.rules import BaseRule, RuntimeErrorRule
+from ansiblelint.config import options
 from ansiblelint.errors import MatchError
 from ansiblelint.file_utils import Lintable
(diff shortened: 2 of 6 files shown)


src/ansiblelint/testing/fixtures.py
```diff
@@ -50,6 +52,15 @@ def rule_runner(request):
     return RunFromText(collection)
 
 
+@pytest.fixture
+def config_options():
+    """Return configuration options that will be restored after testrun."""
+    global options  # pylint: disable=global-statement
```
skarzi, line 58: I am not happy with this fixture, probably the better idea will be to create some `ConfigWrapper` like for instance `pytest-django` do in the `settings` fixture ([reference](https://github.com/pytest-dev/pytest-django/blob/59fdb49fbc380bc09bcac65de8acab767194b553/pytest_django/fixtures.py#L352)).
Do you have any other ideas for a more elegant solution?

.ansible-lint
```diff
@@ -38,5 +38,9 @@ warn_list:
   - git-latest
   - experimetal  # experimental is included in the implicit list
 
-# offline mode disables installation of requirements.yml
+# Offline mode disables installation of requirements.yml
 offline: false
+
+# Define required Ansible's variables to satisfy syntax check
+extra_vars:
+  - foo: bar
```
ssbarnea, line 46: This is not how vars are defined in ansible, you mush use a dictionary, not a list of dictionaries for variables, somethng like
```
extra_vars:
  foo: bar
  zoo: moo
```

src/ansiblelint/rules/AnsibleSyntaxCheckRule.py
```diff
@@ -39,7 +41,15 @@ def _get_ansible_syntax_check_matches(lintable: Lintable) -> List[MatchError]:
             return []
 
         with timed_info("Executing syntax check on %s", lintable.path):
-            cmd = ['ansible-playbook', '--syntax-check', str(lintable.path)]
+            extra_vars_cmd = []
+            if options.extra_vars:
+                extra_vars_cmd = ['--extra-vars', json.dumps(options.extra_vars)]
```
ssbarnea, line 46: This will likely fail to work well for more complex vars, like one containing shell escapable vars, multilines and so.

IMHO, I would instead dump these vars inside `.cache/extra_vars.yml` and pass `@.cache/extra_vars.yml` to ansible, so it will load them from there.

Be sure to add some problematic var example to the list and see if it works. I really do not want to get bug reports about linter failing to load var files, when this is the job of ansible itself.

Maybe we should only support mentioning the extra_vars_filename inside config and force users that want this feature to add another file with them, so we can rely on ansible to process the file.

src/ansiblelint/testing/fixtures.py
ssbarnea: I think that django approach is considerably better than what we have but if you want to do this, please do it as a separated pull-request after we sort the rest.

src/ansiblelint/rules/AnsibleSyntaxCheckRule.py
felixfontein: I don't see why this should fail for more complex variables. The shell is not involved in interpreting this (see `shell=False` below), this is passed directly as-is to ansible-playbook.

ssbarnea: I will be glad to see it as working as is, lets put something like this into the example, which also works as test-file:
```
extra_vars:
  # sample values used for testing
  foo: |
    line1
    line2
  bar: ":{;\t$()"
```

ssbarnea (approved):

PR #1348, opened 2021-02-14
Avoid duplicates when injecting extra ansible lookup paths

joren485:
### Description

Because the default paths are added, without checking if they are already present, they might be added twice:
```Bash
$ ANSIBLE_ROLES_PATH=roles ansible-lint
Added ANSIBLE_ROLES_PATH=roles:roles
```

This PR fixes this behaviour by only adding the default paths if they are not already present.

### DIFF
--- a/src/ansiblelint/_prerun.py
+++ b/src/ansiblelint/_prerun.py
@@ -100,17 +100,15 @@ def _prepare_ansible_paths() -> None:
     if 'ANSIBLE_LIBRARY' in os.environ:
         library_paths = os.environ['ANSIBLE_LIBRARY'].split(':')
 
-    if os.path.exists("plugins/modules") and "plugins/modules" not in library_paths:
-        library_paths.append("plugins/modules")
-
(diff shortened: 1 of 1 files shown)


src/ansiblelint/_prerun.py
```diff
@@ -102,14 +102,16 @@ def _prepare_ansible_paths() -> None:
 
     if os.path.exists("plugins/modules") and "plugins/modules" not in library_paths:
         library_paths.append("plugins/modules")
-
-    if os.path.exists(".cache/collections"):
+    if (
+        os.path.exists(".cache/collections")
+        and ".cache/collections" not in collection_list
```
felixfontein, line 107: Maybe there should be a helper function `add_if_exists_and_not_present` (happy bikeshedding ;) ) which receives the list and the new path.

ssbarnea (approved):

felixfontein (approved):

Issue #790, opened 2020-05-18
pipes: stdin incorrectly triggers detection mode


ssbarnea:
Apparently current logic to trigger auto-detection mode is wrongly triggered when calling the linter using pipes:

```bash
cat test.yml | ansible-lint  # <-- this should lint that file alone
```

Assign it to me.

Note: apparently flake8 has the same behavior, I need to find-out if it is possible to distinguish reliably between being called without any arguments or it the file was fed directly to it.

The tricky part is that the `sys.stdin.isatty()` may not be a reliable way to do it as I suspect that some automation use-cases may return False.

PR #1355, opened 2021-02-15
Adds experimental stdin support

ssbarnea:
Allow ansible-lint to perform linting from stdin when filename arguments as `/dev/stdin` or `-` are given.

Keep in mind:
- no argument still means auto-detect as I found no reliable way to identify that the tool was used with shell pipes
- filename reported is 'stdin', same as flake8 does
- received content is assumed to be a playbook, you cannot lint a task file this way. Others are welcomed to propose improvements for detecting the file type based on its content, because we do not have any `filename` available.

Fixes: #790

### DIFF
--- a/src/ansiblelint/file_utils.py
+++ b/src/ansiblelint/file_utils.py
@@ -3,10 +3,12 @@ import copy
 import logging
 import os
 import subprocess
+import sys
 from argparse import Namespace
 from collections import OrderedDict
 from contextlib import contextmanager
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Set, Union
 
 import wcmatch.pathlib
@@ -83,6 +85,9 @@ def kind_from_path(path: Path):
                 return k
     if path.is_dir():
         return "role"
+
+    if str(path) == '/dev/stdin':
+        return "playbook"
     raise RuntimeError("Unable to determine file type for %s" % pathex)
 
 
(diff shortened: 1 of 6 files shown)


src/ansiblelint/runner.py
```diff
@@ -189,4 +189,12 @@ def _get_matches(rules: "RulesCollection", options: "Namespace") -> LintResult:
     # Assure we do not print duplicates and the order is consistent
     matches = sorted(set(matches))
 
+    # Convert reported filenames into human redable ones, so we hide the
+    # fact we used temporary files when processing input from stdin.
+    for match in matches:
+        for lintable in lintables:
```
webknjaz, line 195: How about using an `itertools.product()` here?

src/ansiblelint/file_utils.py
```diff
@@ -118,12 +127,24 @@ def __init__(
             if role.exists:
                 self.role = role.name
 
-        self.kind = kind or kind_from_path(self.path)
-        # We store absolute directory in dir
-        if self.kind == "role":
-            self.dir = str(self.path.resolve())
+        if str(self.path) in ['/dev/stdin', '-']:
+            self.file = NamedTemporaryFile(mode="w+", suffix="playbook.yml")
```
webknjaz, line 131: Can it be kept in-memory instead?

src/ansiblelint/runner.py
ssbarnea: Not really as I do also have the break.

src/ansiblelint/file_utils.py
ssbarnea: Nope because I do need to feed `ansible-playbook --syntax-check` a file. The rest of the linter code could work fine with in-memory files, but not ansible itself.

PR #1372, opened 2021-02-18
Fix file config extra_vars options loading

skarzi:
In #1342 `extra_vars` option was added to the file config, but it's not properly loaded (sorry, I should add a test for this in the before-mentioned PR :disappointed:).
This PR fixes `extra_vars` option loading by updating `merge_config` function - now all file config options that wasn't declared in `bool`, `scalar_map` or `lists_map` variable are merged into `cli_config`.


ssbarnea (approved):

jrgoldfinemiddleton:
Thanks, this one was biting me.  Thought I was going crazy.

Issue #1354, opened 2021-02-14
5.0.0 fails to recognize modules from within roles tasks files


luizribeiro:
##### Summary

When a role provides a module, (for example, [`kewlfft.aur`](https://github.com/kewlfft/ansible-aur)) and that role is imported through `requirements.yml`, `ansible-lint` fails to find the module provided by the role while parsing a playbook.

This might have a similar root cause to #644. However, it only recently started reproducing after an upgrade to 5.0.0.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
% ansible --version
ansible 2.10.5
  config file = None
  configured module search path = ['/Users/luiz/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/luiz/Library/Caches/pypoetry/virtualenvs/foo-Q681gf9M-py3.9/lib/python3.9/site-packages/ansible
  executable location = /Users/luiz/Library/Caches/pypoetry/virtualenvs/foo-Q681gf9M-py3.9/bin/ansible
  python version = 3.9.1 (default, Feb  3 2021, 07:04:48) [Clang 11.0.0 (clang-1100.0.33.17)]

% ansible-lint --version
ansible-lint 5.0.0
```

- ansible installation method: pip, through poetry
- ansible-lint installation method: pip, through poetry

##### OS / ENVIRONMENT
macOS Mojave

##### STEPS TO REPRODUCE
Setup this `requirements.yml`:

```yaml
---
roles:
  - name: kewlfft.aur
```

With a playbook as simple as:

```yaml
---
- name: Update AUR on arch boxes
  hosts: arch
  remote_user: aur_builder
  tasks:
  - name: ensure apache is at the latest version
    aur:
      upgrade: true
      aur_only: true
```

##### Desired Behaviour

It should be able to find the `aur.py` module that is provided by the `kewlfft.aur` role and `ansible-lint` should succeed.

##### Actual Behaviour

Running `ansible-lint` on this playbook yields:

```console (paste below)
% ansible-lint playbook.yml
Running ansible-galaxy install --roles-path .cache/roles -vr requirements.yml
Running ansible-galaxy collection install -p .cache/collections -vr requirements.yml
Added ANSIBLE_ROLES_PATH=.cache/roles
WARNING  Listing 1 violation(s) that are fatal
[...]

PR #1377, opened 2021-02-18
Disable action validation with tasks by default

ssbarnea:
This change should address mentioned bug by introducing a small risk of not catching some invalid module calls for roles that do not have playbooks inside the same repository.

All users that have at least one playbook that imports the role, will still be able to detect these errors as ansible-playbook --syntax-check finds them.

We plan to add a feature that would create fake temporary playbooks for roles found inside a repository, ones that would help us identify missing or misspelled modules.

Fixes: #1354

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -48,3 +48,7 @@ extra_vars:
     line1
     line2
   complex_variable: ":{;\t$()"
+
+# Uncomment to enforce action validation with tasks, usually is not
+# needed as Ansible syntax check also covers it.
+# skip_action_validation: false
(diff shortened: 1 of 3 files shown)


tadeboro (approved):
Looks good as far as the code goes. I did not test this on an example though.

ssbarnea:
> Looks good as far as the code goes. I did not test this on an example though.

I have some other work in progress to be included in the same patch release, but I wanted to avoid big patches as they are more likely to include bugs. This sorts the most common issue encountered by v5 early adopters.

Issue #1378, opened 2021-02-18
Minimum ansible version check fails with stack dump


JohnMalmberg:
##### Summary
See https://github.com/ansible-community/ansible-lint/blob/master/src/ansiblelint/_prerun.py

The "failed" check block at line 45 is referencing the variable "__version__" which only exists if the except block is thrown.
If the version compare fails this causes a pretty much useless stack trace instead of explaining the version issue.

The version check is also using "<=" instead of "<" for the check, which means that the version set for "ANSIBLE_MIN_VERSION" is not the actual minimum version that will work with the linting program.

This bug appears to have been introduced with ansible-lint 5.0.0.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible 2.9.0
$ ansible-lint --version
ansible-lint 5.0.1
See trace back below,
```

- ansible installation method: one of source, pip
- ansible-lint installation method: one of source, pip

##### OS / ENVIRONMENT
Run in a virtualenv on CentOS 7.9

##### STEPS TO REPRODUCE

ansible-lint --version

```console (paste below)
ansible-lint --version

```

##### Desired Behaviour

 ansible-lint --version
ansible-lint 5.0.1

Possible security bugs should be reported via email to `security@ansible.com`

##### Actual Behaviour

Please give some details of what is actually happening.
Include a [minimum complete verifiable example] with:
- playbook
- output of running ansible-lint
- if you're getting a stack trace, output of
  `ansible-playbook --syntax-check playbook`

```paste below
$ ansible-lint --version
ansible-lint 5.0.1
Traceback (most recent call last):
  File "/home/mypath/roles/test_venv/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
  File "/home/mypath/roles/test_venv/lib64/python3.6/site-packages/ansiblelint/__main__.py", line 254, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
  File "/home/mypath/roles/test_venv/lib64/python3.6/site-packages/ansiblelint/__main__.py", line 155, in main
    initialize_options(argv[1:])
  File "/home/mypath/ansible-lab/roles/test_venv/lib64/python3.6/site-packages/ansiblelint/__main__.py", line 80, in initialize_options
    check_ansible_presence()
[...]

PR #1383, opened 2021-02-19
Rewrite ansible version checking

ssbarnea:
- assure we check that ansible python module and ansible cli versions
  are the same
- print ansible version alongside ansible-lint version
- fix version version comparison from <= to <

Fixes: #1378

### DIFF
--- a/src/ansiblelint/_prerun.py
+++ b/src/ansiblelint/_prerun.py
@@ -3,7 +3,8 @@ import pathlib
 import re
 import subprocess
 import sys
-from typing import List, Optional
+from functools import lru_cache
+from typing import List, Optional, Tuple
 
 from packaging import version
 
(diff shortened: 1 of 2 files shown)


src/ansiblelint/_prerun.py
```diff
@@ -29,18 +29,39 @@
...
+    ansible_cli_version = match.group(1)
     try:
         # pylint: disable=import-outside-toplevel
-        from ansible import release
+        from ansible.release import __version__ as ansible_module_version
 
-        if version.parse(release.__version__) <= version.parse(ANSIBLE_MIN_VERSION):
+        if version.parse(ansible_module_version) < version.parse(ANSIBLE_MIN_VERSION):
             failed = True
     except (ImportError, ModuleNotFoundError, UnboundLocalError) as e:
```
JohnMalmberg, line 62: UnboundLocalErrror probably should not be trapped here, as it should only show up from a coding error, and having it caught in an exception handler could make it harder to track down the issue.

JohnMalmberg (changes requested):
Tests good with out the UnboundLocalError being caught.
Tested with ansible==2.9.0 and ansible==2.8.19

src/ansiblelint/_prerun.py
```diff
@@ -29,30 +29,62 @@
...
-        )
+    ansible_cli_version = ""
+    result = subprocess.run(
+        args=["ansible", "--version"],
+        stdout=subprocess.PIPE,
+        universal_newlines=True,
+        check=False,
+    )
+    if result.returncode != 0:
+        err = "FATAL: Unable to retrieve ansible cli version: %s" % result.stdout
```
felixfontein, line 48: How about moving the main logic into another function which always returns an error; then you can simplify this function like:
```suggestion
        return ansible_cli_version, "FATAL: Unable to retrieve ansible cli version: %s" % result.stdout
```
Then you don't have to indent the remainder of the function.

Issue #1400, opened 2021-02-21
[5.0] A valid role namespace is required, edit meta/main.yml


juju4:
##### Summary

While reviewing an ansible role, I got this message
```
A valid role namespace is required, edit meta/main.yml and assure that
author field is also a valid namespace:

galaxy_info:
  author: your_galaxy_namespace

See: https://galaxy.ansible.com/docs/contributing/namespaces.html#galaxy-namespace-limitations
```
https://github.com/juju4/ansible-lxdconfigure/runs/1946726641?check_suite_focus=true#step:7:1

but author field is already present
https://github.com/juju4/ansible-lxdconfigure/blob/master/meta/main.yml#L2

what was not and missing for ansible-lint is role_name field which historically is implicitly declared but now, maybe required.
Adding role_name removes error.

Message is confusing and should be updated.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible 2.10.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/ansible
  executable location = /opt/hostedtoolcache/Python/3.9.1/x64/bin/ansible
  python version = 3.9.1 (default, Jan 25 2021, 10:38:32) [GCC 9.3.0]

ansible_lint-5.0.2
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
Github Ubuntu 20.04 host

ssbarnea:
That is very weird because I cloned you repository, used the same version (5.0.2) and did not get this error. Unless the author is a non-lowercase-compatible string, this error should not appear, and looking at the code, I see it as having a valid value.

ssbarnea:
@juju4 You may also be interested about https://github.com/ansible-community/ansible-lint/discussions/1403

PR #1401, opened 2021-02-22
Improve standalone role name computing

ssbarnea:
Fixes: #1400

### DIFF
--- a/src/ansiblelint/_prerun.py
+++ b/src/ansiblelint/_prerun.py
@@ -19,16 +19,6 @@ from ansiblelint.constants import (
 )
 from ansiblelint.loaders import yaml_from_file
 
-MISSING_GALAXY_NAMESPACE = """\
-A valid role namespace is required, edit meta/main.yml and assure that
-author field is also a valid namespace:
-
-galaxy_info:
-  author: your_galaxy_namespace
-
-See: https://galaxy.ansible.com/docs/contributing/namespaces.html#galaxy-namespace-limitations
-"""
-
 
 def check_ansible_presence(exit_on_error=False) -> Tuple[str, str]:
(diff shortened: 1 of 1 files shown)


ssbarnea:
@juju4 please test this patch.

src/ansiblelint/_prerun.py
```diff
@@ -162,10 +152,23 @@ def _install_galaxy_role() -> None:
     role_name = yaml['galaxy_info'].get('role_name', None)
     role_author = yaml['galaxy_info'].get('author', None)
     if not role_name:
-        role_name = os.path.dirname(".")
+        role_name = pathlib.Path(".").absolute().name
         role_name = re.sub(r'^{0}'.format(re.escape('ansible-role-')), '', role_name)
-    if not role_author or not re.match(r"[\w\d_]{2,}", role_name):
-        print(MISSING_GALAXY_NAMESPACE, file=sys.stderr)
+    fqrn = f"{role_author}.{role_name}"
+    if not re.match(r"[\w\d_]{2,}\.[a-z][a-z0-9_]+$", fqrn):
```
felixfontein, line 158:
```suggestion
    if not re.match(r"[a-z0-9][a-z0-9_]{2,}\.[a-z][a-z0-9_]+$", fqrn):
```
The document you link to (https://galaxy.ansible.com/docs/contributing/namespaces.html#galaxy-namespace-limitations) specifies explicit restrictions on the galaxy namespace that you are not really checking here.

Also it does not give any restrictions on the role name. Where do you have these from? I think there should also be a reference for them.

ssbarnea: Ouch, we only support one link per rule and the role requirements are listed at https://galaxy.ansible.com/docs/contributing/creating_role.html#role-names

PR #1408, opened 2021-02-23
Add yes and no to truthy check in yamllint

joren485:
### Description

When running `ansible-lint` with `yamllint` installed, `ansible-lint` will complain about the use of the unqouted `yes` and `no` boolean values. However, [`yes` and `no` are accepted boolean values by Ansible](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html).

This PR adds `yes` and `no` to the accepted boolean list.

[Ansible accepts a lot more values as boolean values](https://github.com/ansible/ansible/blob/c22bc46cb960ade62ea9bad240a2ae0cd3a17582/lib/ansible/module_utils/parsing/convert_bool.py#L11-L12). If anyone wants it, I can add the other strings as well.

### DIFF
--- a/src/ansiblelint/rules/YamllintRule.py
+++ b/src/ansiblelint/rules/YamllintRule.py
@@ -24,6 +24,8 @@ YAMLLINT_CONFIG = """
 extends: default
 rules:
   document-start: disable
+  truthy:
+    allowed-values: ['true', 'false', 'yes', 'no']
   # 160 chars was the default used by old E204 rule, but
   # you can easily change it or disable in your .yamllint file.
   line-length:


ssbarnea (changes requested):
I personally oppose this as I used YAML 1.2 spec for very long, even if Ansible still uses the more relaxed YAML 1.1 spec.

If you really want it, you can add you custom .yamllint file with your preferences.

I will not close this right away because I do not want to enforce my oppinated preferences on all users.

If a survey proves that more than ⅔ of users prefer relaxing this check, I will accept it.

Keep in mind that this has effects even outside Ansible files, changing this will affect checks done on ANY YAML file, it may allow you to introduce bugs on some of them.

joren485:
> If you really want it, you can add you custom .yamllint file with your preferences.

It should be noted that this will always [raise a warning](https://github.com/ansible-community/ansible-lint/blob/ed24aa6d75d894f121a36b96792f03009990940f/src/ansiblelint/rules/YamllintRule.py#L59).

A less intrusive solution to the problem this PR tries to fix, might be to only raise a warning if the local `yamllint` configuration conflicts with the `ansible-lint` `yamllint` configuration.

> Keep in mind that this has effects even outside Ansible files, changing this will affect checks done on ANY YAML file, it may allow you to introduce bugs on some of them.

This also applies to [the 160 line length limit that the config has](https://github.com/ansible-community/ansible-lint/blob/ed24aa6d75d894f121a36b96792f03009990940f/src/ansiblelint/rules/YamllintRule.py#L30).

ssbarnea:
You can transform these to warnings by adding them to `warn_list` in config.

joren485:
> You can transform these to warnings by adding them to `warn_list` in config.

Adding `yaml` will turn all `yamllint` violations into warnings. Is it possible to only turn the truthy violations into warnings?.

ssbarnea:
Insted of adding the entire `yaml` rule to warnings, you can only add the specific tag reproted for this issue which is `yaml-truthy` if I remember well, you can see it displayed as red in parentheses after the generic rule name (end of line).

You can use tags/rules-id anywhere, they are interchangeable.

joren485:
Thank you very much, `truthy` works.  This fixes the problem for my specific use case.

I will leave it up to you to close (or accept) this PR, because I believe in general it might be a good idea to support `yes` and `no` as they are supported by Ansible.

As a small side note: some tests and examples in this repo are using `yes`/`no`. If you decide to not want `yes`/`no`, you might want to replace them with `True`/`False`.

klaus-tux:
Hi @ssbarnea and @joren485 !

I was just about to open a new PR with the same change but then I stumbled upon this one.

My argument for introducing this change is the following: The documentation of the ansible modules uses `yes` and `no` as possible choices for boolean parameters at all places. (See [here](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html#parameters) for an example.) Therefore I found it very confusing that the linting tool belonging to ansible tells me that I should use `true` and `false` instead.

I understand the counterargument that this change changes the linting of non-ansible YAML files as well. Nevertheless I would prefer to have a default in ansible-lint that fits to ansible and its documentation more closely. Users who need a more strict checking would have to override this default with their own one in their `.yamllint` file then.

What do you think?

Best regards,
Klaus

ssbarnea:
I will need to repeat it again, a decision about truthy has already being made few months back and Ansible will gradually change all the documentation to advertise only pure `true/false` values. Keep in mind that this will take many months but the process already started. [see](https://github.com/ansible/ansible-lint/discussions/2305) - basically we migrate towards YAML 1.2 specification, which is backwards compatible with 1.1 anyway.

Probably in 6 months from now it will be quite hard to find examples using other values in our documentation.

PS. I will lock this thread to avoid extra noise in the future.

klaus-tux:
Hi @ssbarnea ,

thank you for clarifying. I was not aware, that ansible is moving to YAML 1.2 and `true/false`. Now I see the broader picture. Thank you for the heads-up and sorry for bothering you.

Best,
Klaus

PR #1419, opened 2021-03-02
Avoid creating .cache directory in cwd

ssbarnea:
Instead of using cwd for the .cache folder, we now try to use the project_dir for that, which defaults to the location of
.ansiblelint config file.

Running ansible-lint from a subdirectory of the project, should no longer create .cache folders in new locations.

Execution outside a project or a git repository will make the linter use user home directory as reference,

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -23,6 +23,19 @@ associated tags.
    :cwd: ..
    :returncode: 0
 
+Temporary files
+---------------
+
+As part of the execution, the linter will likely need to create a cache of
+installed or mocked roles, collections and modules. This is done inside
+``{project_dir}/.cache`` folder. The project directory is either given as a
+command line argument, determined by location of the configuration
+file, git project top-level directiory or user home directory as fallback.
+In order to speed-up reruns, the linter does not clean this folder by itself.
+
+If you are using git, you will likely want to add this folder to your
+``.gitignore`` file by running ``git ignore .cache``.
+
 Progressive mode
 ----------------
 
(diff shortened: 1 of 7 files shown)


docs/usage.rst
```diff
@@ -23,6 +23,19 @@ associated tags.
...
+
+As part of the execution, the linter will likely need to create a cache of
+installed or mocked roles, collections and modules. This is done inside
+``{project_dir}/.cache`` folder. The project directory is either given as a
+command line argument, determined by location of the configuration
+file, git project top-level directiory or user home directory as fallback.
+In order to speed-up reruns, the linter does not clean this folder by itself.
+
+If you are using git, you will likely want to add this folder to your
+``.gitignore`` file by running ``git ignore .cache``.
```
aminvakil, line 37:
```suggestion
``.gitignore`` file.
```

aminvakil (commented):
I'm not sure what magic git you're running :) but on latest git installation (2.30.1) from https://git-scm.com/ there is no `git ignore` command as far as I know.

Edit: As far as I can tell other than this little documentation fix, everything else is ok.

ssbarnea:
> I'm not sure what magic git you're running :) but on latest git installation (2.30.1) from https://git-scm.com/ there is no `git ignore` command as far as I know.
> 
> Edit: As far as I can tell other than this little documentation fix, everything else is ok.

I do have few git extensions installed, sometimes I endup with mistakes like this. Thanks for reviewing it.

aminvakil (approved):
Thank you for doing this!

PR #1434, opened 2021-03-05
Add downstream testing pipeline

ssbarnea:
This adss an initial set of repositories which will be tested against regressions.

Some of them are disabled due to known problems that need to be sorted before activating them.

If new linter changes will be reporting new genuine problems we will first attempt to use the downstream contacts in order to fix them. If this does not happen we may decide to drop those repositories from the testing list.

The new eco job will not be required for merging new patches, as we may need to ignore downstream specific issues.

Related: https://github.com/ansible-community/ansible-lint/discussions/1403


greg-hellings (approved):
I like it. Don't know what to make of that CI failure, though.

konstruktoid:
LGTM

geerlingguy (approved):

geerlingguy:
LGTM!

PR #1436, opened 2021-03-06
Avoid looking for config file outside current git repo

ssbarnea:
This fixes a bug where running the linter inside a project that did
not had a config file could endup picking a config file from another
project (was happening only when projects were nested).

From now on, we stop looking at parent directories when we encounter
the .git folder, which can be seen as project boundary.

### DIFF
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -15,6 +15,10 @@ file. Ansible-lint checks the working directory for the presence of this file
 and applies any configuration found there. The configuration file location can
 also be overridden via the ``-c path/to/file`` CLI flag.
 
+When configuration file is not found in current directory, the tool will try
+to look for one in parent directories but it will not go outside current git
+repository.
+
 If a value is provided on both the command line and via a config file, the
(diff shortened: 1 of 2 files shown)


aminvakil:
This is definitely an improvement, but isn't there a better way than looking for `.git` folder in order to be sure `ansible-lint` should not search in parent directory anymore?

ssbarnea:
> This is definitely an improvement, but isn't there a better way than looking for `.git` folder in order to be sure `ansible-lint` should not search in parent directory anymore?

There is but I did not bother using the external command for this particular case because the other approach would have
ended up bigger, and likely a little bit slower.

I do not expect this to be perfect but considering that discovery of config file was not even documented I doubt it will create any regressions. In fact I needed this change do the https://github.com/ansible-community/ansible-lint/pull/1434 which uncovered this bug: mainly I am cloning some repos inside .cache and try to lint them. One of them does not have a linter config and it fails to run because without this change the linter loads the config from the upper level, which is not desired. I doubt anyone was so crazy to nest repos

aminvakil (approved):

PR #1445, opened 2021-03-08
new test for 'import_tasks'

europ:
simple test for #540


test/TestImportTasks.py
```diff
@@ -0,0 +1,73 @@
...
+}
+
+
+@pytest.mark.xfail
+def test_import_tasks(default_rules_collection):
+    """Calling 'import_task' from multiple files in different directories."""
+    root_dir_path = tempfile.mkdtemp()
+
+    for directory in LAYOUT['directories']:
+        os.makedirs(os.path.join(root_dir_path, directory), exist_ok=True)
```
webknjaz, line 64: Using `tempfile` leaves garbage on disk in case of failures. Please use the built-in `tmp_path` fixture that pytest provides instead: https://docs.pytest.org/en/stable/tmpdir.html#the-tmp-path-fixture.

Also, that fixture provides a `pathlib.Path` object so you could use a cleaner interface of `pathlib.Path.write_text()` to save files on disk.

test/TestImportTasks.py
```diff
@@ -0,0 +1,73 @@
...
+                  Something...
+            """
+        ),
+    },
+}
+
+
+@pytest.mark.xfail
+def test_import_tasks(default_rules_collection):
+    """Calling 'import_task' from multiple files in different directories."""
```
webknjaz, line 60: What does it check?

test/TestImportTasks.py
```diff
@@ -0,0 +1,73 @@
+"""Calling 'import_task' from multiple files in different directories."""
```
webknjaz, line 1: From this text, it's not clear what the suite checks for.

test/TestImportTasks.py
```diff
@@ -0,0 +1,73 @@
...
+import os
+import shutil
+import tempfile
```
webknjaz, line 4: These will be unnecessary when you switch to `tmp_path`.

webknjaz:
> simple test for #540

Please adjust the explanation and the title to mention what this test is about.

test/TestImportTasks.py
```diff
@@ -0,0 +1,73 @@
...
+              debug:
+                msg: |
+                  Something...
+            """
+        ),
+    },
+}
+
+
+@pytest.mark.xfail
```
webknjaz, line 58: Every xfail needs a reason which should at least link some clear explanation and/or have words explaining that in place.
```suggestion
@pytest.mark.xfail(reason='https://github.com/ansible-community/ansible-lint/issues/<a_dedicated_issue_explaining_this_specific_problem>')
```

test/TestImportTasks.py
```diff
@@ -0,0 +1,73 @@
...
+        os.makedirs(os.path.join(root_dir_path, directory), exist_ok=True)
+
+    for file_path, file_content in LAYOUT['files'].items():
+        open(os.path.join(root_dir_path, file_path), 'w').write(file_content)
+
+    result = Runner(root_dir_path, rules=default_rules_collection).run()
+
+    shutil.rmtree(root_dir_path)
+
+    assert len(result) == 0
```
webknjaz, line 73: Could you describe this expectation in words?

test/TestImportTasks.py
```diff
@@ -0,0 +1,73 @@
...
+    'directories': ['tasks', 'tasks/subtasks'],
+    'files': {
+        'main.yml': textwrap.dedent(
+            """\
+            ---
+            - hosts: target
+              gather_facts: false
+              tasks:
+                - name: from main import task 1
+                  import_tasks: tasks/task_1.yml
```
webknjaz, line 21: Since you mentioned in the issue that `include_tasks` is also affected, it'd be reasonable to use `@pytest.mark.parametrize` on the test to verify both scenarios.

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,128 @@
...
+    (
+        pytest.param(LAYOUT_IMPORTS, id='using only import_tasks'),
+        pytest.param(LAYOUT_INCLUDES, id='using only include_tasks'),
+    ),
+)
+@pytest.mark.xfail(
+    reason='https://github.com/ansible-community/ansible-lint/issues/1446'
+)
+def test_file_path_evaluation(default_rules_collection, ansible_project_layout):
+    """Test if the usage of import_tasks / include_tasks will end up with false positive load-failure due to incorrect file path evaluation."""
```
webknjaz, line 115: According to PEP 257, docstring titles should fit in 72 chars and any extra description could be written after an empty line.

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,128 @@
...
+    'ansible_project_layout',
+    (
+        pytest.param(LAYOUT_IMPORTS, id='using only import_tasks'),
+        pytest.param(LAYOUT_INCLUDES, id='using only include_tasks'),
+    ),
+)
+@pytest.mark.xfail(
+    reason='https://github.com/ansible-community/ansible-lint/issues/1446'
+)
+def test_file_path_evaluation(default_rules_collection, ansible_project_layout):
```
webknjaz, line 114: This got lost somewhere but please use `tmp_path` instead of managing tmp dirs manually.
```suggestion
def test_file_path_evaluation(default_rules_collection, ansible_project_layout, tmp_path):
```

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,128 @@
...
+
+    for directory in ansible_project_layout['directories']:
+        os.makedirs(os.path.join(root_dir_path, directory), exist_ok=True)
+
+    for file_path, file_content in ansible_project_layout['files'].items():
+        open(os.path.join(root_dir_path, file_path), 'w').write(file_content)
+
+    result = Runner(root_dir_path, rules=default_rules_collection).run()
+
+    shutil.rmtree(root_dir_path)
```
webknjaz, line 126: This will not get executed if something fails before this line. Using `tmp_path` OTOH will take care of the cleanup.

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,128 @@
...
+            - name: from subtask 2 do something
+              debug:
+                msg: |
+                  Something...
+            """
+        ),
+    },
+}
+
+
```
webknjaz, line 103: :point_down: These XPASS entries mean that the test passes under those envs while it's expected to always fail until the code is fixed. This means that the test may be unstable and does not unambiguously catch the bug.

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,128 @@
...
+        os.makedirs(os.path.join(root_dir_path, directory), exist_ok=True)
+
+    for file_path, file_content in ansible_project_layout['files'].items():
+        open(os.path.join(root_dir_path, file_path), 'w').write(file_content)
+
+    result = Runner(root_dir_path, rules=default_rules_collection).run()
+
+    shutil.rmtree(root_dir_path)
+
+    assert '[Errno 2] No such file or directory:' in result[0].message
```
webknjaz, line 128: Oh, this is what's causing XPASS.
```suggestion
    assert '[Errno 2] No such file or directory:' not in result[0].message
```
The test should check for the behavior that would be observed if there's no bug.

test/TestImportTasks.py
europ: Changed. Now it includes more descriptive definition. We are checking the file path evaluation because the use of import_tasks or include_tasks in different directories leads to an incorrectly evaluated file path, which is used by import_tasks or include_tasks.

test/TestImportTasks.py
europ: Result stores a list of `[Errno 2] No such file or directory: /path/to/file` messages because of the bug.

webknjaz: No, the bug is not expected. The expectation should be about the behavior w/o the bug.

test/TestFilePathEvaluation.py
europ: I have created special fixtures that use `tmp_path` above the test function.

webknjaz: I don't see the need to have extra fixtures. Also, fixtures are not supposed to be called directly but injected as test method args.

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,149 @@
...
+            """
+        ),
+    },
+}
+
+
+@pytest.fixture
+def create_root(tmp_path):
+    """Define root path for project."""
+    return str(tmp_path)
```
webknjaz, line 104: There's no need to convert everything into strings everywhere. It's wise versa — everything should be kept as pathlib objects and only converted in a specific place where this is needed (although, it's probably unnecessary at all).

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    root_dir_path = create_root()
+
+    for directory in ansible_project_layout['directories']:
+        create_directory(directory)
+
+    for file_path, file_content in ansible_project_layout['files'].items():
+        create_file(file_path, file_content)
+
+    result = Runner(root_dir_path, rules=default_rules_collection).run()
```
webknjaz, line 147: Here's how to make this cleaner:

```suggestion
def test_file_path_evaluation(default_rules_collection, ansible_project_layout, tmp_path):
    """Test file path evaluation when using import_tasks / include_tasks in the project.

    Usage of import_tasks / include_tasks may introduce false-positive load-failure due
    to incorrect file path evaluation.
    """
    for file_path, file_content in ansible_project_layout['files'].items():
        full_path = tmp_path / file_path
        full_path.parent.mkdir(parents=True, exist_ok=True)
        full_path.write_text(file_content)

    result = Runner(str(tmp_path), rules=default_rules_collection).run()
```
(it also allows to simplify the layout object structure)

webknjaz (approved):
I think it looks great now. But I'll leave it to @ssbarnea to merge this.

ssbarnea (approved):

test/TestFilePathEvaluation.py
```diff
@@ -0,0 +1,120 @@
...
+              Something...
+        """
+    ),
+}
+
+
+@pytest.mark.parametrize(
+    'ansible_project_layout',
+    (
+        pytest.param(LAYOUT_IMPORTS, id='using only import_tasks'),
```
ssbarnea, line 98: Usually I prefer test ids that do not have spaces or special chars, so I can use them with `-k` without trouble but I will likely fix it next time we need to touch the test, hopefully soon for removal of xfail.

webknjaz: FWIW spaced ids work with `-k` just fine. But I prefer concrete specifiers like `tox -- 'test/TestFilePathEvaluation.py::test_file_path_evaluation[using only include_tasks]'`

PR #1447, opened 2021-03-08
Undeprecate numeric rule tags

tadeboro:
The reason behind the undeprecation is fairly simple: right now, it is impossible to create a playbook that uses tags that will not trigger any warnings on ansible-lint 4 or 5.

If we use numeric tags, ansible-lint >= 5 will complain. If we use more descriptive tags, introduced in ansible-lint 5.0.0, ansible-lint
< 4 will complain about the errors we want to ignore.

This is especially important for collection authors that publish their work on Ansible Galaxy. On each import, Ansible Galaxy runs
ansible-lint 4 over all of the roles. Those developers are, if they want to have a clean import, forced to use an older version of ansible lint, or have ansible-lint >= 5 complain about numeric tags.

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -41,7 +41,7 @@ from ansiblelint.color import (
     reconfigure,
     render_yaml,
 )
-from ansiblelint.config import options, used_old_tags
+from ansiblelint.config import options
 from ansiblelint.constants import ANSIBLE_MISSING_RC, EXIT_CONTROL_C_RC
 from ansiblelint.file_utils import cwd
 from ansiblelint.skip_utils import normalize_tag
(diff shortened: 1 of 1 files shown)


felixfontein:
Delaying the deprecation until later in the 5.x release cycle (or even until 6, depending on how fast that happens) sounds like a good idea!

ssbarnea:
I am ok with it but first I want to see some feedback from the galaxy team members. I do not ever remember when it was last time when someone from the galaxy team made a pull request or at least gave feedback on irc or issue tracker. AFAIK there is no proof that any galaxy engineer is watching what we do. Even an ok or not from time to time would be ok.

In fact the version of the linter used by galaxy server is not even documented and we do not have any indication on what is needed to upgrade it.

Without this, I struggle to find motivation to add extra effort/hacks for unsupportive consumers.

src/ansiblelint/__main__.py
```diff
@@ -131,13 +131,20 @@ def report_outcome(result: "LintResult", options, mark_as_success=False) -> int:
...
     msg += "".join(sorted(entries))
 
-    for k, v in used_old_tags.items():
-        _logger.warning(
-            "Replaced deprecated tag '%s' with '%s' but it will become an "
-            "error in the future.",
-            k,
-            v,
-        )
+    # Do not deprecate the odl tags just yet. Why? Because it is not currently feasible
```
felixfontein, line 134:
```suggestion
    # Do not deprecate the old tags just yet. Why? Because it is not currently feasible
```

src/ansiblelint/__main__.py
```diff
@@ -131,13 +131,20 @@ def report_outcome(result: "LintResult", options, mark_as_success=False) -> int:
...
+    # to migrate old tags to new tags. There are a lot of things out there that still
+    # use ansible-lint 4 (for example, Ansible Galaxy and Automation Hub imports). If we
+    # replace the odl tags, those tools will report warnings. If we do not replace them,
```
aminvakil, line 137: @felixfontein You missed this:)
```suggestion
    # replace the old tags, those tools will report warnings. If we do not replace them,
```

ssbarnea:
@tadeboro Can you please rebase so i can merge it? Next time create a feature branch so we can edit it (you have this power).

ssbarnea:
Closed in favour of https://github.com/ansible-community/ansible-lint/pull/1452 as I cannot rebase this one.

PR #1450, opened 2021-03-10
Add optional rule that check for no-same-owner

ssbarnea:
Add a new rule that identifies use of the same owner when transferring files between hosts as a violation.

This rule is disabled by default and user needs to manually activate it inside the config as it is needed only for very particular use-cases.

Adding this rule to the linter should make it much easier for others to reuse it if they seem fit and ease upgrades of the linter, as the custom rule no longer needs to be ported to newer version of the linter.

Reference: https://zuul-ci.org/docs/zuul-jobs/policy.html#preservation-of-owner-between-executor-and-remote

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -28,6 +28,11 @@ skip_list:
   - skip_this_tag
   - git-latest
 
+# Any rule that has the 'opt-in' tag will not be loaded unless its 'id' is
+# mentioned in the enable_list:
+enable_list:
+  - no-same-owner
+
 # Report only a subset of tags and fully ignore any others
 # tags:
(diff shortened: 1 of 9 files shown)


greg-hellings (approved):
I like the addition of opt-in rules. Can we get it mentioned in the docs?

ssbarnea:
> I like the addition of opt-in rules. Can we get it mentioned in the docs?

We already do, because the .ansible-lint file from the root of the repository is included inside the docs at config, section. Basically we documented it by only mentioning it there, is sneaky but practical, look at https://ansible-lint.readthedocs.io/en/latest/configuring.html

Issue #1250, opened 2021-01-19
E306 shouldn't warn when the command never fails


pikrzysztof:
Why would I want pipefail to be set when:

A) the command can't fail because I said so
B) the failure condition is nontrivial, such as "the error message does not contain Y word"

?

Example code that E306 shouldn't warn about:

```
- shell: "pgrep stuff | wc -l" # noqa 306
  failed_when: false
  changed_when: false
  check_mode: false
  register: stuff_is_running
```

ssbarnea:
I guess that the presence of `failed_when: false` should disable this rule. Feel free to make a PR to add this, it does make sense.

PR #1453, opened 2021-03-10
Improve risky-shell-pipe rule matching

ssbarnea:
Removes one false positive case where user set `ignore_errors: true`, case in which the rule does not make much sense.

Fixes: #1250


greg-hellings:
Yes!!

Issue #1454, opened 2021-03-10
role_name check is not skipped when autodetecting standalone role name


geerlingguy:
##### Summary

5.0.3 is _sooo close_ to working with all my roles again, but I've run into this issue:

I have a role named `geerlingguy.apache-php-fpm`. It is on:

  - Galaxy: https://galaxy.ansible.com/geerlingguy/apache-php-fpm
  - GitHub: https://github.com/geerlingguy/ansible-role-apache-php-fpm

When I run ansible-lint I get:

```
Computed fully qualified role name of geerlingguy.apache-php-fpm is not valid.
Please edit meta/main.yml and assure we can correctly determine full role name:

galaxy_info:
  role_name: my_name  # if absent directory name hosting role is used instead
  namespace: my_galaxy_namespace  # if absent, author is used instead

Namespace: https://galaxy.ansible.com/docs/contributing/namespaces.html#galaxy-namespace-limitations
Role: https://galaxy.ansible.com/docs/contributing/creating_role.html#role-names

Error: Process completed with exit code 10.
```

I even have, inside my `.ansible-lint` file:

```yaml
skip_list:
  - 'role-name'
```

But it seems like the computed role name is run against the `role-name` linter _before Ansible Lint even starts linting the role_`. I think, instead, that the role-name lint should not be run as part of setup at all, since roles can have dashes (especially roles that are not on galaxy itself—I hate Ansible's war on dashes of late, and use them extensively in roles, group names, etc.).

See failed run: https://github.com/geerlingguy/ansible-role-apache-php-fpm/runs/2080256861?check_suite_focus=true

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible 2.10.5
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/Users/jgeerling/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.2 (default, Feb 24 2021, 13:26:09) [Clang 12.0.0 (clang-1200.0.32.29)]

ansible-lint 5.0.3 using ansible 2.10.5
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

macOS, Ubuntu (GitHub Actions CI)

##### STEPS TO REPRODUCE

[...]

PR #1459, opened 2021-03-11
Avoid pre-run errors when role-name is disabled

ssbarnea:
- when `role-name` is added to skip_list, just install simple role name
- when `role-name` is added to warn_list, display warning and continue execution.
- `role-name` rule is now moved to default warn_list, so users that do not have custom warn list, will only see its reported violations but the linter will report success.

Fixes: #1454


geerlingguy (approved):
LGTM.

awcrosby (approved):
looks like what was discussed.

PR #1470, opened 2021-03-17
Fixed matching patterns used for mocking

ssbarnea:


felixfontein:
There's another one in https://github.com/ansible-community/ansible-lint/blob/ee28104a47b9672579682693beae12464f338e82/src/ansiblelint/_prerun.py#L229

src/ansiblelint/_prerun.py
```diff
@@ -226,9 +226,14 @@ def _prepare_ansible_paths() -> None:
 
 def _make_module_stub(module_name: str) -> None:
     # a.b.c is treated a collection
-    if re.match(r"\w+\.\w+\.\w+", module_name):
-        namespace, collection, module_file = module_name.split(".")
+    if re.match(r"\w+\.\w+\.[\.\w]+$", module_name):
```
felixfontein, line 229: Shouldn't this be
```suggestion
    if re.match(r"\w+\.\w+(\.\w+)+$", module_name):
```
if you want to match proper names?

.ansible-lint
```diff
@@ -8,8 +8,9 @@ exclude_paths:
 # Mock modules or roles in order to pass ansible-playbook --syntax-check
 mock_modules:
   - zuul_return
-  - fake_namespace.fake_collection.fake_module
   # note the foo.bar is invalid as being neither a module or a collection
+  - fake_namespace.fake_collection.fake_module
+  - fake_namespace.fake_collection.fake_module.fake_submodule
```
felixfontein, line 13:
```suggestion
  - fake_namespace.fake_collection.fake_module
```
I would really avoid advertising that you can have further dots in module names ;-)

src/ansiblelint/_prerun.py
```diff
@@ -226,25 +226,28 @@ def _prepare_ansible_paths() -> None:
 
 def _make_module_stub(module_name: str) -> None:
     # a.b.c is treated a collection
-    if re.match(r"\w+\.\w+\.\w+", module_name):
-        namespace, collection, module_file = module_name.split(".")
-        path = f"{ options.project_dir }/.cache/collections/ansible_collections/{ namespace }/{ collection }/plugins/modules"
+    if re.match(r"^(\w+|\w+\.\w+\.[\.\w]+)$", module_name):
```
felixfontein, line 229:
```suggestion
    if re.match(r"^(\w+|\w+\.\w+(\.\w+)+)$", module_name):
```

ssbarnea: Based on my tests done on https://regex101.com/ mine example worked as intended but the one with nested matches failed.

.ansible-lint
ssbarnea: Is not so much about advertising desired but listing it here assures that we do not choke when we encounter it (there it at least one test making use of it).

src/ansiblelint/_prerun.py
felixfontein: What exactly did fail? I tried several variants, and they all worked.

.ansible-lint
felixfontein: In that case, at least use correct terms: `fake_namespace.fake_collection.fake_directory.fake_module`

PR #1481, opened 2021-03-19
[RFC] src/ansiblelint/_prerun.py: Add extra link to current directory in cache

apatard:
ansible-lint is now checking playbook with --syntax-check, so it needs
to have the current role in cache, otherwise it will fail with error like:

internal-error: the role 'foo' was not found ...

There has been a previous attempt to fix this by creating a link to
current role base directory and the link is named according to galaxy
information. This name has several issues:
- it's specific to galaxy. People not using galaxy doesn't really need
  to respect the galaxy usage. For instance, if they follow
  https://galaxy.ansible.com/docs/contributing/creating_role.html#role-metadata
  they will put 'John Doe' in 'author' so the link will be useless
- it's uselessly breaking something which has been working before. The converge.yml
  molecule playbook is often containing only the role name (without any kind of
  namespace) and everything was fine. With the link name, people
  will have to edit every role having the issue and possibly their playbooks/
  requirements/... files too.

As a solution, this patch is also adding a link to the current role
directory name after the role name found in meta/main.yml or the
current directory name.

### DIFF
--- a/src/ansiblelint/_prerun.py
+++ b/src/ansiblelint/_prerun.py
@@ -193,15 +193,16 @@ def _install_galaxy_role() -> None:
         fqrn = role_name
     p = pathlib.Path(f"{options.project_dir}/.cache/roles")
     p.mkdir(parents=True, exist_ok=True)
-    link_path = p / f"{role_namespace}.{role_name}"
-    # despite documentation stating that is_file() reports true for symlinks,
-    # it appears that is_dir() reports true instead, so we rely on exits().
-    if not link_path.exists():
-        link_path.symlink_to(pathlib.Path("../..", target_is_directory=True))
-    _logger.info(
-        "Using %s symlink to current repository in order to enable Ansible to find the role using its expected full name.",
-        link_path,
-    )
+    for rolepath in [f"{role_namespace}.{role_name}", f"{role_name}"]:
+        link_path = p / rolepath
+        # despite documentation stating that is_file() reports true for symlinks,
+        # it appears that is_dir() reports true instead, so we rely on exits().
+        if not link_path.exists():
+            link_path.symlink_to(pathlib.Path("../..", target_is_directory=True))
+        _logger.info(
+            "Using %s symlink to current repository in order to enable Ansible to find the role using its expected full name.",
+            link_path,
+        )
 
 
 def _prepare_ansible_paths() -> None:


src/ansiblelint/_prerun.py
```diff
@@ -193,15 +193,16 @@ def _install_galaxy_role() -> None:
...
+    for rolepath in [f"{role_namespace}.{role_name}", f"{role_name}"]:
+        link_path = p / rolepath
+        # despite documentation stating that is_file() reports true for symlinks,
+        # it appears that is_dir() reports true instead, so we rely on exits().
+        if not link_path.exists():
+            link_path.symlink_to(pathlib.Path("../..", target_is_directory=True))
+        _logger.info(
+            "Using %s symlink to current repository in order to enable Ansible to find the role using its expected full name.",
+            link_path,
+        )
```
ssbarnea, line 205: I am against a model where we create any kind of variations for role names just to avoid a failure from ansible, mainly because we could hide real issues or at least bad practices.

That is not about "galaxy", i about ansible content "packaging" and how content is suppose to be installed and reused.

On the other hand, I do agree that it would be wise to allow users to slowly adapt their content at their own speed (adapting can take even years for some).

I will wait for some feedback from others in order to see which approach we should take here, no need to worry as we will find a workaround.

@felixfontein @tadeboro What do you think?

IMHO creating both is clearly mistake, I am inclined to support a single one with a fallback mechanism, likely with a warning when the fallback happens, or even an error and a configuration option that could enable the compatibility mode.

apatard: Even if I'm going to fix my few roles not on galaxy to make thing just work, I'm not sure about the bad practice thing for roles not relying on galaxy. People just clone the role directory and things are working without issues.

But I agree that having both links is not nice but I wanted to see a quick fix for that. When I see commits like :
- https://github.com/sio/homelab/commit/fa3ad24a68ae65dad43d27ffd9d7d86245c1f639 (from https://github.com/ansible-community/ansible-lint/issues/1329#issuecomment-781175799) and 
- https://github.com/geerlingguy/ansible-role-mysql/commit/86ffba892e1553b7e6f042ec007eef25daa8f5db#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

Something has to be done with such kind of "regressions" or people will stop using ansible-lint....

apatard: To add one more point: I was testing things to check what would happen by using (for instance) ``rtp.foo`` for a role name, ansible-lint will be happy *but* molecule syntax step will complain now that it can't find the role, since the CI will run git clone and the clone name will be named after the git repo name (for instance) ``foo`` and not ``rtp.foo``. So the idea is to ask people to rename all their git repositories too ?

ssbarnea: Molecule does not do yet any pre-runs in order to help ansible find it and I would like to avoid copy/paste the same logic from the linter.

Ideally we should have a shared library (packagE) that we can use to share stuff like this but I am worried about extra maintenance pains. Alternative would be to add ansible-lint as a dependency of molecule but without installing ansible, effectively using only the pre-runs steps from the linter. I guess you are aware of https://github.com/ansible-community/molecule/pull/2998 which is already 4mo old.

To answer your question: no, I don't want to ask people to rename repos. I hope to get some good feedback today and be able to address this issue over the weekend, for both tools, in the end both of them have a common requirement of needing a valid ansible runtime environment (one that aims to match how the code is supposed to run in production aka 'installed').

ssbarnea:
Closing for the moment but we will reopen the subject based on the feedback we will receive. I doubt we will ever want to create two symlinks. If we are to do it we will likely add an option for determining creation of FQRN or simple names, but not both.

I am saying this because I do not expect a role to be consumed in two variants when deployed and I do want the linter to be able to spot any playbooks that wrongly use the unsupported way.

In the end, we can easily mock roles and avoid errors from ansible, but that does not give any confidence that this code will run in production.

PR #1490, opened 2021-03-23
Fix role name handling in prerun.py

apatard:
This PR tries to address issues with the link made in ``.cache/roles`` for current role name.
The first patch tries to make ``_install_galaxy_role()`` behave in same way as ``ansible-galaxy`` and
the second one is fixing the link name to use the fqrn computed earlier in the function. This will
provide a way to allow using the role short name by putting the 'role-name' check in
``skip_list``.

### DIFF
--- a/playbooks/eco.yml
+++ b/playbooks/eco.yml
@@ -5,7 +5,7 @@
   vars:
     cache_dir: "{{ (lookup('env','HOME'), '.cache', 'ansible-lint-eco') | path_join | realpath }}"
     repos:
-      - name: ansible-role-mysql
+      - name: geerlingguy.mysql
         url: https://github.com/geerlingguy/ansible-role-mysql.git
         contact: geerlingguy
       - name: zuul-jobs
--- a/playbooks/eco.yml
+++ b/playbooks/eco.yml
@@ -5,9 +5,14 @@
   vars:
     cache_dir: "{{ (lookup('env','HOME'), '.cache', 'ansible-lint-eco') | path_join | realpath }}"
     repos:
+      - name: ansible-role-mysql
+        url: https://github.com/geerlingguy/ansible-role-mysql.git
+        contact: geerlingguy
       - name: geerlingguy.mysql
         url: https://github.com/geerlingguy/ansible-role-mysql.git
         contact: geerlingguy
+        # duplicates first one but covers for the case where clone is made in a
+        # folder that uses namespace.rolename pattern (installed galaxy role)
       - name: zuul-jobs
         url: https://opendev.org/zuul/zuul-jobs
         contact: ssbarnea
--- a/src/ansiblelint/prerun.py
+++ b/src/ansiblelint/prerun.py
@@ -6,7 +6,7 @@ import re
 import subprocess
 import sys
 from functools import lru_cache
-from typing import List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from packaging import version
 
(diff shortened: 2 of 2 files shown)


playbooks/eco.yml
```diff
@@ -5,7 +5,7 @@
   vars:
     cache_dir: "{{ (lookup('env','HOME'), '.cache', 'ansible-lint-eco') | path_join | realpath }}"
     repos:
-      - name: ansible-role-mysql
+      - name: geerlingguy.mysql
```
ssbarnea, line 8: Testing that we can lint role folders that use the galaxy naming when installed is a good idea but removing the old test  about `ansible-role-mysql` format is also a very bad idea, as we could easily break it. As this role is small, I do not see any problem in cloning it twice with both names, so we assure both cases are covered.

src/ansiblelint/prerun.py
```diff
@@ -202,7 +202,7 @@ def _install_galaxy_role() -> None:
         fqrn = role_name
     p = pathlib.Path(f"{options.project_dir}/.cache/roles")
     p.mkdir(parents=True, exist_ok=True)
-    link_path = p / f"{role_namespace}.{role_name}"
+    link_path = p / fqrn
```
ssbarnea, line 205: Indeed that is a real bug fix!

src/ansiblelint/prerun.py
```diff
@@ -169,7 +169,7 @@ def _install_galaxy_role() -> None:
     role_namespace = yaml['galaxy_info'].get('namespace', None)
     if not role_namespace:
         role_namespace = yaml['galaxy_info'].get('author', None)
-    if not role_name:
+    if not role_name or os.path.exists(".git") or os.path.exists(".hg"):
```
ssbarnea, line 172: I am afraid that this may not be correct and it also seems unrelated, better to remove it from this patch. We can address it separately if needed.

ssbarnea (changes requested):

playbooks/eco.yml
apatard: the problem is that keeping the current configuration will lead to CI failure as the ``.ansible-lint`` file of this role is configured with ``role-name`` in  ``skip_list``.

src/ansiblelint/prerun.py
apatard: tbh, I've wrote this patch as I found the comment while trying to understand how things were supposed to work. I've not seen any bug about this so I've no problem with dropping this change.

ssbarnea (approved):

ssbarnea:
@geerlingguy Please review this change because it breaks one of your roles. That is because you added, for good reasons, `role-name` to the skip-list. Stil, the proposed logic is to use compatibility mode (short names) when role-name is excluded and full role names when it is not.

As you can see, with this change we get an error on molecule playbooks as they are unable to find the full qualified role name. Obviously that removing the role-name from the skip list would allow the linter to pass.

Still, I do really want your opinion on this matter as I want to minimize the chance of alienating more users. If you are not happy about it, do you have any proposed behaviors in mind?

src/ansiblelint/prerun.py
```diff
@@ -158,23 +158,47 @@ def prepare_environment() -> None:
...
+    # if there's a space in the name space, it's likely author name
+    # and not the galaxy login, so act as if there was no namespace
+    if re.match(r"^\w+ \w+", role_namespace):
+        role_namespace = ""
+    else:
+        role_namespace = f"{role_namespace}."
+    return role_namespace
+
+
+def _get_galaxy_role_name(galaxy_infos: Dict[str, Any]) -> str:
```
ssbarnea, line 175: I am surprised this passes mypy because the correct return type is `Optional[str]` and not str based on the fact I see the default of `None`.

src/ansiblelint/prerun.py
```diff
@@ -158,23 +158,47 @@ def prepare_environment() -> None:
...
 
+def _get_galaxy_role_ns(galaxy_infos: Dict[str, Any]) -> str:
+    """Compute role namespace from meta/main.yml, including trailing dot."""
+    role_namespace = galaxy_infos.get('namespace', None)
+    if not role_namespace:
+        role_namespace = galaxy_infos.get('author', None)
+    # if there's a space in the name space, it's likely author name
+    # and not the galaxy login, so act as if there was no namespace
+    if re.match(r"^\w+ \w+", role_namespace):
+        role_namespace = ""
```
ssbarnea, line 169: This is also going to raise an error: running a match on a None string. Maybe use a default of empty string would be better for all these fields.

apatard: ok. Fixed things to use and empty string and made these two functions to always return a possibly empty string. Will push an update soon

geerlingguy:
@ssbarnea - TBH, for now I have disabled ansible-lint on all my roles because it was just filling up my inbox every day with another 20+ 'CI failure' notifications and there was an awful lot of work I had to do to each role just to get it to pass lint again (trying to configure ansible-lint so it would skip cache directory—I think that's since been fixed), adding in a list of dependencies in yet another location (the 'mock role' list or something), etc.

In the past, I could just install ansible-lint and it would work, and lint my role. Today it is a lot harder since 5.x came out. I am still using ansible-lint on my playbook repositories, where it seems like very little was changed, but for my Galaxy roles I'm not sure when I'll get the time to work on re-adding configuration for all the roles to run with the linter again.

apatard:
> @ssbarnea - TBH, for now I have disabled ansible-lint on all my roles because it was just filling up my inbox every day with another 20+ 'CI failure' notifications and there was an awful lot of work I had to do to each role just to get it to pass lint again (trying to configure ansible-lint so it would skip cache directory—I think that's since been fixed), adding in a list of dependencies in yet another location (the 'mock role' list or something), etc.

The cache thing should be fixed iirc. For the dependency location thing, it's a different issue and if I find time I'll probably look at that too as I got bitten by it too, but in a different PR. No idea what @ssbarnea will think of that.

> 
> In the past, I could just install ansible-lint and it would work, and lint my role. Today it is a lot harder since 5.x came out. I am still using ansible-lint on my playbook repositories, where it seems like very little was changed, but for my Galaxy roles I'm not sure when I'll get the time to work on re-adding configuration for all the roles to run with the linter again.

With this PR, I hope that for roles without dependencies, things should just work again, well except for roles with wrong fqrn and in this case, either fixing or ``.ansible-lint`` is the solution. With my latest changes, ansible-lint is happy with your role but still having feedback from you on how it should work is interesting imho.

src/ansiblelint/prerun.py
apatard: hm. forgot about dropping this. doing now

PR #1505, opened 2021-04-01
Use types for pre-commit

jtcarnes:
updates to pre-commit have allowed you to specify types of files in the hook itself instead of having to have the user specify a files regex (documentation [here](https://pre-commit.com/index.html#filtering-files-with-types)).

By adding `types: [yaml]` your user does not need to put `files: \.(yaml|yml)$` , which makes using it much easier. This is a backwards compatible change as types: [yaml] and `files: \.(yaml|yml)$` will match all the same files.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -7,6 +7,7 @@
     name: Ansible-lint
     description: This hook runs ansible-lint.
     entry: ansible-lint --force-color
+    types: [yaml]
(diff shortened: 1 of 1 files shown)


ssbarnea:
Not including a file type was not a mistake, it was on purpose. There are lots of other files that can affect the result of ansible-lint run and we do not want to miss them. For example soon we will also add tests for `md` files used by docs.

PR #1513, opened 2021-04-06
New Rule - Variable Naming Standards

ansiblejunky:
Add new rule that checks for Variable Naming Standards in the following situations where variables can be declared.

- 'vars' section in the playbook
- 'vars' section in the task
- 'register:' section in the task
- set_fact tasks
- vars files (yaml)

Tested against some local content I had, below are the results of a test:

```
 $ ansible-lint
WARNING  Listing 24 violation(s) that are fatal
my_namespace/my_collection/playbooks/play.yml:4: var-naming Play defines variable 'Another' within 'vars' section that violates variable naming standards
my_namespace/my_collection/playbooks/play.yml:4: var-naming Play defines variable 'Mother' within 'vars' section that violates variable naming standards
my_namespace/my_collection/playbooks/play.yml:4: var-naming Play defines variable 'TEST1' within 'vars' section that violates variable naming standards
my_namespace/my_collection/playbooks/play.yml:4: var-naming Play defines variable 'TEST2' within 'vars' section that violates variable naming standards
my_namespace/my_collection/playbooks/play.yml:4: var-naming Play defines variable 'That' within 'vars' section that violates variable naming standards
my_namespace/my_collection/playbooks/play.yml:4: var-naming Play defines variable 'ThatKey1' within 'vars' section that violates variable naming standards
my_namespace/my_collection/playbooks/play.yml:4: var-naming Play defines variable 'ThoseKeys' within 'vars' section that violates variable naming standards
my_namespace/my_collection/playbooks/play.yml:16: unnamed-task All tasks should be named
my_namespace/my_collection/playbooks/play.yml:16: var-naming Task registers a variable that violates variable naming standards
my_namespace/my_collection/roles/myrole/defaults/main.yml:0: var-naming File defines variable 'DefaultAnotherBadVarName' that violates variable naming standards
my_namespace/my_collection/roles/myrole/defaults/main.yml:0: var-naming File defines variable 'DefaultBadVariableName' that violates variable naming standards
my_namespace/my_collection/roles/myrole/handlers/main.yml:4: var-naming Task defines variables within 'vars' section that violates variable naming standards
my_namespace/my_collection/roles/myrole/handlers/main.yml:7: var-naming Play defines variable 'MyBadVariableName' within 'vars' section that violates variable naming standards
my_namespace/my_collection/roles/myrole/handlers/main.yml:10: var-naming Task uses 'set_fact' to define variables that violates variable naming standards
[...]

PR #1518, opened 2021-04-07
Add rule for variable naming

ssbarnea:
Originally proposed by @ansiblejunky in #1513

As any new rule, that is initially introduced as experimental, which makes it only raise warnings and not errors.

### DIFF
--- a/src/ansiblelint/rules/VariableNamingRule.py
+++ b/src/ansiblelint/rules/VariableNamingRule.py
@@ -1,0 +1,151 @@
+import keyword
+import re
+import sys
+from functools import lru_cache
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Pattern, Union
+
+from ansiblelint.config import options
+from ansiblelint.file_utils import Lintable
+from ansiblelint.rules import AnsibleLintRule
+from ansiblelint.utils import parse_yaml_from_file
+
+if TYPE_CHECKING:
+    from ansiblelint.constants import odict
+    from ansiblelint.errors import MatchError
+
+
+FAIL_PLAY = """
+- hosts: localhost
+  vars:
+    CamelCaseIsBad: false  # invalid
+    this_is_valid:  # valid because content is a dict, not a variable
+      CamelCase: ...
+      ALL_CAPS: ...
+    ALL_CAPS_ARE_BAD_TOO: ...  # invalid
+"""
+
+
+# properties/parameters are prefixed and postfixed with `__`
+def is_property(k: str) -> bool:
+    """Check if key is a property."""
+    return k.startswith('__') and k.endswith('__')
+
+
+class VariableNamingRule(AnsibleLintRule):
+    id = 'var-naming'
+    base_msg = 'All variables should be named using only lowercase and underscores'
+    shortdesc = base_msg
+    description = 'All variables should be named using only lowercase and underscores'
+    severity = (
+        'MEDIUM'  # ansible-lint displays severity when with --parseable-severity option
+    )
+    tags = ['idiom', 'experimental']
+    version_added = 'v5.0.10'
+
+    @lru_cache()
+    def re_pattern(self) -> Pattern[str]:
+        return re.compile(options.var_naming_pattern)
+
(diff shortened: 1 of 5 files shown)


src/ansiblelint/rules/VariableNamingRule.py
```diff
@@ -0,0 +1,117 @@
...
+
+# properties/parameters are prefixed and postfixed with `__`
+def is_property(k: str) -> bool:
+    """Check if key is a property."""
+    return k.startswith('__') and k.endswith('__')
+
+
+def is_invalid_variable_name(text: str) -> bool:
+    """Check if variable name is using right pattern."""
+    patterns = '^[a-z0-9_]*$'
```
MarkusTeufelberger, line 21: This also allows numbers, not only lowercase ASCII letters from a-z and underscores.
It also allows a variable that starts with a number which afair is invalid.

Would it be possible to use Ansible's own functions for this check?

MarkusTeufelberger: e.g. something like `from ansible.utils.vars import isidentifier`

ssbarnea:
@ansiblejunky @MarkusTeufelberger I am unable to find any variable name that passed syntax-check which means this rule does not make any sense, as any invalid variable name would be observed by the syntax check.

Unless someone can come up with at least one value that does not produce syntax-check failure, I will have to drop the idea this PR.

ansiblejunky:
@ssbarnea I'm not sure I understand completely. The ansible syntax checker will detect "invalid" variable names, but this naming standards rule is more intended for detecting variables that do not comply with "good standards".  So while the ansible syntax checker is good, it's not good enough for ensuring good practices with regards to naming your variables. For example, the variable "MyVariable" is acceptable by the ansible syntax checker, however this is a poorly named variable and would/should fail from this lint rule.  It should be "my_variable" or at least all in lowercase. This is just one example. The rule detects this problem in various areas where customers/users often make "lazy" variable names, such as registering variables and set_facts and so on.

I hope this helps.

ssbarnea:
@ansiblejunky AFAIK, `MyVariable` would pass in its current form. Do you want to propose a rule that checks for the `^[a-z0-9_]*$` pattern and allows user to customize the pattern in config file?

I know that a previous revision had that regex, but current code checks only for what Ansible allows nothing else.

I am not against a lowercase pattern, but I am afraid of how much resent it could create among some users. I bet some uses ALL_CAPS in some places. Still, if it passed our eco pipeline I would probably not block it, especially as user could disable or neuter the rule.

ansiblejunky:
@ssbarnea I didn't know the original regex has changed. That makes sense as to why it appears to no longer bring any value-add. My original code and intentions were to have this as an optional rule that people can use to apply this... and if possible, yes, to customize the regex would be slick.

ansiblejunky:
I suppose the question is, whether ansible-lint checks good practices, or whether it checks what is "valid" by ansible documentation. If the former, then this rule is useful. If the latter, then we have to revert to only what the syntax checker looks for.  I know that in various customer implementations this custom lint rule has helped keep YAML code clean and readable, so it's brought a lot of value. However, whether it should be added to ansible-lint repo is the question. 

Perhaps if the rule defaults to the regex that checks only what the syntax checker looks for, and then we provide an example regex to additionally check for mixed case?

MarkusTeufelberger:
Is it possible to hand a regex pattern to a rule and have that enforced instead? Then the current one (`r"^[a-z_][a-z0-9_]*$"`) would be the default one but someone who is really into CAPITAL_LETTERS can use their own instead.

ssbarnea:
> Is it possible to hand a regex pattern to a rule and have that enforced instead? Then the current one (`r"^[a-z_][a-z0-9_]*$"`) would be the default one but someone who is really into CAPITAL_LETTERS can use their own instead.

There is not need to "pass" the custom pattern, that is loaded from .ansible-lint config file when found. Shortly, it already allows users to define their own patterns instead of disabling the rule.

ssbarnea:
BTW, i removed the Codacy, too much false-positives for this project, it appears to be clueless about what we configured within the project.

PR #1520, opened 2021-04-07
Mention pipx as alternative installer

ssbarnea:


webknjaz (commented):
I don't have any strong opinion on this. I'm a bit opposed to recommending third-party tools as you may end up having to deal with people not knowing how to best set up pipx in their systems.

docs/installing.rst
```diff
@@ -25,32 +25,36 @@ related to containers and use the discussions_ forum instead.
...
 
-    * ``core`` - will install latest version of ansible-base 2.10
-    * ``community`` - will install latest version of ansible 2.10 with community collections
+    * ``core`` - will install the latest version of ansible-base 2.10
+    * ``community`` - will install the latest version of ansible 2.10 with community collections
 
 Using Pip
 ---------
 
+While you can use `pip3` to install it, we do recommend using pipx_ as this
```
webknjaz, line 38: Single quotes generate an `any` role that tries to link to any Sphinx object with this identifier. So either add pip to intersphinx and use a proper role from https://webknjaz.github.io/intersphinx-untangled/pip.pypa.io/ or use double backticks for the inline code role.

ssbarnea:
> I don't have any strong opinion on this. I'm a bit opposed to recommending third-party tools as you may end up having to deal with people not knowing how to best set up pipx in their systems.

My guess is that using pip correctly is much harder than using pipx. Once they install pipx, I doubt they can get it to misbehave, like we often see with pip (--user, sudo, ....).

Just to have an idea, just yesterday I had one user complaining that installing using pip was failing under centos-7... yeah pip 9.0.3 from py27, something already not supported.

tadeboro (commented):
If the external tool does not provide an order of magnitude better user experience, then I am against recommending it. When it comes to Python apps, I tend to instruct users to either install them using the distro's package manager or create a dedicated virtual environment (example: https://github.com/sensu/sensu-go-python#installation). Anything else is just asking for troubles in my experience.

In this case, I would say `pipx` is not worth it and that we should stick to standard tools.

ssbarnea:
> If the external tool does not provide an order of magnitude better user experience, then I am against recommending it. When it comes to Python apps, I tend to instruct users to either install them using the distro's package manager or create a dedicated virtual environment (example: https://github.com/sensu/sensu-go-python#installation). Anything else is just asking for troubles in my experience.
> 
> In this case, I would say `pipx` is not worth it and that we should stick to standard tools.

I will wait for additional feedback, so I can see if there is a trend or not. Based on current one I am inclined to rewrite the recommendation differently, listing pip3 in example and only mentioning pipx as an alternative.

greg-hellings (approved):

docs/installing.rst
```diff
@@ -25,32 +25,40 @@ related to containers and use the discussions_ forum instead.
...
-    * ``community`` - will install latest version of ansible 2.10 with community collections
+    * ``core`` - will install the latest version of ansible-base 2.10
+    * ``community`` - will install the latest version of ansible 2.10 with community collections
 
-Using Pip
----------
+Using pip or pipx
+-----------------
+
+You can use either pip3_ or pipx_ to install it, the former one
```
webknjaz, line 38: https://www.grammarly.com/blog/former-vs-latter/

```suggestion
You can use either pip3_ or pipx_ to install it, the latter one
```

webknjaz: @ssbarnea ^

PR #1521, opened 2021-04-07
Avoid duplicating molecule requirements.yml files

apatard:
When running ansible-lint on roles with molecule scenarii, it will
run syntax-check on the scenarii but won't find the needed roles
(if any). The current solution is to write a requirements.yml file
in the top of the directory merging all molecule/*/requirements.yml files
but it's duplication and create an extra file only to get ansible-lint
playbook check working.
As a solution, look for molecule/*/requirements.yml and install
dependencies from them too.


ssbarnea (approved):
It looks ok to me but for this one I want to see more reviews, especially from molecule users.

greg-hellings (approved):

PR #1530, opened 2021-04-13
Utilize yamllint severity levels for rule severity

FloSchwalm:
yamllint allows to categorize found problems as "error" or "warning"  
ansible-lint treats them all as having `VERY_LOW` severity anyway

I think that the rule categorization of yamllint shouldn't be lost, after all it can be configured with a `.yamllint` config file to fit the user's needs.

This PR keeps the `VERY_LOW` severity categorization for problems that yamllint reports as "warning" just as before, but uses `MEDIUM` severity for problems reported as "error".  
Syntax errors reported by yamllint are raised to `VERY_HIGH` severity, though they won't be used as long as ansible-lint skips further rule checking after its own syntax check.

### DIFF
--- a/src/ansiblelint/rules/YamllintRule.py
+++ b/src/ansiblelint/rules/YamllintRule.py
@@ -83,6 +83,11 @@ class YamllintRule(AnsibleLintRule):
             for p in run_yamllint(
                 file.content, YamllintRule.config, filepath=file.path
             ):
+                self.severity = 'VERY_LOW'
+                if p.level == "error":
+                    self.severity = 'MEDIUM'
+                if p.desc.endswith("(syntax)"):
+                    self.severity = 'VERY_HIGH'
                 matches.append(
                     self.create_matcherror(
                         message=p.desc,


ssbarnea (approved):

ssbarnea (approved):
Ideally we should have included some tests.

FloSchwalm:
> Ideally we should have included some tests.

I guess I can try to add some. I'm no python developer, but I think I can figure it out from the existing tests.  
What kind of test were you thinking of? Hardcoded ansible-lint output lines which are then tested if categorized correctly? Or would you rather include a test playbook and actually run ansible-lint in the test?

ssbarnea:
Create one extra file inside examples/playbooks that would raise all 3 types of yamllint severity issues when you lint it. After this you can add one extra test that verifies that the output is the expected one.

lookat `def test_example_custom_module` and copy it to `test_example_yamllint_severity`, test for number of failures reported and their content (especially severity levels). Print the result to see what it contains, or add a `breakpoint()` and run `pytest -k test_example_yamllint_severity` once you add it. Once inside the debugger you will be able to inspect what result contains. 

To save your time: result is a `List[MatchError]`, so you will know how to assert each entry has the expected severity level.

PR #1531, opened 2021-04-14
Update installing.rst

Firefishy:
##### SUMMARY
Add instructions to install using homebrew

##### ISSUE TYPE
- Docs Pull Request

+label: docsite_pr


Firefishy:
The failing CI test appears to be because of a missing PR label, but appears I do not have required permissions to add a label.

ssbarnea (changes requested):
While I am using homebrew daily and like it I am inclined to discourage using it for ansible-lint itself, mainly because it will always be behind version from pip, likely a lot behind.

I am more inclined to suggest something like `brew install pipx`  for those that do not hav e pipx and use pipx to install it.

I will wait for other macos users to comment on this proposal before making a decision, as I want to avoid being biased.

PR #1534, opened 2021-04-18
Allow configurations for each rule stored in ansiblelint.config.options

ssato:
Add ansiblelint.config.options.rules: Dict[str, Any] to store configurations for rules and allow utilizing them from each rule to customize these behaviour more flexibly. 

Because it seems that ansiblelint.cli.{load,merge}_config is able to load and set options.rules.*, users will be able to customize each rules by those configurations automatically like the following only with this change, I think:

```
# .ansible-lint
  ... (snip) ...

rules:
  <rule_id>:
     <rule_config_key_a>: <rule_config_value_a>
     ... (snip) ...
```

Along with the enhancement, I update two rules, RoleNames and YamllintRule, to fix pylint "super-init-not-called" errors.

### DIFF
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -5,7 +5,7 @@ import subprocess
 import sys
 from argparse import Namespace
 from functools import lru_cache
-from typing import Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from packaging.version import Version
 
@@ -84,6 +84,7 @@ options = Namespace(
     extra_vars=None,
     enable_list=[],
     skip_action_validation=True,
+    rules=dict(),  # Placeholder to set and keep configurations for each rule.
 )
 
 # Used to store detected tag deprecations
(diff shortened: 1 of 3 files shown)


ssbarnea:
@ssato Can you please mention one case where this would be needed? So far I used only global options for rules but I so realize that this may be a problem for a custom rule, especially if it needs something that cannot be implemented using magic tags.

ssato:
@ssbarnea Thanks a lot for your comment!

> @ssato Can you please mention one case where this would be needed? So far I used only global options for rules but I so realize that this may be a problem for a custom rule, especially if it needs something that cannot be implemented using magic tags.

Please take a look at such cases of custom rules implemented experimentally need specific options for them below.

- https://github.com/ssato/ansible-lint-custom-rules/blob/per-rule-configs/rules/DebugRule.py#L32 - A custom rule for debug has a boolean option to enable it's check.
- https://github.com/ssato/ansible-lint-custom-rules/blob/per-rule-configs/rules/BlockedModules.py#L30 - A custom rule to restrict ansible modules to be used, has a list option to control which modules to be blocked to use.

Similar thing is already possible if custom rule can load and parse configurations for them by themselves, however, IMHO, it's better to implement a standard way to load and get configurations for each rule in rules.AnsibleLintRule to reduce risks caused by  bugs in custom rules having such code at a minimum.

PR #1539, opened 2021-04-29
Check for missing bool cast after default(<bool>)

Gui13:
This filter checks that any jinja or expression in the code that uses a boolean default is correctly postpended by "| bool".

This is an issue because any non "boolean" value will be truthy.
For instance: `{{ x | default(false) }}` will be 'true' if `x = 'no'`!

We've been burned a lot with this, where we would set a value to 'no' and see it evaluated as true.

### DIFF
--- a/src/ansiblelint/rules/MissingBoolCast.py
+++ b/src/ansiblelint/rules/MissingBoolCast.py
@@ -1,0 +1,33 @@
+from ansiblelint import AnsibleLintRule
+
+import re
+
+# we match anything that has '| default(<bool>)'
+contains_default_bool_re = r'.*\|default\((\'?)(true|false|yes|no)(\'?)\)'
+condition_contains_bool_default = re.compile(contains_default_bool_re, re.IGNORECASE)
+
+# and we check that anything that matches contains a '| bool' immediately after the match
+condition_contains_bool_cast = re.compile(contains_default_bool_re + r'\|bool', re.IGNORECASE)
+
+class MissingBoolCast(AnsibleLintRule):
+    id = 'BOOLFILTER405'
+    shortdesc = 'Missing |bool in `when:` condition with a `default(<boolean>)`'
+    description = '''
+    This filter checks that any jinja or expression in the code that uses a boolean default is correctly
+    postpended by "| bool".
+
+    This is an issue because any non "boolean" value will be truthy.
+    For instance: `{{ x | default(false) }}` will be 'true' if `x = 'no'`!
+    '''
+    tags = ['idiom']
+    severity = 'HIGH'
+
+    def match(self, info, line):
+        return is_bool_filter_missing(line)
+
+def is_bool_filter_missing(string):
+    normalized_string = string.replace(' ', '') # remove all spaces to simplify regex match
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
I may be bit biased here due to using Python as main language for a very long period of time. For me "no" string evaluating as true is something I do expect.

There are two other issues with this change: id should be `no-bool-filter` and it should also contain some tests.

Still, the main issue is that I am not convinced about this rule as adding value but I will have to ask other for feedback.

Gui13:
This stems from a misunderstanding on our side, we usually use `default(true)` or `default(false)` but sometimes the actual value has to be 'no' (the string) by default (for instance, in url queries, and other configuration files).
So some people in our team fell on a `default('no')` in the code  and thought that this meant `default(false)`, and started using it as if it was a boolean value. No linter actually recognizes this as a potentially bad behavior, and we found subtle (and dangerous) bugs due to this.

This is compounded by the fact that ansible (actually, yaml?) handles `yes` and `no` as if they were booleans, so in the code, `default('no')` looks legit. Especially because `default(no)` **doesn't** work, but using `xxx: yes` works. I understand that this might be obvious to somebody deeply ingrained in python/yaml/jinja/ansible, but for a neophyte this makes no sense.

Lastly, since you have to use strings even for boolean values when you work with jinja, Ansible forces you to build a mental model where strings are automatically parsed as booleans; see this for example:

```yaml
---
- hosts: all
  tags: lol
  gather_facts: false
  tasks:
  - name: Set a boolean fact
    vars:
      default_truthy: 'yes'
    set_fact:
      # all these are seen a boolean values even though we use "<my value>"
      false_string: "{{ some_undefined_var | default('false') }}"
      false_bool: "{{ some_undefined_var | default(false) }}"
      false_no: "{{ some_undefined_var | default('no') }}"
      false_default_yes: "{{ default_truthy | default('no') }}"
    delegate_to: localhost
  - name: Debug
    debug:
      var: item
    delegate_to: localhost
    with_items:
    - "{{ false_string }}" # => false
    - "{{ false_bool }}" # => false
    - "{{ false_no }}" # => false
    - "{{ false_default_yes }}" # => true
  # So it seems we can use default('false') ? NO! Not in `when:` statements:
  - name: Debug when condition
    debug:
      msg: "This message should not be printed" # it is printed !
    when: some_undefined_var | default('false')

```

Which displays:
image


This issue could be a warning maybe then: if you are using a `default()` with something inside that looks like a boolean but is a string, you might be doing something wrong.

bcoca:
isn't this what 'strict' option for default is for?

ssbarnea:
Closing as I do not see this getting enough cvorum. Proposed implementation would raise too many false positives. We can reconsider it if we find other approaches because I do understand the need for a better bool testing.

Gui13:
I can understand the false positive issue, I use this filter as an additional filter for ansible lint anyway, I just wanted to upstream it if it had use elsewhere.

I will stand by my statement though, that when you are using default("no") (which evaluates to true) then it is very likely that you are trying to do a bool comparison and this is not going to go well.

@bcoca can you tell me more about the strict option? I cannot find anything relating to that in jinja doc or ansible.

PR #1540, opened 2021-04-29
Add new rule that detects use of blind ignore_errors: true

konstruktoid:
This PR adds a new rule name `ignore-errors` that will trigger if `ignore_errors: true` is used.

The reasoning behind this is that instead of ignoring all errors, the use of `failed_when:` and specified acceptable error codes to reduce the risk of ignoring important failures is recommended.

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>

### DIFF
--- a/src/ansiblelint/rules/IgnoreErrorsRule.py
+++ b/src/ansiblelint/rules/IgnoreErrorsRule.py
@@ -14,10 +14,12 @@ class IgnoreErrorsRule(AnsibleLintRule):
     """Describe and test the IgnoreErrorsRule."""
 
     id = "ignore-errors"
-    shortdesc = 'Use failed_when and specify error conditions instead of using ignore_errors'
+    shortdesc = (
+        'Use failed_when and specify error conditions instead of using ignore_errors'
+    )
     description = (
         'Instead of ignoring all errors, use ``failed_when:`` '
-        'and specify acceptable error codes '
+        'and specify acceptable error conditions '
         'to reduce the risk of ignoring important failures'
     )
     severity = 'LOW'
@@ -28,7 +30,7 @@ class IgnoreErrorsRule(AnsibleLintRule):
         self, task: Dict[str, Any], file: 'Optional[Lintable]' = None
     ) -> Union[bool, str]:
 
-        if task.get("ignore_errors"):
+        if task.get("ignore_errors") and not task.get("register"):
             return True
 
         return False
@@ -51,6 +53,15 @@ if "pytest" in sys.modules:
     - name: run apt-get update
       command: apt-get update
       ignore_errors: false
+'''
+
+    IGNORE_ERRORS_REGISTER = '''
+- hosts: all
+  tasks:
+    - name: run apt-get update
+      command: apt-get update
+      ignore_errors: true
+      register: ignore_errors_register
 '''
 
     FAILED_WHEN = '''
@@ -81,6 +92,14 @@ if "pytest" in sys.modules:
     def test_ignore_errors_false(rule_runner: Any) -> None:
         """The task uses ignore_errors: false, oddly enough."""
         results = rule_runner.run_playbook(IGNORE_ERRORS_FALSE)
+        assert len(results) == 0
+
+    @pytest.mark.parametrize(
+        'rule_runner', (IgnoreErrorsRule,), indirect=['rule_runner']
+    )
+    def test_ignore_errors_register(rule_runner: Any) -> None:
+        """The task uses ignore_errors: but output is registered and managed."""
+        results = rule_runner.run_playbook(IGNORE_ERRORS_REGISTER)
         assert len(results) == 0
 
     @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


ssbarnea (approved):
I am ok with it but I am sure this will be controversial, so I so want extra eyes on it before we add it.

Is good that is added as experimental, which means it will not fail the linting, only acting as a warning, at least for the moment.

src/ansiblelint/rules/IgnoreErrorsRule.py
```diff
@@ -0,0 +1,92 @@
...
+    from typing import Optional
+
+    from ansiblelint.file_utils import Lintable
+
+
+class IgnoreErrorsRule(AnsibleLintRule):
+    """Describe and test the IgnoreErrorsRule."""
+
+    id = "ignore-errors"
+    shortdesc = 'Use failed_when and specify error codes instead of ignore_errors'
```
MarkusTeufelberger, line 17:
```suggestion
    shortdesc = 'Use failed_when and specify error conditions instead of using ignore_errors'
```

src/ansiblelint/rules/IgnoreErrorsRule.py
```diff
@@ -0,0 +1,92 @@
...
+      command: apt-get update
+      ignore_errors: true
+'''
+
+    IGNORE_ERRORS_FALSE = '''
+- hosts: all
+  tasks:
+    - name: run apt-get update
+      command: apt-get update
+      ignore_errors: false
```
MarkusTeufelberger, line 53: Maybe this should be also its own rule? After all, `ignore_errors: false` is a quite nonsensical statement and might be a hint for some change gone wrong that was intended to be made elsewhere.

tadeboro (commented):
I did not look at the code too closely, but adding this rule is OK with me.

We do use `ignore_error: true` a lot in our integration tests because we often expect tasks to fail there. And while it would be possible to use `failed_when` in those cases, it would considerably decrease readability (example of such playbook: https://github.com/sensu/sensu-go-ansible/blob/c000b1f016955047976028805d8ef021773c244e/tests/integration/molecule/module_filter/converge.yml#L8-L19). If we would use ansible-lint on those playbooks, we would need to add a lot of ignores to our playbooks. But luckily we do not ;)

dmsimard:
> The "correct" version if this is failed_when: false which reads a bit more counter-intuitive to me than ignore_errors: true and I don't see the immediate benefit tbh.

To me, ``failed_when: false`` basically means: "this task can never fail and will always return OK"
Whereas  ``ignore_errors: true`` means: "this task can fail and mark it as such but I don't want this failure to fail the entire playbook, I might even handle this failure in a subsequent task"

ssbarnea:
I think that everyone involved in this discussion known the pros and cons of ignore errors. The question is if we can find a way to lower the number of false-positives from this rule?

* How about ignoring the rule if `register` exists on task? That is usually a sign that the recorded value is used later.
* How about ignoring the rule if `(test|molecule)` is part of the file path? That usually indicates that this is testing code, where we can assume the user probably knew.

No need to be afraid about this rule breaking existing code because is added as experimental (warning) only.

As @tadeboro said, we can assume some false-positives with any rule but it would be essential for the number of false positives to be very low as nobody likes polluting the code with ignores.

MarkusTeufelberger:
Please don't make rules dependent on external factors such as other files/folders existing.

Having a "it makes sense in these use cases" exception might be a good hint that this is not a good universal rule to begin with.

Why should `ignore_errors: true` be removed from everywhere except test code in the first place? There are also other reasons why one might want/need to ignore an error happening while still having it show up as "failed" in the logs (something `failed_when` doesn't do).

There might be an argument for different rule sets for roles and test playbooks or some "strict mode" that ensures some certain parameters like "no failing tasks get skipped", but that would be a bigger discussion.

ssbarnea:
Something that some of you may not know is that ignore_errors does ignore errors even if the called ansible module is crashing! I seen multiple cases where modules were called with invalid parameters for months and nobody observed due to use of ignore_errors. Ignore errors is more like a `catch Exception: pass` in python ;)

src/ansiblelint/rules/IgnoreErrorsRule.py
konstruktoid: I agree, it was mostly added as a test case as hinted on https://github.com/ansible-community/ansible-lint/pull/1540/files#diff-8d1811f448323a979183ecf2d278ac0956ac3b60f91cc22eeb1c777b24520537R82.

konstruktoid:
> Something that some of you may not know is that ignore_errors does ignore errors even if the called ansible module is crashing! I seen multiple cases where modules were called with invalid parameters for months and nobody observed due to use of ignore_errors. Ignore errors is more like a `catch Exception: pass` in python ;)

This basically sums up the reasoning for this rule, and it's labeled as experimental which shouldn't cause any issues in existing pipelines.

Ignoring the rule if `register` exists on task seems like a good next step in my opinion as well.

alessfg:
The Ansible docs recommend using `ignore_errors` when using check mode to avoid potential check mode related errors https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html -- Before this rule gets unlabeled as experimental, it'd be great to add an exception for that use case.

ssbarnea:
Make a pr to do that, it makes sense and it should be really easy to do so.

PR #1543, opened 2021-05-03
Allow user to make yamllint rule mandatory

ssbarnea:
Closes: #1538

### DIFF
--- a/src/ansiblelint/rules/YamllintRule.py
+++ b/src/ansiblelint/rules/YamllintRule.py
@@ -3,6 +3,7 @@ import os
 import sys
 from typing import TYPE_CHECKING, List
 
+from ansiblelint.config import options
 from ansiblelint.file_utils import Lintable
 from ansiblelint.rules import AnsibleLintRule
 from ansiblelint.skip_utils import get_rule_skips_from_line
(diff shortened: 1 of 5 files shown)


src/ansiblelint/rules/YamllintRule.py
```diff
@@ -38,6 +43,10 @@
 Specific tag identifiers that are printed at the end of rule name,
 like 'trailing-spaces' or 'indentation' can also be be skipped, allowing
 you to have a more fine control.
+
+By default this rule is not used hwne yamllint library is missing. If you want
```
dosmanak, line 47: Typo "hwne" => "when"

dosmanak:
This change does not solve the non-deterministic behaviour on yamllint presence in the system :-(

ssbarnea:
@dosmanak It does, if you add the now documented rule. On the other hand, if you want deterministic results, you could opt for using a container. Different results were caused by different environments, that was not really a bug in the linter.

dosmanak:
I still do not understand the default behaviour.
When the yamllint is present it is used, whilst if missing it fails silently. Why have you chose such illogical approach? Perhaps I am the only one who feels that way and I will not push the discussion further.

PR #1549, opened 2021-05-12
Fix get_yaml_files to return all configured files

sengaya:
The issue was introduced with #1473 and only affects setups without a
git repo and when custom kinds are used.

### DIFF
--- a/src/ansiblelint/file_utils.py
+++ b/src/ansiblelint/file_utils.py
@@ -227,7 +227,6 @@ def get_yaml_files(options: Namespace) -> Dict[str, Any]:
             os.path.join(root, name)
             for root, dirs, files in os.walk('.')
             for name in files
-            if name.endswith('.yaml') or name.endswith('.yml')
         ]
 
     return OrderedDict.fromkeys(sorted(out))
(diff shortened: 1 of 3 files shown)


ssbarnea:
@sengaya Is not clear from the description but my impression is that this fixed the issue where `get_yaml_files` return only yaml files when run outside git repos and all files outside them.

If so, maybe we should also rename `get_yaml_files` to `get_lintables` in order to avoid confusing us.

ssbarnea (changes requested):
Fix the linting and explain the reasoning a little bit better. Other than this is looks like a good bugfix.

sengaya:
@ssbarnea thanks for the feedback. I would not rename it to `get_lintables` as it looks like the method should return only yaml files that might have a custom kind (e.g. something other than `.yaml` or . `.yml`), but the file itself has still yaml content. Also the method is called by a method `get_lintables` already.
 
But after looking again at `get_yaml_files` I'm a bit lost. I think the detection of yaml files is even more broken. You can create a failing test by adding and comitting a file to e.g. `examples/roles/test-role/tasks/non-lintable.file` and run the test `test_get_yaml_files_silent`. I would expect the file to be ignored, but the test fails. So to me it looks like the method currently just returns any file and neither the default kinds nor custom kinds are applied.

sengaya:
@ssbarnea After digging a bit deeper into the code base I could come up with a solution how to apply the default and custom yaml patterns. Let me know what you think.

ssbarnea:
Obsoleted by #1557

PR #1552, opened 2021-05-17
Ignore empty env variable values when updating them

tadeboro:
Right now, if the variable is not set, prerun routines will prepend the `:` to the variable value, which is not something we want. This commit makes sure we properly handle the aforementioned situation and ignore any empty values.

Should fix the problem described in https://github.com/ansible-community/molecule/pull/3122

### DIFF
--- a/src/ansiblelint/prerun.py
+++ b/src/ansiblelint/prerun.py
@@ -321,7 +321,10 @@ def _write_module_stub(
 def _update_env(varname: str, value: List[str], default: str = "") -> None:
     """Update colon based environment variable if needed. by appending."""
     if value:
-        value = [*os.environ.get(varname, default=default).split(':'), *value]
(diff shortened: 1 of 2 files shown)


ssbarnea:
Please rebase so we get rid of the fixed Ci issues.

ssbarnea (approved):

tadeboro:
Hmm, looks like I broke something since the eco test is failing. This sounds like an interesting problem.

ssbarnea:
> Hmm, looks like I broke something since the eco test is failing. This sounds like an interesting problem.

Yep, I already contacted @greg-hellings on that. It may be possible to be a bug in his role testing that your fix uncovered. Ansible is unable to find that action plugin after this change. Lets see which one finds the root cause for that.

PR #1557, opened 2021-05-19
Improve file detection outside git repositories

ssbarnea:
Address bug related to detection of lintable files when run outside
a git repository.

- extend list of exclusions with few common directiories
- allow linter to identify any file type, not only those with
  yaml/yml extension (matches behavior of git ls-files variant)
- improve logging messages

Closes: #1549


klausenbusk:
I think this change is causing this bug: https://github.com/ansible-community/ansible-lint/issues/1565

klausenbusk:
I also think https://github.com/ansible-community/ansible-lint/issues/1574 is caused by this change.

PR #1558, opened 2021-05-19
Add rule for checking no_log is set

noonedeadpunk:
It's pretty important to ensure, that password won't be logged
and thus exposed in the output. So we add linter rule that will
check that all tasks, that have "password" in argument are not logged.

Signed-Off-By: Dmitriy Rabotyagov <noonedeadpunk@ya.ru>


src/ansiblelint/rules/NoLogPasswordsRule.py
```diff
@@ -0,0 +1,35 @@
...
+
+class NoLogPasswordsRule(AnsibleLintRule):
+    id = "no-log-password"
+    shortdesc = "password should not be logged."
+    description = (
+        "All the modules that take a password argument should fail "
+        "if no_log is not set or set to False in the task."
+    )
+    severity = 'LOW'
+    tags = ["passwords", "experimental"]
```
ssbarnea, line 34: Replace passwords with `security`, describes the category in a more generic way.

ssbarnea: IMHO a description  message like below would be less likely to confuse users
```
When passing password argument you should have no_log configured to a non False value to avoid accidental leaking of secrets.
```

ssbarnea (changes requested):

src/ansiblelint/rules/NoLogPasswordsRule.py
noonedeadpunk: Done

ssbarnea:
You still need to: increase number of rules from 39 to 40 (specific test that fails). For lint I am sure you know how to test locally (tox -e lint).

examples/playbooks/no-log-passwords-failure.yml
```diff
@@ -0,0 +1,18 @@
+- tasks:
```
ssbarnea, line 1: add `hosts: localhost`, without it is not a valid playbook

examples/playbooks/no-log-passwords-success.yml
```diff
@@ -0,0 +1,13 @@
+- tasks:
```
ssbarnea, line 1: add `hosts: localhost`, without it is not a valid playbook

ssbarnea (changes requested):

noonedeadpunk:
Not sure how to satisfy `Codacy Static Code Analysis` though. As it seems other rules are set the same way with same issues, but it's not complaining about them

ssbarnea:
> Not sure how to satisfy `Codacy Static Code Analysis` though. As it seems other rules are set the same way with same issues, but it's not complaining about them

You can fully ignore Codacy is there only as an experiment, but the others must pass, especially the unittesting.  Run locally with `tox -e py39-core` or similar and if it passed you should be fine.

noonedeadpunk:
Yep, also saw things failing, but barely understand why tbh. As how `good_runner.run()` could result in `[no-log-password] (password should not be[270 chars]True`

ssbarnea:
> Yep, also saw things failing, but barely understand why tbh. As how `good_runner.run()` could result in `[no-log-password] (password should not be[270 chars]True`

I fixed them.

ssbarnea (approved):

Issue #1528, opened 2021-04-09
risky-file-permissions rule ignores FQCN


dessert1:
##### Summary
The risky-file-permissions rule is triggered for the `copy` module, but not for `ansible.builtin.copy`.

##### Issue Type
- Bug Report

##### Ansible and Ansible Lint details
```console
$ ansible --version
ansible 2.10.7
  config file = /home/me/.ansible.cfg
  configured module search path = ['/home/me/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.2 (default, Feb 20 2021, 18:40:11) [GCC 10.2.0]
$ ansible-lint --version
ansible-lint 5.0.7 using ansible 2.10.7
```

- ansible installation method: OS package
- ansible-lint installation method: OS package

##### OS / ENVIRONMENT
Arch Linux

##### STEPS TO REPRODUCE
A playbook using `ansible.builtin.copy` without `mode` gives no violations, shortening the module name to just `copy` throws the risky-file-permissions warning as expected.
```console
$ cat site.yml
- hosts: all
  tasks:
    - name: copy that
      ansible.builtin.copy:
$ ansible-lint -pqq site.yml
$ sed -i 's/ansible\.builtin\.//' site.yml
$ ansible-lint -pqq site.yml
site.yml:3: risky-file-permissions
```

##### Desired Behaviour
`ansible-lint` should treat the `ansible.builtin.copy` module exactly like it does `copy`.

##### Actual Behaviour
`ansible-lint` doesn’t seem to work with FQCNs at least in this particular case. There may be other examples where modules are ignored because they’re called with their FCQN.

oldATWiz:
I see this as well with ansible.builtin.command. Adding the FQCN removes the 'no-changed-when: Commands should not change things if nothing needs doing' violation.

```
ansible-lint --version
ansible-lint 5.0.9 using ansible 2.11.0
```

PR #1563, opened 2021-05-20
Make risky-file-permissions rule does not ignore FQCN (#1528)

ssato:
Address the bug that the risky-file-permissions rule ignores the modules passed as FQCN like ansible.builtin.copy.


ssbarnea:
Why not just adding both short and long names to the same lists? The new configuration seems overly complicated. I am curious if others share my opinion on that matter.

ssato:
> Why not just adding both short and long names to the same lists? The new configuration seems overly complicated. I am curious if others share my opinion on that matter.  thinking

I thought that much simple way at first but re-thought there might be a possibility of need to add more modules to take care in this rule. Also, similar fix may be required in other rules listing only short names of modules currently and smarter way to list modules  might be needed, I guess. ;-)

But, of course, I'll make change that if it's better to keep it simple other than taking care of future possibilities. I don't have strong opinion about it.

src/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -33,6 +34,63 @@
...
+    'archive',
+    'assemble',
+    'copy',  # supports preserve
+    'file',
+    'replace',  # implicit preserve behavior but mode: preserve is invalid
+    'template',  # supports preserve
+    # 'unarchive',  # disabled because .tar.gz files can have permissions inside
+}
+for mod in _MODULES.copy():
+    _MODULES.add(f'{_MOD_FQCN_PREFIX_MAP[mod]}.{mod}')
```
tadeboro, line 81: This is overly complex. `_MODULES` should be a (frozen) set with all available options and that is it.

src/ansiblelint/rules/MissingFilePermissionsRule.py
```diff
@@ -33,6 +34,63 @@
...
+_MODULES_WITH_CREATE = {
+    'blockinfile': False,
+    'htpasswd': True,
+    'ini_file': True,
+    'lineinfile': False,
+}
+for mod in _MODULES_WITH_CREATE.copy():
+    _MODULES_WITH_CREATE[f'{_MOD_FQCN_PREFIX_MAP[mod]}.{mod}'] = _MODULES_WITH_CREATE[
+        mod
+    ]
```
tadeboro, line 92: The same thing here. `_MODULES_WITH_CREATE` should be a plain dictionary with all options directly listed.

tadeboro (changes requested):
I like the idea behind the PR, but the implementation is a bit too convoluted for my taste. YAGNI wins for me here.

ssato:
> I like the idea behind the PR, but the implementation is a bit too convoluted for my taste. YAGNI wins for me here.

Thanks for your feedbacks!

I pushed the simple version for the issue.

ssbarnea:
@ssato Now you have write access to the repository. This means you can create branches, perform reviews and even merge changes that received reviews.

I encourage you to start using branches instead of your fork because it allows other cores to collaborate on them, making easier to fix small things and merge things faster. 

Again, thanks for helping with the linter!

ssbarnea (approved):

ssato:
> @ssato Now you have write access to the repository. This means you can create branches, perform reviews and even merge changes that received reviews.
> 
> I encourage you to start using branches instead of your fork because it allows other cores to collaborate on them, making easier to fix small things and merge things faster.

Thanks a lot! I'll do so later.

PR #1564, opened 2021-05-20
Add metadata tag to the role-name rule

noonedeadpunk:
role-name rule is based on the metadata info. Thus, it would be expected
by users to also exclude this task, when `-x metadata` is set in the
config

Signed-Off-By: Dmitriy Rabotyagov <noonedeadpunk@ya.ru>

### DIFF
--- a/src/ansiblelint/rules/RoleNames.py
+++ b/src/ansiblelint/rules/RoleNames.py
@@ -49,7 +49,7 @@ class RoleNames(AnsibleLintRule):
     )
     severity = 'HIGH'
     done: List[str] = []  # already noticed roles list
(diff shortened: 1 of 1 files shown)


noonedeadpunk:
Labels check is frustrating for contributors tbh. Or I just don't see/know the way to set it.

ssbarnea:
You cannot, only cores can add labels and sadly github does not have a diffetent way to prevent a merge while checking the labels.

role-name is usually meta but not necessarily as it can be determined just by folder name.

So it is meta related but not pure meta. Probably ok to add meta label.

ssbarnea (approved):

Issue #1532, opened 2021-04-15
Tests erroneously pass when commented out code is present


Rylon:
##### Summary

I noticed that a task that should be failing is passing when a commented out line of code is present

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console
$ ansible --version
ansible 2.10.8

$ ansible-lint --version
ansible-lint 5.0.3 using ansible 2.10.8
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Docker running Debian 10 image, Python 3.8.5

##### STEPS TO REPRODUCE

Consider this task:

```console
- name: Create directory
  file:
    path:  /my_directory
    state: directory
    owner: root
    group: root
    # mode:  0755
```

Given this task is missing the file mode, I would expect to receive the error `risky-file-permissions: File permissions unset or incorrect`, however the presence of the commented out `mode` seems to cause the the task to pass the test instead of failing.

If I delete the commented out line entirely, the task does indeed fail the test.

##### Desired Behaviour

The task should fail linting because it has no file mode. I wouldn't consider the commented out code to allow the task to pass linting because if it is commented out it isn't going to be considered if Ansible were to run that task.

##### Actual Behaviour

The task does not fail linting, unless the commented out line is removed entirely.

konstruktoid:
Hi @Rylon, and thanks for opening this issue.
Can you try with the latest release?

ssbarnea:
This was was fixed for sure.

Rylon:
Hi, I can't seem to reproduce this on 5.0.11! Thanks!

PR #1568, opened 2021-05-21
Refactor MissingFilePermissionsRule testing

konstruktoid:
This PR:
- was meant to verify #1532, but `# mode: 0755` fails correctly when tested
- merges tests into `src/ansiblelint/rules/MissingFilePermissionsRule.py` and removes `test/TestMissingFilePermissionsRule.py`
- adds a comment test to ` src/ansiblelint/rules/NoLogPasswordsRule.py` as additional verification


konstruktoid:
It could most likely be optimized but I added `noqa: C901` (https://www.flake8rules.com/rules/C901.html), which is triggered due to the number of tests.

ssbarnea (approved):

ssbarnea:
I think that case is ok for an exclusion, especially as I think is impossible to avoid with that approach.

PR #1591, opened 2021-06-01
Remove no-log-password rule.

felixfontein:
Fixes #1589.


felixfontein:
I don't think the failing test is related to this PR.

felixfontein:
Indeed, the failure is caused by a deprecation removal in ansible/ansible's devel branch, see #1592.

ssbarnea (changes requested):
This rule is marked as experimental, which means that by default it will not produce failures, only warnings.

I seen too many users arguing that it can help them avoid leaking secrets and thus I am against its removal.

We should look into improving it and have a broader discussion around it in a community meeting. We could endup making it an opt-in rule if it proves to be problematic even after the recent improvement around looping.

ssbarnea:
For the moment we will keep it as opt-in. If we fail to improve it to make it more useful we will remove it.

PR #1592, opened 2021-06-01
[WIP] Check tests

felixfontein:


felixfontein:
The failing tests were caused by https://github.com/ansible/ansible/pull/74809. `ansible-playbook --syntax-check` now fails when it cannot find the included file in `TestIncludeMissingFileRule.test_include_file_missing`, and ansible-lint never runs the `IncludeMissingFileRule` rule.

felixfontein:
ansible-core's behavior before: succeeds with
```
[DEPRECATION WARNING]: Included file '/path/to/some_file.yml' not found, however since this include is not explicitly marked as 'static: yes', we will try and include it dynamically later. In the future, this will be an error unless 'static: no' is used on the include task. If you do not want missing includes to be considered dynamic, use 'static: yes' on the include or set the global ansible.cfg options to make all includes static for tasks and/or handlers. This feature will be removed from ansible-base in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```
ansible-core's behavior now: fails with
```
ERROR! Unable to retrieve file contents
Could not find or access '/path/to/some_file.yml' on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
```

felixfontein:
Closing in favor of #1599.

PR #1599, opened 2021-06-02
Remove include missing file rule

ssbarnea:
As current versions of Ansible do correctly detect missing imports and raise load-failure errors, we no longer need our own rule for it.

This also fixes CI devel jobs which recently started to fail due to changes made to devel branch. Please note that removal of this rule does not mean the linter would not detect missing files, in fact it already does via ansible loader. Before this change missing files caused two different errors for each missing file on 2.9-2.11 and failed to pass tests on 2.12 (devel).


felixfontein:
This is only suppored by ansible-core 2.12, which is in development. I'm not sure whether it's a good idea to remove the rule already now.

felixfontein (approved):

Issue #1594, opened 2021-06-01
skip_list syntax-check is ignored


Roosterfish:
##### Summary

When trying to disable the `syntax-check` with `skip_list:` in the config file, the linter is completely ignoring it. Adding `syntax-check` to `warn_list:` works like expected.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
# ansible --version
ansible 2.9.20

# ansible-lint --version
ansible-lint 5.0.11 using ansible 2.9.20
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Ubuntu 20

##### STEPS TO REPRODUCE

Create a config file under `~/.ansible-lint.yml`

```yaml
skip_list:
  - syntax-check
kinds:
  - playbook: "**/*.yml"
```

Create a playbook that leads to an error during `ansible-playbook --syntax-check playbook.yml`:

```yaml
- name: Play name
  hosts: "{{ some_unset_variable }}"
```

Run `ansible-lint -c ~/.ansible-lint.yml playbook.yml`

##### Desired Behaviour

The linter should skip the `syntax-check`.

##### Actual Behaviour

```console
INFO     Added ANSIBLE_ROLES_PATH=~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:roles:roles
INFO     Executing syntax check on playbook.yml (1.85s)
WARNING  Listing 1 violation(s) that are fatal
syntax-check: The field 'hosts' has an invalid value, which includes an undefined variable. The error was: 'some_unset_variable' is undefined
playbook.yml:1:3 [WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'
ERROR! The field 'hosts' has an invalid value, which includes an undefined variable. The error was: 'some_unset_variable' is undefined

The error appears to be in '/.../playbook.yml': line 1, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Play name
  ^ here


You can skip specific rules or tags by adding them to your configuration file:
# .ansible-lint
warn_list:  # or 'skip_list' to silence them completely
  - syntax-check  # Ansible syntax check failed

Finished with 1 failure(s), 0 warning(s) on 1 files.
```

ssbarnea:
[...]

PR #1606, opened 2021-06-04
Document syntax-check as unskippable

ssbarnea:
* marks syntax-check rule as unskippable
* extends documentation of the rule in order to explain its special status and hint on which directions the user can take to avoid it
* assures unskippable rules are not advertised for use in skip_list or warn_list

Fixes: #1594

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -118,7 +118,7 @@ def initialize_options(arguments: Optional[List[str]] = None) -> None:
     )
 
 
-def report_outcome(
+def report_outcome(  # noqa: C901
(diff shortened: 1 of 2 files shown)


felixfontein (approved):

tadeboro (approved):
It still feels a bit strange that failing the "syntax" check prevents any other errors from being reported. And I placed the syntax part in quotes because Ansible does a lot more in its syntax check. So even if the playbook is valid as far as the syntax goes, the user does not get any diagnostics back.

But this is only weakly related to this PR, and this PR is indeed accurately described the current state of affairs.

Issue #1587, opened 2021-05-25
Fully qualified collection name automatically stripped away from matchtask


StopMotionCuber:
##### Summary

We are currently using a rule that doesn't exist upstream to check whether we use the fully qualified collection name for builtins, as we recently migrated to fully qualified collection name and have integrated a check to CI to be consistent here.
Code of it is seen below:
```
from typing import Any, Dict, Union, Optional

from ansiblelint.rules import AnsibleLintRule

from ansiblelint.file_utils import Lintable

builtins = [
    "add_host", "apt", "apt_key", "apt_repository",
    "assemble", "assert", "async_status", "blockinfile", "command",
    "copy", "cron", "debconf", "debug", "dnf", "dpkg_selections",
    "expect", "fail", "fetch", "file", "find", "gather_facts",
    "get_url", "getent", "git", "group", "group_by", "hostname",
    "import_playbook", "import_role", "import_tasks", "include",
    "include_role", "include_tasks", "include_vars", "iptables",
    "known_hosts", "lineinfile", "meta", "package", "package_facts",
    "pause", "ping", "pip", "raw", "reboot", "replace", "rpm_key",
    "script", "service", "service_facts", "set_fact", "set_stats",
    "setup", "shell", "slurp", "stat", "subversion", "systemd",
    "sysvinit", "tempfile", "template", "unarchive", "uri", "user",
    "wait_for", "wait_for_connection", "yum", "yum_repository",
]


class FQCNBuiltinsLint(AnsibleLintRule):
    id = "fqcn"
    shortdesc = "Use fqcn for builtins"
    description = "Use fqcn for builtins"
    tags = ["productivity"]

    def matchtask(
        self, task: Dict[str, Any], file: "Optional[Lintable]" = None
    ) -> Union[bool, str]:
        if task["action"]["__ansible_module__"] in builtins:
            return (f"{task['__file__']}:{task['action']['__line__'] - 1}: "
                    f"{task['action']['__ansible_module__']}")

        return False

```
Our problem is now that #1581 got merged, `task["action"]["__ansible_module__"]` will falsely flag basically all of our roles, with https://github.com/ansible-community/ansible-lint/pull/1581/files#diff-d453523679145f511ca4e8fb97be96e7dd664ae88373e2e5e9a90a65fd751cabR562 as the exact line of this problem.

If you would be fine with expanding the normalization to also include the unnormalized way (e.g. as seen in desired behavior) I can happily create a PR for that.

[...]

PR #1614, opened 2021-06-07
Add opt-in rule to check whether FQCN is used for builtins

StopMotionCuber:
As I have written in #1587, we currently have a regression for our custom rule that checks whether FQCNs are used everywhere.
As @ssbarnea suggested there, I added this rule upstream and made it opt-in. To not break current pipelines, it is by default turned _off_ in the CI. I needed to change the `rule_runner` fixture, such that this rule is ran for the `rule_runner` nevertheless (this also seems like the more sensible default to me, as you include a specific rule there and expect it to be run).

This is my first contribution to this project, so feedback is welcome

### DIFF
--- a/src/ansiblelint/rules/FQCNBuiltinsRule.py
+++ b/src/ansiblelint/rules/FQCNBuiltinsRule.py
@@ -80,7 +80,7 @@ builtins = [
 
 class FQCNBuiltinsRule(AnsibleLintRule):
     id = "fqcn-builtins"
-    shortdesc = "Use fqcn for builtins"
+    shortdesc = "Use FQCN for builtin actions"
     description = (
         'Check whether the long version starting with ``ansible.builtin`` '
         'is used in the playbook'
--- a/src/ansiblelint/rules/FQCNBuiltinsRule.py
+++ b/src/ansiblelint/rules/FQCNBuiltinsRule.py
@@ -82,7 +82,7 @@ class FQCNBuiltinsRule(AnsibleLintRule):
     id = "fqcn-builtins"
     shortdesc = "Use fqcn for builtins"
     description = (
-        'Check whether the long version starting with ``ansible-builtin`` '
+        'Check whether the long version starting with ``ansible.builtin`` '
         'is used in the playbook'
     )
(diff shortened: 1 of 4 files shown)


src/ansiblelint/rules/FQCNBuiltinsRule.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def matchtask(
+        self, task: Dict[str, Any], file: Optional[Lintable] = None
+    ) -> Union[bool, str]:
+        if task["action"]["__ansible_module_original__"] in builtins:
+            return True
+            # return (f"{task['__file__']}:{task['action']['__line__'] - 1}: "
+            #         f"{task['action']['__ansible_module__']}")
+
+        return False
```
tadeboro, line 98:
```suggestion
        return task["action"]["__ansible_module_original__"] in builtins
```

tadeboro (changes requested):
I like the rule (for the playbooks I help maintain, we use FQCNs in all tasks). But this rule should definitely be disabled by default because as far as I know, there is no plan to deprecate bare name usage.

I did add a few nitpicks, but apart from that, the rule looks good to me.

StopMotionCuber:
Thanks for your review @tadeboro, I adapted the code to include your comments

tadeboro (approved):

src/ansiblelint/rules/FQCNBuiltinsRule.py
```diff
@@ -0,0 +1,131 @@
...
+    "wait_for",
+    "wait_for_connection",
+    "yum",
+    "yum_repository",
+]
+
+
+class FQCNBuiltinsRule(AnsibleLintRule):
+    id = "fqcn-builtins"
+    shortdesc = "Use fqcn for builtins"
```
felixfontein, line 83:
```suggestion
    shortdesc = "Use FQCN for builtin actions"
```
The rule does not check plugins.

tadeboro (approved):

felixfontein (approved):

ssbarnea (approved):

PR #1616, opened 2021-06-08
Update doc to reference with revision information

ahmedsajid:
The doc should be updated to at least point to the latest version available.


ssbarnea (approved):

ssbarnea:
Nobody can keep the tag updated so I switched to a version that should not need us to update documentation each time a new release is made. pre-commit users should know how to use `autoupdate` command often.

PR #1619, opened 2021-06-09
Add apt_key module to deprecated list

devNan0:
apt_key is deprecated.

See:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html#notes
https://github.com/ansible/ansible/issues/55590

### DIFF
--- a/src/ansiblelint/rules/DeprecatedModuleRule.py
+++ b/src/ansiblelint/rules/DeprecatedModuleRule.py
@@ -63,6 +63,7 @@ class DeprecatedModuleRule(AnsibleLintRule):
         'vsphere_guest',
         'win_msi',
         'include',
+        'apt_key',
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
Based on what I read on https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html#notes this module is not deprecated at all in ansible.

The fact that apt-key command it does use is deprecated at system level is something very different and my impression is that is not up to us to identify its use as a problem, especially as users may have to manage old systems.

devNan0:
Oh ok than i will close the PR again.

Sorry for the noise.

ssbarnea:
@devNan0 No worries, that is why we do reviews.

PR #1620, opened 2021-06-09
Make no-log-password rule an opt-in

ssbarnea:
Moves no-log-password from being experimental to also be an opt-in.

This means it will not be run at all unless user is enabling it
in the config.

Related: https://github.com/ansible-community/ansible-lint/discussions/1605
Closing: #1591


ssbarnea:
@tadeboro Thanks. This should buy us time to improve the rule or to delete it without affecting current users with noise.

PR #1625, opened 2021-06-13
Avoid false positive for async commands without changed_when

### LINKED ISSUE #1588: False positive: async jobs
##### Summary

A `command` module task that is run as an async job is incorrectly treated as a normal sync task.

For async tasks the options like `changed_when` (and `failed_when` and so on) are not given to the async `command` task itself, they are given to the `async_status` module task that is run after the async task.

Ansible-lint does not understand this and complains for rule `no-changed-when` for the `command` task.

Example:
```yaml
---
- name: Asynchronous long task
  command: alongtask.sh
  async: 1000
  poll: 0
  register: job_sleeper

- name: Wait for asynchronous job to end
  async_status:
    jid: '{{ job_sleeper.ansible_job_id }}'
  register: job_result
  until: job_result.finished
  retries: 100
  delay: 10
  changed_when: [....]
```

Note how the `changed_when` is given in the `async_status` task and not in the `command` task.

##### Issue Type

- Bug Report

```console (paste below)
ansible --version
2.9.21

ansible-lint --version
5.0.8

```

##### OS / ENVIRONMENT

EL7.9 all updated

##### STEPS TO REPRODUCE

```yaml
---
- name: Asynchronous yum task
  command: alongtask.sh
  async: 1000
  poll: 0
  register: job_sleeper

- name: Wait for asynchronous job to end
  async_status:
    jid: '{{ job_sleeper.ansible_job_id }}'
  register: job_result
  until: job_result.finished
  retries: 100
  delay: 10
  changed_when: [....]
```

```console (paste below)

```

##### Desired Behaviour

Ansible-lint should not detect `no-changed-when` for `command` module task run as async job since the `changed_when` cannot be given to the `command` module task itself.

It should detect that there is a `changed_when` in the following `async_status` task.

##### Actual Behaviour

Ansible-lint detects false positive `no-changed-when` for `command` module task run as async job even though `changed_when` cannot be correctly given for an async task - the `changed_when` is given for the subsequent `async_status` module task.

```paste below

```

bluikko:
For async tasks changed_when is given to the async task check and not to the actual task.

So ignore tasks missing changed_when if it is an async task.

Closes #1588.


bluikko:
Added the test cases.

ssbarnea (approved):

ssbarnea:
@bluikko you need to update the tests as the expected output has changed. You could run tests locally if you want, with tox.

tadeboro (changes requested):
> For async tasks changed_when is given to the async task check and not to the actual task.

Async tasks with `poll: 0` will always report `changed: true`, so I would say that adding `changed_when` is almost a must for such tasks, but this is not strictly related to this change and should probably be tackled some other time.

What should be discussed here is the change propagation because I think this PR gets things a bit wrong on that front. Let me explain with an example. The following playbook will always report both tasks as changed:

```
---
- hosts: localhost
  gather_facts: false
  tasks:
    - ansible.builtin.command: touch /tmp/123
      async: 600
      poll: 0
      register: status
    - ansible.builtin.async_status:
        jid: "{{ status.ansible_job_id }}"
      register: result
      until: result.finished
      retries: 100
      delay: 5
```

But the next playbook will only report first task as changed and will properly mark the second one as OK if the `/tmp/123` already exists:

```
---
- hosts: localhost
  gather_facts: false
  tasks:
    - ansible.builtin.command: touch /tmp/123
      args:
        creates: /tmp/123
      async: 600
      poll: 0
      register: status
    - ansible.builtin.async_status:
        jid: "{{ status.ansible_job_id }}"
      register: result
      until: result.finished
      retries: 100
      delay: 5
```

The problem with the current implementation is that ansible-lint will silently ignore such scenarios and fail to inform users that the command module should be used in a safer manner.

> So ignore tasks missing changed_when if it is an async task.

This is only partially true as demonstrated by the previous example.

The problem I described cannot be solved on a task level (we would need to analyze the whole playbook if we were to correctly determine all invalid usage patterns), so I would suggest that we leave the rule as is and just amend its documentation: tell users when it is safe to ignore this rule on an async task and what should they do when they wait for result.

bluikko:
> The problem I described cannot be solved on a task level (we would need to analyze the whole playbook if we were to correctly determine all invalid usage patterns), so I would suggest that we leave the rule as is and just amend its documentation: tell users when it is safe to ignore this rule on an async task and what should they do when they wait for result.

I have not considered the `creates` at all. It is not relevant to the case I came across, where nothing is created.

My case is a command that is run that outputs something. That output can only be processed at `async_status` and when done so the tasks are correctly marked as `changed`.

As discussed previously this cannot be resolved "right" when looking at only a single task without knowledge of the following `async_status` task.

If you say that `changed_when: true` must be given to async tasks, wouldn't that then justify its own rule? That would be better than "pointing to documentation" IMHO.

bluikko:
> @bluikko you need to update the tests as the expected output has changed. You could run tests locally if you want, with tox.

Let's see if I got this right...

ssbarnea:
@tadeboro Please take another look at this when you can and if you do not think this is an improvement, feel free to even close the PR and wait for someone to only update the rule docs and mention your workaround.

bluikko:
> @tadeboro Please take another look at this when you can and if you do not think this is an improvement, feel free to even close the PR and wait for someone to only update the rule docs and mention your workaround.

I feel very strongly about this and in my opinion the current behavior is definitely wrong. The "workaround" suggested involves littering the code with invalid `# noqa`s for what is absolutely a false positive.

There is already documentation for async tasks right there in the Ansible documentation. How much documentation does a user need? To be honest in my opinion the suggestion is patronizing to say the least.

At least it should be its own rule so the check can be turned completely off.

ssbarnea:
I already approved the change but I am not convinced that I did the right call here, that is why I asked him for feedback, especially as he had very good feedback.

My personal approach on using of changed_when with async was that I always manually set a value for changed_when that made sense based on what it was doing on remote hosts, sometimes it was `false`, sometimes it was `true`. I never had to use noqa for this rule as I find it much better to just configure a value.

@bluikko Please be patient, we will find a solution. The reason we are so picky is because we want to avoid alienating users with behavior changes, especially potential regressions. Clear bugfixes usually go in without much fuss, but yjay one has some extra risks.

bluikko:
So I actually have to backtrack here a bit.

The Ansible documentation did not actually talk about this at all, it was something I found out by myself when learning Ansible.

I took a quick stab at it, since in my opinion Ansible documentation should be the correct location for that information: https://github.com/ansible/ansible/compare/devel...bluikko:patch-1

bluikko:
> My personal approach on using of changed_when with async was that I always manually set a value for changed_when that made sense based on what it was doing on remote hosts, sometimes it was `false`, sometimes it was `true`.

I see that could be another way to do it. But what if it is unknown at that point if the system was changed? Then you'd be basically putting false information in the task.

As discussed since the beginning in #1588, I totally agree that this cannot be done really "right" when looking at a single task only - and it is simply a judgement call.

But perhaps the Ansible documentation could be improved also. I do not remember anymore how I came to use async tasks exactly but as per https://github.com/ansible-community/ansible-lint/pull/1625#issuecomment-862234870 it apparently didn't come from the Ansible documentation.

Since many more people read the Ansible documentation than use Ansible-lint, perhaps that is a better place for documenting the issue. Unless this kind of practice documenting has no place in it.

ssbarnea:
The recent update to ansible docs convinced me. You really did a good job to update Ansible docs first, the linter docs are more of a fail-over for the case where we do not have a reference to official docs, or a way to update them.

bluikko:
> The recent update to ansible docs convinced me. You really did a good job to update Ansible docs first, the linter docs are more of a fail-over for the case where we do not have a reference to official docs, or a way to update them.

Sorry, I have not updated that documentation yet - not even sent a PR yet. I just made a quick fork after the discussion here.

If you think it is useful I can make a PR for it in `ansible/ansible`. It would probably need plenty of editing, I have no idea of what style is expected in that documentation.

tadeboro:
I do not think merging this PR was the right choice because we just traded one false positive scenario for a false negative one, which is a bad trade. Why is this a bad trade? Because you can always silence false positives, but making false negatives loud is impossible.

For example, ansible-lint will stay silent now if it sees this:

```
- name: Init db
  command: init_db some params
  async: 654
  poll: 0
  register: result
```

But Ansible users would probably benefit here if ansible-lint informed them that the `command` module has a `creates` parameter that they can use not only to properly determine the changed status but also control when that command is executed without needing to use an explicit `when` condition.

ssbarnea:
@tadeboro To understand correctly, you do suggest that in the example above the linter should find a violation and tell user to either manually declare the `changed_when` value or make use of `creates` in order to help ansible figure-out the change status?

If true, I think I can make a patch to do so. I am sure some users may be annoyed having to add `changed_when: true` when the documentations states that this is the default value but at the same time I kinda prefer doing something like this than using annoying noqa comments.

tadeboro:
I actually think the `noqa` comment is the right thing to do here because it clearly marks the complex part of the playbook. Let me give a real concrete example of a thing linter will now miss and is indeed something that should be fixed:

```
- name: Mark something
  command: touch /tmp/my_marker
  async: 654
  poll: 0
```

Why is this problematic? Because the waiting task will not have enough information to determine if something changed or not. So the real fix here is to add the `creates` parameter to the command invocation and the waiter will sort itself out.

Adding `changed_when` on an async command task is something ansible-lint should probably avoid since this will not solve anything. And how to make sure the async task is not always marked as changed is another problem and out of the scope of this discussion.

bluikko:
For what it is worth the ansible documentation now has a note about correct usage of `changed_when` and so on for async tasks: https://github.com/ansible/ansible/pull/75020

We just have to disagree on this then. I hope a line is drawn somewhere to avoid proliferation of "look at the docs, dummy" kind of rules. And they should go to their own rules so that they can be turned off totally.

bluikko:
> Adding `changed_when` on an async command task is something ansible-lint should probably avoid since this will not solve anything. And how to make sure the async task is not always marked as changed is another problem and out of the scope of this discussion.

I want to remind you that the name of the rule is `no-changed-when`. So perhaps it indeed should be its own rule, instead of a revert.

If need I can volunteer to write a new rule to avoid keeping my own patches outside of the tree.

tadeboro:
The documentation update is in ansible/ansible is actually bad and contains a bad suggestion since the `async_status` module has no `changed` parameter. We need to sort the upstream docs now also.

PR #1631, opened 2021-06-16
Make ansible-lint recognize inventory.yml files

ssbarnea:
This fixes a bug that was originally reported on vscode-ansible, where it was observed that inventory files reported problems. From now on we do assume `inventory.yml` named files are inventory files.

Fixes: https://github.com/ansible-community/vscode-ansible/issues/126

### DIFF
--- a/examples/inventory.yml
+++ b/examples/inventory.yml
@@ -1,0 +1,30 @@
+# https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html
+ungrouped: {}
+all:
+  hosts:
+    mail.example.com:
+  children:
+    webservers:
+      hosts:
+        foo.example.com:
+        bar.example.com:
+    dbservers:
+      hosts:
+        one.example.com:
+        two.example.com:
+        three.example.com:
+    east:
+      hosts:
+        foo.example.com:
+        one.example.com:
+        two.example.com:
+    west:
+      hosts:
+        bar.example.com:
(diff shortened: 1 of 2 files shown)


tadeboro:
This PR is just bad because it makes assumptions out of thin air. There are no naming rules for things like inventory and playbook files and this change is overfitted to one specific use case. Plus, if I now create inventory.yml playbook, that playbook does not get linted at all.

I would suggest that we revert this change and find another way of solving the original issue.

ssbarnea:
Let's see if we can find a way to address this issue without the downside you mentioned. The reality is that before that patch ansible-lint would give an error if given an inventory yaml file as argument.

If a file is at the root of the repository we are in a gray area, for example we treat requirements.yml as requirement files. What if you would create a playbook with that name? ;)

I did read the [official docs about inventories](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#id16) again and i noted that I made a mistake, the examples given do mention an `inventory/` folder. This means that I could fix it by only recognizing inventory files from within those folders.

While this approach will still give an error for the user that put the inventory at root of the repo, at least it will give him a different escape route than adding it to the exclude list.

I already considered reading ansible config in order to find inventory filename but that would not work well as people may have any number of inventory files.

What should ansible-lint do if called with an yaml file that is not a playbook? (not hosted under a playbooks/ folder).

PR #1637, opened 2021-06-18
Improve install_collection implementation

ssbarnea:
- assure at default verbosity we do log ansible-galaxy command ran (info level is not logged by default)
- add ability to specify destination for install_collection
- assure we inject our own cache to top of collection path when present
- allow use of require_collection without a cache folder.

Note: this is likely the last change made to prerun code before moving this code into newly created standalone library https://github.com/ansible-community/ansible-compat/issues/4

### DIFF
--- a/src/ansiblelint/prerun.py
+++ b/src/ansiblelint/prerun.py
@@ -97,7 +97,7 @@ def check_ansible_presence(exit_on_error: bool = False) -> Tuple[str, str]:
     return ver, err
 
 
-def install_collection(collection: str) -> None:
+def install_collection(collection: str, destination: Optional[str] = None) -> None:
     """Install an Ansible collection.
 
     Can accept version constraints like 'foo.bar:>=1.2.3'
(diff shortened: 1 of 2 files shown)


src/ansiblelint/prerun.py
```diff
@@ -106,13 +106,13 @@ def install_collection(collection: str) -> None:
...
-        f"{options.cache_dir}/collections",
         "-v",
-        f"{collection}",
     ]
+    if destination:
+        cmd.extend(["-p", destination])
+    cmd.append(f"{collection}")
 
-    _logger.info("Running %s", " ".join(cmd))
+    _logger.warning("Running %s", " ".join(cmd))
```
tadeboro, line 115: This does not sound like a warning since it is a normal operation (there is nothing users can do to get rid of that warning). I would leave it silenced by default or make the default log level INFO.

src/ansiblelint/prerun.py
```diff
@@ -201,7 +201,9 @@ def prepare_environment(required_collections: Optional[Dict[str, str]] = None) -
 
     if required_collections:
         for name, min_version in required_collections.items():
-            install_collection(f"{name}:>={min_version}")
+            install_collection(
+                f"{name}:>={min_version}", f"{options.cache_dir}/collections"
```
felixfontein, line 205: Is it obvious that at this point, `options.cache_dir is not None`?

ssbarnea: Good point! I happen to know it should not be None by now but that is not the right way to deal with that. I will add an extra condition for that.

PR #1642, opened 2021-06-24
collection check for the import-playbook

provenvelocity:
This code implements the collection check for the import-playbook issue and also implements the fix needed for molecule.

https://github.com/ansible-community/molecule/issues/3135

https://github.com/ansible-community/ansible-lint/issues/1641

I'm sure I'm missing a step I have never had the time nor permission to contribute back to public modules... please help where needed.

### DIFF
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -60,7 +60,7 @@ BASE_KINDS = [
 
 
 options = Namespace(
-    cache_dir=None,
+    cache_dir=os.path.expanduser("~/") + ".cache",
     colored=True,
     configured=False,
     cwd=".",
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -113,7 +113,7 @@ def initialize_options(arguments: Optional[List[str]] = None) -> None:
         os.path.abspath(options.project_dir).encode()
     ).hexdigest()[:6]
     options.cache_dir = "%s/ansible-lint/%s" % (
-        os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache")),
+        os.getenv("XDG_CACHE_HOME", os.path.expanduser(options.cache_dir)),
         cache_key,
     )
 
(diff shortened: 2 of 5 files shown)


ssbarnea (changes requested):
* please ensure all CI jobs are green
* add a test that ensures that your change fixes the original problem, at this moment your change does not include any test, so it could regress at any point.

Both points are essential, especially as we have some incoming changes that will remove the `prepare` bit from the linter, without testing we would not be able to know if our change would break your fix.

erikgb:
@provenvelocity Any chance you can fix the issues pinpointed by @ssbarnea? I would like https://github.com/ansible-community/ansible-lint/issues/1641 to be fixed.

paytroff:
Hello, can you do the modifications for checks please ?
I'm interesting by this bugfixes too :)

src/ansiblelint/config.py
```diff
@@ -60,7 +60,7 @@
 
 
 options = Namespace(
-    cache_dir=None,
+    cache_dir=os.path.expanduser("~/") + ".cache",
```
webknjaz, line 63: Can this be something like
```suggestion
    cache_dir=pathlib.Path.home() / ".cache",
```
?

ssbarnea:
@provenvelocity Any chance to update this to make it pass our testing? If it gets stalling for too long we will close the pull request, which sadly would not fix the bug.

ssbarnea:
@provenvelocity Can you please fix this? I will soon close this PR if not fixed.

erikgb:
@provenvelocity Any chance you can fix the requested changes? 🙏 I have been waiting for months now to get some support for collections in `ansible-lint` and `molecule` - and this would at least get me a short step further. 😓 I have stopped using Ansible for new stuff, mainly because of constant issues like this is an example of, but have some "legacy" stuff that still needs my attention.

provenvelocity:
Sorry I had given up and got switched at work.

I will see if I can get this this weekend and also if you want to edit my
forked code please do so…

ssbarnea:
I am closing this for two reasons: it was not updated in a very long time and second, we are aiming to remove prerun module and switch to use of ansible-compat, which makes any change to it undesired. See #1784

PR #1645, opened 2021-07-05
WIP: arch packaging workflow

aminvakil:
This PR retrieves PKGBUILD from archlinux official repositories using `asp export ansible-lint` and patches it to remove its sources using latest tarball from `ansible-lint` repo and retrieves version using `python setup.py --version`.

Follow up on https://github.com/ansible-community/ansible-lint/issues/1523
Partial-Fix: #1646


PKGBUILD
```diff
@@ -0,0 +1,42 @@
+# Maintainer: Amin Vakil <info AT aminvakil DOT com>
+# Contributor: Jelle van der Waa <jelle@archlinux.org>
+# Contributor: Frederik Schwan <freswa at archlinux dot org>
+# Contributor: Sander Boom <sanderboom@gmail.com>
+
+_pkgname=ansible-lint
+pkgname=ansible-lint-test
+pkgver=5.0.7.r73.gfef55b6
```
ssbarnea, line 8: In final version we should generate this based on current version. Usually I use `python setup.py --version` in order to retrieve the version. This has the advantage of re-using the setuptools-scm versioning.

I think it would be very cool to have a `tox -e pkg-arch` command that does build the package locally as it would make it much easier for others to debug/fix it if gets broken.  Having extra logic inside GHA file does not help much as we cannot run the same commands locally easily.

ssbarnea: Alternatively you can add a playbooks that runs these commands, something similar with eco.yml which we already have.

aminvakil: `pkgver` gets replaced using `python setup.py --version` now. (https://github.com/ansible-community/ansible-lint/pull/1645/commits/e6ad474e25753c8184c2f11f4b5c9809126d1475)

aminvakil: I'm also working something like `eco.yml`, will push here when it's ready.

pabelanger:
@aminvakil @ssbarnea I would encourage you to look at https://github.com/ansible/python-builder-image/ for building python packages in containers.  We've standardize this on a lot of projects, even to the point create a suit of testing around it for zuul.

I am happy to give a quick exmaple of how to use it with ansible-lint, your dockerfile would need to look like: https://github.com/ansible/ansible-core-image/blob/main/Dockerfile

aminvakil:
@pabelanger Sorry, I should've mentioned this in PR description, changed it now. This is a follow up on https://github.com/ansible-community/ansible-lint/issues/1523.

I don't know how we can use that docker image for testing archlinux, but I can understand that this is not fine using third-party (my) docker images not owned by ansible, I will try to get it working somehow by official archlinux image.

pabelanger:
Why archlinux specifically for testing? That's something we don't use today, but base everything off centos-8 (moving to stream).

As for the testing part in other parts of ansible (like collections) we're hoping to unify it.

ssbarnea:
Yep, we really need to use official distro images, the only exception would be if Ansible team maintains some tuned images that can speedup testing.

pabelanger:
> Yep, we really need to use official distro images, the only exception would be if Ansible team maintains some tuned images that can speedup testing.

yes, we do. We have ansible-runner images published today https://quay.io/repository/ansible/ansible-runner, which could be used. These are the top layers for execution environments. Otherwise, you can use the python-base layer: https://quay.io/repository/ansible/python-base

This ways, as you say. you don't need to maintain the full stack of layers. Which is exactly how we built out all our network-ee testing in: https://github.com/ansible/network-ee

aminvakil:
> Why archlinux specifically for testing? That's something we don't use today, but base everything off centos-8 (moving to stream).

Because it has been broken once in archlinux and brought up here and @ssbarnea asked if anyone is interested in adding archlinux packaging to CI to prevent breaking it again.

I don't care for stream 8 and won't change this PR to address that OS, but if it's unnecessary I can close this PR for sure.

I'm already building ansible-lint for myself using latest commit from master branch nightly (https://github.com/aminvakil/aur/runs/2994703794) to make sure I will catch any unexpected issue before release.

ssbarnea:
@aminvakil please continue, we will do both but not in a single PR. You needed arch, do it. We will take care of others.

.github/workflows/pkg-arch.yml
```diff
@@ -0,0 +1,40 @@
...
+  # - cron: 2 18 * * 5
+
+jobs:
+  pkg-arch:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out src from Git
+      uses: actions/checkout@v2
+
+    - name: Test on official archlinux.
```
webknjaz, line 27:
```suggestion
    - name: Test on official archlinux
```

pkg-arch.patch
```diff
@@ -0,0 +1,39 @@
...
+ build() {
+-  cd ${pkgname}-${pkgver}
++  cd ..
+   python setup.py build
+ }
+
+ package() {
+-  cd ${pkgname}-${pkgver}
++  cd ..
+   PYTHONHASHSEED=0 python setup.py install --root="${pkgdir}" --optimize=1
```
webknjaz, line 37: Invoking setuptools like this is unrecommended. You should use PEP517 instead. But I wouldn't recommend building sdist from Git. It's better to have it built in a previous job that everything else would depend on. Then, this one should download that artifact and work with it. This way, you'll get to test the thing that actually gets uploaded to PyPI (Git repo may have extra files that won't be included, for example).

pkg-arch.patch
```diff
@@ -0,0 +1,39 @@
...
+-        'a090862e657bfc29c89b23b2ecc8c585e15f098310b997e39600b6bdf0cd433e1045e2da348ce27dc2d3c0f9ef01950cf0ec5e7b071bc43d85161faabd8a1f60')
+
+-prepare() {
+-  cd ${pkgname}-${pkgver}
+-  # stop setuptools from using the scm version which doesn't exist in this context
+-  patch -Np1 < "${srcdir}"/disable_use_scm_version.patch
+-  sed -i "/^\[metadata\]/a version = ${pkgver}" setup.cfg
++pkgver() {
++  cd ..
++  python setup.py --version
```
webknjaz, line 25: Instead of doing this, I'd recommend using an sdist as mentioned here https://github.com/ansible-community/ansible-lint/pull/1645/files#r665175286. It already contains version in the metadata.

webknjaz (commented):
Here's an example with RPMs using an sdist pre-built in a previous job for inspiration: https://github.com/ansible/pylibssh/blob/6de4512/.github/workflows/build-test-n-publish.yml#L679-L863

aminvakil:
cc @Scimmia22 @freswa

pkg-arch.patch
aminvakil: I'm getting confused about what we're trying to test here, this is the official way following best practices of Arch Packaging Guidelines that it's actually getting used in archlinux, I'm not sure what's the purpose of this test if the way package gets built here differs from the build that actually happens for the package in archlinux.

PKGBUILD
aminvakil: `python setup.py --version` or running
```
import setuptools_scm
print (setuptools_scm.get_version())
```
as @webknjaz mentioned (https://github.com/ansible/pylibssh/blob/6de4512/.github/workflows/build-test-n-publish.yml#L141-L148)
is returning 5.0.8.dev98+g04eb0f4 instead of 5.1.0... something.
I guess that's because I've cloned the repository when 5.0.8 was about to come out or something like that.
What should I do about this?

ssbarnea: You are right, if you forget to pull last tags you endup with various versions, is not really a bug. I encounter this issue quite often.

I think that we should better use pre-build sdist or wheel and rely on the version from within them. We can assume that a distribution package is produced from these and not directly from the cloned code and avoid the versioning issue. This should also allow you to drop setuptools-scm from being a dependency for the distro package.

webknjaz: Exactly. Dirty versions are also produced when you mutate some files on disk (which could be ignored with `git update-index --assume-unchanged` in CI).
But yes, sdist is usually considered a canonical source from which you're able to build anything you want.

aminvakil: Very well, I need a couple of days to get familiar with the stuff you're talking about :) I haven't done it before.
@webknjaz Thanks for the links!

webknjaz: By the way, you could get the same SCM output by just using `python -m setuptools_scm` but that's not what you need when using the sdist metadata version anyway.

pkg-arch.patch
webknjaz: Yep, sdist is the best source for OS-level packaging ecosystems. Fedora, for example, uses sdists as a source and has a set of macros to build it using PEP517. I know most of the distros do something similar.

As for `setup.py`, consider it being there by accident, it may as well be missing. You must use a PEP517 compliant tool which would determine what hooks to run to build a project (it could be something other than setuptools — flit, poetry, a totally custom in-tree build backend etc.)
`setup.py` only exists in this project to accommodate editable installs in old pip versions. I'm hearing that it's no longer necessary in the latest versions so at some point we may consider getting rid of it (because it's becoming just a junk file + external packagers abuse it and hardcode the build backend when they shouldn't).

felixfontein:
Citing @aminvakil from abd1f92d3bfa9020405057b2a04421b5ef4c28a0:

> Because of ansible-collections/community.general@518ace2, I cannot comment on this PR, so here it goes. I will not continue working on this for obvious reasons, but feel free to use this PR if you want and close this.

PR #1648, opened 2021-07-08
Use packit to produce rpms for fedora-35

ssbarnea:


ansible-lint.spec
```diff
@@ -0,0 +1,82 @@
...
+
+%description
+TBD Ansible-lint description
+
+%prep
+%autosetup
+
+
+%build
+%py3_build
```
webknjaz, line 49: For Fedora it's better to use https://src.fedoraproject.org/rpms/pyproject-rpm-macros

TomasTomecek:
Did you make any progress on the rpm packaging in the meantime? Is there anything we can help with?

ssbarnea:
No progress as I got other priorities, not sure when I will be able to return to this one. Feel free to work on it.

TomasTomecek:
I can see that all builds on Fedora and el8 are passing: what's left to be done?

.packit.yaml
```diff
@@ -0,0 +1,29 @@
+specfile_path: ansible-lint.spec
+actions:
+  create-archive:
+    - "python3 setup.py sdist --dist-dir ."
```
webknjaz, line 4: Let's not call this directly. If we let some envs bypass PEP 517, it just adds more maintenance burden in the end. Also, you basically don't run the same thing that is checked in the CI potentially making the output different w/o realizing that.
```suggestion
    - python3 -m pip install build --user
    - python3 -m build --sdist --outdir .
```

.packit.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - "sh -c 'echo ansible-lint-$(python3 setup.py --version).tar.gz'"
```
webknjaz, line 5: @TomasTomecek would something like this work?
```suggestion
    - "sh -c 'echo ${PACKIT_PROJECT_NAME_VERSION}.tar.gz'"
```
Is this var populated from `get-current-version`? (The docs imply that it's only coming from `git-describe` which I want to clarify)

.packit.yaml
```diff
@@ -0,0 +1,29 @@
...
+  get-current-version:
+    - "python3 setup.py --version"
+  post-upstream-clone:
+    - |
+      bash -c "sed -i -e 's/Version: VERSION_PLACEHOLDER/Version: $(python3 setup.py --version)/' ansible-lint.spec"
+jobs:
+  - job: copr_build
+    metadata:
+      targets:
+        # Unable to build centos-stream due to missing python3-setuptools_scm_git_archive
```
webknjaz, line 15: Why not patch it out in the spec file? It's not like there's nowhere to get the version from.

.packit.yaml
```diff
@@ -0,0 +1,29 @@
+specfile_path: ansible-lint.spec
+actions:
+  create-archive:
+    - "python3 setup.py sdist --dist-dir ."
+    - "sh -c 'echo ansible-lint-$(python3 setup.py --version).tar.gz'"
+  get-current-version:
+    - "python3 setup.py --version"
```
webknjaz, line 7: Can this be replaced with something like
```suggestion
    - python3 -m setuptools_scm
```
?

.packit.yaml
```diff
@@ -0,0 +1,29 @@
...
+  post-upstream-clone:
+    - |
+      bash -c "sed -i -e 's/Version: VERSION_PLACEHOLDER/Version: $(python3 setup.py --version)/' ansible-lint.spec"
```
webknjaz, line 10:
```suggestion
      bash -c "sed -i -e 's/Version: VERSION_PLACEHOLDER/Version: $(python3 -m setuptools_scm | awk '{print $3}')/' ansible-lint.spec"
```

.packit.yaml
```diff
@@ -0,0 +1,29 @@
+specfile_path: ansible-lint.spec
+actions:
+  create-archive:
+    - "python3 setup.py sdist --dist-dir ."
+    - "sh -c 'echo ansible-lint-$(python3 setup.py --version).tar.gz'"
+  get-current-version:
+    - "python3 setup.py --version"
+  post-upstream-clone:
```
webknjaz, line 8: @ssbarnea _I think_ this is the wrong hook to be doing this. There's `fix-spec-file` that seems to be a better fit.
```suggestion
  fix-spec-file:
```

.packit.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - |
+      bash -c "sed -i -e 's/Version: VERSION_PLACEHOLDER/Version: $(python3 setup.py --version)/' ansible-lint.spec"
+jobs:
```
webknjaz, line 11: If we modify files in the repo, this will confuse setuptools-scm. So it's best to pretend like nothing happened:
```suggestion
    - git update-index --assume-unchanged ansible-lint.spec
jobs:
```

.packit.yaml
TomasTomecek: yup, you're correct, the content of the env var is by default coming from `git describe` but you can override with the `get-current-version` action

.packit.yaml
TomasTomecek: `fix-spec-file` action also changes the %autosetup line to make sure the tarball is unpacked and accessed properly which you may not need: it should be fine to use `post-upstream-clone` action for that, depends what works best for you

.config/ansible-lint.spec
```diff
@@ -0,0 +1,87 @@
...
+
+%files
+%{python3_sitelib}/ansiblelint/
+%{python3_sitelib}/ansible_lint-*.dist-info/
+%{_bindir}/ansible-lint
+%license COPYING
+%doc docs/* README.rst
+
+%changelog
+Available at https://github.com/ansible-community/ansible-lint/releases
```
TomasTomecek, line 87: There is an expected format here so this should read:
```
%changelog
* Tue Mar 22 2022 Name <email@asd.xyz> - 1.2.3-1
  - Available at https://github.com/ansible-community/ansible-lint/releases
```
https://rpm-packaging-guide.github.io/

ssbarnea: Is the conformant format going to break something or not? If not I would keep it as now and defer a fix for a follow-up, especially as it would involve scripting to convert the markdown from github releases into the format listed above.

priyamsahoo (approved):
LGTM !

.config/ansible-lint.spec
ssbarnea: I will do something about the changelog in follow-up.

PR #1649, opened 2021-07-08
Update .pre-commit-config.yaml pylint repo

clickthisnick:
According to https://github.com/pre-commit/mirrors-pylint readme `This mirror repository is deprecated, use pylint directly.`

The result is the mirrors-pylint repo is missing a bunch of tags that pylint has.


ssbarnea:
@clickthisnick That is failing due to other issues which are not easy to fix at all. Using the mirror is not really such a big problem, at least at his moment. Unless you can make it pass, I will endup closing this change.

clickthisnick:
@ssbarnea Disabling the single rule in the test should allow this to move forward which makes the latest stable pylint versions available but feel free to close, not a big deal atm as you said either way

test/local-content/collections/ansible_collections/testns/testcoll/plugins/modules/test_module_2.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+# pylint: disable=cyclic-import
```
webknjaz, line 2: This should have a note pointing to the upstream issue as the motivation for adding an ignore

ssbarnea: @clickthisnick Can you please address that? If not I will close it because there are not so many benefits from doing this location swap.

PR #1655, opened 2021-07-11
Refactor use of pip-compile

ssbarnea:
* rename `test-requirements.txt` to `constraints.txt` to better describe its purpose
* move test dependencies into test extra so rule authors can easily install them
* inside tox.ini make use of `PIP_CONSTRAINT` variable for pinning dependencies instead of `-c` pip argument. This has the advantage of reaching all subprocesses, like pre-commit. Also this avoids a bug where tox fails to run if deps key contains only constraints. On `tox -e deps` we explicitly set the var to /dev/null in order to allow job to upgrade all deps.
* some minor fixes on pip-compile arguments and file matchers

This change does affect only testing of ansible-test, it should not affect the use of the library in any way.


.github/workflows/tox.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-        - 3.8
+        - 3.9
```
webknjaz, line 29: This should probably be submitted separately.

setup.cfg
```diff
@@ -103,6 +103,12 @@ core =
 # yamllint must remain optional
 yamllint =
   yamllint >= 1.25.0  # GPLv3
+test =
```
webknjaz, line 106: I don't think these belong in extras. It's absolutely useless for the end-users while being exposed as a public API for installing loosened test deps.

.github/workflows/tox.yml
ssbarnea: Yep, I can do this switch before the rest, in one that also updates pip-compile version. It will make the history easier to read.

tox.ini
```diff
@@ -67,17 +66,16 @@ whitelist_externals =
...
-  pip-tools>=5.4.0
+  pip-tools>=6.2.0
   setuptools>=51.1.1
 skip_install = true
 commands =
-  # --no-annotate is used in order to prevent conflicts with PRs produced by dependency-bot
-  pip-compile --no-annotate --output-file=test-requirements.txt setup.py test-requirements.in
-  pip-compile --no-annotate --output-file=docs/requirements.txt setup.py docs/requirements.in
-  {envpython} -m pre_commit run {posargs:--all-files --hook-stage manual -v}
+  {envpython} -m pre_commit run {posargs:--all-files --show-diff-on-failure}
```
webknjaz, line 78: The extra option is useful but it doesn't feel like it belongs in this PR.

tox.ini
```diff
@@ -87,21 +85,21 @@ description = Bump all test depeendencies
...
+  {envpython} -m pre_commit run {posargs:--all-files --show-diff-on-failure
+ --hook-stage manual}
 
 [testenv:docs]
 description = Builds docs
 basepython = python3
 deps =
-  -r{toxinidir}/docs/requirements.in
-  -c{toxinidir}/docs/requirements.txt
+  -r{toxinidir}/docs/requirements.txt
```
webknjaz, line 102: Having unpinned docs deps isn't the brightest idea: this hurts reproducibility/predictability.

constraints.txt
```diff
@@ -0,0 +1,168 @@
+#
+# This file is autogenerated by pip-compile with python 3.9
+# To update, run:
+#
+#    pip-compile --extra=test --extra=yamllint --output-file=constraints.txt docs/requirements.txt setup.cfg
```
webknjaz, line 5: Mixing the docs pins up with the test ones doesn't seem right. Those envs are unrelated and must be kept strictly separate.

webknjaz (changes requested):
I don't think that it's a good idea to couple things like this. Strong :-1: from me.

tox.ini
ssbarnea: That is not removing it, we use PIP_CONSTRAINT env var to achieve the same locking, is defined inside root test env. I know is not obvious but is more reliable as it does apply everywhere including on pre-commit and on build stage when installing packages, something that `-c` is not able to achieve. Also there is another special case where I found -c not to be working: if you have an empty `deps =`  with only -c, tox will fail to run, but when using PIP_CONSTRAINT it will work (you can guess why).

PR #1659, opened 2021-07-12
Run pip-compile as a hook

ssbarnea:
- simplify execution logic of pip-compile
- remove pip-compile duplication between lint and deps tox envs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,6 +18,32 @@ repos:
     language: system
     files: >-
       ^setup\.py$
+  - id: pip-compile
+    name: pip-compile
+    entry: pip-compile -q --no-annotate --output-file=test-requirements.txt setup.py test-requirements.in
+    language: system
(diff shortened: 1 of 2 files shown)


webknjaz (approved):

.pre-commit-config.yaml
```diff
@@ -18,6 +18,32 @@ repos:
...
+    files: ^(test-requirements\.(txt|in)|)$
+    pass_filenames: false
+    stages: [manual]
+  - id: pip-compile-docs-upgrade
+    name: pip-compile-docs-upgrade
+    entry: pip-compile -q --upgrade --no-annotate --output-file=docs/requirements.txt docs/requirements.in setup.py
+    language: system
+    files: ^(docs\/requirements\.(txt|in)|)$
+    pass_filenames: false
+    stages: [manual]
```
ssbarnea, line 46: @webknjaz I am not really proud of the duplication here but I was not able to find a better way so far. If you ever find a better approach let me know.

PR #1660, opened 2021-07-12
fix: deterministic behaviour of cache installation in prerun

### LINKED ISSUE #1654: Current working directory affects execution environment generated on prerun
##### Summary

Not sure if this should be a bug, a proposal or even if it is the intended behaviour.

When running on a collection repository, `ansible-lint` caches it on a `~/.cache/ansible-lint/<HASH>/collections/ansible_collections` directory as part of the `prerun` process. When being launched on a directory other than the root one, such as a role, it does not detect that the role is part of a collection, and only caches the role directory, not the collection one.

I think that this behaviour is wrong because `ansible-lint` always tries to find the project root directory and work from there, except during the `prerun`. In fact, the hash is generated based on the root directory even when being launched on a subdirectory. After #1483, `prerun` was made public so other packages can use it, and this can cause issues to those packages.

I already have a PR ready to be submitted in case this is accepted.

##### Issue Type

- Bug Report

```console (paste below)
$ ansible --version
ansible [core 2.11.2]
  config file = None
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/.pyenv/versions/3.9.1/envs/ansible-tests/lib/python3.9/site-packages/ansible
  ansible collection location = ~/.ansible/collections:/usr/share/ansible/collections
  executable location = ~/.pyenv/versions/ansible-tests/bin/ansible
  python version = 3.9.1 (default, Feb 21 2021, 17:37:26) [Clang 12.0.0 (clang-1200.0.31.1)]
  jinja version = 3.0.1
  libyaml = True

$ ansible-lint --version
ansible-lint 5.1.0a1 using ansible 2.11.2
```

##### OS / ENVIRONMENT

Can be reproduced both in `ansible-lint` `5.0.12` and `5.1.0a1`.

##### STEPS TO REPRODUCE

```console (paste below)
$ ansible-galaxy collection init xabinapal.test
$ cd xabinapal/test
$ touch .ansible-lint

$ cd roles
$ ansible-galaxy role init test

$ cd ..
$ git init
$ git add -A
```

Linting from the repository root detects it as a collection (as it finds a `galaxy.yml` file) and sets the `ANSIBLE_COLLECTIONS_PATH` environment variable:
```console (paste below)
$ ansible-lint -v
INFO     Added ANSIBLE_COLLECTIONS_PATH=~/.cache/ansible-lint/b9d075/collections
INFO     Added ANSIBLE_ROLES_PATH=~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:roles:~/.cache/ansible-lint/b9d075/roles
INFO     Added ANSIBLE_COLLECTIONS_PATH=~/.cache/ansible-lint/b9d075/collections:~/.cache/ansible-lint/b9d075/collections
INFO     Added ANSIBLE_ROLES_PATH=~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:roles:~/.cache/ansible-lint/b9d075/roles:roles:~/.cache/ansible-lint/b9d075/roles
```

Executing in a role directory (or, any other directory other than the root) also includes the `ANSIBLE_COLLECTIONS_PATH` environment variable, as the cached directory already contains the `collections/ansible_collections` directory:
```console (paste below)
$ cd roles/test
$ ansible-lint -v
INFO     Using ~/.cache/ansible-lint/b9d075/roles/test symlink to current repository in order to enable Ansible to find the role using its expected full name.
INFO     Added ANSIBLE_COLLECTIONS_PATH=~/.cache/ansible-lint/b9d075/collections
INFO     Added ANSIBLE_ROLES_PATH=~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:~/.cache/ansible-lint/b9d075/roles
```

After removing the cache, running from a diferent directory does not treat the repository as a collection one:
```console (paste below)
$ rm -rf ~/.cache/ansible-lint/b9d075

$ cd roles/test
$ ansible-lint -v
INFO     Using ~/.cache/ansible-lint/b9d075/roles/test symlink to current repository in order to enable Ansible to find the role using its expected full name.
INFO     Added ANSIBLE_ROLES_PATH=~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:~/.cache/ansible-lint/b9d075/roles:~/.cache/ansible-lint/b9d075/roles
```

##### Desired Behaviour

Behaviour should be deterministic instead of depending on execution path and cache contents. It should determine the repository type based on where the `.ansible-lint` file is located and not on the current directory.

##### Actual Behaviour

Follow the steps to reproduce. Current working directory and previous executions affect the environment of the `prerun` process.

xabinapal:
Ensure cache directory and environment are populated from `project_dir` and not from current working directory in `prerun` process.

Fixes: #1654

### DIFF
--- a/src/ansiblelint/prerun.py
+++ b/src/ansiblelint/prerun.py
@@ -263,7 +263,7 @@ def _symlink_galaxy_install(link_path: pathlib.Path) -> None:
 
 def _install_galaxy_role() -> None:
     """Detect standalone galaxy role and installs it."""
-    galaxy_path = "%s/meta/main.yml" % options.project_dir
+    galaxy_path = f"{options.project_dir}/meta/main.yml"
     if not os.path.exists(galaxy_path):
         return
     yaml = yaml_from_file(galaxy_path)
@@ -274,9 +274,8 @@ def _install_galaxy_role() -> None:
 
     if 'role-name' not in options.skip_list:
         if not re.match(r"[a-z0-9][a-z0-9_]+\.[a-z][a-z0-9_]+$", fqrn):
-            msg = (
-                """\
-Computed fully qualified role name of %s does not follow current galaxy requirements.
+            msg = f"""
+Computed fully qualified role name of {fqrn} does not follow current galaxy requirements.
 Please edit meta/main.yml and assure we can correctly determine full role name:
 
 galaxy_info:
@@ -288,8 +287,6 @@ Role: https://galaxy.ansible.com/docs/contributing/creating_role.html#role-names
 
 As an alternative, you can add 'role-name' to either skip_list or warn_list.
 """
-                % fqrn
-            )
             if 'role-name' in options.warn_list:
                 _logger.warning(msg)
             else:
@@ -315,7 +312,7 @@ As an alternative, you can add 'role-name' to either skip_list or warn_list.
 
 def _install_galaxy_collection() -> None:
     """Detect standalone galaxy collection and installs it."""
-    galaxy_path = "%s/galaxy.yml" % options.project_dir
+    galaxy_path = f"{options.project_dir}/galaxy.yml"
     if not os.path.exists(galaxy_path):
         return
     yaml = yaml_from_file(galaxy_path)
--- a/src/ansiblelint/prerun.py
+++ b/src/ansiblelint/prerun.py
@@ -334,10 +334,10 @@ def _prepare_ansible_paths() -> None:
     roles_path: List[str] = []
 
     for path_list, path in (
-        (library_paths, "plugins/modules"),
+        (library_paths, f"{options.project_dir}/plugins/modules"),
         (library_paths, f"{options.cache_dir}/modules"),
         (collection_list, f"{options.cache_dir}/collections"),
-        (roles_path, "roles"),
+        (roles_path, f"{options.project_dir}/roles"),
         (roles_path, f"{options.cache_dir}/roles"),
     ):
         if path not in path_list and os.path.exists(path):
--- a/src/ansiblelint/prerun.py
+++ b/src/ansiblelint/prerun.py
@@ -258,9 +258,10 @@ def _symlink_galaxy_install(link_path) -> None:
 
 def _install_galaxy_role() -> None:
     """Detect standalone galaxy role and installs it."""
-    if not os.path.exists("meta/main.yml"):
+    galaxy_path = "%s/meta/main.yml" % options.project_dir
+    if not os.path.exists(galaxy_path):
         return
-    yaml = yaml_from_file("meta/main.yml")
+    yaml = yaml_from_file(galaxy_path)
     if 'galaxy_info' not in yaml:
         return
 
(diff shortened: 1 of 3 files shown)


ssbarnea:
@xabinapal I see you have two open PRs that seem to address the same bug. Can you close one?

xabinapal:
@ssbarnea I see them as two different issues. This one fixes `prerun` not using the `project_dir` option and #1661 fixes the guessing of the `project_dir` (not only during `prerun`) to fallback to CWD. If you prefer a single PR, I can add that commit to this one.

ssbarnea:
There are several problems with this change:

* the symlink utility function added had no docstring explaining its purpose and even after reading its content I find its signature at least weird as it does receive only one argument and it inferst the other one from tool options, that does not go well with the reusability of the prerun module, which is reused by molecule.
* there is no real need to pass custom directory location to ansible-galaxy install command because that command will respect what is configured in ansible configured collection/role paths. Probably you seen that we alter these variables in order to inject our custom cache location at the beginning of the path. That ensures that ansible-galaxy command will install them at out desired location, without us having to pass it as an argument.

Keep in mind that we should avoid using `options` inside prerun module. Instead we should pass things like project_dir on each method. That is because there is no config when used in molecule.

xabinapal:
The symlink function was a refactor of already existing code that was duplicated twice, just to enforce DRY: https://github.com/ansible-community/ansible-lint/blob/master/src/ansiblelint/prerun.py#L409-L412 and https://github.com/ansible-community/ansible-lint/blob/master/src/ansiblelint/prerun.py#L295-L298

I changed it a little bit to check if `link_path` is not a symlink and raise a friendlier error than the default exception. Admittedly, giving it a second thought, it could be more readable with EAFP. Also, I forgot to add the docstring, thanks for noticing. Refactoring this function could be like this:

```python
def _symlink_galaxy_install(link_path: pathlib.Path) -> None:
    """Helper function to symlink project directory in a custom location"""
    target = str(pathlib.Path(options.project_dir).absolute())
    try:
        if os.readlink(str(link_path.absolute())) != target:
            link_path.unlink()
        else:
            return
    except FileNotFoundError:
        pass
    except OSError:
        raise RuntimeError(
            "Can't replace non-symlink path '%s'" % link_path.absolute()
        ) from None
    link_path.symlink_to(target, target_is_directory=True)
```

Regarding the usage of `options` inside this function, I agree that it’s not the cleanest way of doing things. I noticed that `options.project_dir` and `options.cache_dir` are used extensively throughout the entire file and not being passed as arguments, so I just followed that pattern. If required, I could refactor this function adding a second parameter to it or, if required, I could try to refactor the entire file.

On the other hand, I don’t really understand what you mean with the second problem. I’ve not changed neither the `install_collection` function nor the`install_requirement` one, the only ones that launch `ansible-galaxy install`.

ssbarnea:
I need to close this because it is in direct conflict with #1666 as we are moving all the bootstraping of the testing environment out of the linter, so we can use it with molecule and also to enable use of execution environments. If the problem is still valid after we make this change, we will have to alter [ansible-compat](https://github.com/ansible-community/ansible-compat) library.

PR #1666, opened 2021-07-14
Adopt ansible-compat runtime

ssbarnea:
As part of decoupling molecule from ansible-lint, the functionality from ansiblelint.prerun was moved in a standalone generic library named ansible-compat.

This change is marked as major because it will break compatibility for anyone that relied on use of prerun module from the linter. The effective removal of prerun submodule will be done
in a follow-up, just to allow for an easier review.

Fixes: #1712
Fixes: #1784

### DIFF
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -60,6 +60,7 @@ options = Namespace(
     cwd=".",
     display_relative_path=True,
     exclude_paths=[],
+    format="rich",
     lintables=[],
     listrules=False,
     listtags=False,
--- a/src/ansiblelint/testing/__init__.py
+++ b/src/ansiblelint/testing/__init__.py
@@ -7,7 +7,8 @@ import sys
 import tempfile
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
-from ansiblelint.prerun import prepare_environment
+from ansiblelint.app import get_app
+from ansiblelint.rules import RulesCollection
 
 if TYPE_CHECKING:
     # https://github.com/PyCQA/pylint/issues/3240
@@ -18,11 +19,10 @@ else:
 
 # Emulate command line execution initialization as without it Ansible module
 # would be loaded with incomplete module/role/collection list.
-prepare_environment()
+app = get_app()
 
 # pylint: disable=wrong-import-position
 from ansiblelint.errors import MatchError  # noqa: E402
-from ansiblelint.rules import RulesCollection  # noqa: E402
 from ansiblelint.runner import Runner  # noqa: E402
 
 
--- a/src/ansiblelint/__init__.py
+++ b/src/ansiblelint/__init__.py
@@ -18,7 +18,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 """Main ansible-lint package."""
-# prerun must run before any other imports
 from ansiblelint.version import __version__
 
 __all__ = ("__version__",)
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -35,12 +35,11 @@ from ansible_compat.config import ansible_version
 from enrich.console import should_do_markup
 
 from ansiblelint import cli
-from ansiblelint.app import App
+from ansiblelint.app import get_app
 from ansiblelint.color import console, console_options, reconfigure, render_yaml
 from ansiblelint.config import options
 from ansiblelint.constants import EXIT_CONTROL_C_RC
 from ansiblelint.file_utils import abspath, cwd, normpath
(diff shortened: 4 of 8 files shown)


setup.cfg
```diff
@@ -71,6 +71,7 @@ install_requires =
   # ansible-lint does not list ansible as direct dependency in order to
   # allow user to choose his target version. Be sure that you mention one
   # of the extras or install a specific version of ansible yourself.
+  ansible-compat>=0.3.0
```
felixfontein, line 74: I really think we should stick to a stable version, and not an initial development version which can easily have breaking changes.
```suggestion
  ansible-compat>=1.0.0,<2.0.0
```

felixfontein (changes requested):

setup.cfg
webknjaz: For 0.x it's usually capped with the next minor:
```suggestion
  ansible-compat >=0.3.0, <0.4.0
```
No need to force a major release just yet — we have a perfect opportunity to iterate on the API and locking it up at this point is a bad idea.
Having the same people in charge of the both projects makes it rather safe. OTOH it may be reasonable to have some downstream tests integrated into the compat's CI to be on the safe side.
Also, zerover does not mean development. Dev versions have suffices like `.dev0`. 0ver just means that the project is not yet fully settled on the API which is not necessarily a bad thing.

webknjaz:
Do you want to add `Closes #1712`?

setup.cfg
```diff
@@ -71,6 +71,7 @@ install_requires =
   # ansible-lint does not list ansible as direct dependency in order to
   # allow user to choose his target version. Be sure that you mention one
   # of the extras or install a specific version of ansible yourself.
+  ansible-compat >=0.3.0, <0.4.0
```
webknjaz, line 74: Let's try this, it's equivalent to the dep spec above but that is semantically closer to SemVer expectations:
```suggestion
  ansible-compat ~= 0.3.0
```

ssbarnea: In fact I will completely remove the capping and we will add it only when we will break the API with compat. We know very well who are to be the consumers. That is because capping it too soon it would require extra maintenance for us.

src/ansiblelint/testing/__init__.py
```diff
@@ -18,7 +18,8 @@
 
 # Emulate command line execution initialization as without it Ansible module
 # would be loaded with incomplete module/role/collection list.
-prepare_environment()
+runtime = Runtime(isolated=True)
+runtime.prepare_environment()
```
webknjaz, line 22: I don't really like this pattern of initializing a single behaviour with two separate things/invocations which must always be one after another with no code causing side-effects in between.

ssbarnea: You will have to get used to it because I remember @tadeboro suggesting months ago that doing extra optional stuff inside constructor was a bad idea. The two step process is there by design and not accident, so we have a simple constructor.

src/ansiblelint/rules/__init__.py
```diff
@@ -39,6 +40,10 @@
 
 
 class AnsibleLintRule(BaseRule):
+    """Based class used by most rules."""
```
cognifloyd, line 43:
```suggestion
    """Base class used by most rules."""
```

src/ansiblelint/config.py
```diff
@@ -60,6 +60,7 @@
     cwd=".",
     display_relative_path=True,
     exclude_paths=[],
+    format="rich",
```
cognifloyd, line 63: What does this have to do with removing prerun?

src/ansiblelint/testing/__init__.py
```diff
@@ -7,7 +7,10 @@
 import tempfile
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
-from ansiblelint.prerun import prepare_environment
+from ansible_compat.runtime import Runtime
```
cognifloyd, line 10: Unused import?

src/ansiblelint/testing/__init__.py
```diff
@@ -7,7 +7,10 @@
...
+
+from ansiblelint.app import get_app
+from ansiblelint.rules import RulesCollection  # noqa: E402
```
cognifloyd, line 13: I believe you can remove the noqa comment since this moved up, right?
```suggestion
from ansiblelint.rules import RulesCollection
```

cognifloyd (commented):
Looking awesome! Just a few nitpicky comments/questions.

src/ansiblelint/config.py
ssbarnea: It is needed, otherwise I will get a runtime errors when I try to access .format in main. In fact I am not sure how that it worked before this but I suspect it has something to do on config initialization order.

cognifloyd (approved):

PR #1698, opened 2021-08-11
Continuously check for the broken links across the project

webknjaz:


webknjaz:
It's a separate check, I don't want it to confuse the build docs output.

ssbarnea:
Nobody reads the output unless something does wrong and tox will stop at the first command that fails, so you will not get extra output.

The reasons why I am advocating for avoiding creation of extra test environment and jobs:
* extra CI load and delays due to increase in number of nodes used
* harder for an individual to run tests locally (tox -e docs is bit of universal), with this building docs would require two operations instead of one.
* scalability: what if we would have a different tox environment for each linter we use? how easy it would be to run testing? I know few projects that did this, not fun.

For this reason I prefer to chain thing-tests to bigger ones, this one in particular could easily be included in either linting or docs building.

webknjaz:
In general, GHA (as a platform) is better at orchestrating the jobs that run on it. If you stick too much into the same job, you break GH's ability to schedule runs in a more optimal manner. Also, it's not visible in the list of checks and nobody knows what runs and succeeds.
This is why I'm strongly against piling many distinct tasks into the same env.

ssbarnea:
> In general, GHA (as a platform) is better at orchestrating the jobs that run on it. If you stick too much into the same job, you break GH's ability to schedule runs in a more optimal manner. Also, it's not visible in the list of checks and nobody knows what runs and succeeds. It's hard to know/introspect what actually got checked other than "something is green here, who knows if anything got skipped, though".
> This is why I'm strongly against piling many distinct tasks into the same env.

How about a compromise for GHA? `TOXENV: docs,docs-links` and we can avoid adding new job while still having separated tox envs. Also list it inside `envlist`  so those calling just `tox` will run it.

webknjaz:
> How about a compromise for GHA? `TOXENV: docs,docs-links` and we can avoid adding new job while still having separated tox envs. Also list it inside `envlist` so those calling just `tox` will run it.

But that's exactly what I don't like — separate check results should show up distinctively.

ssbarnea (approved):
It the job proves problematic I will know who to look for! ;)

PR #1699, opened 2021-08-11
Use pypa/build to produce dists in tox

webknjaz:


tox.ini
```diff
@@ -134,7 +134,7 @@ basepython = python3
 description =
   Build package, verify metadata, install package and assert behavior when ansible is missing.
 deps =
-  pep517 >= 0.7.0
+  build >= 0.6.0, < 0.7.0
```
ssbarnea, line 137: I would personally do not bother to put a cap there. In fact I would prefer to findout if newer versions breaks CI instead of having someone raise a bug for it.

tox.ini
```diff
@@ -145,10 +145,8 @@ commands =
   os.path.isdir(dist_dir) or sys.exit(0); \
   print("Removing \{!s\} contents...".format(dist_dir), file=sys.stderr); \
   shutil.rmtree(dist_dir)'
-  {envpython} -m pep517.build \
-    --source \
-    --binary \
-    --out-dir {toxinidir}/dist/ \
+  {envpython} -m build \
```
ssbarnea, line 148: Thanks! Yet another project where I missed to upgrade the pkg building.

tox.ini
webknjaz: For now, I prefer stability here. There are enough end-users of build to report issues upstream.

Issue #1681, opened 2021-08-01
no-changed-when: improve documentation and testing


Pant:
##### Summary

When using the `command:` module as in the example bellow, ansible-lint gives error `Commands should not change things if nothing needs doing`

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.11.3] 
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/pant/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /home/pant/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.5 (default, May 24 2021, 12:50:35) [GCC 11.1.0]
  jinja version = 2.11.3
  libyaml = True

$ ansible-lint --version
ansible-lint 5.1.2 using ansible 2.11.3
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Linux Anonymous 5.12.13-arch1-1 - x86_64 GNU/Linux

##### STEPS TO REPRODUCE

Example (from https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html):

```
- name: Safely use templated variable to run command. Always use the quote filter to avoid injection issues
  ansible.builtin.command: cat {{ myfile|quote }}
  register: myoutput
```

Problem message: 

```
no-changed-when: Commands should not change things if nothing needs doing
```

<!--- ```console (paste below)

```
-->
<!--- HINT: You can paste gist.github.com links for larger files -->

##### Desired Behaviour
<!--- Describe what you expected to happen when running the steps above -->

There should not be an problem running a command with the `register` option.

##### Actual Behaviour
<!--- Describe what actually happened. If possible run with extra verbosity (-vvvv) -->

Problem occurs when running a command with the `register` option.

---

Also there are cases where a command needs to be run but `creates` files are uknown. For example:

```
- name: Install composer dependencies
  command: composer install -d /var/www/attendize
  args: 
    creates: /var/www/attendize/????unknown_files????
```

Shouldn't commands like these (in both cases) be passed from ansible-lint without lint errors/warnings?



> Note 1: In example, directory `/var/www/attendize/` already exists and is not empty.

[...]

PR #1706, opened 2021-08-25
no-changed-when: improve testing and documentation

konstruktoid:
- adds tests to the `no-changed-when` rule
- improve documentation

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>
Fixes: #1681

### DIFF
--- a/src/ansiblelint/rules/CommandHasChangesCheckRule.py
+++ b/src/ansiblelint/rules/CommandHasChangesCheckRule.py
@@ -32,12 +32,30 @@ if TYPE_CHECKING:
 class CommandHasChangesCheckRule(AnsibleLintRule):
     id = 'no-changed-when'
     shortdesc = 'Commands should not change things if nothing needs doing'
-    description = (
-        'Commands should either read information (and thus set '
-        '``changed_when``) or not do something if it has already been '
-        'done (using creates/removes) or only do it if another '
-        'check has a particular result (``when``)'
-    )
+    description = """
+Commands should either read information (and thus set
+``changed_when``) or not do something if it has already been
+done (using ``creates`` or ``removes`` argument) or only do it if another
+check has a particular result (``when``).
+
+An example where the ``shell`` output is registered and the return code
+is used to define when the task has changed.
+
+.. code:: yaml
+
+    - name: handle shell output with return code
+      ansible.builtin.shell: cat {{ myfile|quote }}
+      register: myoutput
+      changed_when: myoutput.rc != 0
+
+The following example will trigger the rule since the task does not
+handle the output of the ``command``.
+
+.. code:: yaml
+
+    - name: does not handle any output or return codes
+      ansible.builtin.command: cat {{ myfile|quote }}
+    """
     severity = 'HIGH'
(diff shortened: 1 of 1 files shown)


src/ansiblelint/rules/CommandHasChangesCheckRule.py
```diff
@@ -31,12 +32,30 @@
...
+done (using ``creates`` or ``removes`` argument) or only do it if another
+check has a particular result (``when``).
+
+For example, this task registers the ``shell`` output and uses the return code
+to define when the task has changed.
+
+.. code:: yaml
+
+    - name: handle shell output with return code
+      ansible.builtin.shell: cat {{ myfile|quote }}
```
ssbarnea, line 47: This needs quotes, lets avoid examples that would not pass linting.

src/ansiblelint/rules/CommandHasChangesCheckRule.py
```diff
@@ -31,12 +32,30 @@
...
+      register: myoutput
+      changed_when: myoutput.rc != 0
+
+The following example will trigger the rule since the task does not
+handle the output of the ``command``.
+
+.. code:: yaml
+
+    - name: does not handle any output or return codes
+      ansible.builtin.command: cat {{ myfile|quote }}
```
ssbarnea, line 57: DITTO, add quotes to this due to use of jinja braces.

MarkusTeufelberger:
The existing `when` case is wrong by the way (the presence of `when: true` does not influence the idempotence of a task result), but that likely needs a separate PR to be fixed. It could need a test case though, if you're already writing tests.

src/ansiblelint/rules/CommandHasChangesCheckRule.py
konstruktoid: Except that it does.

```sh
$ ansible-lint --version
ansible-lint 5.1.3.dev22 using ansible 2.11.4
$ cat /tmp/issue1681.yml
---
- hosts: all
  tasks:
    - name: command without quotes
      ansible.builtin.command: cat {{ myfile|quote }}
      register: myoutput
      changed_when: myoutput.rc != 0

    - name: command should fail
      ansible.builtin.command: cat {{ myfile|quote }}
...
$ ansible-lint /tmp/issue1681.yml
Loading custom .yamllint config file, this extends our internal yamllint config.
WARNING  Overriding detected file kind 'yaml' with 'playbook' for given positional argument: /tmp/issue1681.yml
no-changed-when: Commands should not change things if nothing needs doing
../../../../tmp/issue1681.yml:9 Task/Handler: command should fail

# .ansible-lint
  - no-changed-when  # Commands should not change things if nothing needs doing

Finished with 1 failure(s), 0 warning(s) on 1 files.
$
```

konstruktoid:
> 
> 
> The existing `when` case is wrong by the way (the presence of `when: true` does not influence the idempotence of a task result), but that likely needs a separate PR to be fixed. It could need a test case though, if you're already writing tests.

Yeah, but I believe that is out-of-scope for this PR.

PR #1709, opened 2021-08-30
Expand coverage of path globbing

greg-hellings:
Exclude path handling does not pick up very many glob matches. This
improves the matching by adding in fnmatch module usage

Fixes: #1561

### DIFF
--- a/test/TestRunner.py
+++ b/test/TestRunner.py
@@ -77,20 +77,21 @@ def test_runner_exclude_paths(default_rules_collection: RulesCollection) -> None
 
 
 @pytest.mark.parametrize(
-    ('exclude_path'), ('**/example.yml', 'examples/**/example.yml')
+    ('exclude_path'), ('**/playbooks/*.yml',)
 )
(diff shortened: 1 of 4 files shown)


ssbarnea (approved):

test/TestRunner.py
```diff
@@ -76,12 +76,17 @@ def test_runner_exclude_paths(default_rules_collection: RulesCollection) -> None
     assert len(matches) == 0
 
 
-def test_runner_exclude_globs(default_rules_collection: RulesCollection) -> None:
+@pytest.mark.parametrize(
+    ('exclude_path'), ('**/example.yml', 'examples/**/example.yml')
```
tadeboro, line 80: Those samples are not the best since they, while being valid globs, are too similar to the extended (wildcard) matches that are supported by the wcmatch library for example. Maybe we should add some more bash-like expressions here? Things like `*.yml`, etc.

ssbarnea (approved):

Issue #1717, opened 2021-09-06
Add a packaging extra for pulling `ansible-core`


webknjaz:
# Issue Type
- Feature request

# Desired Behaviour

There's extras in `setup.cfg` that depend on `ansible` and `ansible-base`. We need a similar one for `ansible-core`.

ssbarnea:
I think that is a bugfix and we should just fix it at https://github.com/ansible-community/ansible-lint/blob/main/setup.cfg#L97-L98

Basically we will not have an extra for 2.10, but if someone needs 2.10 in particular they can install it without an extra. 

Just replace ansible-base with ansible-core and we can call this fixed.

PR #1720, opened 2021-09-07
Make core extra use ansible-core

ssbarnea:
Instead of installing `ansible-base` when `core` extra is
used, we will require `ansible-core`. When original change was
introduced few months back we already knew that the future
proof name was `core` so that is with the extra was called as
such.

Users that want to explicitly use ansible 2.10 should avoid using the extra and mention ansible as a separated dependency. 

If previous environments already existed with ansible-base installed on them an installation failure will appear as ansible-base needs to be removed before ansible-core is installed. Due to this, this change will only be include in the next `5.2.0` release instead of being considered a simple bugfix.

Fixes: #1717

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -29,8 +29,8 @@ related to containers and use the discussions_ forum instead.
     specific version of Ansible. You need to either install the desired version
     of Ansible yourself or mention one of the helper extras:
 
-    * ``core`` - will install the latest version of ansible-base 2.10
-    * ``community`` - will install the latest version of ansible 2.10 with community collections
+    * ``core`` - will install the latest version of ansible-core 2.11+
+    * ``community`` - will install the latest version of ansible community edition
 
 Using pip or pipx
 -----------------
--- a/setup.cfg
+++ b/setup.cfg
@@ -95,7 +95,7 @@ community =
   ansible>=2.10  # GPLv3+
 # this will move to latest stable ansible as soon that is released
 core =
-  ansible-base>=2.10  # GPLv3+
+  ansible-core>=2.11.4  # GPLv3+
 # will install ansible from devel branch, may break at any moment.
 ; Disabled due to https://github.com/pypa/twine/issues/726
 ; devel =
--- a/tox.ini
+++ b/tox.ini
@@ -14,8 +14,7 @@ description =
   Run the tests under {basepython} and
   devel: ansible devel branch
   ansible29: ansible 2.9
-  core: ansible-base 2.10
-  py: ansible-core 2.11
+  core: ansible-core 2.11+
 extras =
   yamllint
   core: core
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,7 +84,7 @@ repos:
     args: ["--strict"]
     additional_dependencies:
     - Sphinx>=3.1.2
-    - ansible-base
+    - ansible-core
     - enrich
     - flaky
     - pytest
@@ -107,7 +107,7 @@ repos:
   hooks:
   - id: pylint
     additional_dependencies:
-    - ansible-base
+    - ansible-core
     - docutils
     - enrich
     - flaky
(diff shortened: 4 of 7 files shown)


setup.cfg
```diff
@@ -95,7 +95,7 @@ community =
   ansible>=2.10  # GPLv3+
 # this will move to latest stable ansible as soon that is released
 core =
-  ansible-base>=2.10  # GPLv3+
+  ansible-core>=2.11.4  # GPLv3+
```
webknjaz, line 98: Note that this will break envs of people who upgrade their envs with `pip install -U ansible-lint[core]` since upgrading from `ansible-base` to `ansible-core` in the pre-existing env is broken.

docs/installing.rst
```diff
@@ -29,7 +29,7 @@ related to containers and use the discussions_ forum instead.
     specific version of Ansible. You need to either install the desired version
     of Ansible yourself or mention one of the helper extras:
 
-    * ``core`` - will install the latest version of ansible-base 2.10
+    * ``core`` - will install the latest version of ansible-core 2.11+
     * ``community`` - will install the latest version of ansible 2.10 with community collections
```
webknjaz, line 33: FTR this statement is outdated too.

tox.ini
```diff
@@ -14,7 +14,7 @@ description =
   Run the tests under {basepython} and
   devel: ansible devel branch
   ansible29: ansible 2.9
-  core: ansible-base 2.10
+  core: ansible-core 2.11+
   py: ansible-core 2.11
```
webknjaz, line 18: ?

tox.ini
```diff
@@ -14,7 +14,7 @@ description =
   Run the tests under {basepython} and
   devel: ansible devel branch
   ansible29: ansible 2.9
-  core: ansible-base 2.10
+  core: ansible-core 2.11+
```
webknjaz, line 17: looks like `py` already incorporates this.

setup.cfg
ssbarnea: I can live with that. Upgrade is not really a supported path and ansible itself have decent error messages about that. We could mark this change as behavior change in order to increase linter minor version when we merge it.

webknjaz: + maybe document it in the changelog?

ssbarnea: That is automatically done at https://github.com/ansible-community/ansible-lint/releases once we merge it. I made it "enhancement" which makes it bump the second version number instead of the 3rd.

tox.ini
ssbarnea: That is the description text and is correct, now "core" translates to ansible-core.

setup.cfg
webknjaz: No, I mean document it properly. Just dumping the current PR title in a changelog is pointless because it doesn't contain any useful information wrt what the effect is for the end users and what to do with it.

tox.ini
webknjaz: Yes, but this seems to duplicate `core` which is weird.

setup.cfg
ssbarnea: I modified that PR description to mention implications and reasoning. All changelog entries do have links to the PRs. I also ensured that it will be included in 5.2.0 due to potentially breaking change.

This release/changelog was used unchanged for more than two years (since v4.3.0) and changing it is would clearly be outside the scope of this review. I hope that is addressing current concerns.

tox.ini
ssbarnea: Is does not duplicate anything, the line you mentioning is using the "core" substring from tox env name in order to build a meaningful doc description for the environment. Not including it or updating it will mislead users when they run `tox -va`.

The reality is right now we are not testing the linter with 2.11, we only test it with `2.9`, `2.10` and `devel` while 2.10 is mislabeled as core. Once we merge this PR we will officially test latest core.

We could restore 2.10 testing if we want but there is no real incentive as the change of specific issue with it is close to zero if both others passed.

tox.ini
ssbarnea: In fact `py` was a mistake and I removed it.

tadeboro (approved):

PR #1730, opened 2021-09-30
Prevent parseable option from changing JSON output format

ssbarnea:
This bug was originally reported against ansible-language-server but
it proved to be caused by a bug inside the linter.

Fixes: https://github.com/ansible/ansible-language-server/issues/43

### DIFF
--- a/src/ansiblelint/app.py
+++ b/src/ansiblelint/app.py
@@ -73,10 +73,10 @@ def choose_formatter_factory(
         r = formatters.QuietFormatter
     elif options_list.parseable_severity:
         r = formatters.ParseableSeverityFormatter
-    elif options_list.parseable or options_list.format == 'pep8':
-        r = formatters.ParseableFormatter
     elif options_list.format == 'codeclimate':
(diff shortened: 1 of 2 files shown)


webknjaz (approved):

tadeboro (approved):
This change will fix the immediate issue at hand, but the process of selecting the output format is a mess. As far as I can tell, there are at least three command-line flags that affect the output: `-p`, `-f`, and `--parseable-severity`. And there is not information on how should those things interact.

ssbarnea:
> This change will fix the immediate issue at hand, but the process of selecting the output format is a mess. As far as I can tell, there are at least three command-line flags that affect the output: `-p`, `-f`, and `--parseable-severity`. And there is not information on how should those things interact.

@tadeboro I totally agree with you on that aspect. I am aware of it and I want to remove both parseable and severity options, keeping only format w/o any modifiers but this would require a major version bump. Part of the same change is renaming "codeclimate" to "json", while keeping the format the same (likely with an alias). I do remember 2-3 feature request for json output added after we implemented it.

PR #1731, opened 2021-09-30
Avoid testing with ansible devel on py36/py37

ssbarnea:
As recently ansible devel branch dropped support for py36/py37 we no longer test it on those versions.

This is fixing the CI which is currently broken since https://github.com/ansible/ansible/pull/74013 was merged.


dmsimard (approved):
lgtm

PR #1764, opened 2021-11-29
docs: Add missing type and parameter in custom-rules.rst

SIPR-octo:
Hello there! :wave: 

When trying to write some new custom rules, we found out a couple of minor mistakes in the documentation, so figured out we will open a PR to fix those.

I didn't add the type of `file` as it requires another import (`from ansiblelint.file_utils import Lintable`) but I can do that if it is necessary for you!

### DIFF
--- a/docs/custom-rules.rst
+++ b/docs/custom-rules.rst
@@ -46,7 +46,7 @@ An example rule using ``matchtask`` is:
 
 .. code-block:: python
 
-    from typing import Any, Dict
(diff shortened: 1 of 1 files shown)


ssbarnea:
Feel free to add the extra import for having full typed signature, especially for the docs we should try to recommend best practices.

SIPR-octo:
Makes sense, I pushed a new commit, hope this does the trick!

ssbarnea (approved):

PR #1770, opened 2021-11-30
Avoid errors processing symlinked files

ssbarnea:
Fixes bug where linter will fail to process symlinked files because
it failed to retrieve the content of the lintable.

This bug was observed on our "eco" pipeline.
Related: https://github.com/devroles/ansible_collection_system/pull/8#pullrequestreview-817978598


greg-hellings (approved):
:+1:

PR #1773, opened 2021-12-01
Configure root logger instead of package logger

relrod:
Change:
- This was discussed internally; the goal is to make it so that rules
  which set up their own logger are still subject to the global rules
  of ansible-lint particularly with regard to our verbosity/quietness
  flags.

- To accomodate for using the root logger, the logging levels of -q and
  -qq had to change slightly. This is because the default log level of
  the root logger is WARNING, and we want to avoid setting it to NOTSET
  which would allow everything by default. To work around this, we now
  make our default be WARNING as well. -q becomes ERROR, and -qq
  becomes CRITICAL which we previously didn't have.

Test Plan:
- New integration tests with various levels of -v and -q.

Signed-off-by: Rick Elrod <rick@elrod.me>


webknjaz:
Affecting an interpreter-global logger seems weird. Why not have a common module with loggers that the rules and other modules could import from?

relrod:
@webknjaz The primary reason is because people might develop rules that set up their own logger, and this could break apps that parse output (such as navigator or ALS). Configuring the root logger helps prevent that. @ssbarnea might have other thoughts here - I spoke with him in DM a few days ago and he thought this was a good direction to go.

webknjaz:
@relrod how about providing a way to get a logger in the base class for rules? I think it's worth standardizing on.

cidrblock (approved):

ssbarnea (approved):

PR #1786, opened 2021-12-17
Allow rpm --nodeps command-instead-of-module

bluikko:
Running `command rpm -e [...]` causes a failure due to rule `command-instead-of-module`. But the `yum` module does not support "expunge" action, so the suggestion to use it is invalid.

Edit `command-instead-of-module` rule to add `rpm -e` to commands that skip the suggestion to use `yum` module instead of `command`.


bluikko:
Quite the brain fart here...

It was of course not the `rpm -e` command that cannot be done with the `yum` module (it can), the problem is `-e --nodeps`.

The `--nodeps` functionality does not exist in the `yum` module. But also ansible-lint cannot be simply patched like this to make it take both `-e` and `--nodeps` into account (?). So instead ignore the `rpm` command if it includes `--nodeps`.

It would be better if the rule would take into account both `-e` and `--nodeps` together I guess.

ssbarnea (approved):

PR #1797, opened 2022-01-10
Make code compatible with rich 11

rabi:
This has been dropped in rich 11.0.0.

Fixes https://github.com/ansible-community/ansible-lint/issues/1795

### DIFF
--- a/src/ansiblelint/generate_docs.py
+++ b/src/ansiblelint/generate_docs.py
@@ -3,7 +3,13 @@ import logging
 from typing import Iterable
 
 from rich import box
-from rich.console import render_group
+
+# Remove this compatibility try-catch block once we drop support for rich < 10.7.0
+try:
+    from rich.console import group
+except ImportError:
+    from rich.console import render_group as group
+
 from rich.markdown import Markdown
 from rich.table import Table
 
@@ -47,7 +53,7 @@ def rules_as_rst(rules: RulesCollection) -> str:
     return r
 
 
-@render_group()
+@group()
 def rules_as_rich(rules: RulesCollection) -> Iterable[Table]:
     """Print documentation for a list of rules, returns empty string."""
     width = max(16, *[len(rule.id) for rule in rules])


setup.cfg
```diff
@@ -71,7 +71,7 @@ install_requires =
   enrich>=1.2.6
   packaging
   pyyaml
-  rich>=9.5.1
+  rich>=10.7.0
```
tadeboro, line 74: Version 10.7.0 is barely four months old, so I would rather see that we keep the old constraints and add a compatibility layer to the code that uses console groups.

src/ansiblelint/generate_docs.py
```diff
@@ -3,7 +3,7 @@
 from typing import Iterable
 
 from rich import box
-from rich.console import render_group
+from rich.console import group
```
tadeboro, line 6: Maybe we should do something like this:

```suggestion
# Remove this compatibility try-catch block once we drop support for rich < 10.7.0
try:
    from rich.console import group
except ImportError:
    from rich.console import render_group as group
```

tadeboro (changes requested):
I like the general idea of making sure the linter works with the latest stable version of rich. But I am not 100% sold on raising the required rich version at this point since the version that introduced incompatibility is only four months old.

src/ansiblelint/generate_docs.py
rabi: I don't like do the try/except dance. But, if that's what we want I'll change.

tadeboro: I am also not a fan of this, but there is not much we can do here since upstream is not taking backward compatibility seriously enough.

ssbarnea (approved):

tadeboro (approved):
Thank you, @rabi, for your contribution!

src/ansiblelint/generate_docs.py
ssbarnea: At least they bumped the major version. Good part is that this did not affect normal usage of the linter because that module is used only to generate docs.

aaronraimist: It definitely broke our usage of the linter

https://github.com/spantaleev/matrix-docker-ansible-deploy/runs/4757647392?check_suite_focus=true
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7d96526b531a6ffdda79c081663a81e8ffc234ab/.github/workflows/ansible-lint.yml

And I'm guessing all of the other people who 👍ed https://github.com/ansible-community/ansible-lint/issues/1795 weren't all trying to generate docs.

tadeboro: Actually, the rich version bump did screw us over in all cases because we load that module at https://github.com/ansible-community/ansible-lint/blob/main/src/ansiblelint/__main__.py#L214 and not inside the relevant `if` clause at https://github.com/ansible-community/ansible-lint/blob/main/src/ansiblelint/__main__.py#L219

PR #1805, opened 2022-01-11
Add ability to automatically fix simple errors

ssbarnea:


cognifloyd:
I could see a separate `ansible-fmt` entry point. I don't think you'll want to run both, however, because a fmt run requires a full lint run to identify issues before the transforms can fix them.

ssbarnea:
> I could see a separate `ansible-fmt` entry point. I don't think you'll want to run both, however, because a fmt run requires a full lint run to identify issues before the transforms can fix them.

Interesting. Now that you said it it does make sense as you can easily decide to fix errors if you recognize any of those you know how to fix. If any fatal error occurs (like syntax check failure), you can skip attempting to fix anything. This makes the fixation smart and likely to nicely integrate as a pre-commit hook.

ssbarnea:
@cognifloyd Considering the huge number of changes and modified files it would awesome if you can start splitting this into bits that can be reviewed and merged gradually.

As each transformation (fix) rule seems to be directly linked to a validation run, I wonder if it would not be easier to just extend the API of Rule and allow implementations to provide ability to perform fixes, basically we would avoid the need to have two types of rules, ensuring that a rule class may also include methods for fixation. WDYT about this?

cognifloyd:
Maybe I could add a `Transform` mixin for rules. If `isinstance(match.rule, Transform)` then it can run the fixes. That'll let me keep some of the base functionality separate but allow combining the interrelated rule+transform code.

That could work. I'll work on that later today.

cognifloyd:
I like how that turned out. Combining Rule and Transform reduced a bunch of duplication so it is much cleaner now.

PR #1806, opened 2022-01-11
MatchError's should include the filename

cognifloyd:
Whenever possible, we should include the filename and linenumber in the MatchError.

I also used the `LINE_NUMBER_KEY` const var instead of using `"__line__"`.

This is minor and should not impact anyone now, but it makes fixing identified errors easier in #1805 (this was extracted from that PR).

### DIFF
--- a/src/ansiblelint/rules/BecomeUserWithoutBecomeRule.py
+++ b/src/ansiblelint/rules/BecomeUserWithoutBecomeRule.py
@@ -22,6 +22,7 @@ from functools import reduce
 from typing import TYPE_CHECKING, Any, List
 
 from ansiblelint.rules import AnsibleLintRule
+from ansiblelint.utils import LINE_NUMBER_KEY
 
 if TYPE_CHECKING:
     from ansiblelint.constants import odict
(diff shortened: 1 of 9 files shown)


cognifloyd:
These rules do not provide linenumber because identifying the line would be fiddly or misleading:

- Meta rules can identify where `galaxy_info` is, but calculating line numbers for the contents of `galaxy_info` would be annoying and passing a linenumber for `galaxy_info` would be misleading.
  - MetaMainHasInfoRule
  - MetaTagValidRUle
  - MetaVideoLinksRule

cognifloyd:
I also used the `LINE_NUMBER_KEY` const var instead of using `"__line__"`.

cognifloyd:
rebased on main

ssbarnea (approved):

ssbarnea:
@cognifloyd Please search for `__line__` and `__file__` and replace the few remaining occurences with the newly added key so we have it in a single place, consistent.

cognifloyd:
Yup. I searched for `__line__` and `__file__`. There is one other use of `"__line__"` but it is semantically different (an attribute instead of a dict key), and its use is isolated to one function in utils, so I didn't change that.

cognifloyd:
The other instances of `__file__` are the magic python var, not string constants.

cognifloyd:
Here are the other uses of the string `'__line__'` (referring to an attribute instead of a key):
https://github.com/ansible-community/ansible-lint/blob/6df35484845627cb117320a221c7f3aac07f85b4/src/ansiblelint/utils.py#L715-L739

tadeboro (approved):
Thank you, @cognifloyd, for your contribution!

PR #1809, opened 2022-01-11
Recursively extract tasks from blocks

cognifloyd:
allow matchtask() to match embedded tasks

```
- block:
  - block:
    - template:
        src: ...
```

See for example:
https://github.com/ansible-community/ansible-lint/blob/ff90dd4a5b1f105028a4c4d5d70a9cd2bcb69bdb/examples/roles/role_for_no_same_owner/tasks/fail.yml#L13-L19

The `no-same-owner` rule, added in #1450, uses that test tasks file. It does its own task parsing to ensure it can recursively find tasks. We should allow all rules to recursively find tasks.

### DIFF
--- a/src/ansiblelint/rules/CommandHasChangesCheckRule.py
+++ b/src/ansiblelint/rules/CommandHasChangesCheckRule.py
@@ -65,7 +65,8 @@ handle the output of the ``command``.
     def matchtask(
         self, task: Dict[str, Any], file: 'Optional[Lintable]' = None
     ) -> Union[bool, str]:
-        if task["__ansible_action_type__"] == 'task':
+        # tasks in a block are "meta" type
+        if task["__ansible_action_type__"] in ['task', 'meta']:
             if task["action"]["__ansible_module__"] in self._commands:
                 return (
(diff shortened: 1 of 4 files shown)


ssbarnea (changes requested):
This change needs a modification of at least one test that will verify the fixed bug or we risk breaking it in the future.

Can you modify one test or add a new one that ensures that this change fixes a bug? The test should fail without it and pass after this change goes in.

If I understood correctly it should be ok to modify one example and add a deep bug that is not found without this change.

cognifloyd:
LOL. I updated `CommandHasChangesCheckRule` to detect command tasks in blocks... and now lots of tests are failing because we don't have `changed_when` in most of the examples.

ssbarnea:
Updating one rule/test to make use of it should be enough to prevent regression. Feel free to add `changed_when: false` or to update the tests to count for the new number of failures.

cognifloyd:
I only had to update 1 playbook. 😌  Tests are green now.

ssbarnea (approved):

PR #1810, opened 2022-01-12
Refactor coverage implementation

ssbarnea:
Adopts the same coverage configuration we use on ansible-compat, which includes:

- avoid using pytest-cov
- integration with codecov.io
- avoid noisy messages from codecov on pull requests, as   we have a GHA check to report its status

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -177,7 +177,7 @@ jobs:
     #       ${{ runner.os }}-
     - name: Install tox
       run: |
-        python3 -m pip install --upgrade tox
+        python3 -m pip install --upgrade tox 'coverage[toml]'
     - name: Log installed dists
       run: >-
         python3 -m pip freeze --all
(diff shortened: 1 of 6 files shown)


webknjaz:
This looks like a bad idea. Why didn't you want anybody to review the
change before merging?

.github/workflows/tox.yml
```diff
@@ -211,6 +211,15 @@ jobs:
         python3 -m tox
       env:
         TOXENV: ${{ matrix.tox_env }}-devel
+    - name: Combine coverage data
+      # produce a single .coverage file at repo root
+      run: coverage combine .tox/.coverage.*
+    - name: Upload coverage data
+      uses: codecov/codecov-action@v1
```
webknjaz, line 218: Note: we need to use v2 because v1 is deprecated (not sure for how long it'll be around).

pyproject.toml
```diff
@@ -7,6 +7,16 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[tool.coverage.run]
```
webknjaz, line 10: This should've gone to `.coveragerc` which is already present in the repo.

PR #1814, opened 2022-01-16
Integrate WPS into flake8

webknjaz:
This is essentially a patch integrating a strict set of linters (WPS) while adding a number of ignores and adjusting the code consistency/style to fix some of the violations. Pre-commit additions are made for convenience to some of the rules added.

*(Should I split this into a series of PRs? Some of the changes are a bit coupled, though)*

### DIFF
--- a/.darglint
+++ b/.darglint
@@ -1,3 +1,8 @@
 [darglint]
+# NOTE: All `darglint` styles except for `sphinx` hit ridiculously low
+# NOTE: performance on some of the in-project Python modules.
+# Refs:
+# * https://github.com/terrencepreilly/darglint/issues/186
+# * https://github.com/wemake-services/wemake-python-styleguide/issues/2287
 docstring_style = sphinx
 strictness = full
--- a/.darglint
+++ b/.darglint
@@ -1,3 +1,3 @@
 [darglint]
-docstring_style=sphinx
-strictness=long
+docstring_style = sphinx
+strictness = full
--- a/.darglint
+++ b/.darglint
@@ -1,4 +1,3 @@
 [darglint]
 docstring_style=sphinx
-enable=DAR104
 strictness=long
--- a/.flake8
+++ b/.flake8
@@ -48,10 +48,6 @@ extend-ignore =
   B014,  # Redundant exception types in `except (ImportError, ModuleNotFoundError) as e:`.  Write `except ImportError as e:`, which catches exactly the same exceptions.
   C405,  # Unnecessary list literal - rewrite as a set literal.
   C408,  # Unnecessary dict call - rewrite as a literal.
-  C812,  # missing trailing comma
-  C813,  # missing trailing comma in Python 3
-  C815,  # missing trailing comma in Python 3.5+
-  C816,  # missing trailing comma in Python 3.6+
   DAR101,
   DAR201,
   DAR301,
--- a/.flake8
+++ b/.flake8
@@ -41,7 +41,11 @@ extend-ignore =
   E501,
   H,
 # WPS:
-  B,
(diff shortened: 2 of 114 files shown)


ssbarnea (dismissed):
I am not even sure how to put this in a nice form: this is NOT going to happen.

After the "joy" that a similar change caused the navigator project, I become a strong opposer of WPS and what it brings.

Should I also mention that this change does change black default formatting we use

That is not only only an undesired change but it is also coming at a very inappropriate time, causing more stress instead of dealing with real bugfixes, not that we run out of important things to fix.

webknjaz:
> I am not even sure how to put this in a nice form: this is NOT going to happen.

This is definitely not nice.

> After the "joy" that a similar change caused the navigator project, I become a strong opposer of WPS and what it brings.

This is too abstract and doesn't mean anything. If you want to make a review, put in constructive meaningful comments.
 
> Should I also mention that this change does change black default formatting we use

No, you shouldn't because it's a lie. Black does not change quotes in this repository.

> That is not only only an undesired change but it is also coming at a very inappropriate time, causing more stress instead of dealing with real bugfixes, not that we run out of important things to fix.

This is as important. We need to start improving the quality of this codebase.

ssbarnea:
@webknjaz If you really want to improve quality of the code, I would suggest one of these two options: either fix known bugs or address some of the rules what are already skipped. Is not that you are not aware of the list of linting rules excludes from https://github.com/ansible-community/ansible-lint/blob/main/.pylintrc#L11-L32 and https://github.com/ansible-community/ansible-lint/blob/main/.flake8#L45-L99 as you were directly involved in adding them while introduced new lintes more than an year ago.

Adding more tools that report problems does not increase the quality by themselves. As someone that spent a lot of time addressing some of these I don't want to see more of these, not until I see work done to address existing ones.

For those reasons I am against this move and I want to close this PR but I will allow people that actively contributed to the extension to review this proposal.

ssbarnea (changes requested):
Please do not reset my -1 on this.

webknjaz:
@ssbarnea please reply to the scope of this PR, not some other out-of-the-scope activities. The importance of addressing existing suppressed rules has nothing to do with the effort of making other improvements I deeply care about.
To reiterate: _I hear you_, I _understand_ that you think it's important to make other improvements — but there are other things that are important too (and I'm not meaning to judge whether your suggestions are less or more important, they are just not in the scope _right here_).
Since I've already invested so much time improving things, it only makes sense for me not to lose the improvements. This is why I chose to submit them, not because I think that some other improvements are not important in some way. Otherwise, parallel changes in the repository will create more conflicts meaning that even more effort would be required to change things for the better. I would've preferred to spend that extra effort on other impactful things, maybe even on the stuff, you pointed out about the older ignores.

As I hinted in the original post, the commits here are mostly atomic. And the improvements are real: they do make the project better than it was before.
By mentioning the possibility of splitting the patch, I'm
1) welcoming improvement suggestions
2) making a case for a possible suggestion to add a separate vetting invocation that does not affect people like it's done in the navigator (I could've probably done this right away but _I forgot — there is no evil intent here —_ because this effort has been started months ago, it lived on my laptop until I've got to spend time to finish it, it wasn't in the scope originally and I didn't think about it — still, this shouldn't prevent people from looking into the patch)
3) inviting to make related _in-the-scope suggestions_ that are _actionable_

P.S. I'm only dismissing the reviews that are unrelated and/or toxic. If you think you can behave and be constructive, you're more than welcome to post suggestions even though I didn't originally ask for your review (nor did GitHub auto-request it).

:peace_symbol:

ganeshrn:
As discussed in team meeting, please make this changes in multiple PR's and cleanup the code before enforcing the linter.

Issue #1820, opened 2022-01-21
Redefine yaml formatting rules to match other tools


ssbarnea:
I came recently to my attention that some decisions that we made in the past regarding the formatting rules ansible-lint checks for on YAML files may be in conflict with other formatters or yaml dumpers.

## Tools evaluated so far

* [prettier](https://prettier.io/) - A very popular code formatter, supporting multiple file format.
* [vccode-yaml](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) - vscode extension for editing YAML files, which seems to be re-using prettier for formatting. I am not aware of any divergence between them but afaik, it does not need prettier to be installed to work.
* [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) - a python library that can also dump yaml and format it in various ways, we already use it for loading yaml with comments. One particular feature is that it can preserve comments, something that pyyaml cannot.

### too few spaces before comment

That rule is originating from yamllint and we could easily just disable it in our custom defaults for yamllint, those that are used as long user does not create his own yamllint config.

Apparently all 3 tools mentioned above do use only one space before an inline comment. Thus I am considering change our settings to match those tools, and avoid conflicts. I seen it myself how annoying is to observe that your editor reformatting on save may produce changes that later are undone by project CI.

## TBD

Feel free to report other tools or conflicts that you know.

PR #1821, opened 2022-01-21
Neuter min space requirement for inline comments

ssbarnea:
Update our yamllint config defaults to disable the requirement of adding two spaces before each inline comments. This does not affect users that already set a different value in their yamllint config files.

This change will not produce extra errors after is applied, it will only reduce the number of errors reported in general as `foo: bar # that is cool!` will no longer complain about missing an additional space, using one or two will be fine.

The expected behavior is that ansible-lint will no longer report problems like:
```
warning  too few spaces before comment  (comments)
```

This will improve the compatibility with various code formatters, as mentioned in #1820


greg-hellings (approved):
Agreed, wholeheatedly.

src/ansiblelint/rules/YamllintRule.py
```diff
@@ -28,6 +28,9 @@
 YAMLLINT_CONFIG = """
 extends: default
 rules:
+  comments:
+    # https://github.com/prettier/prettier/issues/6780
+    min-spaces-from-content: 1
```
webknjaz, line 33: Why not 2, like in Python?

cognifloyd: Because this conflicts with other yaml linters that don't allow specifying multiple spaces before the comment. Prettier is one example (see prettier/prettier#6780 as in the comment)

Another example is the `fmt` mode in #1805 (aside: I'm working on splitting that behemoth into smaller PRs) which uses `ruamel.yaml` to edit files to fix issues identified by ansible-lint's rules. `ruamel.yaml` does not have a way for me to specify how many spaces should be inserted before comments in any automated fashion. I would have to manually iterate through the document to check every comment and adjust the columns... Oh pain! (if that would even work) I really hope we don't have to do that.

cognifloyd (approved):

webknjaz (approved):

cidrblock (approved):
thank you!

tadeboro (changes requested):
Does this change mean that previously "green" YAML files will now become "red"? If yes, then I am -1 on this change during the 5.x.y series of releases. Breaking people's CI/CD setups is not something we should do lightly.

And I know introducing new rules can also "break" the CI/CD, but those situations are different since in most cases they indicate that something can be improved. Changing comment formatting brings no direct benefit to linter users.

ssbarnea:
I did not see any breakage reported by eco job but I do realise that someone could be surprised by the new setup. Maybe we can do this in two steps, first disabling that check and delaying the new preference for a major version change.

tadeboro (approved):

PR #1822, opened 2022-01-21
Add ability to load .config/ansible-lint.yml

ssbarnea:
This will allow users to declutter their project root directory by allowing users to put configuration in `.config/ansible-lint.yml` instead of `.ansible-lint` file.

It matches a feature already implemented by molecule and supported by https://dot-config.github.io -- the goal being to reduce cluttering of root project directories.

This does not change the ability to load the file from its current location.

### DIFF
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -10,10 +10,11 @@ Configuring
 Configuration File
 ------------------
 
-Ansible-lint supports local configuration via a ``.ansible-lint`` configuration
-file. Ansible-lint checks the working directory for the presence of this file
-and applies any configuration found there. The configuration file location can
-also be overridden via the ``-c path/to/file`` CLI flag.
+Ansible-lint supports local configuration via a ``.ansible-lint`` or
+``.config/ansible-lint.yml`` configuration files. Ansible-lint checks the
+working directory for the presence of this file and applies any configuration
+found there. The configuration file location can also be overridden via the
+``-c path/to/file`` CLI flag.
 
 When configuration file is not found in current directory, the tool will try
 to look for one in parent directories but it will not go outside current git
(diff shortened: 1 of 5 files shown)


webknjaz (changes requested):
I've never seen other projects using a trailing dot when the config is in a subdirectory. In that case, it's usually only the directory itself that is dotted.
I'd go for `.config/ansible-lint`, then. Alternatively, there could be support for specifying a different path via `pyproject.toml` instead of hardcoding one string — because it's not a widely accepted standard and some projects may already have some other directory for config adopted.
Also, I would expect tests for this kind of patch.

ssbarnea:
> I've never seen other projects using a leading dot when the config is in a subdirectory. In that case, it's usually only the directory itself that is dotted. I'd go for `.config/ansible-lint`, then. Alternatively, there could be support for specifying a different path via `pyproject.toml` instead of hardcoding one string — because it's not a widely accepted standard and some projects may already have some other directory for config adopted. Also, I would expect tests for this kind of patch.

I am not going to go for the configure the config file location rabbit hole.

I considered a different name without the dot but that would add extra complexity and I prefer to KISS. Telling people that just move existing config filename without asking them to rename them is easier. Also, changing the filename would require changes in other places like json schemas that detect the configuration of ansible-lint. 

But I will add tests, that is indeed a genuine comment. Anything else should be considered as something for the future, like having a proper yml file extension would not be a bad idea, just outside the scope.

tadeboro:
I think this is not needed. Ansible Lint already consults too many locations when it searches for the config file. It is bad enough that the configuration loader can retrieve configuration from a file that has nothing to do with the current project.

As for configuring the config file path: we already allow this (`-c` command-line switch). So adding support for this would probably be a lesser evil compared to another magic search we are trying to introduce here.

cognifloyd (approved):

PR #1823, opened 2022-01-21
Add utility function to initialize ansible Templar

copart-jafloyd:
This commit was cherry-picked from #1805. There are a variety of situations where it is helpful to get a `Templar` instance instead of delegating template parsing. So, this splits the `ansible_template` function into two functions: `ansible_template` (no change in functionality) and `ansible_templar` (the function I need).

PS: I'm @cognifloyd - I'm filing this PR with my work account so github stops complaining when I push commits using that account once this PR is merged.


cognifloyd:
@ssbarnea could you run tests and review this please?

ssbarnea (approved):

PR #1824, opened 2022-01-21
Add utils.nested_items_path and Deprecate utils.nested_items

cognifloyd:
This was extracted from #1805.

There are a lot of cases where I need to not only iterate over the contents of nested dicts/lists, but I also need the path to each item's parent.
For example, this would allow for more descriptive error messages about the location of rule matches and it facilitates the transforms in #1805.

- Add new util function nested_items_path
- Reformat with black
- Add more typehints to satisfy mypy
- convert all uses of nested_items() to nested_items_path()
- add test for nested_items_path utility
- add TODO about marking utils.nested_items deprecated
- Mark utils.nested_items deprecated

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -76,8 +76,7 @@ def initialize_logger(level: int = 0) -> None:
     # Unknown logging level is treated as DEBUG
     logging_level = VERBOSITY_MAP.get(level, logging.DEBUG)
     logger.setLevel(logging_level)
-    # Also pass all warnings.warn() messages through the logger
-    logging.captureWarnings(True)
+    logging.captureWarnings(True)  # pass all warnings.warn() messages through logging
     # Use module-level _logger instance to validate it
     _logger.debug("Logging initialized to level %s", logging_level)
 
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -76,6 +76,8 @@ def initialize_logger(level: int = 0) -> None:
     # Unknown logging level is treated as DEBUG
     logging_level = VERBOSITY_MAP.get(level, logging.DEBUG)
     logger.setLevel(logging_level)
+    # Also pass all warnings.warn() messages through the logger
+    logging.captureWarnings(True)
     # Use module-level _logger instance to validate it
     _logger.debug("Logging initialized to level %s", logging_level)
 
--- a/src/ansiblelint/rules/NoTabsRule.py
+++ b/src/ansiblelint/rules/NoTabsRule.py
@@ -4,7 +4,7 @@ import sys
 from typing import TYPE_CHECKING, Any, Dict, Union
 
 from ansiblelint.rules import AnsibleLintRule
-from ansiblelint.utils import nested_items_path
+from ansiblelint.yaml_utils import nested_items_path
 
 if TYPE_CHECKING:
     from typing import Optional
--- a/src/ansiblelint/rules/NoTabsRule.py
+++ b/src/ansiblelint/rules/NoTabsRule.py
@@ -4,7 +4,7 @@ import sys
 from typing import TYPE_CHECKING, Any, Dict, Union
 
 from ansiblelint.rules import AnsibleLintRule
-from ansiblelint.utils import nested_items
+from ansiblelint.utils import nested_items_path
 
 if TYPE_CHECKING:
     from typing import Optional
@@ -29,10 +29,15 @@ class NoTabsRule(AnsibleLintRule):
     def matchtask(
         self, task: Dict[str, Any], file: 'Optional[Lintable]' = None
     ) -> Union[bool, str]:
-        for k, v, parent in nested_items(task):
+        for k, v, parent_path in nested_items_path(task):
             if isinstance(k, str) and '\t' in k:
                 return True
-            if (parent, k) not in self.allow_list and isinstance(v, str) and '\t' in v:
+            if (
+                parent_path
+                and (parent_path[-1], k) not in self.allow_list
+                and isinstance(v, str)
+                and '\t' in v
+            ):
                 return True
         return False
 
--- a/src/ansiblelint/rules/NoTabsRule.py
+++ b/src/ansiblelint/rules/NoTabsRule.py
@@ -32,9 +32,9 @@ class NoTabsRule(AnsibleLintRule):
         for k, v, parent_path in nested_items_path(task):
             if isinstance(k, str) and '\t' in k:
                 return True
+            parent_key = "" if not parent_path else parent_path[-1]
             if (
-                parent_path
-                and (parent_path[-1], k) not in self.allow_list
+                (parent_key, k) not in self.allow_list
                 and isinstance(v, str)
                 and '\t' in v
             ):
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -27,17 +27,7 @@ from argparse import Namespace
 from collections.abc import ItemsView
 from functools import lru_cache
 from pathlib import Path
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    Generator,
-    List,
-    Optional,
-    Sequence,
-    Tuple,
-    Union,
-)
+from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tuple, Union
 
 import yaml
 from ansible import constants
@@ -889,7 +879,7 @@ def convert_to_boolean(value: Any) -> bool:
 
 def nested_items(
     data: Union[Dict[Any, Any], List[Any]], parent: str = ""
-) -> Generator[Tuple[Any, Any, str], None, None]:
+) -> Iterator[Tuple[Any, Any, str]]:
     """Iterate a nested data structure."""
     if isinstance(data, dict):
         for k, v in data.items():
@@ -906,7 +896,7 @@ def nested_items(
 def nested_items_path(
     data: Union[Dict[Any, Any], List[Any]],
     parent_path: Optional[List[Union[str, int]]] = None,
-) -> Generator[Tuple[Any, Any, List[Union[str, int]]], None, None]:
+) -> Iterator[Tuple[Any, Any, List[Union[str, int]]]]:
     """Iterate a nested data structure."""
     if parent_path is None:
         parent_path = []
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -33,13 +33,11 @@ from typing import (
     Callable,
     Dict,
     Generator,
-    Iterator,
     List,
     Optional,
     Sequence,
     Tuple,
     Union,
-    cast,
 )
 
 import yaml
@@ -896,7 +894,7 @@ def nested_items(
     """Iterate a nested data structure."""
     warnings.warn(
         "Call to deprecated function ansiblelint.utils.nested_items. "
-        "Use ansiblelint.utils.nested_items_path instead.",
+        "Use ansiblelint.yaml_utils.nested_items_path instead.",
         category=DeprecationWarning,
         stacklevel=2,
     )
@@ -910,85 +908,3 @@ def nested_items(
             yield "list-item", item, parent
             for k, v, p in nested_items(item):
                 yield k, v, p
-
-
-def nested_items_path(
-    data: Union[Dict[Any, Any], List[Any]],
-    parent_path: Optional[List[Union[str, int]]] = None,
-) -> Iterator[Tuple[Any, Any, List[Union[str, int]]]]:
-    """Iterate a nested data structure, yielding key/index, value, and parent_path.
-
-    This is a recursive function that calls itself for each nested layer of data.
-    Each iteration yields:
-
-    1. the current item's dictionary key or list index,
-    2. the current item's value, and
-    3. the path to the current item from the outermost data structure.
-
-    For dicts, the yielded (1) key and (2) value are what ``dict.items()`` yields.
-    For lists, the yielded (1) index and (2) value are what ``enumerate()`` yields.
-    The final component, the parent path, is a list of dict keys and list indexes.
-    The parent path can be helpful in providing error messages that indicate
-    precisely which part of a yaml file (or other data structure) needs to be fixed.
-
-    For example, given this playbook:
-
-    .. code-block:: yaml
-
-        - name: a play
-          tasks:
-          - name: a task
-            debug:
-              msg: foobar
-
-    Here's the first and last yielded items:
-
-    .. code-block:: python
-
-        >>> playbook=[{"name": "a play", "tasks": [{"name": "a task", "debug": {"msg": "foobar"}}]}]
-        >>> next( nested_items_path( playbook ) )
-        (0, {'name': 'a play', 'tasks': [{'name': 'a task', 'debug': {'msg': 'foobar'}}]}, [])
-        >>> list( nested_items_path( playbook ) )[-1]
-        ('msg', 'foobar', [0, 'tasks', 0, 'debug'])
-
-    Note that, for outermost data structure, the parent path is ``[]`` because
-    you do not need to descend into any nested dicts or lists to find the indicated
-    key and value.
-
-    If a rule were designed to prohibit "foobar" debug messages, it could use the
-    parent path to provide a path to the problematic ``msg``. It might use a jq-style
-    path in its error message: "the error is at ``.[0].tasks[0].debug.msg``".
-    Or if a utility could automatically fix issues, it could use the path to descend
-    to the parent object using something like this:
-
-    .. code-block:: python
-
-        target = data
-        for segment in parent_path:
-            target = target[segment]
-
-    :param data: The nested data (dicts or lists).
-    :param parent_path: Do not use this param. It is used internally to recursively
-                        build the yielded parent path (list of keys, indexes).
-
-    :returns: each iteration yields the key (of the parent dict) or the index (lists)
-    """
-    if parent_path is None:
-        parent_path = []
-    convert_to_tuples_type = Callable[
-        [Union[Dict[Any, Any], List[Any]]],
-        Iterator[Tuple[Union[str, int], Any]],
-    ]
-    if isinstance(data, dict):
-        # dict subclasses can override items() so don't use dict.items directly
-        convert_to_tuples = cast(convert_to_tuples_type, data.__class__.items)
-    elif isinstance(data, list):
-        convert_to_tuples = cast(convert_to_tuples_type, enumerate)
-    else:
-        raise TypeError(
-            f"Expected a dict or a list but got {data!r} of type '{type(data)}'",
-        )
-    for key, value in convert_to_tuples(data):
-        yield key, value, parent_path
-        if isinstance(value, (dict, list)):
-            yield from nested_items_path(data=value, parent_path=parent_path + [key])
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -955,7 +955,7 @@ def nested_items_path(
     you do not need to descend into any nested dicts or lists to find the indicated
     key and value.
 
-    If a rule were designed to prohimit "foobar" debug messages, it could use the
+    If a rule were designed to prohibit "foobar" debug messages, it could use the
     parent path to provide a path to the problematic ``msg``. It might use a jq-style
     path in its error message: "the error is at ``.[0].tasks[0].debug.msg``".
     Or if a utility could automatically fix issues, it could use the path to descend
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -905,7 +905,7 @@ def nested_items(
 
 def nested_items_path(
     data: Union[Dict[Any, Any], List[Any]],
-    parent_path: Optional[List[Union[str, int]]] = None
+    parent_path: Optional[List[Union[str, int]]] = None,
 ) -> Generator[Tuple[Any, Any, List[Union[str, int]]], None, None]:
     """Iterate a nested data structure."""
     if parent_path is None:
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -915,7 +915,63 @@ def nested_items_path(
     data: Union[Dict[Any, Any], List[Any]],
     parent_path: Optional[List[Union[str, int]]] = None,
 ) -> Iterator[Tuple[Any, Any, List[Union[str, int]]]]:
-    """Iterate a nested data structure."""
+    """Iterate a nested data structure, yielding key/index, value, and parent_path.
+
+    This is a recursive function that calls itself for each nested layer of data.
+    Each iteration yields:
+
+    1. the current item's dictionary key or list index,
+    2. the current item's value, and
+    3. the path to the current item from the outermost data structure.
+
+    For dicts, the yielded (1) key and (2) value are what mydict.items() yields.
+    For lists, the yielded (1) index and (2) value are what enumerate(mylist) yields.
+    The final component, the parent path, is a list of dict keys and list indexes.
+    The parent path can be helpful in providing error messages that indicate
+    precisely which part of a yaml file (or other data strucutre) needs to be fixed.
+
+    For example, given this playbook:
+
+    .. code-block:: yaml
+
+        - name: a play
+          tasks:
+          - name: a task
+            debug:
+              msg: foobar
+
+    Here's the first and last yielded items:
+
+    .. code-block:: python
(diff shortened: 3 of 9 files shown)


src/ansiblelint/utils.py
```diff
@@ -901,3 +898,22 @@ def nested_items(
...
+    if isinstance(data, dict):
+        for key, value in data.items():
+            yield key, value, parent_path
+            for k, v, p in nested_items_path(value, parent_path + [key]):
+                yield k, v, p
+    if isinstance(data, list):
+        for index, item in enumerate(data):
+            yield index, item, parent_path
+            for k, v, p in nested_items_path(item, parent_path + [index]):
+                yield k, v, p
```
webknjaz, line 919: Looking at this function more, I see a huge opportunity to get rid of the duplication and some nesting:
```suggestion

    if isinstance(data, dict):
        convert_to_tuple = dict.items
    elif isinstance(data, list):
        convert_to_tuple = enumerate
    else:
        raise TypeError

    for key, value in convert_to_tuple(data):
        yield key, value, parent_path
        yield from nested_items_path(value, parent_path + [key])
```
The only thing I don't understand is that why the second loop inverts key and value — I suppose it's a bug.

src/ansiblelint/utils.py
```diff
@@ -901,3 +898,22 @@ def nested_items(
             yield "list-item", item, parent
             for k, v, p in nested_items(item):
                 yield k, v, p
+
+
+def nested_items_path(
```
webknjaz, line 903: I wonder if we can take this opportunity to move it to a dedicated `pathutils` module instead of keeping it in this wholesale utils dump.

test/TestUtils.py
```diff
@@ -526,7 +526,34 @@ def test_nested_items() -> None:
         ("list-item", "apple", "fruits"),
         ("list-item", "orange", "fruits"),
     ]
-    assert list(utils.nested_items(data)) == items
+    with pytest.deprecated_call():
```
webknjaz, line 529: It's always recommended to make use of the `match=` argument in such cases so that the test is accurate.

src/ansiblelint/utils.py
```diff
@@ -901,3 +898,22 @@ def nested_items(
...
+    data: Union[Dict[Any, Any], List[Any]],
+    parent_path: Optional[List[Union[str, int]]] = None,
+) -> Iterator[Tuple[Any, Any, List[Union[str, int]]]]:
+    """Iterate a nested data structure."""
```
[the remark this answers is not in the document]
cognifloyd: I added a docstring that should help clarify this.

test/TestUtils.py
cognifloyd: done.

src/ansiblelint/__main__.py
```diff
@@ -76,6 +76,8 @@ def initialize_logger(level: int = 0) -> None:
     # Unknown logging level is treated as DEBUG
     logging_level = VERBOSITY_MAP.get(level, logging.DEBUG)
     logger.setLevel(logging_level)
+    # Also pass all warnings.warn() messages through the logger
```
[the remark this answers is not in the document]
cognifloyd: done.

src/ansiblelint/utils.py
```diff
@@ -901,3 +898,22 @@ def nested_items(
...
+    if parent_path is None:
+        parent_path = []
+    if isinstance(data, dict):
+        for key, value in data.items():
+            yield key, value, parent_path
+            for k, v, p in nested_items_path(value, parent_path + [key]):
```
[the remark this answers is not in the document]
cognifloyd: Agreed. Changed.

src/ansiblelint/utils.py
```diff
@@ -901,3 +898,22 @@ def nested_items(
...
+                yield k, v, p
+    if isinstance(data, list):
+        for index, item in enumerate(data):
+            yield index, item, parent_path
+            for k, v, p in nested_items_path(item, parent_path + [index]):
```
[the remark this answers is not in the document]
cognifloyd: done

src/ansiblelint/utils.py
cognifloyd: Nice! Yeah, I'll adjust it like this.

cognifloyd: done.

src/ansiblelint/utils.py
```diff
@@ -901,3 +909,80 @@ def nested_items(
...
+    elif isinstance(data, list):
+        convert_to_tuple = enumerate
+    else:
+        raise TypeError(
+            f"nested_items_path expects a dict or a list but got {type(data)}"
+        )
+    for key, value in convert_to_tuple(data):
+        yield key, value, parent_path
+        if isinstance(value, (dict, list)):
+            yield from nested_items_path(value, parent_path + [key])
```
cognifloyd, line 988: > The only thing I don't understand is that why the second loop inverts key and value — I suppose it's a bug.

@webknjaz You said this might be a bug - but it is not a bug.
This recursive call is passing in `data=value` setting a new parent_path for the children of this `value`. That is how it creates the parent_path for each level of the nested data structure.

Maybe it would help to see just the recursive call lines in what I originally had (before switching to `yield from` and de-duplicating the loops):
```python
            for k, v, p in nested_items_path(value, parent_path + [key]):
            for k, v, p in nested_items_path(item, parent_path + [index]):
```
value is like item
key is like index

In any case, hopefully it's much clearer without the two separate loops.

src/ansiblelint/utils.py
```diff
@@ -901,3 +909,80 @@ def nested_items(
...
+    elif isinstance(data, list):
+        convert_to_tuples = enumerate
+    else:
+        raise TypeError(
+            f"nested_items_path expects a dict or a list but got {type(data)}"
+        )
+    for key, value in convert_to_tuples(data):
+        yield key, value, parent_path
+        if isinstance(value, (dict, list)):
+            yield from nested_items_path(data=value, parent_path=parent_path + [key])
```
cognifloyd, line 988: mypy didn't like this refactor.
```
src/ansiblelint/utils.py:980: error: Incompatible types in assignment (expression has type "Type[enumerate[Any]]", variable has type "Callable[[Dict[_KT, _VT]], ItemsView[_KT, _VT]]")
src/ansiblelint/utils.py:985: error: Argument 1 has incompatible type "Union[Dict[Any, Any], List[Any]]"; expected "Dict[_KT, _VT]"
src/ansiblelint/utils.py:988: error: List item 0 has incompatible type "_KT"; expected "Union[str, int]"
```

Any suggestions on how to clean this up for mypy? Or should I go back to how it was?

cognifloyd: OK. I found a `cast()` that works for this.

src/ansiblelint/utils.py
```diff
@@ -901,3 +898,22 @@ def nested_items(
             yield "list-item", item, parent
             for k, v, p in nested_items(item):
                 yield k, v, p
+
+
+def nested_items_path(
+    data: Union[Dict[Any, Any], List[Any]],
```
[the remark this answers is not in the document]
webknjaz: I guess we can keep it with a docstring that makes it clearer what it is. OTOH, I'm a bit conflicted about it because a variable name should normally reflect its specific semantic meaning within the given context and I felt like it's overly generic: `data=1` and `data=b'some-binary-blob` are also valid examples of the generic context of data, for example.

src/ansiblelint/utils.py
```diff
@@ -889,8 +880,14 @@ def convert_to_boolean(value: Any) -> bool:
 
 def nested_items(
     data: Union[Dict[Any, Any], List[Any]], parent: str = ""
-) -> Generator[Tuple[Any, Any, str], None, None]:
+) -> Iterator[Tuple[Any, Any, str]]:
```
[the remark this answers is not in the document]
webknjaz: I normally prefer not to include refactoring in the feature PRs, unless it's absolutely necessary. It'd be fine if submitted separately but in this context, it prevents the patch from being atomic.

src/ansiblelint/utils.py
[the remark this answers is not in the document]
webknjaz: > The parent_path is used when recursively calling the function. I don't expect users of the function to use the parent_path param, just the child elements.

Really? This opens up a question, then, why does that argument leak into the public API? I'd make the main implementation function private in this case and have a thin wrapper exposing only the argument that is supposed to be a part of the API.

src/ansiblelint/utils.py
webknjaz: @cognifloyd could you confirm that the swapped arguments in the last recursive call were indeed a bug? Because I was reviewing from my phone and might've missed something. Just want to be sure that this guess was correct.

src/ansiblelint/utils.py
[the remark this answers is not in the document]
ssbarnea: How about using `json` instead of `data because that is really what is supposed to contain, a data structure that is JSON-compatible. Anyway, I am also ok with keeping it data too, the comments are already good-enough.

src/ansiblelint/utils.py
ssbarnea: While I do not love the `utils` kitchensink, I do not think that the move is needed now, especially as we do not have an already existing module that would better encapsulate that type of code. That is not about file or paths, is about data parsing. Anyway let's avoid scope creeping, keeping the location is better, at least for now.

ssbarnea:
I think that you still need to pass `tox -e lint` but other than this LGTM.

src/ansiblelint/utils.py
webknjaz: > So, you're proposing `ansiblelint.path_utils` (or maybe `ansiblelint.data_utils` would be more descriptive?).

Something like that. It should be specific and accurate.
The problem with generic non-specific namespaces like `utils` or `helpers` is the same as with variables like `data`, `things` and `stuff` — they don't provide any clarity/insight as to why they exist, what's inside and what is their relation to the current level of abstraction. Over time, these such modules become dumps for helpers that do something needed somewhere but have no specific purpose.

Here's one of the explanations of this problem: https://tonsky.me/blog/utils/.

> Would you want to move any other functions to such a new module?

In general, yes. But not within this PR. It shouldn't contain things not related to the reimplementation of this one function.

> I think this function fits well with many of these other utils. It helps in finding/iterating through playbooks and tasks files. Is it worth it to split them into more modules?

See the comments above.

src/ansiblelint/utils.py
webknjaz: I think, for this, it's possible to solve with custom type definitions but it'd require some extra effort. This is caused by `Any` being too generic.

src/ansiblelint/utils.py
webknjaz: Ah, never mind. I see you replied at https://github.com/ansible-community/ansible-lint/pull/1824#pullrequestreview-860164964.

src/ansiblelint/utils.py
webknjaz: Interesting. I guess we need good test coverage for this since it's untrivial to follow and the tests would act as a demonstration of the usage/invocations + expected result example.

src/ansiblelint/utils.py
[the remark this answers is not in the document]
webknjaz: Maybe `node` could be more accurate, I don't know. But I hope you've got the idea of why I think this could be improved :)

src/ansiblelint/utils.py
webknjaz: By the way, one of the ways to narrow types is type guards: https://mypy.readthedocs.io/en/latest/type_narrowing.html#user-defined-type-guards.

webknjaz: Oh, I see. mypy gets confused by `if isinstance(value, (dict, list)):` — if you were to check for just one type, it'd understand what's going on better, I think. But that's probably not what you want...

src/ansiblelint/rules/NoTabsRule.py
```diff
@@ -29,10 +29,15 @@ class NoTabsRule(AnsibleLintRule):
...
         self, task: Dict[str, Any], file: 'Optional[Lintable]' = None
     ) -> Union[bool, str]:
-        for k, v, parent in nested_items(task):
+        for k, v, parent_path in nested_items_path(task):
             if isinstance(k, str) and '\t' in k:
                 return True
-            if (parent, k) not in self.allow_list and isinstance(v, str) and '\t' in v:
+            if (
+                parent_path
+                and (parent_path[-1], k) not in self.allow_list
```
webknjaz, line 37: I'd suggest putting `parent_path[-1]` into a variable. It's currently unclear what this is on this level of abstraction

src/ansiblelint/utils.py
```diff
@@ -901,3 +910,85 @@ def nested_items(
...
+        >>> next( nested_items_path( playbook ) )
+        (0, {'name': 'a play', 'tasks': [{'name': 'a task', 'debug': {'msg': 'foobar'}}]}, [])
+        >>> list( nested_items_path( playbook ) )[-1]
+        ('msg', 'foobar', [0, 'tasks', 0, 'debug'])
+
+    Note that, for outermost data structure, the parent path is ``[]`` because
+    you do not need to descend into any nested dicts or lists to find the indicated
+    key and value.
+
+    If a rule were designed to prohimit "foobar" debug messages, it could use the
```
webknjaz, line 958: I suppose it's
```suggestion
    If a rule were designed to prohibit "foobar" debug messages, it could use the
```

src/ansiblelint/utils.py
```diff
@@ -901,3 +910,85 @@ def nested_items(
...
+
+    This is a recursive function that calls itself for each nested layer of data.
+    Each iteration yields:
+
+    1. the current item's dictionary key or list index,
+    2. the current item's value, and
+    3. the path to the current item from the outermost data structure.
+
+    For dicts, the yielded (1) key and (2) value are what mydict.items() yields.
+    For lists, the yielded (1) index and (2) value are what enumerate(mylist) yields.
```
webknjaz, line 929: I wouldn't use the invented my-prefixed identifiers as they add go the confusion, not being declared anywhere.
```suggestion
    For dicts, the yielded (1) key and (2) value are what ``dict.items()`` yields.
    For lists, the yielded (1) index and (2) value are what ``enumerate()`` yields.
```

src/ansiblelint/rules/NoTabsRule.py
cognifloyd: done. Is it clearer now?

src/ansiblelint/utils.py
```diff
@@ -901,3 +910,85 @@ def nested_items(
...
+        Iterator[Tuple[Union[str, int], Any]],
+    ]
+    if isinstance(data, dict):
+        # dict subclasses can override items() so don't use dict.items directly
+        convert_to_tuples = cast(convert_to_tuples_type, data.__class__.items)
+    elif isinstance(data, list):
+        convert_to_tuples = cast(convert_to_tuples_type, enumerate)
+    else:
+        raise TypeError(
+            f"nested_items_path expects a dict or a list but got {type(data)}"
```
webknjaz, line 989: Maybe this would be better:
```suggestion
            f"Expected a dict or a list but got {data!r} of type '{type(data)}'",
```
(thinking about the stdlib exceptions that usually don't specify the called interface in the exceptions)

src/ansiblelint/rules/NoTabsRule.py
webknjaz: Yep, looks better.

src/ansiblelint/utils.py
webknjaz: As for "in the scope" vs not — I think it is because there is a deprecation warning that points at the new importable. It would be better for it to point at the final location so we wouldn't need to go through several rounds of deprecations as we decide where to better put this thing.

Alternatively, the deprecation bits could be put in a dedicated atomic PR where it could be coupled with the new proper selection.

src/ansiblelint/utils.py
```diff
@@ -901,3 +910,85 @@ def nested_items(
...
+    :returns: each iteration yields the key (of the parent dict) or the index (lists)
+    """
+    if parent_path is None:
+        parent_path = []
+    convert_to_tuples_type = Callable[
+        [Union[Dict[Any, Any], List[Any]]],
+        Iterator[Tuple[Union[str, int], Any]],
+    ]
+    if isinstance(data, dict):
+        # dict subclasses can override items() so don't use dict.items directly
```
webknjaz, line 983: Could do `functools.partial(data.items)` and `functools.partial(enumerate, data)`. Then, in the loop it would be just `convert_to_tuples()`. If that helps...

src/ansiblelint/utils.py
```diff
@@ -901,3 +910,85 @@ def nested_items(
...
+        convert_to_tuples = cast(convert_to_tuples_type, data.__class__.items)
+    elif isinstance(data, list):
+        convert_to_tuples = cast(convert_to_tuples_type, enumerate)
+    else:
+        raise TypeError(
+            f"Expected a dict or a list but got {data!r} of type '{type(data)}'",
+        )
+    for key, value in convert_to_tuples(data):
```
webknjaz, line 991: Also, you could always cast this instead:
```suggestion
    for key, value in cast(whateversexpected, convert_to_tuples(data)):
```

src/ansiblelint/utils.py
```diff
@@ -901,3 +910,85 @@ def nested_items(
...
+        yield key, value, parent_path
+        if isinstance(value, (dict, list)):
+            yield from nested_items_path(data=value, parent_path=parent_path + [key])
```
webknjaz, line 994: Or maybe just cast `value` and `key` here?

test/TestUtils.py
```diff
@@ -526,7 +526,45 @@ def test_nested_items() -> None:
...
+    ]
+    assert list(utils.nested_items_path(data)) == items
+
+
+def test_nested_items_path_raises_typeerror() -> None:
+    """Verify non-dict/non-list types make nested_items_path() raises TypeError."""
+    unexpected_data_types = ["string", 42, 1.234, None, ("tuple",), {"set"}]
+
+    for data in unexpected_data_types:
+        with pytest.raises(TypeError):
```
webknjaz, line 566: Same as with the deprecated call — it's strongly recommended to use the `match=` arg here too.

test/TestUtils.py
```diff
@@ -526,7 +526,45 @@ def test_nested_items() -> None:
...
+        (2, "everything", ["answer", 0, "forty-two"]),
+    ]
+    assert list(utils.nested_items_path(data)) == items
+
+
+def test_nested_items_path_raises_typeerror() -> None:
+    """Verify non-dict/non-list types make nested_items_path() raises TypeError."""
+    unexpected_data_types = ["string", 42, 1.234, None, ("tuple",), {"set"}]
+
+    for data in unexpected_data_types:
```
webknjaz, line 565: Instead of having an in loop, it's preferable to generate separate tests with `@pytest.mark.parametrize`. This way, it won't stop on the first failure and keep testing other cases. And every case will show up separately in the test report.

test/TestUtils.py
```diff
@@ -526,7 +526,45 @@ def test_nested_items() -> None:
...
+        (0, {"forty-two": ["life", "universe", "everything"]}, ["answer"]),
+        ("forty-two", ["life", "universe", "everything"], ["answer", 0]),
+        (0, "life", ["answer", 0, "forty-two"]),
+        (1, "universe", ["answer", 0, "forty-two"]),
+        (2, "everything", ["answer", 0, "forty-two"]),
+    ]
+    assert list(utils.nested_items_path(data)) == items
+
+
+def test_nested_items_path_raises_typeerror() -> None:
```
webknjaz, line 561: It'd look like this:
```suggestion
@pytest.mark.parametrize(
    'invalid_data_input',
    (
        "string",
        42,
        1.234,
        None,
        ("tuple",),
        {"set"},
    )
)
def test_nested_items_path_raises_typeerror(invalid_data_input) -> None:
```

test/TestUtils.py
webknjaz, line 565:
```suggestion
```

test/TestUtils.py
```diff
@@ -526,7 +526,45 @@ def test_nested_items() -> None:
...
+    """Verify non-dict/non-list types make nested_items_path() raises TypeError."""
+    unexpected_data_types = ["string", 42, 1.234, None, ("tuple",), {"set"}]
+
+    for data in unexpected_data_types:
+        with pytest.raises(TypeError):
+            list(utils.nested_items_path(data))  # type: ignore
```
webknjaz, line 567: Instead of ignoring every possible mypy type rule, it's best to only ignore a specific type violation. For this, add the respective rule id in square brackets.
```suggestion
            list(utils.nested_items_path(data))  # type: ignore[only-that-rule-here]
```
(I don't know what the exact violation was but you can find out, I suppose)

src/ansiblelint/utils.py
cognifloyd: OK. I'm fine with moving it - `utils` is way too big. So what should we name it? Maybe `yaml_utils` would be most appropriate since this is primarily for iterating through yaml data and at least a few of the other utils relate to that as well (to be moved in a follow-up).

src/ansiblelint/utils.py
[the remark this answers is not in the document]
cognifloyd: `collection`?

src/ansiblelint/utils.py
webknjaz: Works for me. As long as it's accurate. Maybe mention the expectations in the module docstring so that non-matching things wouldn't appear there.

src/ansiblelint/utils.py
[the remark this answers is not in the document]
webknjaz: I suppose collection or container match the purpose per https://docs.python.org/dev/library/collections.abc.html.

`data_collection` I guess. Because just `collection` does not set the expectation of its semantics. `"qwerty"` is a collection of characters, right? `{'q', 'w', 'e', 'r', 't', 'y'}` too.

test/TestUtils.py
cognifloyd: It was complaining that data isn't a dict or a list, which is exactly the point.

src/ansiblelint/utils.py
cognifloyd: Sure. That's a bit nicer. Done.

src/ansiblelint/utils.py
cognifloyd: Nope, that doesn't work. mypy complains that the two assignments of `convert_to_tuples` have different types, so we have to cast at the assignment.

src/ansiblelint/utils.py
cognifloyd: I don't understand this suggestion.

test/TestUtils.py
cognifloyd: done

test/TestUtils.py
cognifloyd: OK. Switching to the parametrized list with the `Any` typehint satisfied it. Whatever. I dropped the comment.

test/TestUtils.py
cognifloyd: done

src/ansiblelint/utils.py
cognifloyd: oh. right. No, this is too late to cast them here. mypy was complaining that the two assignments had different types, so casting at the assignment corrects that.

src/ansiblelint/utils.py
cognifloyd: K. I moved it to `yaml_utils`.

src/ansiblelint/utils.py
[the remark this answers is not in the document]
cognifloyd: ok. I split it into inner and outer functions.

src/ansiblelint/utils.py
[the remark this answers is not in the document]
cognifloyd: done

cognifloyd:
K. I believe I've addressed all of the comments.

src/ansiblelint/utils.py
cognifloyd: > While I do not love the `utils` kitchensink, I do not think that the move is needed now, especially as we do not have an already existing module that would better encapsulate that type of code. That is not about file or paths, is about data parsing. Anyway let's avoid scope creeping, keeping the location is better, at least for now.

Oh. I missed this comment. What do you think of where it is now: `yaml_utils` ? Or should I move it back.

ssbarnea (approved):

src/ansiblelint/utils.py
ssbarnea: He fixe it.

PR #1825, opened 2022-01-22
Refactor `matchtasks()` to reduce complexity

cognifloyd:
Split out of #1805.

This extracts some of the logic from `AnsibleLintRule.matchtasks()` to a new iterator: `ansiblelint.utils.iter_tasks_in_file()`.
This refactor:
- is backwards compatible,
- reduces the complexity of `matchtasks()`,
- makes that logic reusable - I needed it for one of the transformations I added in #1805,
- prepares for another feature I'll extract from #1805 in a follow-up PR that slightly increases the complexity of `matchtasks()` (that feature will add a new `matchrawtask()` that is like `matchtask()` but passes the un-normalized/raw task as well).


src/ansiblelint/utils.py
```diff
@@ -759,6 +760,36 @@ def construct_mapping(
     return data
 
 
+def iter_tasks_in_file(lintable: Lintable, rule_id: str):
```
ssbarnea, line 763: This method yields but the signature does not report that. I suspect something like this should do the trick https://stackoverflow.com/a/57916822/99834

In fact I was curious why our linters did not catch this, probably badly configured.

src/ansiblelint/utils.py
```diff
@@ -759,6 +760,36 @@ def construct_mapping(
...
+    """Iterate over tasks in file."""
```
ssbarnea, line 764: Be sure that you also document the return values inside the docstring, that is essential when you return more than one value. In that case we yield 4 fields. I checked on typing about the subject as I was not able to find anything using google or SO, got suggested use of https://docs.python.org/3/library/typing.html#typing.NamedTuple as an alternative. 

Still, if that requires many other extra changes we can leave it out for follow-up.

src/ansiblelint/utils.py
cognifloyd: Fixed.

The linter did catch it, but I hadn't configured pre-commit yet.

I just ran `pre-commit install` to make sure I don't push changes like this again.

src/ansiblelint/utils.py
cognifloyd: Documented. Is that enough or do you think we still need a named tuple?

src/ansiblelint/utils.py
```diff
@@ -759,6 +761,60 @@ def construct_mapping(
...
+        normalized due to an AnsibleParserError.
+
+    :param lintable: The playbook or tasks/handlers yaml file to get tasks from
+    :param rule_id: The current rule id to allow calculating skipped.
+
+    :return: raw_task, normalized_task, skipped, error
+    """
+    data = parse_yaml_linenumbers(lintable)
+    if not data:
+        return
```
cognifloyd, line 792: codecov doesn't like this. I could increase coverage by indenting the rest of this function under `if data:` ... but that's kind  of gross. Can we take the hit in coverage? Or do you have a recommended way to add a test for this?

ssbarnea: Considering the nature of the change it would be wise to add a test for it. I guess that adding some `foo: none` in a playbook would do the trick. You may be able to do this even if you modify an existing test, i guess.

src/ansiblelint/utils.py
```diff
@@ -759,6 +761,60 @@ def construct_mapping(
...
+
+    raw_tasks = get_action_tasks(data, lintable)
+
+    for raw_task in raw_tasks:
+        # An empty `tags` block causes `None` to be returned if
+        # the `or []` is not present - `task.get('tags', [])`
+        # does not suffice.
+        if 'skip_ansible_lint' in (raw_task.get('tags') or []) or (
+            rule_id in raw_task.get('skipped_rules', ())
+        ):
```
webknjaz, line 803: Look how we can add context via "label-variables":
```suggestion
        skip_with_a_yaml_comment = 'skip_ansible_lint' in (raw_task.get('tags') or [])
        skip_with_a_task_tag = rule_id in raw_task.get('skipped_rules', ())
        if skip_with_a_yaml_comment or skip_with_a_task_tag:
```

cognifloyd: yeah... pylint+flake8 has taught me to avoid creating new vars wherever possible to keep the complexity low. But I can add some better vars here.

cognifloyd: done

src/ansiblelint/utils.py
cognifloyd: ok. I added a test.

src/ansiblelint/rules/__init__.py
```diff
@@ -107,24 +108,17 @@ def matchtasks(self, file: Lintable) -> List[MatchError]:
...
             return matches
 
-        yaml = ansiblelint.utils.parse_yaml_linenumbers(file)
-        if not yaml:
-            return matches
-
-        yaml = ansiblelint.skip_utils.append_skipped_rules(yaml, file)
-
-        try:
-            tasks = ansiblelint.utils.get_normalized_tasks(yaml, file)
```
cognifloyd, line 117: After removing this line, `get_normalized_tasks` is no longer used anywhere. That means that code coverage decreases because nothing is using it. Should we deprecate or remove `get_normalized_tasks`? Or perhaps add a test for `get_normalized_tasks` in case any 3rd party rules are using it (not sure why they would)?

cognifloyd: I removed `get_normalized_tasks`. If we need to deprecate it or add a test instead, then let me know and I can revert the delete commit.

cognifloyd: Oh. But removing it still decreases overall coverage (albeit slightly less).

cognifloyd: Oh. never mind. I was looking at an old codecov report. Codecov is green now.

cognifloyd:
Yeah! Everything is green, even codecov.

ssbarnea:
@cognifloyd Rebase joy... :p

cognifloyd:
Rebased

ssbarnea (approved):

cognifloyd:
Tests are green :)

PR #1826, opened 2022-01-24
Extract list logic to reduce complexity of __main__.main()

cognifloyd:
Extracted from #1805

When we start adding the fmt options, `main()` becomes too complex so linters start yelling.
This extracts the list specific options to lower that complexity.


cognifloyd:
Hmm. So apparently there are no tests for either list option.

ssbarnea (approved):

ssbarnea:
@cognifloyd please add some. I observed that we lacked some coverage on this area two days ago. See https://app.codecov.io/gh/ansible-community/ansible-lint

PR #1827, opened 2022-01-24
Move ansiblelint.testing.RunFromText import into test code

cognifloyd:
Some of the rules were importing test code even outside of the testing block. This fixes that so there are no unintended side-effects of importing the testing code.

In particular, importing that testing code should be delayed so that `prepare_environment()` doesn't get called when importing the rule:
https://github.com/ansible-community/ansible-lint/blob/main/src/ansiblelint/testing/__init__.py#L19-L21
It probably doesn't make a difference, but it was a code smell, so this cleans that up.


ssbarnea:
@cognifloyd I will check, from time to time I need to ignore coverage result. Rounding errors can make it go down.

ssbarnea (approved):

PR #1828, opened 2022-01-25
New Feature: Reformat YAML files with ruamel.yaml

cognifloyd:
**This is ready for review / feedback! It is marked draft until the UX is finalized. Then I'll mark it ready for the final pre-merge reviews.**

### Overview

I want to introduce a feature to `ansible-lint`: It should be able to fix issues identified by rules. The feature in this PR is a pre-requisite to introducing that.

_NB: This is and must be **opt-in**. `ansible-lint` will not modify any files unless you ask it to._

In order to fix any issues identified by rules, we will have to read each YAML file with `ruamel.yaml` and then write it out. Reading/writing YAML files with `ruamel.yaml` standardizes spacing and quotes while preserving comments. In other words, it does very basic YAML reformatting which might fix some of the issues identified by yamllint. Fixing most of the issues identified by `ansible-lint` (including many of the yamllint-identified issues) requires transforming the files in some way. This PR only implements the read+write part, leaving the actual rule transforms for future PRs.

To use this **(opt-in)** mode, run `ansible-lint --fmt-yaml` (UX is not final! Please [bikeshed](https://en.wikipedia.org/wiki/bikeshedding) the interface). That makes ansible-lint pass all YAML files it checks through `ruamel.yaml`.

#### In summary:

- This PR provides very basic YAML formatting like standardizing yaml quotes and indentation.
- It will not fix all issues identified by yamllint.
- This is separate from the transforms that fix ansible-lint identified issues. 
- This is a pre-requisite to those transforms because we cannot fix anything without rewriting yaml files.

### Please bikeshed the interface for this and planned future PRs:

UX discussion moved to https://github.com/ansible-community/ansible-lint/discussions/1910

The current winning interface is `--write`. So that's what this PR implements.

### TODO

- [ ] add a test that uses each of the block chomping indicators to show how it reformats things.
- [ ] reformat everything with prettier to make sure the `ruamel.yaml` settings here do not conflict.


cognifloyd:
Please note: 1000 lines of this commit are in `examples/playbooks/lots_of_warnings.transformed.yml`. That particular test makes sure it the read/dump process has no issues with that size of file. So, this PR isn't as big as it looks.

cognifloyd:
> I would expect ansible-fmt to be a "beautifier" which uses best practices for both the YAML and the included Jinja2 expressions. That's why I would avoid using this name for now.

OK. So maybe we add `ansible-fmt` later after we add the jinja [template dumper](https://github.com/ansible-community/ansible-lint/blob/fmt/src/ansiblelint/transform_utils.py) (takes a Jinja AST and dumps it back to a template string).

cognifloyd:
Should we use `ansible-lint --fmt-yaml` to clarify that this only reformats yaml files?

felixfontein:
`--fmt-yaml` sounds good to me. (No idea what the others think though :) )

cognifloyd:
Maybe a chart would make this easier to grok

| Feature |  | ansible-lint | ansile-lint<br>--fmt | ansible-lint<br>--fmt --skip-rules | ansile-lint<br>--transform |
|---|---|---|---|---|---|
| Lint ansible content | YAML | Y | Y |  | Y* |
| Lint ansible content | Jinja2 | Y | Y |  | Y* |
| Reformat files | YAML |  | Y | Y | Y^+ |
| Refomrat files | Jinja2 |  |  |  | Y^- |
| Transform files<br>(fix or simplify fixing lint issues) | YAML |  |  |  | Y |
| Transform files<br>(fix or simplify fixing lint issues) | Jinja2 |  |  |  | Y |

- `*` Transforms require lint results to identify what to fix

- `^` Transforms must reformat any transformed files

- `+` Reformatting YAML files without lint errors is optional

- `-` Reformatting JInja Templates can be lossy with whitespace because AST does not preserve chomped whitespace (like `{%-` or `-%}`). So, only Jinja templates with identified issues get reformatted.

examples/playbooks/become.transformed.yml
```diff
@@ -0,0 +1,15 @@
+---
```
cognifloyd, line 1: [This](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/become.transformed.yml) is a reformatted copy of [become.yml](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/become.yml)

examples/playbooks/contains_secrets.transformed.yml
```diff
@@ -0,0 +1,15 @@
+---
```
cognifloyd, line 1: [This](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/contains_secrets.transformed.yml) is a reformatted copy of [examples/playbooks/contains_secrets.yml](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/contains_secrets.yml)

examples/playbooks/lots_of_warnings.transformed.yml
```diff
@@ -0,0 +1,1000 @@
+---
```
cognifloyd, line 1: [This](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/lots_of_warnings.transformed.yml) is a reformatted copy of [examples/playbooks/lots_of_warnings.yml](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/lots_of_warnings.yml)

examples/playbooks/nomatchestest.transformed.yml
```diff
@@ -0,0 +1,9 @@
+---
```
cognifloyd, line 1: [This](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/nomatchestest.transformed.yml) is a reformatted copy of [examples/playbooks/nomatchestest.yml](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/nomatchestest.yml)

examples/playbooks/unicode.transformed.yml
```diff
@@ -0,0 +1,5 @@
+---
```
cognifloyd, line 1: [This](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/unicode.transformed.yml) is a reformatted copy of [examples/playbooks/unicode.yml](https://github.com/cognifloyd/ansible-lint/blob/yaml-fmt/examples/playbooks/unicode.yml)

cognifloyd (commented):
Here are some links to help with reviewing the diff between the test fixtures before and after `fmt`.

cognifloyd:
After asking around at work I got the following feedback (in no particular order; numbered to facilitate comments about each of these):

1. Make a chart.
    - @cognifloyd: [_Done_](https://github.com/ansible-community/ansible-lint/pull/1828#issuecomment-1021485392)

2. `fmt`/`transform` should be in something other than `ansible-lint`:
    - > Did you consider extracting this new formatting/transforming functionality into a new entity? so it's not a part of the linter?
well, obviously ansible-fmt doesn't really reflect the purpose of the transformer/autofixer you are working on
maybe something like `ansible-transform` will do
same code base, but its own entry point and flags
    - @cognifloyd: _That's the point of the bikeshedding on this PR: to see what UX/naming would be best for these features._

3. concern that reformatting the yaml might reorder the yaml dict keys.
    - @cognifloyd: _docs might need to clarify that this is not an issue. w/ py3.5+ dicts retain insertion order (and guaranteed ordered for py3.6+). And `ruamel.yaml` also goes to great lengths to preserve the order since it used to be written for py2._

4. Add a dry-run for `--yaml-fmt`
    - > [by] dry-run [I mean] do a diff on what changed and which files changed
I would like to see what is going to be updated (what is wrong with that yaml) and maybe i will update manually rather than having the whole file rewritten
    - @cognifloyd: _I assumed people would keep their ansible stuff in git. Then `git diff` would show what was updated. But, I can see how people might prefer the pure-lint approach. To facilitate this, we could add a rule that formats all the files in a temporary directory and warns if the files would be updated with `--fmt-yaml`._

5. Want to know what `--fmt-yaml` changes.
    - > print out what was updated after the `--fmt-yaml` is done
    - @cognifloyd: _This was an oversight. We don't report anywhere how many files were reformatted. We should probably at least print a list of the files that changed due to `--fmt-yaml`._

6. Backup files for files changed by `--fmt-yaml`
    - > or maybe generate a backup file as an option and have them do diff manually
    - @cognifloyd: _This is a very `ansible`-esque way to handle things. Should we add a backup option especially for those that do not have their playbooks in version control?_

cognifloyd:
Hmm. we might need to silence the `yaml` rule when we run with `--fmt-yaml` because that might resolve the issues. We won't know until the next lint run if the yaml issues were fixed.

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,125 @@
...
+                yaml.dump(data, file.path, transform=self._final_yaml_transform)
+
+    def _final_yaml_transform(self, text: str) -> str:
+        """
+        Ensure that top-level sequences are not over-indented.
+
+        In order to make that work, we cannot delegate adding the yaml explict_start
+        to ruamel.yaml or dedent() won't have anything to work with.
+        Instead, we add the explicit_start here.
+        """
```
cognifloyd, line 109: I just ran `ansible-lint --fmt-yaml` on a few thousand yaml files... And this function did not do very well. I ran into a variety of edge cases with comments and one case where a yaml block (`|`) had a line that started with `#`, so we got a false positive on comment handling.

So, this approach doesn't work well enough for widespread use.
But I found a way to delegate this formatting to ruamel.yaml! Using a subclass of `ruamel.yaml.emitter.Emitter` I was able to customize the sequence indents so that it does not indent root-level sequences.

I'll clean that up and then push it later.

cognifloyd: OK. I replaced this function with a `FormattedEmitter` class.

lilatomic:
Hi!
I think this will not work with [yaml block scalar folded styles](https://yaml.org/spec/1.2.1/#id2796251). One usecase for this is to separate command-line arguments. For example:
```yaml
- name: backup influxdb ee
  command:
    cmd: >-
      influxd-ctl 
      -bind {{ influx_leader_hostname }}:{{ influx_meta_port }}
      -bind-tls -k
      -auth-type basic
      -user backup
      -pwd {{ backup_password }}
      backup
      -strategy {{ backup_strategy }}
      {{ backup_dir }}
```
I think this is a limitation of ruamel, it doesn't provide a CST (concrete syntax tree), so we can't know that the text came from a folded scalar.

cognifloyd:
@lilatomic ok. I will add a test that uses each of the block chomping indicators to show how it reformats things.

(aside: it might take me a bit to do that. I'm on vacation atm, so I won't be next to a computer much)

Do you know of any other python library that does use a CST for YAML?

webknjaz:
@cognifloyd was it decided to have a CLI option and not a separate entry-point? I can't keep up with all of the discussions.

cognifloyd:
@webknjaz  I used a `--fmt-yaml` option, but I'm still open to change. I'm still asking for more feedback on the interface.
https://github.com/ansible-community/ansible-lint/pull/1828#issuecomment-1021485392

What would you prefer?

cognifloyd:
Please see the chart in https://github.com/ansible-community/ansible-lint/pull/1828#issuecomment-1021485392 and think about the UX for both interrelated features: fmt for YAML (this PR) and transforms for ansible-lint content (future PRs).

@webknjaz what entry point (and/or options) do you think would be best for each of these?

lilatomic:
> I will add a test that uses each of the block chomping indicators to show how it reformats things.

I actually discovered my error while trying to work up a test case for it. I've pushed it to my fork [here](https://github.com/lilatomic/ansible-lint/commit/aaf235a285c64828086c32f3524591a8dc51c873).

I don't think we need to test for chomping. The chomping indicators will impact the contents of the string scalar, so I would expect yaml libraries to handle that correctly. Folding, OTOH, allows for many representations of the same string scalar, so reformats which pull in only the AST or which use other formatting (eg max line length) can impact it.

lilatomic:
Have we considered configuring transforms in a way similar to the way we configure the warn or skip lists? For example, we could have `remediate_list: ['yaml']` to enable the autoformat. We could also use this to supply configuration options (for example, a default umask or owner for risky-file-perms or whether to just use "preserve").

Also, maybe this was hashed already, but have we considered implementing the transform as part of the rules themselves? I think this would allow us to better associate the identification of lints with their potential resolution. For example, a general YAML linter could also run the transform itself. I think the real benefit would be for things like command-instead-of-module. We can reuse the machinery which already  whether the command can become the module, and then propose how. As well, if we determine that some subcommands can't become their modules, we would modify the lint rule by modifying the same file.

ssbarnea:
I investigated the list of already existing command line arguments and I think we should use `--write`. Prettier has the same option, alongside the short version but we cannot use that.

I do not want to use file format in command name because in the future we may want reformatting for other file formats, and we already have a cluttered list of cli options.

Regarding making the reformatting configurable, lets keep this change as a MVP that does formatting and keep it out, for now. My expectation that this initial version will reformat only when you pass `--write`. An extra benefit is that the write argument is also useful for rules that would fix simple errors, not only reformatting feature.

cognifloyd:
@ssbarnea Where would `--write` go in my chart: https://github.com/ansible-community/ansible-lint/pull/1828#issuecomment-1021485392

Are you saying `--write` would replace just `--fmt` (`--fmt-yaml` in the chart)? Or both `--fmt` and `--transform`?

ssbarnea:
IMHO, --write is all we need but don't change it before someone else agrees. 

Reformatting and fixing is not a black and white area, many things could be between them.

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,164 @@
...
+        for match in self.matches:
+            try:
+                lintable = lintables[match.filename]
+            except KeyError:
+                # we shouldn't get here, but this is easy to recover from so do that.
+                lintable = Lintable(match.filename)
+                self.matches_per_file[lintable] = []
+            self.matches_per_file[lintable].append(match)
+
+    def run(self) -> None:
```
lilatomic, line 94: I think this function does several things:
- sets up the transformer
- loads the file (call to `load_data`)
- checks if we should operate on the file
- runs the transform (first half of the call to `yaml.dump`)
- writes the file to disk (second half of the call to `yaml.dump`)

Two of these tasks are IO/side-effecting. I think this makes it harder to test. For example, if we want to test that it still works with like 1000 lines of YAML, we have to have a file with 1000 lines of YAML. We couldn't just create it in place, with something like `['expr']*1000`.

I think bundling these together will also lead to duplication between transformers. I suspect that all transformers will need to read and write the file, so it would be worth perhaps extracting to a template method. I think that would also make it easier to create new transformer, by simply needing to subclass transformer and override a `def transform(file: TextIO)->TextIO`

I think we should address this now (as opposed to once this is merged) to make testing easier and so the initial interface is easier to use.

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,164 @@
...
+        yaml.map_indent = 2  # type: ignore[assignment]
+        yaml.sequence_indent = 4 if self.indent_sequences else 2  # type: ignore[assignment]
+        yaml.sequence_dash_offset = yaml.sequence_indent - 2  # type: ignore[operator]
+
+        if self.indent_sequences:  # in the future: or other formatting options
+            # For root-level sequences, FormattedEmitter overrides sequence_indent
+            # and sequence_dash_offset to prevent root-level indents.
+            yaml.Emitter = FormattedEmitter
+
+        # explicit_start=True + map_indent=2 + sequence_indent=2 + sequence_dash_offset=0
```
lilatomic, line 126: I think the information in this comment is valuable, but I don't think that an inline comment is a good place for it. I can't determine which line it "belongs" to. It feels like it's a [doctest](https://docs.python.org/3/library/doctest.html) for a method like `configure_yaml_formatter`

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,164 @@
...
+    _sequence_dash_offset = 0  # Should be _sequence_indent - 2
+
+    # NB: mypy does not support overriding attributes with properties yet:
+    #     https://github.com/python/mypy/issues/4125
+    #     To silence we have to ignore[override] both the @property and the method.
+
+    @property  # type: ignore[override]
+    def best_sequence_indent(self) -> int:  # type: ignore[override]
+        """Return the configured sequence_indent or 2 for root level."""
+        return 2 if self.column < 2 else self._sequence_indent
```
lilatomic, line 48: I'm not sure I know what `self.column < 2` means. I think it's determining whether we're at the root? We might want to add a comment or extract a function for that.

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,164 @@
...
+class Transformer:
+    """Transformer class performs the fmt transformations."""
+
+    def __init__(self, result: LintResult):
+        """Initialize a Transformer instance."""
+        # TODO: options for explict_start, indent_sequences
+        self.explicit_start = True
+        self.indent_sequences = True
+
+        self.matches: List[MatchError] = result.matches
```
lilatomic, line 76: It looks like this to the end of the function is calculating the matches per file. Should computing that be the responsibility of the Transformer?

lilatomic (commented):
Hiya! I'm excited for where this could go!
I left a few comments about places I thought we could improve testability. There were also a few places I wasn't sure why something was happening and I think could be clarified.

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,164 @@
...
+            except KeyError:
+                # we shouldn't get here, but this is easy to recover from so do that.
+                lintable = Lintable(match.filename)
+                self.matches_per_file[lintable] = []
+            self.matches_per_file[lintable].append(match)
+
+    def run(self) -> None:
+        """Execute the fmt transforms."""
+        # ruamel.yaml rt=round trip (preserves comments while allowing for modification)
+        yaml = YAML(typ="rt")
```
cognifloyd, line 97: I can extract this into it's own function that configures ruamel.yaml. That should make the I/O of this method more obvious.

cognifloyd (commented):
Woohoo! Thanks for the feedback @lilatomic . I've got a couple questions before I make adjustments to clarify the code or remove unnecessary bits.

src/ansiblelint/transformer.py
cognifloyd: I moved this into a docstring for a new `yaml` property. Does that help?

src/ansiblelint/transformer.py
cognifloyd: I made a new `yaml` property to handle instantiating `ruamel.yaml.YAML` when needed.

src/ansiblelint/transformer.py
lilatomic: Ah, yes, this makes sense. I was thinking that the design should look like that, and it does!
I would still suggest that we extract the actual reformatting of the file (and all that ruamel configuration) into a separate function so it's easier to test. It would also be easier to implement it as a TransformMixin.transform later

src/ansiblelint/transformer.py
lilatomic: I would disregard this comment now. It makes more sense for the "singleton Transformer that runs all the TransformerMixin.transform" to keep track of this.

src/ansiblelint/transformer.py
lilatomic: I think that the comparison is a good way of explaining why all the fiddly knobs are the way they are. I saw you pushed it to the docstring, I think that's a good place for it. My personal style would be to extract a function and then have these as test cases. I find I usually don't regret breaking things up like that.

src/ansiblelint/transformer.py
cognifloyd: done

src/ansiblelint/transformer.py
cognifloyd: > I would still suggest that we extract the actual reformatting of the file (and all that ruamel configuration) into a separate function so it's easier to test. It would also be easier to implement it as a TransformMixin.transform later

OK. So, I pulled out the ruamel.yaml config, but I'm not clear on how I might reorganize the `run()` method for testability. Could you take a look at the latest code and see if you see a good way to break it up?

src/ansiblelint/transformer.py
lilatomic: neat, looks good!

src/ansiblelint/transformer.py
cognifloyd: So, test cases instead of or in addition to the docstring?

lilatomic: I think if we intend for this configuration to be exposed publicly, we would probably need both (or doctests, which is already both). But if it's mostly internal, either is fine, I think it's a personal style thing. Even if we were planning on exposing this configuration publicly, I think we can safely push it off to another PR.

src/ansiblelint/transformer.py
lilatomic: I think the key thing would be separating the "formatting yaml" part from the "reading/writing files". I think we could just separate the `yaml.dump` into `yaml.dumps` and `file.write`, and then scoot the `yaml.write` into a function like 
```python
def reformat(self, unformatted):
    return self.yaml.dumps(unformatted)
```
Then we can have a few integration tests for actually reading and writing files, and then many more unit tests for covering weird YAML things.

We could get really into breaking up the other things (separating the is_yaml filtering), but I think the above is a good step.

cognifloyd: Ah. The joys of `ruamel.yaml`. It does not provide a `dumps`. It will only write to a stream pushing people to let it write the file directly. So, to get around that I would have to create a `StringIO` and then handle writing it myself.
I'm doing that for the test I'm writing, but it's awkward.

src/ansiblelint/transformer.py
cognifloyd: I just added a parametrized test for this. :)

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,227 @@
...
+                data = ""
+                file_is_yaml = False
+
+            if file_is_yaml:
+                # ruamel.yaml only preserves empty (no whitespace) blank lines.
+                # So, drop spaces in whitespace-only lines. ("\n  \n" -> "\n\n")
+                data = _whitespace_only_lines_re.sub("", cast(str, data))
+                # load_data has an lru_cache, so using it should be cached vs using YAML().load() to reload
+                data = load_data(data)
+                self.yaml.dump(data, file.path)
```
cognifloyd, line 227: Note that `yaml.dump()` has optimizations around creating the string & dumping to file all at once. The docs really push people away from separating creating a string and then writing it to a file:
https://yaml.readthedocs.io/en/latest/example.html#output-of-dump-as-a-string

There is no `dumps()` function w/ `ruamel.yaml`. Though a `dumps()` function might make tests that avoid IO easier, it's going to be a bunch of work to get around the design decisions of `ruamel.yaml` to write to a `StringIO()` instead of directly to the file. If there's a clean way to do this, I'm all ears.

lilatomic: I think it's good. Sure, it's a bit less nice than `dumps`, but it's only a few extra lines in a test.
I think the way you've used the StringIO in the tests provides a good seam for testing. In my head I was thinking that we should have a function to handle the formatting and dump to File or StringIO, and then I wrote it out as
```python
class Transformer:
    ...
    def dumps(self, data, dest)->None:
        self.yaml.dump(data, dest)
```
which really isn't adding anything of value.

So I think the way you've done it here is good.

lilatomic:
Thanks for incorporating my feedback! It looks good to me!

lilatomic:
2 considerations I have with the name of the command:
1. We will probably want to have a way of enabling certain transforms, the same way we can with skip-list, warn-list, enable-list &c.
2. We will probably want to support several output formats (ex write directly, export git patches, print diffs to screen, summarise for CI)

`--write` seems like a good option, but I'm not so sure of how it combines with those considerations: `--write-list` doesn't make sense to me, but `--write-format` seems ok. I think `--transform-list` is a good substitute for `--write-list`. 

I can't really come up with anything better. `--fix` seems too glib, and my personal vote for `--remediate` is definitely bad.

cognifloyd:
I just added a count of the number of modified files to the ansible-lint ouptut. It would look approximately like this:
```
Modified 42 files.
Finished with 0 failure(s), 0 warning(s), on 100 files.
```

AndreasDavour:
I just followed a link in the Bullhorn and feel like I'm tossed into a context I really feel unfamiliar about. But, I wanted to toss in a question, and a few cents on the idea of ansible-lint modifying and not only linting. It will be an option, so it does not just suddenly start modifying, right? Run ansible-lint and only when run with ansible-lint -fmt (or something) will it actually do changes?

ssbarnea:
> I just followed a link in the Bullhorn and feel like I'm tossed into a context I really feel unfamiliar about. But, I wanted to toss in a question, and a few cents on the idea of ansible-lint modifying and not only linting. It will be an option, so it does not just suddenly start modifying, right? Run ansible-lint and only when run with ansible-lint -fmt (or something) will it actually do changes?

The plan is to make the `--write` option an opt-in, so by default it will not perform reformatting, at least for the next major version.

AndreasDavour:
> > I just followed a link in the Bullhorn and feel like I'm tossed into a context I really feel unfamiliar about. But, I wanted to toss in a question, and a few cents on the idea of ansible-lint modifying and not only linting. It will be an option, so it does not just suddenly start modifying, right? Run ansible-lint and only when run with ansible-lint -fmt (or something) will it actually do changes?
> 
> The plan is to make the `--write` option an opt-in, so by default it will not perform reformatting, at least for the next major version.

Excellent. I will then sleep soundly. Note, I like the suggestion *a log*, as the idea is sound.

cognifloyd:
@AndreasDavour Could you re-read the PR description? I just edited it to so it provides more context and highlights that this will be **opt-in**. What else can I clarify?

rndmh3ro:
> @AndreasDavour Could you re-read the PR description? I just edited it to so it provides more context and highlights that this will be **opt-in**. What else can I clarify?

Very clear to me!
And I like this change.

cognifloyd:
Rebased on main. Also includes #1884 which should be merged first.

AndreasDavour:
> @AndreasDavour Could you re-read the PR description? I just edited it to so it provides more context and highlights that this will be **opt-in**. What else can I clarify?

I think that looks perfectly fine. Thanks.

cognifloyd:
#1884 was merged into main. I rebased a bit to include that.

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,253 @@
...
+                # ruamel.yaml only preserves empty (no whitespace) blank lines.
+                # So, drop spaces in whitespace-only lines. ("\n  \n" -> "\n\n")
+                data = _whitespace_only_lines_re.sub("", cast(str, data))
+                # load_data has an lru_cache, so using it should be cached vs using YAML().load() to reload
+                data = load_data(data)
+                with StringIO() as stream:
+                    # ruamel.yaml does not have `dumps`. We have to give it a path or a stream.
+                    self.yaml.dump(data, stream)
+                    file.content = self._final_yaml_transform(stream.getvalue())
+                file.write()
```
cognifloyd, line 253: @lilatomic How's this? Setting the content is now separate from writing the file. What else would you do to this?

src/ansiblelint/transformer.py
```diff
@@ -0,0 +1,253 @@
...
+
+        # NB: ruamel.yaml does not have typehints, so mypy complains about everything here.
+
+        # Ansible uses PyYAML which only supports YAML 1.1. ruamel.yaml defaults to 1.2.
+        # So, we have to make sure we dump yaml files using YAML 1.1.
+        # We can relax the version requirement once ansible uses a version of PyYAML
+        # that includes this PR: https://github.com/yaml/pyyaml/pull/555
+        yaml.version = (1, 1)  # type: ignore[assignment]
+        # Sadly, this means all YAML files will be prefixed with "%YAML 1.1\n" on dump
+        # We'll have to drop that using a transform so people don't yell too much.
```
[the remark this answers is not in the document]
ssbarnea: AFAIK, can't handle YAML 1.2 is not true because YAML 1.2 is a stricter set. Any tool that can load 1.1 can also load 1.2 the same.

1.1 gave users too much artistic freedom and one major fix in 1.2 is less freedom. This means that 1.1 documents can be converted to 1.2 without being worried about breaking them.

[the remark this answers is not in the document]
cognifloyd: For booleans, yes, it is a stricter set (which is great!). But numbers, especially octals, and some time parsing had their regexes modified in the 1.2 spec. So an octal in 1.2 (`0o755`) is considered a string in 1.1. And 1.2 does not recognize the 1.1 style octal (`0755`), and treats it as an int.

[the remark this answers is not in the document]
bcoca: I didn't say the issues were due to YAML 1.2, just that ruamel had generated YAML that Ansible (using libyaml) was unable to parse.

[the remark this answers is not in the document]
bcoca: the case of octals would have created different errors, but those would be seen after parsing (which did happen, but that was in addition to the other issue)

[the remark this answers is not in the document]
bcoca: in any case i was time constrained and just opted to loose comments and used pyyaml (a sure bet) vs trying to debug ruamel issues, so no idea how easy/hard it would be to handle them.

cognifloyd:
I just summarized the UX options and discussion in https://github.com/ansible-community/ansible-lint/discussions/1910. Please discuss the UX there.

bcoca:
for ux, `--conform-to <profile>` might be my preferred, assumes linting profiles and if no profile, use default linting profile/rules.

src/ansiblelint/transformer.py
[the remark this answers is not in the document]
cognifloyd: K. Well, when someone runs across one of these issues (assuming this YAML 1.1 fix doesn't prevent them), I will debug it.

`ansible-lint` depends on comments to skip some rules, so people would get very angry if we started discarding their skip comments. So, discarding comments is not really an option for adding these features.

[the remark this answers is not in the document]
bcoca: not suggesting you do so, I'm just explaining the tradeoff I had to make during migration and the lack of investigation into the causes since I really did not have time to spare. I just gave a friendly warning that I had encountered these issues back then and that they merit investigation.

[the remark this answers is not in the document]
cognifloyd: Makes sense! Thank you!

cognifloyd:
I switched from `--fmt-yaml` to `--write` since 2 people like it (@ssbarnea and @lilatomic), I'm fine with that, and no one has said they dislike it. Every other suggestion has been disliked by at least one person.

ssbarnea:
I think it would be a good idea to implement the `--write` option in a separated PR and make it rewrite files while not doing anything to them. That should also lower the size of the current PR.

cognifloyd:
@ssbarnea done in #1926

ssbarnea:
Is this still needed or we can close it now?

cognifloyd:
I need to rebase/refactor to use the components already merged. This is a very noisy PR so I'll submit that as a new PR.

PR #1833, opened 2022-01-26
Add `AnsibleLintRule.matchrawtask()`

cognifloyd:
This allows rules to opt-in to an alternate version of `matchtask()`.
To use it, rules set `needs_raw_task = True` at the class level.
That triggers using `matchrawtask()` instead of `matchtask()`.

For one of the rules I'm developing, I need to inspect the raw (not normalized via ansible) task to identify the target issue.
This rule flags tasks that use action shorthand. `normalize_task()`, however, parses the shorthand and puts it in `task['action']`, so there's no good way to detect the action shorthand with the current `matchtask()` interface. There are a variety of awkward workarounds like parsing the playbooks myself, but I would really like to take advantage of all the work `ansible-lint` already does to find the relevant tasks and skip them as needed.
https://github.com/cognifloyd/ansible-lint/blob/fmt/src/ansiblelint/rules/TaskNoActionShorthand.py

For more discussion about this no-action-shorthand rule, see: https://github.com/ansible-community/ansible-lint/discussions/1609

This is extracted from #1805 (the TaskNoActionShorthand rule will also be extracted from #1805).

I created an alternative implementation in #1834. Only one of the PRs should be merged.

### DIFF
--- a/src/ansiblelint/_internal/rules.py
+++ b/src/ansiblelint/_internal/rules.py
@@ -22,6 +22,7 @@ class BaseRule:
     severity: str = ""
     link: str = ""
     has_dynamic_tags: bool = False
+    needs_raw_task: bool = False
 
     def getmatches(self, file: "Lintable") -> List["MatchError"]:
         """Return all matches while ignoring exceptions."""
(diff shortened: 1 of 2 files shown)


ssbarnea:
I am wondering if it would not be better to avoid adding a new method to base Rule class and just a configurable property for the class that defines the preference for receiving raw tasks.

If we go this route it would need only one line to make a rule run on raw instead of processed task and avoid increasing the number of methods that may be needed.

Don't take this as a request to change, is only a question. @webknjaz @tadeboro WDYT? Do you think we will ever need both methods on the same rule? If not it are likely better with a single method.

cognifloyd:
When matching rawtask, I still need the normalized task to simplify extracting the module name. So, I need both.

cognifloyd:
@ssbarnea you asked:
> Do you think we will ever need both methods on the same rule?

***No.*** When I need the raw_task, I also need the normalized task to help in inspecting the raw_task (eg I need `task["action"]["__ansible_original_module__"]` when I'm looking up module params in the raw_task.) So, any rule that needs raw_task should also have the normalized task. Therefore, rules should not define both methods.

> avoid adding a new method to base Rule class and just a configurable property for the class that defines the preference for receiving raw tasks.
> If we go this route it would need only one line to make a rule run on raw instead of processed task and avoid increasing the number of methods that may be needed.

This sounds a lot like the alternate implementation in #1834. Have you checked that one out yet?

cognifloyd:
Closing in favor of #1834

PR #1834, opened 2022-01-26
Allow rules to request original raw task data

cognifloyd:
This allows a rule to opt-in to receiving a "raw" copy of the task (as it is before normalization) via `task["__raw_task__"]`.
To use it, rules set `needs_raw_task = True` at the class level.

For one of the rules I'm developing, I need to inspect the raw (not normalized via ansible) task to identify the target issue.
This rule flags tasks that use action shorthand. `normalize_task()`, however, parses the shorthand and puts it in `task['action']`, so there's no good way to detect the action shorthand with the current `matchtask()` interface. There are a variety of awkward workarounds like parsing the playbooks myself, but I would really like to take advantage of all the work `ansible-lint` already does to find the relevant tasks and skip them as needed.
https://github.com/cognifloyd/ansible-lint/blob/fmt/src/ansiblelint/rules/TaskNoActionShorthand.py

For more discussion about this no-action-shorthand rule, see: https://github.com/ansible-community/ansible-lint/discussions/1609

The TaskNoActionShorthand rule will be extracted from #1805.

This is an alternative implementation to #1833 which was extracted from #1805. Only one of the PRs should be merged.


ssbarnea (approved):

ssbarnea:
I like this approach more than the alternative but you need to also add a simple test for it, one that verifies that a rule that requests this does receive it. This will also address the problem of decreasing the code coverage.

I think we already have a custom rule inside tests that you may reuse for testing this, or you can add a new one.

test/TestLintRule.py
```diff
@@ -18,23 +18,28 @@
...
+@pytest.fixture
+def lintable() -> Lintable:
+    """Return a playbook Lintable for use in this file's tests."""
+    return Lintable('examples/playbooks/ematcher-rule.yml', kind="playbook")
 
-    def test_rule_postmatching(self) -> None:
-        rule = UnsetVariableMatcherRule.UnsetVariableMatcherRule()
-        lintable = Lintable('examples/playbooks/bracketsmatchtest.yml', kind="playbook")
-        matches = rule.matchlines(lintable)
-        assert len(matches) == 2
```
cognifloyd, line 40: This used to be a test for `postmatch()` which doesn't exist any more. https://github.com/ansible-community/ansible-lint/commit/a4c6725aa4d38e7dca33b428f61cf1c885adc4f6#diff-d64893433e4ff6f626cc9b23e20b781c3ebb65b4df183c75dd536461f1076dc4R15

So, this `test_rule_postmatching` test duplicates `test_rule_matching`. So I deleted it.

cognifloyd:
K. I added a test Rule that uses this and adjusted the counts for RulesCollection since that uses the new rule now too. codecov is happier now.

cognifloyd:
The test errors are failures to upload codecov reports.

ssbarnea:
I seen it and did a recheck. If it happens again let's make the uploading optional. Even if we fail to upload half of the files we will likely see any coverage drop do no problems.

cognifloyd:
Tests are green :)

PR #1837, opened 2022-01-28
Strip ./ from discover_lintables results

sebix:
calling WcMatch('.', ...) causes the resulting file names to be prefixed
with `./`.
If the tests are run in a git-repository, this is not an issue, as git
is preferred of WcMatch in the code.

Therefore remove the `./` prefix from the WcMatch return value to get
the identical expected output.

fixes ansible-community/ansible-lint#1836

### DIFF
--- a/src/ansiblelint/file_utils.py
+++ b/src/ansiblelint/file_utils.py
@@ -261,13 +261,20 @@ def discover_lintables(options: Namespace) -> Dict[str, Any]:
     if out is None:
         exclude_pattern = "|".join(str(x) for x in options.exclude_paths)
         _logger.info("Looking up for files, excluding %s ...", exclude_pattern)
-        out = set(
-            WcMatch(
+        # remove './' prefix from output of WcMatch
+        out = {
+            strip_dotslash_prefix(fname)
+            for fname in WcMatch(
                 '.', exclude_pattern=exclude_pattern, flags=RECURSIVE, limit=256
             ).match()
-        )
+        }
 
     return OrderedDict.fromkeys(sorted(out))
+
+
+def strip_dotslash_prefix(fname: str) -> str:
+    """Remove ./ leading from filenames."""
+    return fname[2:] if fname.startswith('./') else fname
 
 
 def guess_project_dir(config_file: Optional[str]) -> str:


sebix:
I'm not sure what's the best way to make mypy happy. `out` will still always contain only strings, not anything else. Applying a function (with string -> string) does not change the type.

ssbarnea (approved):

src/ansiblelint/file_utils.py
```diff
@@ -261,10 +271,15 @@ def discover_lintables(options: Namespace) -> Dict[str, Any]:
     if out is None:
         exclude_pattern = "|".join(str(x) for x in options.exclude_paths)
         _logger.info("Looking up for files, excluding %s ...", exclude_pattern)
+        # remove './' prefix from output of WcMatch
+        strip_suffix: Callable[[str], str] = lambda fname: fname[2:]
```
felixfontein, line 275: What happens if for whatever reason `fname` does **not** start with `./`? Then this will chomp off something important.

I think it's better to make this a proper function and only cut the first two characters off if the string really starts with `./`.

src/ansiblelint/file_utils.py
```diff
@@ -261,10 +271,15 @@ def discover_lintables(options: Namespace) -> Dict[str, Any]:
...
         out = set(
-            WcMatch(
-                '.', exclude_pattern=exclude_pattern, flags=RECURSIVE, limit=256
-            ).match()
+            map(
+                strip_suffix,
+                WcMatch(
+                    '.', exclude_pattern=exclude_pattern, flags=RECURSIVE, limit=256
+                ).match(),
+            )
```
felixfontein, line 282: Why not a set comprehension instead, something like
```.py
out = {strip_suffix(path) for path in WcMatch(...).match()}
```
?

sebix:
Thanks for the feedback. I refactored the code as suggested with a separate function and a set comprehension. That also obsoleted the import of `typing.Callable` which I reverted therefore.

felixfontein (commented):
The changes in src/ansiblelint/file_utils.py look good to me.

I'm not sure what the other changes are about though. They seem to be unrelated to this PR.

sebix:
> I'm not sure what the other changes are about though. They seem to be unrelated to this PR.

Now rebase on updated main. Looks better :)

felixfontein (approved):

sebix:
... and now also fixed this check fail:
```
src/ansiblelint/file_utils.py:276:1: D400 First line should end with a period
```

PR #1838, opened 2022-01-30
chore: remove black custom config

cognifloyd:
Make it more consistent!

I use black formatting on a lot of projects. So I was surprised when this project used black but had inconsistent strings.


webknjaz (changes requested):
I'm against it because we need a better set of formatters than black to replace it with something sane. I'd think twice about forcing people to do double quotes, for example. It would be reasonable to have https://github.com/asottile/add-trailing-comma.git, though.

ssbarnea (approved):

cidrblock (approved):
Seems sane to me.  

Although I was not part of the original black discussion, this appears to be a switch from single to double quotes, not a PR related to the use of black.

Can another Pr be done to remove `skip-string-normalization = false` as it is the defualt?

ssbarnea:
> Seems sane to me.
> 
> Although I was not part of the original black discussion, this appears to be a switch from single to double quotes, not a PR related to the use of black.
> 
> Can another Pr be done to remove `skip-string-normalization = false` as it is the defualt?

I made a push to remove the custom config as is no longer needed.

webknjaz:
> Although I was not part of the original black discussion

This is because there never was a discussion, it was imposed as Sorin's personal opinion right when I burnt out and gave up maintaining lint (which was very much related).

greg-hellings:
Consistency is good, and consistency that is checked, fixed, and enforced by a static tool like `black` is best of all. Even if black does not fit everyone's exact preference, it presents a powerful and useful set of tools that easily tie in with CI systems to ensure that code is consistent. I've almost come to appreciate Python formatting when Black is applied to a code base. As it has become pretty standard in the community I think it's good for us to use it and even better to push towards the defaults when we do not have a compelling reason to avoid them.

:+1:  from me.

webknjaz:
I think preventing readable code in some cases and standing in a way of development is a compelling enough reason not to use it. There are better ways of doing similar things that can be configured and accepted with good settings.
Also, it's not "default", it's just popular. For many reasons. But it doesn't mean that its good qualities overweight its terrible sides.
Its "opinionated" repressive ways of bullying coders are what keeps driving me away from black projects, for example (including but not limited to this one). Aside from other dangerous properties like encouraging formatting over quality/readability/maintainability.
It also prevents people from writing some of PEP 8 compliant code, even though it is actually a widely accepted standard.
Yes, consistency is good. _Very_ good. But when it comes to encouraging bad practices and turning a blind eye, "somebody else uses it and doesn't complain" is never a good enough excuse to do such things.
I keep seeing people judging whether code is good by checking whether it's formatted. While the autoformatting bit is rather dangerous with such a mindset because it prevents useful linting alerts from hinting people about bad architectural/structural decisions.
Long line? No problem — now it's formatted. And who cares that it's 10 levels deep in nesting and nobody will ever be able to understand that code being scared to change a byte in it.
Need to change formatting a little bit _within the limits of PEP 8_ to make code actually convey the intent of a human behind it? Nope, a good intent will become trashed and turned into a mess anyway. Who cares about understanding why a block of code does what it does? Nobody, right?
Such changes should be mindful and make actual sense but I've only seen a few people thinking ahead, trying to follow the clean code practices and keep the projects maintainable.
So no, having _black_ specifically is not a good idea at all. There are other tools that can be combined to do this job better and allow the required level of control over making better decisions for the sake of quality, not just formatting.

apatard (approved):
personally, I don't care that much about quotes but, as other says, for consistency, it's better to not have a special black configuration.

ssbarnea:
I will merge this as we already have at least 5 supporters and lock the thread to avoid more noise.

@webknjaz Regarding the debate about use or not of black in our projects, feel free to open a proposal in devtools project and add it to our **team meeting agenda**.

The reality is that merging this change [all devtools projects](https://github.com/ansible-community/devtools/blob/main/config/devtools.yml) (~30) will be in sync when it comes to black use with default settings.

Anyone is welcome to use the kind of quotes they can find easier to type when they make a pull request, as our pre-commit.ci job will just fix them. They don't even have to worry about doing any leg-work.

PR #1850, opened 2022-02-04
Remove support for py36 and py37

ssbarnea:
As part of reduce testing complexity and getting ansible-lint in sync with `ansible-core`, we will require `python_version>=3.8`.

This change marks the start of `6.x` and once is merged we will create a `stable/5.x` for the last comment before it. That branch will be used to backport changes needed for the older versions, at least until we can make a full release of 6.x.

One additional reason for making this move is a recent explosion of dependencies dropping support for py36 or even py37: `pytest`, `coverage`, `pip-compile`, `tomli`.

Please mark your support or lack of on this change as we are aware that it may upset some users that are still stuck on older platform as it will be harder for them to get bugfixes.

Related: https://github.com/ansible-community/molecule/pull/3436

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -112,8 +112,6 @@ jobs:
       matrix:
         python-version:
         # keep list sorted as it determines UI order too
-        - 3.6
-        - 3.7
         - 3.8
         - 3.9
         - "3.10"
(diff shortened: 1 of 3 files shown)


webknjaz:
I think this change should add a config for Patchback, then. And maybe for `cherry-pick` too.

ssbarnea:
> I think this change should add a config for Patchback, then. And maybe for `cherry-pick` too.

That could count as **scope creeping**. Anything that is not directly needed should go in follow-ups.

In fact, I suppose that backports will likely be done manually as I expect to see the codebase diverge.

tadeboro (approved):

ganeshrn (approved):

cidrblock (approved):

greg-hellings:
+1 from me.

Issue #1615, opened 2021-06-07
Inconsistent behaviour builtin vs. FQCN import_tasks


devNan0:
##### Summary
A unnamed `import_tasks` task will pass ansible-lint
A unnamed `ansible.builtin.import_tasks` task will raise an `unnamed-task` failure

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```
ansible-lint 5.0.12 using ansible 2.10.10
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
Debian 10

##### STEPS TO REPRODUCE
```
# Will raise 'unnamed-task' failure
- ansible.builtin.import_tasks: foobar.yml

# Will pass
- import_tasks: foobar.yml

```

##### Desired Behaviour
Both ways should pass without an error

##### Actual Behaviour
```
unnamed-task: All tasks should be named
tasks/main.yml:3 Task/Handler: import_tasks  foobar.yml
```

ryaner:
This looks to be caused because the tasks are normalized after `get_action_tasks`(https://github.com/ansible-community/ansible-lint/blob/0837dce8466331dd8e3996dd0785348756fd4370/src/ansiblelint/utils.py#L705) is called, resulting in the functions not matching.
Extending the list in there to include the FQCN imports makes everything consistent again. I can't see a safe way to normal before that function.

PR #1854, opened 2022-02-07
Make FQCN import tasks include their relevant files for checking

ryaner:
Tasks can come in in either the short format or the FQCN format
as this is called before noramlize_task functions.


ryaner:
Another option for fixing would be to split the string down on L328 `(k, v) = item`.
`k` can either be `import_tasks` or `ansible.builtin.import_tasks`.
Including the cases where the functions have FQCNs is slightly faster given the number of times the function is called.

ssbarnea (changes requested):
The fix seems logic and thanks for creating the PR.

Also, please update the tests in order to include a regression prevention measure. Think about a change, even to existing test covering the same functionality, that would fail if we remove your fix.

ryaner:
I combined in the fix for #1615 as adding the FQCN into the tests caused the `All tasks should be named` to trigger during the tests. If you'd prefer I separate things I can do so, making this dependant on a PR for #1615.

This does not fix the fact that `import_` calls do not trigger the FQCN ruleset. I don't see a way to pass the tasks to only a subset of the rulesets.

ssbarnea (approved):

PR #1857, opened 2022-02-09
test: replace eco playbook with pytest tests

ssbarnea:
Replace `playbooks/eco.yml` with pure pytest implementation. This improves the output and out ability to debug it.

Needed-By: #1854


cidrblock (approved):

test/eco/colsystem.result
```diff
@@ -0,0 +1,6 @@
+load-failure: [Errno 2] No such file or directory: '/Users/ssbarnea/.cache/ansible-lint-eco/colsystem/tests/ansible-lint.yml' (filenotfounderror)
```
cognifloyd, line 1: Is this going to cause false positives when running in other people's medicines?

Issue #1507, opened 2021-04-05
Ansible-lint does not look for ansible binary inside venv


noonedeadpunk:
##### Summary
When ansible-lint is installed inside venv, it is expected that ansible-lint will try to find ansible binary inside venv as well, even when venv is not activated. Instead it looks inside $PATH only.
This results in the following error, when you have ansible 2.9 installed by system packages, and ansible 2.10 along with ansible-lint inside the venv:
`ERROR    FATAL: Ansible CLI (2.9.6) and python module (2.10.7) versions do not match. This indicates a broken execution environment.`

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ~/.virtualenvs/test-lint/bin/ansible-lint --version
ansible-lint 5.0.6 using ansible 2.9.6
FATAL: Ansible CLI (2.9.6) and python module (2.10.7) versions do not match. This indicates a broken execution environment.

$ ~/.virtualenvs/test-lint/bin/ansible --version
ansible 2.10.7
  config file = /home/dmitriy/.ansible.cfg
  configured module search path = ['/home/dmitriy/.ansible/roles/config_template/library', '/home/dmitriy/.ansible/plugins/library', '/home/dmitriy/.ansible/roles/ceph-ansible/library']
  ansible python module location = /home/dmitriy/.virtualenvs/test-lint/lib/python3.8/site-packages/ansible
  executable location = /home/dmitriy/.virtualenvs/test-lint/bin/ansible
  python version = 3.8.5 (default, Jan 27 2021, 15:41:15) [GCC 9.3.0]
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
Python 3.8.5
Ubuntu 20.04

##### STEPS TO REPRODUCE

```console (paste below)
$ virtualenv ~/.virtualenvs/test-lint/
$ ~/.virtualenvs/test-lint/bin/pip install ansible-lint[community,yamllint]
$ ~/.virtualenvs/test-lint/bin/ansible-lint --version
ansible-lint 5.0.6 using ansible 2.9.6
FATAL: Ansible CLI (2.9.6) and python module (2.10.7) versions do not match. This indicates a broken execution environment.
$ source ~/.virtualenvs/test-lint/bin/activate
(test-lint) $ ~/.virtualenvs/test-lint/bin/ansible-lint --version
ansible-lint 5.0.6 using ansible 2.10.7
(test-lint) $
```

[...]

PR #1860, opened 2022-02-10
Allow use without activation of virtualenv

ssbarnea:
Allow users to call ansible-lint with a full path even when it is installed inside an inactive virtualenv.

This change also includes a fail-safe test that prevents future regressions.

Fixes: #1507
Closes: #1866 (duplicate)

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -25,6 +25,7 @@ import hashlib
 import logging
 import os
 import pathlib
+import shutil
 import subprocess
 import sys
 from contextlib import contextmanager
@@ -138,6 +139,9 @@ def _do_list(rules: "RulesCollection") -> int:
 
 def main(argv: Optional[List[str]] = None) -> int:
     """Linter CLI entry point."""
+    # alter PATH if needed (venv support)
+    path_inject()
+
     if argv is None:
         argv = sys.argv
     initialize_options(argv[1:])
@@ -248,6 +252,27 @@ def _run_cli_entrypoint() -> None:
         sys.exit(EXIT_CONTROL_C_RC)
     except RuntimeError as e:
         raise SystemExit(str(e))
+
+
+def path_inject() -> None:
+    """Add python interpreter path to top of PATH to fix outside venv calling."""
+    # This make it possible to call ansible-lint that was installed inside a
+    # virtualenv without having to pre-activate it. Otherwise subprocess will
+    # either fail to find ansible executables or call the wrong ones.
+    #
+    # This must be run before we do run any subprocesses, and loading config
+    # does this as part of the ansible detection.
+    paths = [x for x in os.environ.get("PATH", "").split(os.pathsep) if x]
+    ansible_path = shutil.which("ansible")
+    if ansible_path:
+        ansible_path = os.path.dirname(ansible_path)
+    py_path = os.path.dirname(sys.executable)
+    # Determine if we need to manipulate PATH
+    if ansible_path not in paths or py_path != ansible_path:  # pragma: no cover
+        # tested by test_call_from_outside_venv but coverage cannot detect it
+        paths.insert(0, py_path)
+        os.environ["PATH"] = os.pathsep.join(paths)
+        print(f"WARNING: PATH altered to include {py_path}", file=sys.stderr)
 
 
 if __name__ == "__main__":
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -1,0 +1,33 @@
+"""Tests related to ansiblelint.__main__ module."""
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+
+@pytest.mark.parametrize(
+    ("expected_warning"),
+    (False, True),
+    ids=("normal", "isolated"),
+)
+def test_call_from_outside_venv(expected_warning: bool) -> None:
+    """Asserts ability to be called w/ or w/o venv activation."""
+    env = None
+    if expected_warning:
+        env = {"HOME": Path.home()}
+    py_path = os.path.dirname(sys.executable)
+    # Passing custom env prevents the process from inheriting PATH or other
+    # environment variables from the current process, so we emulate being
+    # called from outside the venv.
+    proc = subprocess.run(
+        [f"{py_path}/ansible-lint", "--version"],
+        check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+        env=env,
+    )
+    warning_found = "PATH altered to include" in proc.stderr
+    assert warning_found is expected_warning


cognifloyd (approved):

src/ansiblelint/_preimport.py
```diff
@@ -0,0 +1,11 @@
+"""Code that must run before other imports."""
+import os
+import sys
+
+# Hack to allow a venv installed ansible-lint be called from outside the venv
+paths = [x for x in os.environ.get("PATH", "").split(":") if x]
```
webknjaz, line 6: Use `os.path.sep` instead.

src/ansiblelint/_preimport.py
```diff
@@ -0,0 +1,11 @@
...
+py_path = os.path.dirname(sys.executable)
+if py_path not in paths:
+    paths.insert(0, py_path)
+    os.environ["PATH"] = ":".join(paths)
```
webknjaz, line 10: Same, use `os.path.sep`

src/ansiblelint/__main__.py
```diff
@@ -32,11 +32,15 @@
 
 from enrich.console import should_do_markup
 
+# pylint: disable=unused-import
+import ansiblelint._preimport  # must be kept before other ansiblelint imports!
```
webknjaz, line 36: This is a dirty hack. Instead, expose a function and call it at the beginning of the runtime.

ssbarnea: I refactored it to load from config, at the same location where we do something similar, as that command does call ansible.

cidrblock (approved):

src/ansiblelint/config.py
```diff
@@ -164,5 +165,26 @@ def ansible_version(version: str = "") -> Version:
...
+        ansible_path = os.path.dirname(ansible_path)
+    py_path = os.path.dirname(sys.executable)
+    # Determine if we need to manipulate PATH
+    if ansible_path not in paths or py_path != ansible_path:
+        paths.insert(0, py_path)
+        os.environ["PATH"] = os.pathsep.join(paths)
+        print(f"WARNING: PATH altered to include {py_path}", file=sys.stderr)
+
+
+path_inject()
```
webknjaz, line 188: I don't like this hack of patching runtime stuff during import-time. Why can't it be the first thing called in the entry point?

test/test_config.py
```diff
@@ -0,0 +1,31 @@
...
+        universal_newlines=True,
+    )
+    assert "PATH altered to include" not in proc.stderr
+
+    # env -i is used in order to prevent the process from inheriting PATH or
+    # other environment variables from the current process, so we emulate being
+    # called from outside the venv.
+    proc = subprocess.run(
+        f'env -i HOME="{Path.home()}" {py_path}/ansible-lint --version',
+        shell=True,
```
webknjaz, line 26: I'd just use the `env` arg for this instead of requiring shell mode.
```suggestion
        (f"{py_path}/ansible-lint", "--version"),
        env={
            "HOME": Path.home(),
        },
```

src/ansiblelint/config.py
ssbarnea: That is because we have lots of imports before that https://github.com/ansible-community/ansible-lint/blob/main/src/ansiblelint/__main__.py#L38 

You are welcomed to find a better way to do it in a follow-up. At least this change keeps a single place where we run code at import, not adding a new one.

src/ansiblelint/config.py
```diff
@@ -164,5 +165,27 @@ def ansible_version(version: str = "") -> Version:
...
+    ansible_path = shutil.which("ansible")
+    if ansible_path:
+        ansible_path = os.path.dirname(ansible_path)
+    py_path = os.path.dirname(sys.executable)
+    # Determine if we need to manipulate PATH
+    if ansible_path not in paths or py_path != ansible_path:  # pragma: no cover
+        # tested by test_call_from_outside_venv but coverage cannot detect it
+        paths.insert(0, py_path)
+        os.environ["PATH"] = os.pathsep.join(paths)
+        print(f"WARNING: PATH altered to include {py_path}", file=sys.stderr)
```
webknjaz, line 186: Could you use a warning subsystem instead?
```suggestion
        warnings.warn(f"PATH altered to include {py_path}", UserWarning, stacklevel=2)
```

src/ansiblelint/config.py
webknjaz: Following this approach makes the problem of relying on multiple unobvious side effects even worse than it is now. Also, I don't see why putting this in `main()` wouldn't work for the current PR.

test/test_config.py
```diff
@@ -0,0 +1,31 @@
+"""Tests related to ansiblelint.config module."""
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+
+def test_call_from_outside_venv() -> None:
```
webknjaz, line 8: I think the copy-paste could be reduced with `@pytest.mark.parametrize` since it's essentially one test. This would also help make the test variants more accurate — the first one relies on unstable side-effects, for example.

src/ansiblelint/config.py
webknjaz: Per https://docs.python.org/3/library/warnings.html:
> Warning messages are normally written to [sys.stderr](https://docs.python.org/3/library/sys.html#sys.stderr), but their disposition can be changed flexibly, from ignoring all warnings to turning them into exceptions.

and `warnings.showwarning()`:
> Write a warning to a file. The default implementation calls `formatwarning(message, category, filename, lineno, line)` and writes the resulting string to file, which defaults to [sys.stderr](https://docs.python.org/3/library/sys.html#sys.stderr). You may replace this function with any callable by assigning to `warnings.showwarning`.

src/ansiblelint/config.py
```diff
@@ -164,5 +165,26 @@ def ansible_version(version: str = "") -> Version:
...
+def path_inject() -> None:
+    """Add python interpreter path to top of PATH to fix outside venv calling."""
+    # This make it possible to call ansible-lint that was installed inside a
+    # virtualenv without having to pre-activate it. Otherwise subprocess will
+    # either fail to find ansible executables or call the wrong ones.
+    #
+    # This must be run before we do run any subprocesses, and loading config
+    # does this as part of the ansible detection.
+    paths = [x for x in os.environ.get("PATH", "").split(os.pathsep) if x]
+    ansible_path = shutil.which("ansible")
```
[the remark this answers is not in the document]
ssbarnea: Not needed, always returns full paths.

test/test_config.py
ssbarnea: Done, but it was not really necessary.

test/test_config.py
```diff
@@ -0,0 +1,31 @@
...
+    # environment variables from the current process, so we emulate being
+    # called from outside the venv.
+    proc = subprocess.run(
+        [f"{py_path}/ansible-lint", "--version"],
+        check=True,
+        capture_output=True,
+        universal_newlines=True,
+        env=env,
+    )  # type: ignore # env confuses mypy
+    assert ("PATH altered to include" in proc.stderr) == isolated
```
webknjaz, line 31: For bool/None types, an identity check is recommended:
```suggestion
    warning_emitted = "PATH altered to include" in proc.stderr
    assert warning_emitted is isolated
```

ssbarnea:
@webknjaz I had a talk with @cidrblock and I will attempt to remove the code running at import from linter before doing this change. If it works we may be able to ditch that anti-pattern.

src/ansiblelint/config.py
```diff
@@ -164,5 +165,27 @@ def ansible_version(version: str = "") -> Version:
...
+    if ansible_path not in paths or py_path != ansible_path:  # pragma: no cover
+        # tested by test_call_from_outside_venv but coverage cannot detect it
+        paths.insert(0, py_path)
+        os.environ["PATH"] = os.pathsep.join(paths)
+        print(f"WARNING: PATH altered to include {py_path}", file=sys.stderr)
+
+
+path_inject()
 if ansible_collections_path() in os.environ:
     collection_list = os.environ[ansible_collections_path()].split(":")
```
cognifloyd, line 191: @webknjaz The import-time side effects are here. `ansible_collections_path()` calls `ansible_version()` which runs `ansible --version` in a subprocess. So, running `path_inject()` in `main()` is too late. Changing things so that `collection_list` is not created at import time is out-of-scope for this PR.

cognifloyd (approved):

src/ansiblelint/config.py
ssbarnea: Use of warnings module is outside the scope.

PR #1864, opened 2022-02-11
Simplify `App.report_outcome()`

cognifloyd:
Simplify `App.report_outcome()` to facilitate expanding reported output in #1828.
Plus it makes the logic just a bit clearer.


ssbarnea (approved):
LGTM, @webknjaz WDYT?

PR #1866, opened 2022-02-11
Run in an un-activated virtual environment, fixes #1507

### LINKED ISSUE #1507: Ansible-lint does not look for ansible binary inside venv

kpinc:
Fix so the command line ansible-lint will run in an un-activated virtual environment, via a pathname supplied to the shell.  Fixes issue #1507.

THEORY:
The foundational observation is that because activating a virtual environment does only one thing, add a path to the venv's $PATH, the only thing we know about an un-activated venv is that the venv[0] of it's process is either not in a $PATH directory or, if it is, then the venv is, in effect, activated.  Therefore the only way to possibly test for virtual environment activation is to check venv[0] against $PATH. 

The underlying premise is that if it is necessary to specify a path, one that contains directories, to execute the python which runs ansible-lint, then the other ansible-* commands run by ansible-lint reside in the os.path.basedir() of that directory.  Alternatively, given that the operational test to determine if ansible-lint is running in an un-activated virtual environment is whether a path containing a directory designation is required to execute the running ansible-lint, then, having decided that this is the case, use the directory designation to execute the other ansible-* commands of the ansible-lint's virtual environment.

The approach taken here is a heuristic, one which I believe is explicit and clear.  The alternative, as far as I've been able to imagine, is to fallback to trying to execute in an un-activated venv as a fallback after the search through $PATH fails.   (Or vice-versa.) This also is a heuristic.  Comparison without code is uninteresting, but it seems to me that the corner cases that fail using either heuristic would involve the case of someone running some ansible-* commands in some system/virtual environments and some in another.  It might make sense to simply declare this a configuration that is not supported.

PRACTICE:
Obtains directory portion of sys.executable and checks if it is an element of $PATH.  If so the system can rely on $PATH to discover all executables; the original codepath is used and basenames are the argv[0] given to subprocess.run().  If not then the Python binary was run with a non-trivial pathname, as when executing a binary in an un-activated virtual environment's bin directory; the directory portion of sys.executable is prefaced to argv[[0] when executing ansible-related commands.

Testing is done by monkeypatching to remove the element of $PATH which tox adds to activate the virtual environment.  Existing tests which run the ansible-lint executable were modified to run twice, once as-is and a 2nd time with the virtual environment deactivated.  

Further, to directly test what a user would do, one test was also modified to, in addition to existing codepaths, run the ansible-lint executable installed in the tox virtual environment.  The existing test runs via 'python -m ansiblelint'.

### DIFF
--- a/src/ansiblelint/venv_utils.py
+++ b/src/ansiblelint/venv_utils.py
@@ -1,0 +1,54 @@
+"""Utilities supporting virtual environments."""
+
+# We run various ansible command line tools (and detect the ansible version
+# by executing `ansible --version`).
+# The executable we want to run may not be in $PATH, especially if we
+# are running in a virtual environment that's not been activated.
+
+import os
+import pathlib
+import subprocess
+import sys
+from functools import lru_cache
+from typing import TYPE_CHECKING, Any, List
+
+if TYPE_CHECKING:
+    # https://github.com/PyCQA/pylint/issues/3240
+    # pylint: disable=unsubscriptable-object
+    CompletedProcess = subprocess.CompletedProcess[Any]
+else:
+    CompletedProcess = subprocess.CompletedProcess
+
+
+def _normalize(path: str) -> pathlib.Path:
+    # pathlib.Path takes care of MS Windows case-insensitivity
+    return pathlib.Path(os.path.realpath(os.path.normpath(os.path.expanduser(path))))
+
+
+def _get_search_path() -> List[pathlib.Path]:
+    return [_normalize(path) for path in os.get_exec_path()]
+
+
+@lru_cache()
+def _get_venv_path() -> str:
+    venv_dir = _normalize(os.path.dirname(sys.executable))
+    search_path = _get_search_path()
+    if venv_dir in search_path or search_path and search_path[0] == pathlib.Path("."):
+        return ""
+    return str(venv_dir)
+
+
+def add_venv_path(cmd: str) -> str:
+    """Prepend the path to the venv, if any, to the supplied command."""
+    return os.path.join(_get_venv_path(), cmd)
+
+
+def run_ansible_version() -> CompletedProcess:
+    """Run `ansible --version` and return the result."""
+    return subprocess.run(
+        args=[add_venv_path("ansible"), "--version"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+        check=False,
+    )
--- a/test/test_prerun.py
+++ b/test/test_prerun.py
@@ -1,7 +1,9 @@
 """Tests related to prerun part of the linter."""
 import os
+import re
 import subprocess
-from typing import List
+import sys
+from typing import Callable, List
 
 import pytest
 from _pytest.monkeypatch import MonkeyPatch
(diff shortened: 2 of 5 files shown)


kpinc:
I see that the patch proposed in issue #1860 also fixes #1507.  That  thread might be interested.

ssbarnea:
@kpinc Why do you think that approach is better than the one from https://github.com/ansible-community/ansible-lint/pull/1860 ?

I am bit surprised about the number of lines of code involved in as I was able to achieve (the same?) using far less. Please take a look at my PR and see if you can spot some fatal flaws.

test/test_prerun.py
```diff
@@ -11,35 +13,113 @@
...
+
+    def deactiveator(deactivate: bool) -> str:
+        if deactivate:
+            search_path = os.get_exec_path()
+            try:
+                pos = search_path.index(EXECDIR)
+            except ValueError:
+                return ""
+            del search_path[pos]
+            monkeypatch.setenv("PATH", ";".join(search_path))
```
webknjaz, line 34: I believe this meant to be
```suggestion
            monkeypatch.setenv("PATH", os.path.sep.join(search_path))
```
*nix separator is `:` normally while windows uses `;`

test/test_prerun.py
```diff
@@ -11,35 +13,113 @@
...
 from ansiblelint.testing import run_ansible_lint
 
+EXECDIR = os.path.dirname(sys.executable)
+
+Deactivator = Callable[[bool], str]
+DeactivateVenv = Callable[[], Deactivator]
+
+
+@pytest.fixture
+def deactivate_venv(monkeypatch: MonkeyPatch) -> Deactivator:
```
webknjaz, line 23: I like the use of `monkeypatch` but returning a callable seems like an overengineering: since all the tests call it at the beginning of their execution, this fixture could do this when injected.

Assuming you're worried/don't know how this works with `parametrize` — it's quite simple: just inject a `request` fixture here and access the parameter via `request.param`. And on the test function side, add `indirect=('deactivate_venv', )` argument into the `parametrize()` decorator. That's it!

test/test_prerun.py
```diff
@@ -11,35 +13,113 @@
...
+    """Deliver a function to deactivate the current venv, if any, if requested, by removing it's bin dir from $PATH."""
+
+    def deactiveator(deactivate: bool) -> str:
+        if deactivate:
```
webknjaz, line 27: Pro tip: this clause can be turned into a "guard expression" by flipping the check. Use  `if not request.param: return ""` to exit early and this will allow dedenting a whole bunch of code making the control flow look more linear.

test/test_prerun.py
```diff
@@ -11,35 +13,113 @@
...
+    # as-it-is before the fix of issue #1507 and after the fix.
+    # (See: https://github.com/ansible-community/ansible-lint/issues/1507)
+    # But these assertations are written as much like the "before" version
+    # of this test as possible, to demonstrate that the assertion
+    # itself is not substantively changed by the fix.
+    # When these asserations fail, without the #1507 patch,
+    # it demonstrates that ansible-lint does not run in a virtual environment
+    # unless that environment is activated.
+    assert re.search(
+        "Running [^ ]*ansible-galaxy role install", result.stderr
```
webknjaz, line 111: Why not do a more precise comparison with
```suggestion
        f"Running {execdir}/ansible-galaxy role install", result.stderr
```
?

test/test_prerun.py
ssbarnea: @webknjaz And you know very well that ansible-lint does not run on Windows.

src/ansiblelint/venv_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    return str(venv_dir)
+
+
+def add_venv_path(cmd: str) -> str:
+    """Prepend the path to the venv, if any, to the supplied command."""
+    return os.path.join(_get_venv_path(), cmd)
+
+
+def run_ansible_version() -> CompletedProcess:
+    """Run `ansible --version` and return the result."""
```
webknjaz, line 47: This will fix the failing docs build:
```suggestion
    """Run ``ansible --version`` and return the result."""
```

test/test_prerun.py
ssbarnea: And your suggestion is misleading as `os.path.sep` is `/` while `os.pathsep` is `:`.-- not sure if just using the `:` not not just be much easier to read.

A simple dot... totally different outcome.

PR #1875, opened 2022-02-12
Restore WSL1 testing with windows-2019

ssbarnea:
As we recently seen WSL job timing out on tox run without
giving any feedback we enforce a max run time, which is already
more than double the average run time.


ssbarnea:
I wonder if that has anything to do with current migration of `windows-latest` from `windows-2019` to `windows-2022`, still going and estimated to finish first week of March.

Maybe only one of these images is affected by this bug.

PR #1882, opened 2022-02-12
Clarify GPLv3 license of the distribution

ssbarnea:
- Ensure that GitHub advertises the project as GPLv3 in order to avoid misleading some users by adding COPYING file, matching ansible repository.
- Keep MIT status for its own-source code and future contributions
- Ensure that pypi metadata lists only GPLv3 as a license because the distribution includes direct GPL dependencies.
- From the technical point of view that involved move of MIT license file from `/LICENSE` to `docs/licenses/LICENSE.MIT.txt`. We tried to keep both licenses at repository root buy this is breaking GitHub licensing detection heuristic and prevents it from displaying "GPLv3" license hyperlink on the right sidebar, which was the main source of confusions so far.
- Corrects mistake in setup.cfg metadata where we used `GPLv3+` instead of `GPLv3`, while Ansible itself is only `GPLv3`, so the plus clause does not apply.

This is not a relicensing of the project and should be seen as an esthetic change aimed to clarify recurring confusions. The need to do this now is more important as we made `ansible-core`and yamllint direct dependencies via #1888 and #1881.

Please read https://github.com/ansible-community/ansible-lint/discussions/1832 for more details.


cidrblock (approved):

tadeboro (approved):

setup.cfg
```diff
@@ -19,7 +19,7 @@ author_email = will@thames.id.au
 maintainer = Ansible by Red Hat
 maintainer_email = info@ansible.com
 license = MIT
-license_file = LICENSE
+license_file = COPYING
```
webknjaz, line 22: Singular `license_file` is deprecated. Instead, use

```suggestion
license_files =
  COPYING
  licenses/*
```

Also note that if this option is omitted, it will fall back to including all the files matching well-known patterns meaning that our case may be covered by that as well.

README.rst
```diff
@@ -44,17 +44,12 @@ Please read `Contribution guidelines`_ if you wish to contribute.
 Licensing
 =========
 
-The code in the ansible-lint repository is licensed under the MIT_ license. If
-you contribute to this repository, that license applies to your contributions.
+The ansible-lint project is distributed as GPLv3_ due to use of GPLv3_ runtime
```
webknjaz, line 47:
```suggestion
The ansible-lint project is distributed under GPLv3_ due to use of GPLv3_ runtime
```

webknjaz (commented):
Also, I'd drop that `.txt` suffix from the file name. It could be called just `LICENSE.MIT`.

ssbarnea:
> Also, I'd drop that `.txt` suffix from the file name. It could be called just `LICENSE.MIT`.

I prefer to be explicit here. I also see @cidrblock did the same on https://github.com/ansible/ansible-navigator/tree/main/licenses

setup.cfg
```diff
@@ -18,8 +18,10 @@ author = Will Thames
 author_email = will@thames.id.au
 maintainer = Ansible by Red Hat
 maintainer_email = info@ansible.com
-license = MIT
-license_file = LICENSE
+license = GPLv3
+license_files =
+  COPYING
+  docs/licenses/LICENSE.MIT.txt
```
webknjaz, line 24: Why not use a wildcard?

ssbarnea: Done!

README.rst
```diff
@@ -66,5 +61,5 @@ ansible-lint was created by `Will Thames`_ and is now maintained as part of the
 .. _Will Thames: https://github.com/willthames
 .. _Ansible: https://ansible.com
 .. _Red Hat: https://redhat.com
-.. _MIT: https://github.com/ansible-community/ansible-lint/blob/main/LICENSE
-.. _GPLv3: https://github.com/ansible/ansible/blob/devel/COPYING
+.. _MIT: https://github.com/ansible-community/ansible-lint/blob/main/docs/licenses/LICENSE.mit.txt
+.. _GPLv3: https://github.com/ansible-community/ansible-lint/blob/main/COPYING
```
webknjaz, line 65: Given the future move back to @ansible, do we really want to update this?

webknjaz: Oh, it's the core repo. Then, it's fine, I guess..

webknjaz (approved):

PR #1884, opened 2022-02-13
Enable Lintables to be modified

cognifloyd:
We need a way to write files and keep track of how we've changed them.

This prepares for formatting in transforms in #1828. Feedback from that PR suggests that we'll need at least a count of the files that have changed and possibly a diff view that shows how files have changed.

This adds a setter for `Lintable.content` which monitors the content for changes..
When the content is changed, it sets `Lintable.updated = True`.
Original content is also preserved to facilitate further comparison or resetting the content if needed.

There is also a deleter for `Lintable.content` so that if something needs to modify a file externally, the developer can trigger reloading the content from disk with `del lintable.content`.

_NB: I originally used a stream/file-like object because that's what `ruamel.yaml`'s API expects. But that made the interface more confusing. The Transformer will have to create a StringIO instance on the fly when using `ruamel.yaml`._

Replaces #1865 (an alternative implementation)


webknjaz:
Where are these streams closed? If they aren't, this will cause `ResourceWarning`s.

cognifloyd:
The current implementation relies on StringIO so closing them is not an issue. If we need to switch to temporary files then we'd probably need to auto open/close them or something. I'm not sure where.

lilatomic:
At least to me, it would be clearer to rename `Lintable._content` to `original_content`, and `Lintable._writable` to `Lintable._content`. That's the spirit of the comments next to them, and I think it would help remove confusion like "Why do we never update the content?". I would also in `Lintable.__init__` assign both of them the value of `content`, I think that would eliminate a few null checks.

Could you also provide some guidance on when we would use `content` vs `writable`? That is, if I have a rule that can rewrite, would I always request a `writable`? or would I request the content, see if I need to change it, and then request the writable? I noticed that the tests seem to truncate the contents of the writable, I'm not sure if that's illustrative of the expected usage.

Could you also talk about why you went with a method to expose a writable stream, rather than a method to set the updated content? I understand why we're using streams over string; but why have  `def writable(self) -> TextIO` rather than `def update(self, contents: TextIO) -> None`? I think an advantage of an `update` method is that it would allow for linters and lintables to better track their updates, eg with `def update(self, contents: TextIO, linter_class: Type[AnsibleLintRule], comment: str = "") -> None`. I think this would also simplify the usage of `content` vs `writable`, since a rule would always use `content` for the content and only invoke `update` if it needed to do that.

I'm excited about rules being able to remediate Ansible, and I think this lays important groundwork for that feature.

cognifloyd:
OK. I changed the implementation from a `Lintable.writable` stream to a `Lintable.content` setter. I kept the same tests (modifying them to use `Lintable.content` instead of `Lintable.writable`).

ssbarnea (approved):
LGTM but I will need another pair of eyes on this one.

src/ansiblelint/file_utils.py
```diff
@@ -199,12 +200,57 @@ def get(self, key: Any, default: Any = None) -> Any:
...
+        """Update ``content`` and calculate ``updated``."""
+        if not isinstance(value, str):
+            raise TypeError(f"Expected str but got {type(value)}")
+        if self._original_content is None:
+            if self._content is not None:
+                self._original_content = self._content
+            elif not self.path.exists():
+                # new file
+                self._original_content = ""
+        self.updated = self._original_content != value
```
cognifloyd, line 222: Hmm. In `content.setter`, should we try to read the file to populate `_original_content` if the file exists and it has not been read yet?

```suggestion
        if self._original_content is None:
            if self._content is not None:
                self._original_content = self._content
            elif self.path.exists():
                # use self.content to populate self._original_content
                _ = self.content
            else:
                # new file
                self._original_content = ""
        self.updated = self._original_content != value
```

webknjaz:
Why are the comments minimized?

cognifloyd:
> Why are the comments minimized?

Because they were about the old implementation. You asked about streams, for example, which this does not use any more. So I marked the comments as out-of-date.

webknjaz:
I don't think we need to do this with legitimate comments. It's harder to find them this way. We mostly hide spam / outdated bot comments but doing this doesn't generate any notifications decreasing the chances of finding out what happened and why. So it's also bad for transparency. If the conversations need to be disregarded, just leave a comment about that — it's probably the best way to keep the discussions/decisions traceable.

cognifloyd:
okey dokey. unhid

ssbarnea:
I do mark comments as outdated every-time they ni longer apply to current implementation, just to keep the review size easy to read to other reviewers, including myself. The moment they become outdated, they become spam, and that is why github added these particular reasons when you hide them.

AFAIK, @webknjas is the only one that was intrigued by this. Maybe it is time to check with others from the team and see what think.

I personally find long conversations hard to follow, hide option helps fighting that.

webknjaz:
Well, if there's disregarded comments without replies for no reason, it doesn't look good. I wasn't intrigued, I was only shocked by the disrespect. There is no indication informing me that everything has been updated and needs rechecking. The only indication is that there are some new commits without any context explaining that they contain different content. Anyway, I don't see why you disregard what I said — I merely described what signals I was getting and what I saw with my own eyes. I don't think you know better what I saw unless I explain, which I did.

webknjaz:
> okey dokey. unhid

@cognifloyd thanks! For the future — it may be fine to hide things but then, take the responsibility of notifying the affected parties about this with an explicit comment and mentions that would trigger notifications and give people a chance to see what happened without forcing them to guess whether our judgment about their comments was reasonable. We can't always know upfront if what you thought another person meant is the same thing you perceived. Besides, we don't know/can't assume if the other party is caught up with the updates and making things disappear would inevitably cause extra cognitive load for those who care to understand how the context they may remember got transformed into a new one.

test/test_file_utils.py
```diff
@@ -212,3 +213,194 @@ def test_guess_project_dir(tmp_path: Path) -> None:
...
+
+@pytest.mark.parametrize(
+    "updated_content", ((None,), (b"bytes",)), ids=("none", "bytes")
+)
+def test_lintable_content_setter_with_bad_types(updated_content: Any) -> None:
+    """Validate ``Lintable.updated`` when setting ``Lintable.content``."""
+    lintable = Lintable("bad_type.yaml", BASIC_PLAYBOOK)
+    assert lintable.content == BASIC_PLAYBOOK
+
+    with pytest.raises(TypeError):
```
webknjaz, line 283: It is strongly recommended to always use the `match=` argument for all `pytest.raises()`/`pytest.warns()`/`pytest.deprecated()` uses to prevent accidental illegitimate passes.

test/test_file_utils.py
```diff
@@ -212,3 +213,194 @@ def test_guess_project_dir(tmp_path: Path) -> None:
...
+
+
+@pytest.fixture
+def tmp_updated_lintable(
+    tmp_path: Path, path: str, content: str, updated_content: str
+) -> Lintable:
+    """Create a temp file Lintable with a content update that is not on disk."""
+    lintable = Lintable(tmp_path / path, content)
+    with lintable.path.open("w", encoding="utf-8") as f:
+        f.write(content)
```
webknjaz, line 234: Why didn't you just do 
```suggestion
    lintable.path.write_text(content)
```
?

test/test_file_utils.py
cognifloyd: Doing that feels like relying on implementation details. Had I done that with the old implementation, I would have had to modify it with this implementation because the TypeError was being raised in my code. So, I don't like using the `match=` argument.

This test is not checking for a message. It is checking that it raises a TypeError for non-str things. I don't care why or what message is used, so adding a `match=` here is misleading about my intents when writing the code.

src/ansiblelint/file_utils.py
```diff
@@ -199,12 +200,60 @@ def get(self, key: Any, default: Any = None) -> Any:
...
+        content has not changed. For example:
+
+        .. code:: python
+
+            lintable.write(force=True)
+        """
+        if not force and not self.updated:
+            # No changes to write.
+            return
+        with open(self.path.resolve(), mode="w", encoding="utf-8") as file:
```
webknjaz, line 254: Why use an old-fashioned low-level API for this and not pathlib-native one?

src/ansiblelint/file_utils.py
```diff
@@ -199,12 +200,60 @@ def get(self, key: Any, default: Any = None) -> Any:
...
-        """Retried file content, from internal cache or disk."""
+        """Retrieve file content, from internal cache or disk."""
         if self._content is None:
             with open(self.path.resolve(), mode="r", encoding="utf-8") as f:
                 self._content = f.read()
+            if self._original_content is None:
+                self._original_content = self._content
         return self._content
 
+    @content.setter
```
webknjaz, line 211: The use of a setter seems weird to me here. A natural goal of setters is to set values mostly but this one also does I/O and shuffles the internal state. It feels like this sort of thing would require an explicit method call.

test/test_file_utils.py
cognifloyd: Reasoning is in https://github.com/ansible-community/ansible-lint/pull/1884#discussion_r807134835

src/ansiblelint/file_utils.py
webknjaz: OTOH the whole idea of having a mutable object appears questionable — would it be possible to have some transformation API that would take one lintable and construct/return a new one in an immutable manner?

src/ansiblelint/file_utils.py
```diff
@@ -199,12 +200,60 @@ def get(self, key: Any, default: Any = None) -> Any:
...
+    def content(self, value: str) -> None:
+        """Update ``content`` and calculate ``updated``."""
+        if not isinstance(value, str):
+            raise TypeError(f"Expected str but got {type(value)}")
+        if self._original_content is None:
+            if self._content is not None:
+                self._original_content = self._content
+            elif self.path.exists():
+                # use self.content to populate self._original_content
+                _ = self.content
```
webknjaz, line 221: No need to do a dummy assignment for the getattr to happen:
```suggestion
                self.content
```

cognifloyd: Doing that silences the pycharm linter which says the line has no effect.

webknjaz: Although, that problem above is minor — a more serious problem is that an attribute access is used to perform an implicit action. This is bad — this sort of code does not carry any context, nor has the right semantic. If we see a function/method call, it's immediately obvious — some action is requested to be performed, the details can be inferred from that method name that almost certainly starts with a descriptive verb explaining that action.
One would say that if there is a need to explain what code does (not "why") with a code comment, then there is something wrong with it.
A possible improvement could be putting such an action into a private method, for example. Wouldn't it be clearer if the code looked like `self._populate_content_cache()`?

cognifloyd: Thank you for suggesting an interface for such a method. Sure, that makes sense.

ssbarnea:
@cidrblock Can you please help with this one?

cidrblock (approved):
Looks good.

The only comment I have, which is totally subjective and could go in this PR, in another PR or in the trash..........  and it's probably because I'm less familiar with the patterns in the repo.

I (personally) prefer to see all the class self.* attributes defined and typed at the top of the init, even without default values. I tend to look there first to see what something is when used later and it's type.... especially when I'm seeing code for the first time.  Your call.

src/ansiblelint/file_utils.py
webknjaz: > Doing that silences the pycharm linter which says the line has no effect.

Which is why I suggested a follow up improvement :) It was a code smell and silencing the linter just made it shut up but not fix the underlying design issue. Although, if it was really needed, I'd opt for using a "noqa" comment with an excessive code comment explaining why the hack can't be avoided.

src/ansiblelint/file_utils.py
webknjaz: You're right, it may be hard to make it immutable with the current codebase, I won't insist on refactoring the whole project, of course :)

Have you considered exposing a method? (just thinking out loud)

cognifloyd:
> The only comment I have, which is totally subjective and could go in this PR, in another PR [snip]
> 
> I (personally) prefer to see all the class self.* attributes defined and typed at the top of the init, even without default values.

I think that would be a fine follow-up PR. I'm trying to minimize the refactoring in this PR.

ssbarnea:
As none of the current opened comments is critical, I will merge it as is. We can address these in follow-ups, especially as it is likely that we would need to tune the implementation while we start using it.

test/test_file_utils.py
webknjaz: Imagine you refactor that code and make a mistake in a function signature you call inside of the setter or down the stack. It will cause a `TypeError` and the test will still pass with a bug merged in 🤷‍♂️

webknjaz (commented):
Apparently some of the comments were not posted yesterday. Submitting them now.

src/ansiblelint/file_utils.py
```diff
@@ -197,13 +198,67 @@ def get(self, key: Any, default: Any = None) -> Any:
...
+
+        .. code:: python
+
+            lintable.write(force=True)
+        """
+        if not force and not self.updated:
+            # No changes to write.
+            return
+        with open(self.path.resolve(), mode="w", encoding="utf-8") as file:
+            file.write(self._content or "")
```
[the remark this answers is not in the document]
cognifloyd: Huh. I thought `resolve()` returned a string, not another Path object. Yeah, this would work.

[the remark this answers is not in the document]
cognifloyd: See #1913

PR #1887, opened 2022-02-13
Enable cspell

ssbarnea:
Related: https://github.com/ansible-community/devtools/issues/33


cidrblock (approved):
+1

I'm curious to see what @cognifloyd and/or @tadeboro think too.

Issue #1732, opened 2021-10-01
breaking: drop parseable and severity format modifiers


ssbarnea:
##### Summary
We should remove `-p` and `--parseable-severity` options in next major version. First step to deprecate them ASAP and on next release removing them.

Originially pointed on https://github.com/ansible-community/ansible-lint/pull/1730#pullrequestreview-768170608

Part of the same set of changes would be to also rename `codeclimate` output format to `json` to make it easier to discover by users.

PR #1898, opened 2022-02-14
Removed --parseable-severity feature

ssbarnea:
Removes the unorthodox CLI option and output format as we have other standard file output formats that also include the same information and they are considerably more flexible.

This will simplify both the implementation and also the list of command line arguments the tool accepts. This will be followed up by another change that will also remove the `-p` argument as that was kept as alias for `-f pep8` for long time.

Fixes: #1732


cidrblock (approved):
looks good.

PR #1909, opened 2022-02-16
Moved some pytest options to tox

ssbarnea:
- Ensures that direct call of pytest will run without xdist
- Runs with xdist only when tox or pytest do not receive extra args
- Avoid extra reporting options when run directly

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,22 +1,9 @@
 # spell-checker:ignore filterwarnings norecursedirs optionflags
 [pytest]
 addopts =
-    # `pytest-xdist`:
-    -n auto
-
-    # `pytest-mon`:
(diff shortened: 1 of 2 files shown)


pytest.ini
```diff
@@ -1,22 +1,9 @@
 # spell-checker:ignore filterwarnings norecursedirs optionflags
 [pytest]
 addopts =
-    # `pytest-xdist`:
-    -n auto
```
cognifloyd, line 5: Hooray! Now I don't have to comment this out when I'm running a single test in pycharm.

I still have to comment out `--pyargs` and `test ansiblelint.rules` so I can select a single test to run, but this is still much better.

cognifloyd (approved):

cidrblock (approved):
more cleaner.

PR #1912, opened 2022-02-16
Do not force pytest to always run all tests

cognifloyd:
I run pytest within pycharm. PyCharm handles automatically adding the src dir to the PYTHONPATH so that I do not have to install the lib and have duplicate files. That also simplifies navigating the code as the IDE never gets confused about which copy I want to jump to.

However, if I leave `test ansiblelint.rules` in pytest.ini then pytest always tries to run all tests. So, if I select a single test I'm trying to run during development, pytest runs all of the tests instead.

So, I've had to consistently comment that out, which makes git complain that my checkout is not clean, and just makes my development experience worse.

Also, we currently have `--doctest-modules` in the args, but pytest doesn't find one of the doctest modules because it is not under `tests` or `ansiblelint.rules`. It's in the `ansiblelint.yaml_utils`. Getting rid of the hard-coded `tests ansiblelint.rules` modules allows pytest to truly discover all relevant tests.

Note: I built this on top of #1909. It should be rebased on main once #1909 is merged.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,22 +1,9 @@
 # spell-checker:ignore filterwarnings norecursedirs optionflags
 [pytest]
 addopts =
-    # `pytest-xdist`:
-    -n auto
-
-    # `pytest-mon`:
-    # useful for live testing with `pytest-watch` during development:
-    # --testmon
-
-    # flaky:
-    --no-success-flaky-report
-
-    --durations=10
     -ra
     --showlocals
     --doctest-modules
-    --junitxml=.test-results/pytest/results.xml
-
     # interpret all the target args as things that can be imported:
     --pyargs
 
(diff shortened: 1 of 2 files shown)


cognifloyd:
After further testing, I found that `--pyargs` is not a problem. Specifying the modules to test in pytest.ini is the source of my issue.

pytest.ini
```diff
@@ -1,27 +1,14 @@
...
     --doctest-modules
-    --junitxml=.test-results/pytest/results.xml
-
     # interpret all the target args as things that can be imported:
     --pyargs
 
-    # importable packages for test lookup:
-    test ansiblelint.rules
+	# pytest will collect most tests from either `tests` or `ansiblelint.rules`.
+	# It also collects doctests from other modules.
```
cognifloyd, line 11:
```suggestion
	# It also collects doctest from other modules.
```

ssbarnea (approved):

ssbarnea:
This is not enough for fixing vscode debugging but is a step towards the right direction. I will do a follow-up.

PR #1914, opened 2022-02-17
Fix test integration with IDEs

ssbarnea:
This should fix debugging tests in both vscode and pycharm. We need to be sure that we disable pytest-cov inside `pytest.ini` and enable it only within `tox.ini` file.

This ensures that we can:

- [x] run/debug with pytest from console (outside tox)
- [x] run/debug from vscode
- [x] run/debug from pycharm
- [x] capture coverage on CI
- [x] capture coverage with simple `tox -e py` (no posargs)

Related: #1912 #1909

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -112,6 +112,7 @@ prerun
 psutil
 pyargs
 pycache
+pycharm
 pyenv
 pygments
 pylint
@@ -187,6 +188,7 @@ xdist
 xfail
 xunit
 zuul
+
 # Contributors
 Sorin
 Sbarnea
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,14 +1,14 @@
 # spell-checker:ignore filterwarnings norecursedirs optionflags
 [pytest]
+# do not add options here as this will likely break either console runs or IDE
+# integration like vscode or pycharm
 addopts =
-    -ra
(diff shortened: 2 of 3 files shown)


webknjaz (changes requested):
The base args must be in `pytest.ini`. Maintaining them only in tox will create an inconsistent experience. `tox.ini` is only there for conditional overrides.

ssbarnea:
> The base args must be in `pytest.ini`. Maintaining them only in tox will create an inconsistent experience. `tox.ini` is only there for conditional overrides.

If you have an alternative that is **fixing** all the points noted in the description, I would be happy to see it. So far I was not the only contributor that found these issues inconvenient and I do not want to force anyone willing to contribute to alter his environment to make it work, especially as that might be even impossible without altering tracked files.

cidrblock (approved):
Anything we can do to ensure the first time test running/debugging experience is smooth I'll +1

ssbarnea:
I that brings any regressions, we will fix them in follow-ups.

@cognifloyd Please test and report if this is improving the experience for you.

cognifloyd:
Nice. Works great in pycharm.
I setup my args in pycharm's settings: `-ra -l --tb=short --strict-markers`
(ie "Additional Arguments" in the run templates for "autodetect" and "pytest")

PR #1924, opened 2022-02-17
Normalize format of yaml examples

ssbarnea:
Related: #1820


examples/playbooks/tasks/x.yml
```diff
@@ -1,4 +1,4 @@
-- # nothing, checks bug #849
+-  # nothing, checks bug #849
```
cognifloyd, line 1: This is surprising. It added an extra space?

cognifloyd (approved):

examples/playbooks/tasks/x.yml
ssbarnea: I observed too. We will see. I do have a theory why this happens and I hope we will be able to replicate the behavior with our formatter.

PR #1927, opened 2022-02-17
Configure ruamel.yaml dumping with custom `Emitter`

cognifloyd:
This custom `Emitter` ensures that root-level sequences (playbooks, task files, etc) are not over-indented without sacrificing indents at other levels.

This custom `Emitter` also changes ruamel.yaml scalar quoting preference from single- to double-quotes.

This PR also adds the `Transformer.yaml` property that configures a `ruamel.yaml.YAML` instance to use our emitter (along with configuring a variety of other dump options).

This builds on #1926 which should be merged first.
Extracted from #1828.


cognifloyd:
OK. I think this is done and ready for review.

The changes to the `formatting-after` fixture files highlight the differences with prettier's formatting. And, since lint is passing, prettier is not trying to change the format created by `FormattedYAML`.

PR #1934, opened 2022-02-18
Add formatting fixtures

ssbarnea:
These are needed for formatting feature, so we can keep compatibility with prettier formatting.


cidrblock (approved):
Tricky one, but seems sane.

cognifloyd (approved):

PR #1936, opened 2022-02-21
Add tool to generate formatting fixtures

cognifloyd:
Add tools/generate-formatting-fixtures.py which uses prettier to generate the fixture files.

Rename test/fixtures/formatting-after -> test/fixtures/formatting-prettier.
Add new test/fixtures/formatting-ruamel-yaml which is a copy of the prettier files for now.
In #1927, the diff will show where our ruamel.yaml formatter differs from prettier.

Also, add new fmt-2.yml formatting test fixture to cover a few more yaml oddities.


cognifloyd:
aargh. galaxy seems to be on the fritz.

webknjaz:
Consider implementing this as a pytest plugin activated by a custom CLI option.

cognifloyd:
@webknjaz done. I converted it to a "test" + pytest plugin. If `--regenerate-formatting-fixtures`, the plugin skips all tests except the one that regenerates the fixtures. Otherwise it skips just `test_regenerate_formatting_fixtures`.

ssbarnea (approved):
OK. I made a fix myself to keep using "after" in naming of formatting fixtures as we should not use implementation detail in them.

Keep in mind that we might start to format something else than yaml or to swap for another yaml library. The fact that we use ruamel-yaml is an implementation detail.

cognifloyd:
Works for me. Thanks for reviewing!

PR #1943, opened 2022-02-26
Add support for reformatting YAML files

cognifloyd:
I want to introduce a feature to `ansible-lint`: It should be able to fix issues identified by rules. The feature in this PR is a pre-requisite to introducing that.

In order to fix any issues identified by rules, we will have to read each YAML file with ruamel.yaml and then write it out. Reading/writing YAML files with ruamel.yaml standardizes spacing and quotes while preserving comments. In other words, it does very basic YAML reformatting which might fix some of the issues identified by `yamllint`. Fixing most of the issues identified by ansible-lint (including at least some of the `yamllint`-identified issues) requires transforming the files in some way. This PR only implements the read+write part, leaving the actual rule transforms for future PRs.

To use this **(opt-in)** mode, run `ansible-lint --write` (UX discussion summarized in #1910). That makes ansible-lint pass all YAML files it checks through our customized `ruamel.yaml.YAML` instance.

`ansible-lint --write` reports a count of the number of modified files, which looks approximately like this:

```
Modified 42 files.
Finished with 0 failure(s), 0 warning(s), on 100 files.
```

Notes:
- _This is and must be **opt-in**. ansible-lint will not modify any files unless you ask it to._
- The tests here revealed a YAML formatting issue not accounted for in #1927 that moved full line comments instead of just eol comments. 808e050fce2294411cd3b40e3aa70dd3b8b1d230 fixes that issue.

Other merged PRs prepared the basis for this one (which replaces #1828):
- https://github.com/ansible-community/ansible-lint/pull/1884
- https://github.com/ansible-community/ansible-lint/pull/1926
- https://github.com/ansible-community/ansible-lint/pull/1934
- https://github.com/ansible-community/ansible-lint/pull/1936
- https://github.com/ansible-community/ansible-lint/pull/1927
- https://github.com/ansible-community/ansible-lint/pull/1944
- https://github.com/ansible-community/ansible-lint/pull/1946
- https://github.com/ansible-community/ansible-lint/pull/1947
- https://github.com/ansible-community/ansible-lint/pull/1948
- https://github.com/ansible-community/ansible-lint/pull/1950
- https://github.com/ansible-community/ansible-lint/pull/1951
- https://github.com/ansible-community/ansible-lint/pull/1952


src/ansiblelint/transformer.py
```diff
@@ -17,12 +20,48 @@ class Transformer:
...
+                data: str = file.content
+            except (UnicodeDecodeError, IsADirectoryError):
+                # we hit a binary file (eg a jar or tar.gz) or a directory
+                data = ""
+                file_is_yaml = False
+
+            if file_is_yaml:
+                ruamel_data: Union[CommentedMap, CommentedSeq] = yaml.loads(data)
+                file.content = yaml.dumps(ruamel_data)
+                file.write()
```
cognifloyd, line 67: This is the meat of this change. Everything else is to support this.

cognifloyd:
Merged in main

ssbarnea:
While playing with it I found one undesired behavior: when `yaml` was in skip-list, I would have not expected `--write` to have any effect, but it still did.

I think we should respect the skip list and bypass any yaml formatting when the skip is present. WDYT?

ssbarnea (changes requested):
See my previous comment

cognifloyd:
1. If `yaml` is in the skip_list, then should `--write` throw an error? No other transforms will be able to have any affect without reformatting the yaml file. In any case, I'm looking at how to account for the skip list.
2. Oh. I thought I had a safety on there for that issue. Looks like I missed handling that in this iteration. If the file is not a map or a list, then it should not try to write it because `ruamel.yaml` cannot preserve comments when the file loads as "None" (which an empty vars file will do). I'll push the fix in a bit.
3. The empty flow issue `{ }` vs `{}` is addressed in #1946.

cognifloyd:
1. If `--write` and `yaml` is in `skip_list`, then the Transformer no longer runs. Instead a message gets printed to explain why nothing was modified.
2. I added the empty file safety check.

src/ansiblelint/app.py
```diff
@@ -138,11 +147,19 @@ def report_outcome(
         #         v,
         #     )
 
-        if result.matches and not self.options.quiet:
+        if self.options.write and "yaml" in self.options.skip_list:
+            console_stderr.print(
```
ssbarnea, line 151: Please use logging.warning for this as it gives more control.

In fact I doubt we ever need to use console_stderr other than with logging.

cognifloyd: done

test/test_transformer.py
```diff
@@ -0,0 +1,93 @@
...
+            "examples/playbooks/nomatchestest.yml", 0, False, id="nomatchestest"
+        ),
+        pytest.param("examples/playbooks/unicode.yml", 1, False, id="unicode"),
+        pytest.param(
+            "examples/playbooks/lots_of_warnings.yml", 992, False, id="lots_of_warnings"
+        ),
+        pytest.param("examples/playbooks/become.yml", 0, True, id="become"),
+        pytest.param(
+            "examples/playbooks/contains_secrets.yml", 0, False, id="contains_secrets"
+        ),
```
cognifloyd, line 55: As we've reformatted all of the files, almost all of these tests are not actually transforming anything anymore. Only one is changing something now: in `become.yml` it changes the quotes on a string.

ssbarnea (approved):

PR #1947, opened 2022-03-01
Configure FormattedYAML using YamllintConfig

cognifloyd:
This makes a few of the FormattedYAML settings configurable based on the yamllint settings.
Where the yamllint setting is ambiguous (a rule is not enabled, or it allows "any"), then default to something that is (hopefully, but debatably) sane.

- move YAMLLINT_CONFIG from rules.YamllintRule to yaml_utils
- Configure FormattedYAML using YamllintConfig

### DIFF
--- a/src/ansiblelint/rules/YamllintRule.py
+++ b/src/ansiblelint/rules/YamllintRule.py
@@ -1,34 +1,17 @@
 import logging
-import os
-import sys
 from typing import TYPE_CHECKING, List
 
-from yamllint.config import YamlLintConfig
 from yamllint.linter import run as run_yamllint
 
 from ansiblelint.file_utils import Lintable
 from ansiblelint.rules import AnsibleLintRule
 from ansiblelint.skip_utils import get_rule_skips_from_line
+from ansiblelint.yaml_utils import load_yamllint_config
 
 if TYPE_CHECKING:
     from ansiblelint.errors import MatchError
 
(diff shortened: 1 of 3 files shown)


cognifloyd:
I'm guessing this is a `skip-changelog` change, but I'm not sure, so I didn't label it.

src/ansiblelint/rules/YamllintRule.py
```diff
@@ -48,24 +31,8 @@ class YamllintRule(AnsibleLintRule):
     severity = "VERY_LOW"
     tags = ["formatting", "yaml"]
     version_added = "v5.0.0"
-    config = None
+    config = load_yamllint_config()
     has_dynamic_tags = True
-    if "yamllint.config" in sys.modules:
```
cognifloyd, line 53: This import is no longer conditional, so checking this didn't make sense. I dropped the check while moving the config loading.

src/ansiblelint/rules/YamllintRule.py
```diff
@@ -79,24 +46,23 @@ def matchyaml(self, file: Lintable) -> List["MatchError"]:
         if str(file.base_kind) != "text/yaml":
             return matches
 
-        if YamllintRule.config:
```
cognifloyd, line 82: I dropped this conditional check as well since yamllint loading is no longer conditional.

ssbarnea (approved):

PR #1949, opened 2022-03-03
Apply MissingFilePermissionsRule to get_url module

nre-ableton:
The `get_url` module can create files and directories similar to the file and copy modules, so the mode parameter should be enforced for this module as well.


ssbarnea (approved):

ssbarnea (changes requested):
Thanks for the fix. Please update the tests or the fixtures so they would not fail.

nre-ableton:
@ssbarnea Yes, of course, sorry. I agree that having a test would be appropriate. However, I'm not exactly sure where to add such a test... sorry, I'm unfamiliar with the `ansible-lint` codebase and this is my first contribution. I looked in several places but couldn't find similar tests for the `archive` and `assemble` modules (for example).

I pushed a [fixup](https://github.com/ansible-community/ansible-lint/pull/1949/commits/2297f13c01322d3eeef78d3f32bbde53cfb68158), please let me know if this would be sufficient for test coverage, and if so, I'll squash. Thanks!

ssbarnea (approved):

nre-ableton:
@ssbarnea Thanks for fixing the tests! I would have liked to have squashed my fixup, but I guess it's not a problem. Looking forward to the next release!

ssbarnea:
No need to worry about squash, we (almost) always do it, just to keep the history clean. We had cases with 100 commits, nobody wants to see each of them.

PR #1952, opened 2022-03-04
Misc YAML formatting fixes

cognifloyd:
Misc bits to wrap up the MVP of YAML formatting.

This replaces the full line comment indentation logic with some post-processing as recommended in https://stackoverflow.com/a/71355688/1134951

And it 
- standardizes examples indent
- Includes blank lines in yaml preamble
- Ensures that a line with a null list item does not end with a space

This is much more reliable than the previous implementation. There are still some quirks, but it is much better.

### DIFF
--- a/examples/playbooks/with-skip-tag-id.yml
+++ b/examples/playbooks/with-skip-tag-id.yml
@@ -1,7 +1,7 @@
 ---
 - hosts: all
   tasks:
-  - name: trailing whitespace on this line         
-    git:
-      repo: "{{ archive_services_repo_url }}"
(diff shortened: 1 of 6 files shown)


cognifloyd:
Aargh. This mangles multi line strings like `|` that have a line that starts with `#`. Marking as draft for now.

cognifloyd:
OK. I have a way to hopefully detect multi line strings. Since I'm not actually parsing the yaml in the post-processor, there are plenty of edge cases where this might still mangle the string.

ssbarnea (approved):

PR #1954, opened 2022-03-05
Added missing 'truthy' setting in YAMLLINT_CONFIG

ggustafsson:
By default yamllint does not consider "yes" and "no" to be valid truthy values (only "true" or "false" is allowed by default). Since "yes" and "no" is very common in Ansible playbooks (e.g "become: yes") it should be allowed in the default yamllint config used by ansible-lint.

Alternative solution to extending the "truthy" setting would be to disable it completely but I figured just adding the missing common values is a little bit nicer.

### DIFF
--- a/src/ansiblelint/yaml_utils.py
+++ b/src/ansiblelint/yaml_utils.py
@@ -50,6 +50,9 @@ rules:
   # you can easily change it or disable in your .yamllint file.
   line-length:
     max: 160
+  # By default only "false" or "true" is considered truthy.
+  truthy:
+    allowed-values: ["true", "True", "false", "False", "yes", "no"]
 """
 
 


ssbarnea (changes requested):
I am afraid that I will be against this change. While Ansible is very relaxed regarding what kind of booleans it accepts, the linter is not and allows only the YAML 1.2-compatible forms, as one of its goals is to normalize the code and make it future proof.

At some point in the future Ansible will allow or even require YAML 1.2, so using pure booleans is the right path.

There were talks about updating Ansible docs and to popularize pure boolean values `true` and `false` but I am not sure exactly when this will happen.

ggustafsson:
Oh, I thought it was just yamllint that was opinionated about it and that it didn't matter in YAML. If the YAML spec says true/false is the way to go then I am all for it. It's unfortunate that Ansible/YAML/documentation is not consistent at the moment.

Anyways thanks for the info. Let's close this thing and get on with our lives! 🙂

PR #1956, opened 2022-03-05
Better protect strings with '#' when dumping YAML

cognifloyd:
I'm using a string of unicode chars (that happen to look like `#`) to make sure no string can have a line that starts with `#`. That makes sure that those lines won't be mis-identified as full line comments in post-processing.

Anthon suggested moving all full line comments so they start in column 0, but that prevented me from preserving the indentation in a few cases (that I didn't know about when I asked him).
However, doing something earlier, when information about what is and isn't a comment, was a fantastic idea. I ended up modifying the strings instead of the comments, but I'm happy with the (much more reliable) result.

So, thanks @AvdN for helping me find a great way to do this with the current version of ruamel.yaml!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,7 +52,7 @@ src/ansiblelint/_version.py
 # Unformatted fixtures, need to be added with force as we need to exclude
 # them from being reformatted (.prettierignore is a symlink to .gitignore)
 test/fixtures/formatting-before/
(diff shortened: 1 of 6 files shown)


cognifloyd:
I marked this as `skip-changelog` but if that's not right, please feel free to change it.

ssbarnea (approved):

PR #1960, opened 2022-03-07
Refactored fixtures regeneration

ssbarnea:


tox.ini
```diff
@@ -23,6 +23,8 @@ deps =
 commands =
   # safety measure to assure we do not accidentally run tests with broken dependencies
   {envpython} -m pip check
+  # Ensure that our fixtures are kept updated
+  {envpython} tools/regenerate-fixtures.py
```
cognifloyd, line 27: Doing this will mean that people that use `tox` would need to have `prettier` on their PATH (vs only installed by pre-commit)

ssbarnea: Yep. Precommit does not allow us to create a hook that uses two languages (node and python) so we need to take care of installing prettier ourselves.

tools/regenerate-fixtures.py
```diff
@@ -0,0 +1,45 @@
...
+from pathlib import Path
+
+from ansiblelint.yaml_utils import FormattedYAML
+
+"""Re-generate formatting fixtures with prettier and internal formatter.
+
+Pass ``--regenerate-formatting-fixtures`` to run this and skip all other tests.
+This is a "test" because once fixtures are regenerated,
+we run prettier again to make sure it does not change files formatted
+with our internal formatting code.
```
cognifloyd, line 13:
```suggestion
Once fixtures are regenerated, we run ``prettier`` again to make sure that
``prettier`` does not change files formatted with our internal formatting code.
```

tox.ini
cognifloyd: I would rather not force people to install npm + prettier for most local development, but let it run in CI to catch issues when people don't have it installed locally.

cognifloyd: (pre-commit hides the fact that it's getting installed)

PR #1963, opened 2022-03-09
packit: build SRPM in Copr

TomasTomecek:
and install dependencies from the distro instead of PyPI

Details: https://packit.dev/posts/copr-srpms/

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -146,6 +146,7 @@ signoff
 skippable
 skiptasks
 skiputils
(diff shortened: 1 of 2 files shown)


ssbarnea (changes requested):
I like the first bit but not the removal of fedora-35 build, especially as we also need to make it work for fedora-34. Can you please try to send me a message on matrix and see how we can progress with this?

ssbarnea (approved):

PR #1970, opened 2022-03-10
Add YAML path (to play/task) discovery utils for Transforms to use

cognifloyd:
This PR is another step towards adding Rule Transforms.

The Transforms will need the path to a play or task from the outermost sequence so that that play/task can be modified in the transform. That is somewhat complex because we have to use the line number to find that play/task.

Note: There are a variety of places where I had to convert between 0-based and 1-based ints. I tried to make it clear which is which by using `*_number` for 1-based ints and `*_index` for 0-based ints along with a variety of comments to highlight which is which and when I do a conversion.


cognifloyd:
There, I only added 97 test cases for these functions :wink:

Now, I've got 98.10% diff coverage (the `if TYPE_CHECKING` block is not "covered" but does not matter).

test/test_yaml_utils.py
```diff
@@ -212,6 +213,7 @@ def yaml_formatting_fixtures(fixture_filename: str) -> Tuple[str, str, str]:
     (
         "fmt-1.yml",
         "fmt-2.yml",
+        "fmt-3.yml",
```
cognifloyd, line 216: I noticed this test case was missing, so I included it (even though it's not needed for the play/task path discovery utils)

src/ansiblelint/yaml_utils.py
```diff
@@ -225,6 +233,211 @@ def _nested_items_path(
...
 
+def get_path_to_play(
+    lintable: Lintable,
+    line_number: int,  # 1-based
+    ruamel_data: Union[CommentedMap, CommentedSeq],
+) -> List[Union[str, int]]:
+    """Get the path to the play in the given file at the given line number."""
+    if lintable.kind != "playbook" or not isinstance(ruamel_data, CommentedSeq):
+        return []
+    # TODO: raise if line_number < 1 or just return [] ?
```
cognifloyd, line 244: This is one open question I have. If for some reason, someone passes a `line_number < 1`, then they're probably passing in an index, not a number which is wrong.

Should I raise an error? Or just `return []` because anything no play or task can be at lines less than 1?

cognifloyd: I just made it raise ValueError. If that ends up being annoying, we can change it in a follow-up.

ssbarnea (approved):

PR #1978, opened 2022-03-11
Load rule classes from any modules

ssbarnea:
This change is preparing for #1977 so we can rename the module names that include rule classes. We no longer make the assumption that the module name must match that rule class name.


cognifloyd (commented):
Oh wow. This is beautiful. I've looked at fixing this before but didn't see an elegant way to resolve this. 👍 💯

PR #1981, opened 2022-03-12
Remove html extension from documentation

ssbarnea:
Makes URLs nicer by using dirhtml build instead of html. The effect
is that you will no longer see .html as part of URLs.


cognifloyd (approved):

webknjaz (changes requested):
Most of the changes are fine but these things below must be fixed before merging.

PR #1993, opened 2022-03-12
Address pylint missing-class-docstring (old)

ssbarnea:
Most of classes without a docstring were the rules and for these we
moved the description into the docstring.

This change needs extra attention regarding on how the documentation
is built, so we do not mess its rendering.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -16,7 +16,6 @@ disable =
     dangerous-default-value,
     duplicate-code,
     fixme,
-    missing-class-docstring,
     missing-function-docstring,
     missing-module-docstring,
     no-member,
--- a/src/ansiblelint/cli.py
+++ b/src/ansiblelint/cli.py
@@ -111,6 +111,8 @@ def get_config_path(config_file: Optional[str] = None) -> Optional[str]:
 
 
 class AbspathArgAction(argparse.Action):
+    """Argparse action to convert relative paths to absolute paths."""
+
     def __call__(
         self,
         parser: argparse.ArgumentParser,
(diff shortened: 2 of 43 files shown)


cidrblock (approved):

cognifloyd (commented):
Can we wrap the long first lines of the doc strings more?

cognifloyd:
I also wonder if we'll need to do `dedent(__doc__)` for some of these.

cognifloyd:
Yes. Looks like the multiline doc strings need `dedent`. See the read the docs build: https://ansible-lint--1993.org.readthedocs.build/en/1993/default_rules.html#command-instead-of-module

cidrblock:
thinking aloud, rather than `__doc__`, if these need to manipulated, would a @property in the base class call `description` work? (that returned __doc__ with extra lines removed etc if needed?)

ssbarnea:
Main reason for attempt to use docstring was that pylint required it and we did not want to add yet another text. We might go for using shortdesc instead which is not supposed to be more than a line.

I am also considering using markdown but i do not yet know how to make sphinx read docstring as md.

cognifloyd:
Markdown in doc strings: https://stackoverflow.com/questions/56062402/force-sphinx-to-interpret-markdown-in-python-docstrings-instead-of-restructuredt

cognifloyd:
Instead of explicitly doing `description = __doc__`, what if we used a [classmethod property](https://stackoverflow.com/a/1800999/1134951) to return `cls.__doc__`? If someone overwrites the property with a `description` attribute, then that will naturally take precedence (ie, effectively how things are now).

PR #1994, opened 2022-03-14
Changed rules to use docstring as shortdesc

ssbarnea:
Simplifies rule by ensuring that we do use the docstring as their short description. This also enables us to address most pylint no-class-docstring excludes.

That is an alternative to another linked PR I made yesterday where I tried to put the description in a docstring, but that proved to cause several problems related to formatting
of docstrings with rst. This approach is simple and without undesired effects.

Related: https://github.com/ansible-community/ansible-lint/pull/1993


cidrblock (approved):
short_desc?  I'm surprised it passed cspell :)

PR #1995, opened 2022-03-14
Minor documentation improvements

ssbarnea:
- use special link property for reference links
- improve display of field list


src/ansiblelint/rules/meta_no_info.py
```diff
@@ -60,7 +60,7 @@ class MetaMainHasInfoRule(AnsibleLintRule):
     id = "meta-no-info"
     str_info = META_STR_INFO
     info = META_INFO
-    description = "meta/main.yml should contain: ``{}``".format(", ".join(info))
+    description = "meta/main.yml should contain: {}".format(", ".join(info))
```
cidrblock, line 63: any reason not to fstring it?

PR #2003, opened 2022-03-14
Address pylint no-member

ssbarnea:


test/test_transformer.py
```diff
@@ -15,15 +17,19 @@
 
 @pytest.fixture(name="copy_examples_dir")
 def fixture_copy_examples_dir(
-    tmpdir: py.path.local, config_options: Namespace
-) -> Iterator[Tuple[py.path.local, py.path.local]]:
+    tmp_path: pathlib.Path, config_options: Namespace
+) -> Iterator[Tuple[pathlib.Path, pathlib.Path]]:
```
cognifloyd, line 21: This is not right. The `tmp_path` fixture returns `py.path.local`, not `pathlib.Path`.
https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=tmpdir#std-fixture-tmpdir

If we want to make it return `pathlib.Path`, that's cool, but the type hint on `tmp_path` should be `tmp_path: py.path.local`.

cognifloyd: Oh. You switched from tmpdir to tmp_path. I missed that. Doh.

cognifloyd (approved):

Issue #1879, opened 2022-02-12
importing playbooks using FQCN doesn't work


jcgruenhage:
##### Summary
When importing a playbook from a collection using `import: <namespace>.<collection>.<playbook>`, ansible happily runs the playbook, but ansible-lint complains that it can't find the files.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```text
fqcn_import_bug on  main 
❯ ansible-lint --version
ansible-lint 5.4.0 using ansible 2.12.2

fqcn_import_bug on  main 
❯ ansible --version     
ansible [core 2.12.2]
  config file = /home/jcgruenhage/dev/fqcn_import_bug/ansible.cfg
  configured module search path = ['/home/jcgruenhage/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/jcgruenhage/dev/fqcn_import_bug
  executable location = /usr/bin/ansible
  python version = 3.10.2 (main, Jan 15 2022, 03:11:32) [GCC 10.2.1 20201203]
  jinja version = 3.0.3
  libyaml = True

```

- ansible installation method: OS package
- ansible-lint installation method: OS package

##### OS / ENVIRONMENT
VoidLinux (void package maintainer of ansible, ansible-core and anisble-lint here o/)

##### STEPS TO REPRODUCE

I've built a reproducible example at https://github.com/jcgruenhage/fqcn_import_bug, which contains a two small playbooks. One of them is a debug task to print hello world, and the other one imports the hello world playbook using it's FQCN.

A quick demonstration of the example follows below:

```
/tmp 
❯ git clone https://github.com/jcgruenhage/fqcn_import_bug.git
Cloning into 'fqcn_import_bug'...
remote: Enumerating objects: 10, done.
remote: Counting objects: 100% (10/10), done.
remote: Compressing objects: 100% (7/7), done.
Receiving objects: 100% (10/10), done.
remote: Total 10 (delta 0), reused 10 (delta 0), pack-reused 0

/tmp 
❯ cd fqcn_import_bug 

fqcn_import_bug on  main 
❯ ansible-playbook jcgruenhage.fqcn_import_bug.site
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the
implicit localhost does not match 'all'
[...]

PR #2018, opened 2022-03-18
Allow import_playbook from collections

ssbarnea:
Fixes: #1879


ganeshrn (approved):
Can you add test case for this scenario?

ssbarnea:
> Can you add test case for this scenario?

I already did. That modified example is part of a test, like all of them. Most bugfixes do not require adding new tests, just modifying the fixtures.

Issue #1753, opened 2021-10-23
Ansible Lint misclassifies template folders as roles


tadeboro:
##### Summary

When linting roles, Ansible Lint will misclassify subdirectories in the `templates` directory as roles.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible [core 2.11.6]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/tadej/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/tadej/ansible-community/ansible-lint/venv/lib64/python3.9/site-packages/ansible
  ansible collection location = /home/tadej/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/tadej/ansible-community/ansible-lint/venv/bin/ansible
  python version = 3.9.7 (default, Aug 30 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
  jinja version = 3.0.2
  libyaml = True

ansible-lint --version
ansible-lint 5.2.1 using ansible 2.11.6
```

- ansible installation method: pip
- ansible-lint installation method: editable install from main branch

##### OS / ENVIRONMENT

Fedora 34

##### STEPS TO REPRODUCE

```
cd /tmp
ansible-galaxy role init test
cd test
curl https://raw.githubusercontent.com/aminvakil/ansible-role-node-exporter/master/meta/main.yml -o meta/main.yml 
rm .travis.yml 
ansible-lint 
git init
git add .
git commit -m "test"
ansible-lint 
for i in *; do cd $i && git submodule add https://github.com/ansible/distro-test-containers/ distro-test && cd ../; done
```

The output we get is something like this:

```
WARNING  Listing 1 violation(s) that are fatal
role-name: Role name distro-test does not match ``^[a-z][a-z0-9_]+$`` pattern
templates/distro-test:1
```

##### Desired Behaviour

Ansible Lint should ignore anything in the templates directory (and probably also in all other standard role directories).

##### Actual Behaviour

Ansible Lint treats the subdirectory as a role and outputs all kinds of nonsense when trying to lint unstructured files.

##### Additional data

[...]

PR #2021, opened 2022-03-18
Avoid processing files under templates

ssbarnea:
Fixes: #1753

### DIFF
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -9,6 +9,7 @@ DEFAULT_KINDS = [
     # Do not sort this list, order matters.
     {"jinja2": "**/*.j2"},  # jinja2 templates are not always parsable as something else
     {"jinja2": "**/*.j2.*"},
+    {"text": "**/templates/**/*.*"},  # templates are likely not validable
     {"inventory": "**/inventory/**.yml"},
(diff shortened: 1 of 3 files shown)


src/ansiblelint/config.py
```diff
@@ -9,6 +9,7 @@
     # Do not sort this list, order matters.
     {"jinja2": "**/*.j2"},  # jinja2 templates are not always parsable as something else
     {"jinja2": "**/*.j2.*"},
+    {"text": "**/templates/**/*.*"},  # templates are likely not validable
```
cognifloyd, line 12: Do we need to add `text` (and maybe others here) to `ansiblelint.constants.FileType`?

ssbarnea: If we don't see an error, not really. I tested and without adding this line the tests would fail (only by adding the new example). One I added this line, it passed.

cognifloyd: Looks like `FileType` is already missing:
- `jinja2`
- `inventory`
- `galaxy`
- `reno` _what an unfortunate/confusing abbreviation for "release notes"_
- `vars`

cognifloyd: OK. That might be something to test in a future PR then.

PR #2022, opened 2022-03-18
--write: Avoid line-wrapping YAML flow objects

cognifloyd:
When flow-objects (JSON-like YAML maps or lists) get dumped and their line length is too long, they can get wrapped in some very odd places.

This is a flow-style map: `{ foo: bar }`
This is a flow-style list: `[0, 1, 2, 3]`

So, we calculate how long we expect the line-length might be for flow-style maps and lists.
If it is short enough to fit on one line, we leave it in flow-style.
If the line will (probably) be wrapped, we make that map/list use block-style instead of flow-style.


src/ansiblelint/yaml_utils.py
```diff
@@ -886,13 +887,58 @@ def loads(self, stream: str) -> Any:
...
+    def _prevent_wrapping_flow_style(self, data: Any) -> None:
+        if not isinstance(data, (CommentedMap, CommentedSeq)):
+            return
+        for key, value, parent_path in nested_items_path(data):
+            if not isinstance(value, (CommentedMap, CommentedSeq)):
+                continue
+            fa: Format = value.fa  # pylint: disable=invalid-name
+            if fa.flow_style():
+                predicted_indent = self._predict_indent_length(parent_path, key)
+                predicted_width = len(str(value))
```
cognifloyd, line 907: Generally, this should be longer than the dumped width because strings will all be quoted with `'` using python's repr on dicts/lists, but many of the strings will end up without quotes when dumped.
However, if it includes any complex strings with a bunch of escaped characters, then it could be too short. In that case, switching to a block will make dumping such a complex string clearer/more distinct, so I think that is an okay trade-off.

src/ansiblelint/yaml_utils.py
```diff
@@ -1,4 +1,5 @@
 """Utility helpers to simplify working with yaml-based data."""
+# pylint: disable=too-many-lines
```
cognifloyd, line 2: Uhm. I'm not sure how else to organize this better.
The YAML formatting bits (`ruamel.yaml` extensions) make up over half of this file (over 550 lines).
Then there's another large chunk of utils used to navigate through YAML files.
But, I'm not sure what to name any new modules.

ssbarnea (approved):

PR #2023, opened 2022-03-19
--write: Introduce TransformMixin for Rules

cognifloyd:
If a Rule wants to opt-in to transforming, or modifying, files, it can add the `TransformMixin` and implement the `transform()` method.

```python
from ansiblelint.rules import AnsibleLintRule, TransformMixin


class FooBarRule(AnsibleLintRule, TransformMixin):
    ...
    def transform(self, match: MatchError, lintable: Lintable, data: list|dict|str):
        """Edit `data` to address the `match` identified by this rule in the `lintable` file."""
```

This wires up the Transformer to look for `TransformMixin` rules and run their `transform()` method.

NOTE: This is a minimized version of `TransformMixin` to highlight the primary `TrnasformMixin` interface and show how the `Transformer` will use it. The Transforms need other features and utils which will be added in follow-up PRs.

Commits:
- Stub TransformMixin for Rules
- Run transforms on non-YAML files (eg Jinja)
- Refactor running rule transforms


src/ansiblelint/transformer.py
```diff
@@ -62,17 +63,37 @@ def run(self) -> None:
...
+    def _do_transforms(
+        file: Lintable,
+        data: Union[CommentedMap, CommentedSeq, str],
+        matches: List[MatchError],
+    ) -> None:
+        """Do Rule-Transforms handling any last-minute MatchError inspections."""
+        for match in sorted(matches):
+            if not isinstance(match.rule, TransformMixin):
+                continue
+            match.rule.transform(match, file, data)
```
cognifloyd, line 99: Though this is simple enough to inline in `run()`, it will get more logic in follow-up PRs.

ssbarnea (approved):

PR #2031, opened 2022-03-22
Make sure all tasks get evaluated by matchtask including block/always/rescue and nested tasks

### LINKED ISSUE #1895: Variable naming standards different for tasks and blocks
##### Summary
Using uppercase variables in tasks are reported as `var-naming: Task defines variables within 'vars' section that violates variable naming standards´. The same variables for blocks are not reported.

##### Issue Type

- Bug Report

```console (paste below)
$ ansible --version
ansible [core 2.12.2]

$ ansible-lint --version
ansible-lint 5.4.0 using ansible 2.12.2
```

##### OS / ENVIRONMENT
Fedora 35

##### STEPS TO REPRODUCE
Using ansible-lint on this playbook

```console (paste below)
---
- name: Test
  hosts: localhost

  tasks:
  - block:
      - name: foo1
        debug: msg="{{ FOO }}"
    vars:
      FOO: fooo1

  - name: foo2
    debug: msg="{{ FOO }}"
    vars:
      FOO: fooo2
```

results in 

```
Failed to guess project directory using git: 
WARNING  Overriding detected file kind 'yaml' with 'playbook' for given positional argument: test.yml
var-naming: Task defines variables within 'vars' section that violates variable naming standards
test.yml:12 Task/Handler: foo2

# .ansible-lint
  - experimental  # all rules tagged as experimental

Finished with 0 failure(s), 1 warning(s) on 1 files.
```

##### Desired Behavior
The issue needs to be reported for the task and the block.

##### Actual Behavior
The issue is only reported for the task, but not the block.

nishipy:
Fix #1895.


ssbarnea:
My impression is that the correct implementation would be to use `matchtask` as this is expected to be called for each task and avoid trying to traverse tasks from match play, especially as blocks can be nested.

matchplay should only look at play level stuff and not go into tasks/pre_tasks/post_tasks, as these are supported to be handled by matchtask.

Note: I am not 100% sure that is correct, but that I how i remember it to be aimed to work.

nishipy:
Thank you for your point. I did some testing and confirmed that `matchplay` do not go into tasks/pre_tasks/post_taks or roles section. As you mentioned, we need to improve `matchtask`.

However, it seems that the `task` which is one of the inputs of `macthtask` doesn't include information related to block and block vars. e.g. for the following playbook, the inputted `task` will be like:
```yaml
- hosts: 'localhost'
  gather_facts: true
  pre_tasks:
    - block:
      - name: 'Pre task'
        debug: msg="{{ FOO }}"
      vars:
        FOO: fooo1
```
```
{'delegate_to': <class 'ansible.utils.sentinel.Sentinel'>, 'name': 'Pre task', '__line__': 5, '__file__': 'check_test_role.yml', 'skipped_rules': [], 'action': {'__ansible_module__': 'debug', '__ansible_module_original__': 'debug', '__ansible_arguments__': [], 'msg': '{{ FOO }}'}, '__ansible_action_type__': 'meta'}
```

nishipy:
Therefore we might have to overwrite the matchtasks (not but `matchtask`).

cognifloyd:
Yes. Please `matchtask` when evaluating per-task vars. Any rules that use `matchplay` to access tasks need to be refactored. Several bug fixes for nested blocks have been merged, so those NIH rules are not benefiting from those fixes, and probably have blind spots.

If `matchtask` is not emitting block tasks, then we should modify the base `matchtasks` so that it does. If it was skipping the block tasks on purpose, them we should add an attribute for rules to opt-in to receiving the block/rescue/always task.

nishipy:
Thanks, understood. I will not override it but modify the base `matchtasks`.
Let me implement that and get back to you later.

ssbarnea:
> Thanks, understood. I will not override it but modify the base `matchtasks`. Let me implement that and get back to you later.

If you succeed you will make lots of people happy because it will fix similar bugs on lots of rules. Do not get scared if you see some test failing, you will likely have to update the tests when you implement it.

nishipy:
The `iter_tasks_in_file()`, which is called by `matchtasks`, seems not to return the block information. I think of two ideas to resolve it:
1. Modify `iter_tasks_in_file()` to accomplish that
2. Create a new function (like `get_blocks`) that will return a list of blocks
I like 2nd one and just started to implement but could you give me your opinion or other ideas if you have?

cognifloyd:
Hmm. What about modifying `ansiblelint.utils.get_action_tasks()`?

`get_action_tasks()` is only used in `ansiblelint.yaml_utils.iter_tasks_in_file()`.
`iter_tasks_in_file()` is only used in `ansiblelint.rules.AnsibleLintRule.matchtasks()`.

cognifloyd:
Or even `ansiblelint.utils.extract_from_list()` which is called from `get_action_tasks()`.

cognifloyd:
I'm not sure which approach will be most promising. Thank you for working on this / experimenting to figure out how it might be fixed!

nishipy:
Thank you for the information! Let me dig into these functions.

nishipy:
FYI, I found `ansiblelint.skip_utils._get_task_blocks_from_playbook` which seems to return a list of blocks from playbook. This is private function but maybe useful for us.

nishipy:
Updated `matchtasks` to handle the blocks in pre_tasks/tasks/post_tasks of playbooks.
If we will input the following playbook, we get no warnings.
```
- hosts: 'localhost'
  gather_facts: true
  pre_tasks:
    - name: 'block in pre_tasks'
      block:
        - name: 'pre_task 1'
          ansible.builtin.debug: msg='{{ VAR_PRETASKS }}'
      vars:
        VAR_PRETASKS: 'this is pre_task'
  tasks:
    - name: 'block in tasks'
      block:
        - name: 'task 1'
          ansible.builtin.debug: msg='{{ VAR_TASKS }}'
      vars:
        VAR_TASKS: 'this is task'
  post_tasks:
    - name: 'block in post_tasks'
      block:
        - name: 'post task 1'
          ansible.builtin.debug: msg='{{ VAR_POSTTASK }}'
      vars:
        VAR_POSTTASK: 'this is post_task'
```
These commit fix it and we can see some violations as follows:
```
WARNING  Listing 3 violation(s) that are fatal
var-naming: Task defines variables within 'vars' section that violates variable naming standards
check_test_role.yml:4 Block: block in pre_tasks

var-naming: Task defines variables within 'vars' section that violates variable naming standards
check_test_role.yml:11 Block: block in tasks

var-naming: Task defines variables within 'vars' section that violates variable naming standards
check_test_role.yml:18 Block: block in post_tasks
```

nishipy:
Hello @ssbarnea,
Should we add anything else?

src/ansiblelint/rules/__init__.py
```diff
@@ -176,6 +179,16 @@ def matchtasks(self, file: Lintable) -> List[MatchError]:
             )
             match.task = task
             matches.append(match)
+
+        # If playbook contains blocks in pre_tasks/tasks/post_tasks
+        if self.needs_block and file.kind == "playbook":
```
ssbarnea, line 184: Do we really need the needs_block? IMHO, i would prefer to see matchtasks to have a recursive behavior naturally, calling matchpath for each found deep task.

Am I missing something there? Would enabling this break other existing rules?

(Just a question, not a request to change anything)

nishipy: Thanks for the comment.
I added it with reference to `needs_raw_task` for applying only to the specific rule.
Actually I have not tried it for other rules. I just did it and got some errors on pytest. If it is necessary to remove needs_block, then I will dig into and fix.

cognifloyd: I would rather not restrict this to only `file.kind == "playbook"` as well. Tasks files can also have nested blocks.

nishipy: I see. Got it. I'll then try to remove the the `if` line, and pass the tests.

nishipy:
`ansible-lint.utils.get_action_tasks()` seems to remove the block/rescue/always elements from tasks list. So if we evaluate the nested part in `matchtasks()`, we have to delete this line but it might break other rules.
https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/utils.py#L663-L664

src/ansiblelint/rules/__init__.py
nishipy: Hi, I added some commits and removed this statement. So let me close this review.

src/ansiblelint/utils.py
```diff
@@ -521,60 +521,83 @@ def _sanitize_task(task: Dict[str, Any]) -> Dict[str, Any]:
...
+
+def normalize_task_v2(task: Dict[str, Any]) -> Dict[str, Any]:
+    """Ensure tasks have a normalized action key and strings are converted to python objects."""
+    if include_nested_task_key(task):
+        result = _get_dict_with_existing_key(task)
+        # Add dummy action for block/always/rescue statements
+        result["action"] = dict(
+            __ansible_module__="block/always/rescue",
+            __ansible_module_original__="block/always/rescue",
+        )
```
cognifloyd, line 545: The GitHub UI makes reviewing the rest of this function more difficult as it looks a bit messy.

It looks like this block could return early:
```suggestion
        result["action"] = dict(
            __ansible_module__="block/always/rescue",
            __ansible_module_original__="block/always/rescue",
        )
        return result
```


Returning early allows dedenting the rest of this function which should minimize the diff and make it easier to review the rest of this. Wdyt? Would an early return work here, or did I miss something?

cognifloyd (commented):
Ok. This is much much closer! Thank you. I have some suggestions and some questions on ways we could slightly refactor / make things clearer / make it easier to review.

src/ansiblelint/utils.py
```diff
@@ -521,60 +521,83 @@ def _sanitize_task(task: Dict[str, Any]) -> Dict[str, Any]:
     return result
 
 
-def normalize_task_v2(task: Dict[str, Any]) -> Dict[str, Any]:
-    """Ensure tasks have a normalized action key and strings are converted to python objects."""
+def _get_dict_with_existing_key(task: Dict[str, Any]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
nishipy: Thank you. It looks more understandable for us!

src/ansiblelint/utils.py
```diff
@@ -875,3 +896,13 @@ def nested_items(
...
+
+
+def include_nested_task_key(task: Dict[str, Any]) -> bool:
+    """Check if task include block/always/rescue."""
+    include_key = False
+    for key in NESTED_TASK_KEYS:
+        if task.get(key):
+            include_key = True
+
+    return include_key
```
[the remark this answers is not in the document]
nishipy: Thank you for correcting it. It is smarter.

src/ansiblelint/rules/unnamed_task.py
```diff
@@ -45,4 +46,5 @@ class TaskHasNameRule(AnsibleLintRule):
     def matchtask(
         self, task: Dict[str, Any], file: "Optional[Lintable]" = None
     ) -> Union[bool, str]:
-        return not task.get("name")
+        # Ignore nonamed block/always/rescue
+        return not (task.get("name") or include_nested_task_key(task))
```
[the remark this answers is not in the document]
nishipy: Because this rule returns a match error with the unnamed block/rescue/always if it is as it was.
If I understand correctly, we don't warn such a case currently. For e.g. in [this example](https://github.com/ansible/ansible-lint/blob/main/examples/playbooks/blockincludes.yml#L6) the blocks have no name but it is not judged as violations now. Please find [this test code](https://github.com/ansible/ansible-lint/blob/535a5b5ed3eebf9099f72ef81c629d21f29d02cb/test/test_task_includes.py#L12) as well.
```
(main)$ ansible-lint examples/playbooks/blockincludes.yml 
WARNING  Loading custom .yamllint config file, this extends our internal yamllint config.
WARNING  Listing 2 violation(s) that are fatal
unnamed-task: All tasks should be named.
examples/playbooks/tasks/directory with spaces/main.yml:3 Task/Handler: assert fail_msg=tasks in directory with spaces included

unnamed-task: All tasks should be named.
examples/playbooks/tasks/simple_task.yml:3 Task/Handler: assert fail_msg=foo

# .config/ansible-lint.yml
  - unnamed-task  # All tasks should be named.

Finished with 2 failure(s), 0 warning(s) on 4 files.
```

src/ansiblelint/utils.py
nishipy: Thanks, it works here so I added and refactored this part in the commit [fff8973](https://github.com/ansible/ansible-lint/pull/2031/commits/fff89733875d05232a128aef7fd3381bbe3d58ab).

src/ansiblelint/rules/unnamed_task.py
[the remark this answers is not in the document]
nishipy: BTW, it seems that the blocks have names and always/rescue don't in the [Ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html).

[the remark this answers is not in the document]
cognifloyd: The name is for the combined block/rescue/always, whether or not all of them are defined. I don't know if you can have a `rescue` or an `always` without a `block`, but they all work together, not as separate tasks.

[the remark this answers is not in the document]
cognifloyd: @ssbarnea should we leave a TODO or an issue trip expand this role later? Or expand that in this PR or a follow-up?

I think it's a bug for the rule to not flag blocks without names, but I'm not sure how we have to deal with backwards compatibility here.

src/ansiblelint/utils.py
```diff
@@ -521,12 +521,36 @@ def _sanitize_task(task: Dict[str, Any]) -> Dict[str, Any]:
...
+    """Return a dict with existing key in task."""
+    result = {}
+    for (k, v) in list(task.items()):
+        if k in ("action", "local_action", "args", "delegate_to"):
+            # we don't want to re-assign these values, which were
+            # determined by the ModuleArgsParser() above
+            continue
+        result[k] = v
+
+    return result
```
cognifloyd, line 534: Now that you've extracted this little utility, can we also use it for the action tasks (after the `is_nested_task` block)? Maybe one of the args could be the tuple of keys to extract. Wdyt?

cognifloyd (commented):
Oh sweet! This diff is much smaller. Thank you!

I have one more question about a possible improvement.

And then we'll need input from @ssbarnea on the last issue (backwards compatibility and the unnamed task rule).

src/ansiblelint/utils.py
nishipy: Thank you! I see. How about this?: [f0c431b](https://github.com/ansible/ansible-lint/pull/2031/commits/f0c431b76e941470fe211c246320cc418053eeda)

nishipy: Does this change make sense? or should I update the name or args?

src/ansiblelint/rules/unnamed_task.py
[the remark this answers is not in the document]
nishipy: If it is a bug, we also have to update some test code for example:
- test_no_same_owner_rule()
- test_included_tasks()
- test_runner_exclude_globs()
They don't expect match errors with unnamed blocks so `AssertionError` will come.

cognifloyd (commented):
Doh! I drafted a review but never sent it.

Here are some more cleanup suggestions.

Also, it looks like tests are failing. I don't remember why (I looked a few days ago - and I need to haul myself to bed right now). Could you check it out and try to get all tests green?

As for updating tests, I'm fine with correcting the tests now that we're fixing the behavior of the rules to include all tasks.

cognifloyd:
To clarify: Make sure `tox / docs` and `tox / py*` are passing. If `tox / eco` isn't passing - that's fine. we'll probably need to update that, but I can push that into your branch just before we merge.

src/ansiblelint/utils.py
```diff
@@ -521,12 +521,40 @@ def _sanitize_task(task: Dict[str, Any]) -> Dict[str, Any]:
...
 
+def _extract_ansible_parsed_keys_from_task(
+    result: Dict[str, Any],
+    task: Dict[str, Any],
+    keys: Tuple[str, ...],
+    action: Optional[str] = None,
+) -> Dict[str, Any]:
+    """Return a dict with existing key in task."""
+    for (k, v) in list(task.items()):
+        if k in keys or k == action:
```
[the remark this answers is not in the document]
nishipy: Thank you for the quick review!

nishipy:
Thank you for your help. Looking back at my local log of tox, I seemed to have faced some `500 Server Error` as follows on `tox/docs` last week. However all tests except `tox / eco` are passed now!

```
500 Server Error: Internal Server Error for url: https://docs.oasis-open.org/sarif/sarif/v2.1.0/
```

cognifloyd:
Thanks! I believe Sorin is on vacation - and I need input on parts of this. In the meantime, would you please add some tests?
Looks like we need tests for `normalize_task_v2` (which strangely I don't see any tests for :frowning_face: ) and `is_nested_task`.
Those tests should go in: https://github.com/ansible/ansible-lint/blob/main/test/test_utils.py

Please and thank you!

nishipy:
Thanks, sure! Let me try doing that.

nishipy:
Hi @cognifloyd, I've added tests for `normalize_task_v2` and `is_nested_task` so could you review that? Thanks!

ssbarnea (approved):

src/ansiblelint/rules/unnamed_task.py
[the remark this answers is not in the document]
cognifloyd: OK. I put the bug about `unnamed-tasks` in #2104.

#2105 also lists several rules with issues that would be resolved by just using the standard `matchtask()` thanks to this PR.

cognifloyd (approved):
I just expanded the test for `var-naming` to make sure this addresses #1895. And I separated out the remaining issue into #2104 and #2105.

PR #2041, opened 2022-03-24
--write: Allow Transforms to mark MatchErrors as fixed

cognifloyd:
This adds `MatchError.fixed` so that transforms can report when they have fixed a given issue.

When a match is marked as "fixed", it is not included in the list of failures or warnings at the end of the run.
That should ensure that any remaining issues reported are actionable.

It also reports how many issues were fixed at the end of an `ansible-lint --write` run.

When nothing was fixed, the report stays the same:

```
Finished with 10 failure(s), 7 warning(s) on 12 files.
```

If nothing is fixed, but files were reformatted, the report looks like this:
```
Modified 7 files.
Finished with 10 failure(s), 7 warning(s) on 12 files.
```

If any matches were marked fixed then the report looks like this (anything in the "fixed" count is not counted in failures or warnings):

```
Modified 7 files.
Finished with 5 failure(s), 4 warning(s), and fixed 8 issue(s) on 12 files.
```


cognifloyd:
I don't understand what the eco failure is telling me, or how this change could have caused that

ssbarnea (approved):

PR #2044, opened 2022-03-25
Update IRC channel in CONTRIBUTING.rst and add matrix

bluikko:
The old IRC channel has "moved to" topic so update docs to follow that.

Also add the Matrix channel link.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -53,7 +53,7 @@ Talk to us
 ----------
 
 Use Github `discussions`_ forum or for a live chat experience try
-``#ansible-lint`` IRC channel on libera.chat.
+``#ansible-devtools`` IRC channel on libera.chat.
 
 For the full list of Ansible IRC and Mailing list, please see the
 `Ansible Communication`_ page.
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -53,7 +53,8 @@ Talk to us
 ----------
 
 Use Github `discussions`_ forum or for a live chat experience try
-``#ansible-devtools`` IRC channel on libera.chat.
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTING.rst
```diff
@@ -53,7 +53,8 @@ Talk to us
 ----------
 
 Use Github `discussions`_ forum or for a live chat experience try
-``#ansible-lint`` IRC channel on libera.chat.
+``#ansible-devtools`` IRC channel on libera.chat or Matrix room
+`#devtools:ansible.com <https://matrix.to/#/#devtools:ansible.com>`_.
```
cognifloyd, line 57: Looks like this link is confusing the sphinx `linkcheck`: https://github.com/ansible/ansible-lint/runs/5685992864?check_suite_focus=true#step:7:166

You can probably add a matrix.to regex to `docs/conf.py`. Here are the docs: https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=linkcheck#options-for-the-linkcheck-builder

bluikko: I see. Thank you for the pointers, I'll take a look at it over the next few days then.

bluikko: Not sure if I should use `linkcheck_anchors_ignore` or follow the note in https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=linkcheck#confval-linkcheck_anchors_ignore with `linkcheck_ignore`.

I think it would be preferable to do some checking on the links (not familiar with Sphinx what checks it does exactly, for example checking its not a 404 seems good) and at least not _totally_ ignore those links.

bluikko:
Luckily the linter has good linting in the pipeline ;) a bit messy attempt but seems it is now working.

cognifloyd (approved):
Nice. Ignoring just the anchor works for me!

PR #2051, opened 2022-03-28
--write: Optionally pass rule names to --write in cli

cognifloyd:
Allow users to control which transforms `ansible-lint --write` will run.

Introduces `write_list` config file setting to complement the `--write` command line option.
Each of these accept special keywords `all` and `none` as well as a list of rule ids and rule tags.

- `--write` defaults to whatever is in the config file. If `write_list` is not populated in the config file, then `--write` is effectively the same thing as `--write=all`.
- `--write=all` allows all transforms to run.
- `--write=none` resets the list of transforms so that people can reformat YAML without running any of the transforms.
You can also use the `none` keyword to reset the list so you can run only a different set of rules than what is in the config file (as in `--write=none,rule-id` will run just rule-id even if `all` is in the config file).
- `--write=rule-id` adds `rule-id` to `write_list` from the config file. If the config file does not populate `write_list`, then this means only run the transform for rule `rule-id`.
- `--write=rule-tag` works just like `--write=rule-id` but references rule tags instead of rule ids.

Here are some examples that ignore `write_list` in the config file and selects rules tagged with `ansible-2-11-upgrade`:
```
ansible-lint --write none --write ansible-2-11-upgrade
ansible-lint --write none,ansible-2-11-upgrade
```

This example just adds `ansible-2-11-upgrade` to `write_list` from the config file:
```
ansible-lint --write ansible-2-11-upgrade
```


src/ansiblelint/cli.py
```diff
@@ -196,10 +228,20 @@ def get_cli_parser() -> argparse.ArgumentParser:
...
+        nargs="?",  # either 0 (--write) or 1 (--write=a,b,c) argument
+        action=WriteArgAction,
+        help="Allow ansible-lint to modify files. "
+        "'--write' will reformat YAML files to standardize spacing, quotes, etc. "
+        "'--write=all' will also run all available rule transforms on files "
+        "(A rule transform can fix or simplify fixing issues identified by that rule). "
+        "'--write=rule-tag,rule-id' will limit running transforms to rules that match "
+        "the given comma-separated list of rule ids or rule tags. "
+        "'--write=none' behaves the same as '--write' to reformat without running "
+        "transforms.",
```
cognifloyd, line 244: Should `--write` be equivalent to `--write=none` or `--write=all`?

cognifloyd:
#2053 should be merged first

src/ansiblelint/__main__.py
```diff
@@ -141,7 +141,7 @@ def _do_transform(result: "LintResult", opts: "Namespace") -> None:
     from ansiblelint.transformer import Transformer
 
     # future: maybe pass options to Transformer
-    transformer = Transformer(result)
+    transformer = Transformer(result, options.write_list)
```
ssbarnea, line 144: How about making it receive the entire options, as we might want to add other options that control the transformation soon. That should make it future proof.

src/ansiblelint/cli.py
ssbarnea: Write without arguments should behave as `all`.

cognifloyd:
```suggestion
        help="Allow ansible-lint to modify files. "
        "'--write' will reformat YAML files to standardize spacing, quotes, etc. "
        "''--write' will also run all available rule transforms on files "
        "(A rule transform can fix or simplify fixing issues identified by that rule). "
        "'--write=all' behaves the same as '--write' to reformat and run all transforms."
        "'--write=none' reformats without running any of the rule transforms."
        "'--write=rule-tag,rule-id' will reformat and run a subset of transforms.
        "The transforms subset includes rules that match "
        "the given comma-separated list of rule ids or rule tags. "
```

src/ansiblelint/__main__.py
cognifloyd: done

src/ansiblelint/cli.py
cognifloyd: done.

src/ansiblelint/cli.py
```diff
@@ -196,10 +275,22 @@ def get_cli_parser() -> argparse.ArgumentParser:
...
+        "A rule transform can fix or simplify fixing issues identified by that rule). "
+        "You can limit the effective rule transforms (the 'write_list') by passing a "
+        "keywords 'all' or 'none' or a comma separated list of rule ids or rule tags. "
+        "YAML reformatting happens whenever '--write' or '--write=' is used. "
+        "'--write' and '--write=all' are equivalent: they allow all transforms to run. "
+        "The effective list of transforms comes from 'write_list' in the config file, "
+        "followed whatever '--write' args are provided on the commandline. "
+        "'--write=none' resets the list of transforms to allow reformatting YAML "
+        "without running any of the transforms (ie '--write=none,rule-id' will "
+        "ignore write_list in the config file and only run the rule-id transform).",
```
cognifloyd, line 293: This is too long for a help text. It presents as this wall of text:
```
  --write [WRITE_LIST]  Allow ansible-lint to reformat YAML files and run rule transforms (Reformatting YAML
                        files standardizes spacing, quotes, etc. A rule transform can fix or simplify fixing
                        issues identified by that rule). You can limit the effective rule transforms (the
                        'write_list') by passing a keywords 'all' or 'none' or a comma separated list of
                        rule ids or rule tags. YAML reformatting happens whenever '--write' or '--write=' is
                        used. '--write' and '--write=all' are equivalent: they allow all transforms to run.
                        The effective list of transforms comes from 'write_list' in the config file,
                        followed whatever '--write' args are provided on the commandline. '--write=none'
                        resets the list of transforms to allow reformatting YAML without running any of the
                        transforms (ie '--write=none,rule-id' will ignore write_list in the config file and
                        only run the rule-id transform).
```

I guess the next thing to do is figure out where this might go in the docs and then post a link from the help text.

ssbarnea (approved):

Issue #2057, opened 2022-03-31
Invalid task file syntax in imported tasks prevents all non-playbook checks from running


wbh1:
##### Summary

If there's an invalid task file anywhere in the list of files that `ansible-lint` checks, it silently fails to do any linting besides syntax-check on playbooks.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible [core 2.12.4]
  config file = /Users/whegedus/.ansible.cfg
  configured module search path = ['/Users/whegedus/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  ansible collection location = /Users/whegedus/bits/ansible_collections:/Users/whegedus/github:/Users/whegedus/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.9.10 (main, Jan 15 2022, 11:48:00) [Clang 13.0.0 (clang-1300.0.29.3)]
  jinja version = 2.10.1
  libyaml = False


ansible-lint --version
ansible-lint 6.0.2 using ansible 2.12.4
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Tested on MacOS Monterey and Debian 10

##### STEPS TO REPRODUCE

Put invalid syntax in an imported/included file and observe with verbose logging that only playbooks are linted.

See https://github.com/wbh1/ansible-lint-test for a working example.

##### Desired Behavior

`ansible-lint` alerts me to the the invalid syntax. This is what happens if the invalid block is in `tasks/main.yml`, but does not happen on imported tasks.

This is what happens when the invalid syntax block is commented out:
```console
❯ ansible-lint -vvv
WARNING: PATH altered to include /usr/local/opt/python@3.9/bin
DEBUG    Logging initialized to level 10
[...]

PR #2058, opened 2022-03-31
Propagate error message to user on systemexit

wbh1:
A [previous PR](https://github.com/ansible/ansible-lint/pull/1984) added 'raise x from y" syntax, but consequently stopped raising
the useful error message content up the stack to the user.

This keeps the 'raise from' syntax, but just adds the error message back in.

Fixes: https://github.com/ansible/ansible-lint/issues/2057

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -257,7 +257,7 @@ def _run_cli_entrypoint() -> None:
     except KeyboardInterrupt:
         sys.exit(EXIT_CONTROL_C_RC)
     except RuntimeError as exc:
-        raise SystemExit from exc
+        raise SystemExit(exc) from exc
 
 
 def path_inject() -> None:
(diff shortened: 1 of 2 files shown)


cognifloyd (approved):

ssbarnea:
My impression was that `raise x from y` was keeping that information. Is there something special about SystemExit that makes it fail to do so?

Can someone include a link about this? I am asking as I want to read more about the subject and avoid similar problem in the future.

wbh1:
@ssbarnea it's because `SystemExit` doesn't print out a traceback [[1](https://docs.python.org/3/library/exceptions.html#SystemExit)], which is when `__cause__` is actually used. 

`raise x from y` sets the `__cause__` variable on an object inheriting from `BaseException`. That `__cause__` is used for chaining exceptions in the traceback printed to console [[2](https://docs.python.org/3/library/exceptions.html#exception-context)], but since `SystemExit` never displays a stacktrace the `__cause__` never ends up mattering.

I put [this gist](https://gist.github.com/wbh1/f74b61a82acabcc8b97b9377e7ed3f07) together to show the different behaviors.

ssbarnea (approved):

PR #2060, opened 2022-04-01
Add only-builtins rule to check compatibility with core

evgeni:
sometimes it is useful to check if a playbook can be executed on plain ansible-core with no collections installed

type: feature

### DIFF
--- a/src/ansiblelint/rules/only_builtins.py
+++ b/src/ansiblelint/rules/only_builtins.py
@@ -44,12 +44,10 @@ if "pytest" in sys.modules:
     FAIL_PLAY = """
 - hosts: localhost
   tasks:
-  - name: fake_namespace.fake_collection.fake_module
-    community.general.ini_file:
-      path: /etc/conf
-      section: drinks
-      option: fav
-      value: lemonade
+  - name: sysctl
+    ansible.posix.sysctl:
+      name: vm.swappiness
+      value: '5'
     """
 
     @pytest.mark.parametrize(
--- a/src/ansiblelint/rules/only_builtins.py
+++ b/src/ansiblelint/rules/only_builtins.py
@@ -1,0 +1,71 @@
+"""Rule definition for usage of builtin actions only."""
+import sys
+from typing import Any, Dict, Optional, Union
+
+from ansiblelint.file_utils import Lintable
+from ansiblelint.rules import AnsibleLintRule
+
+# fqcn_builtins was added in 5.1.0 as FQCNBuiltinsRule, renamed to fqcn_builtins in 6.0.0
+from ansiblelint.rules.fqcn_builtins import builtins
+
+
+class OnlyBuiltinsRule(AnsibleLintRule):
+    """Use only builtin actions."""
+
+    id = "only-builtins"
+    severity = "MEDIUM"
+    description = "Check whether the playbook uses anything but ``ansible.builtin``"
+    tags = ["opt-in", "experimental"]
+
+    def matchtask(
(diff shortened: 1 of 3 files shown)


src/ansiblelint/rules/only_builtins.py
```diff
@@ -0,0 +1,69 @@
...
+  tasks:
+  - name: shell (fqcn)
+    ansible.builtin.shell: echo This rule should not get matched by the only-builtins rule
+    """
+
+    FAIL_PLAY = """
+- hosts: localhost
+  tasks:
+  - name: sysctl (fqcn)
+    ansible.posix.sysctl:
```
evgeni, line 48: this freaks CI out, as it can't resolve the module, but I don't see a way to make that work?

evgeni: using non-FQCN notation also doesn't work (even if other tests do this, but they don't check *which* error is raised, so probably bad tests?)

evgeni: I've dropped the assert on the message for now :woman_shrugging:

src/ansiblelint/rules/only_builtins.py
```diff
@@ -0,0 +1,69 @@
...
+from ansiblelint.rules.fqcn_builtins import builtins
+
+
+class OnlyBuiltinsRule(AnsibleLintRule):
+    """Use only builtin actions."""
+
+    id = "only-builtins"
+    severity = "MEDIUM"
+    description = "Check whether the playbook uses anything but ``ansible.builtin``"
+    tags = ["opt-in", "formatting", "experimental"]
```
ssbarnea, line 18: Remove formatting, there is nothing about formatting in this rule.

src/ansiblelint/rules/only_builtins.py
```diff
@@ -0,0 +1,69 @@
...
+  tasks:
+  - name: fake_namespace.fake_collection.fake_module
+    fake_namespace.fake_collection.fake_module:
+      name: just a fake
+    """
+
+    @pytest.mark.parametrize(
+        "rule_runner", (OnlyBuiltinsRule,), indirect=["rule_runner"]
+    )
+    def test_only_builtin_fail(rule_runner: RunFromText) -> None:
```
ssbarnea, line 55: That is not a proper test, also test for message but in order to not have
syntax error, try to use something like https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html -- which is not included in ansible-core but is likely already installed.

The idea is to to ensure that the rule can be triggered.

evgeni: As noted in https://github.com/ansible/ansible-lint/pull/2060/files/7e6b349128bb2b676ad937a691b256fa5e754a25#diff-8b028ab60fefe40d188162c38df91f412a790f6ceae38c194a61a66992c7e5a4, I couldn't find a module that would not end up being not found at all.

I've pushed a version with `ini_file` but I don't think this will make the tests any happier. CI runs in a clean environment, where no collections are installed.

evgeni: Okay, I've explicitly added `ansible.posix` to the `requirements.yml` and am now using that.

If you have a better suggestion for a collection to pull in -- or know one that would be present (`community.general` is not, see https://github.com/ansible/ansible-lint/runs/6211280681?check_suite_focus=true) -- I'll happily change the test to use that.

ssbarnea (approved):

ganeshrn (approved):
Thank you!

PR #2063, opened 2022-04-02
Apply var-spacing tests to vars files

notok:
Currently, var-spacing rules are applied only to variables on tasks.
I think the rule should also be applied to vars files (like files in role/defaults/).

I wrote the code referring var-naming rule code.
~~Line numbers could not be obtained when using `matchyaml`, so~~ added jq-style path to the variable.
But the code may be a bit tricky. Any suggestions are welcome.

Fixed code to get line number from `AnsibleUnicode`.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -156,7 +156,7 @@ jobs:
       WSLENV: FORCE_COLOR:PYTEST_REQPASS:TOXENV:TOX_PARALLEL_NO_SPINNER
       # Number of expected test passes, safety measure for accidental skip of
       # tests. Update value if you add/remove tests.
-      PYTEST_REQPASS: 605
+      PYTEST_REQPASS: 606
 
     steps:
       - name: Activate WSL1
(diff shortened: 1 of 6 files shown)


notok:
eco test is failing, but this is correctly detected [lint error](https://github.com/debops/debops/blob/5c0399188b5999ae49b1dfcb9420b9440271f16b/ansible/roles/rspamd/defaults/main.yml#L318).

I sent [PR](https://github.com/debops/debops/pull/2063) to fix this to debops repo.

ssbarnea (changes requested):
This looks ok but it does not include a positive and negative test, something essential for adding it. Can you please add one?

notok:
@ssbarnea Thank you for your comment.

I'm adding tests, but I found a problem that `noqa` does not work for vars files. (This is the same for `var-naming` rule..)
I'll re-check the code.

notok:
I added test for vars file, and fixed the code to make `noqa` working.

@ssbarnea Could you please review again?

ssbarnea (approved):

ssbarnea:
@notok @cognifloy Whem making behavior change it is likely for the eco pipeline to fail as the output of linting on those repos will be different. What you need to do is to run locally `tox -e eco` and to include the modified files with your change.

I did it for you but I wanted to mention it for future changes. This gives us the opportunity to spot if the changed behavior is buggy or not.

notok:
@ssbarnea Thanks for your help. I now understood how to. I will do it next time.

PR #2072, opened 2022-04-13
Add additional default search paths for the config file

MarkusTeufelberger:
Resolves the feature request in https://github.com/ansible/ansible-lint/discussions/2056

I didn't change https://github.com/ansible/ansible-lint/blob/c058c6de39774411cf7905ed5efe320d6c48f844/src/ansiblelint/app.py#L167 and rather added a line in the documentation that this is the canonical location so other than the cli help text for `-c` it should be enough to only reference that location.

### DIFF
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -11,7 +11,9 @@ Configuration File
 ------------------
 
 Ansible-lint supports local configuration via a ``.ansible-lint`` or
-``.config/ansible-lint.yml`` configuration files. Ansible-lint checks the
+``.config/ansible-lint`` configuration file. It will also look for these files
+with ``.yml`` or ``.yaml`` extensions. The canonical location is
+``.config/ansible-lint.yml``. Ansible-lint checks the
 working directory for the presence of this file and applies any configuration
 found there. The configuration file location can also be overridden via the
 ``-c path/to/file`` CLI flag.
--- a/src/ansiblelint/cli.py
+++ b/src/ansiblelint/cli.py
@@ -94,7 +94,14 @@ def get_config_path(config_file: Optional[str] = None) -> Optional[str]:
     if config_file:
         project_filenames = [config_file]
     else:
-        project_filenames = [".ansible-lint", ".config/ansible-lint.yml"]
+        project_filenames = [
+            ".ansible-lint",
+            ".ansible-lint.yml",
+            ".ansible-lint.yaml",
+            ".config/ansible-lint",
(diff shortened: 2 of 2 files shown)


ssbarnea (changes requested):
To be honest I am against this change because it adds complexity and affects loading speed. I do not think that we need to support some many variations.

Keep in mind that any change in that are is likely to need updates for other tools that are looking for config, like IDE extensions. It can have a cascading effect.

Still, I will not use any veto powers to block it, so I will wait for more feedback before making a decision.

MarkusTeufelberger:
I'll update https://github.com/ansible/schemas next (the path there is wrong anyways)

septatrix:
> To be honest I am against this change because it adds complexity and affects loading speed. I do not think that we need to support some many variations.

Honestly I am very confused why `$PWD/.config/ansible-lint.yml` is even checked as tools usually only look into `~/.config` for global stuff and repo-local config is always in the repo root. Regarding the speed there are many tools which look for a multitude of config files like flake8 (`setup.cfg`, `tox.ini`, `.flake8`), isort (`.isort.cfg`, `pyproject.toml`, `setup.cfg`, `tox.ini`, `.editorconfig`) etc. though I am not sure if they also traverse the parent directory until they found a file or only look into the current directory (which I think would be sufficient most of the time).

If you still think that these are too many I would also be fine with only supporting `.ansible-lint`, `.ansible-lint.yml` and `.config/ansible-lint.yml`. I very much have an aversion to the `.config` dir in the repo root and my editor either does not detect the language in `.ansible-lint` or thinks it is Markdown.

> Keep in mind that any change in that are is likely to need updates for other tools that are looking for config, like IDE extensions. It can have a cascading effect.

I do not know of any tool to which this would apply. IDE extensions simply call ansible-lint from the root of the opened directory and parse its result.

MarkusTeufelberger:
As an additional data point: `yamllint` looks into the following locations for its config file (taking the first one it finds):

* .yamllint, .yamllint.yaml or .yamllint.yml in the current working directory
* the file referenced by $YAMLLINT_CONFIG_FILE, if set
* $XDG_CONFIG_HOME/yamllint/config
* ~/.config/yamllint/config

Source: https://yamllint.readthedocs.io/en/stable/configuration.html

MarkusTeufelberger:
Why did you not tell me this in the feature request that was open for about 2 weeks before I did this pull request? What changed in the past 2 weeks that this gets now denied? This is frustrating.

Issue #2076, opened 2022-04-27
Write mode introduces too many spaces inside braces


ziegenberg:
##### Summary

`ansible-lint --write` introduces extra spaces inside braces and complains about it afterwards

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console
ansible --version
ansible [core 2.12.4]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/daniel/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/daniel/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/daniel/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/daniel/.local/bin/ansible
  python version = 3.8.13 (default, Mar 17 2022, 00:00:00) [GCC 11.2.1 20220127 (Red Hat 11.2.1-9)]
  jinja version = 3.1.1
  libyaml = True

ansible-lint --version
WARNING: PATH altered to include /usr/bin
ansible-lint 6.0.2 using ansible 2.12.4
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Run `ansible-lint --write` on a document already adhering to standardized spacing, quotes, etc.

##### Desired Behavior

 `ansible-lint --write` should not format a document already adhering to standardized spacing, quotes, etc.

The following document should not be changed.

```yml
---
- hosts: web
  vars:
    config_ini:
      - {section: General, option: force_ssl, value: 1}
      - {section: General, option: "proxy_client_headers[]", value: HTTP_X_FORWARDED_FOR}

```

##### Actual Behavior

Using the following example playbook:

```yml
---
- hosts: web
  vars:
    config_ini:
      - {section: General, option: force_ssl, value: 1}
      - {section: General, option: "proxy_client_headers[]", value: HTTP_X_FORWARDED_FOR}

```
and running just `ansible-lint --write` it reports:

```console
ansible-lint --write
WARNING: PATH altered to include /usr/bin

Modified 1 files.
Finished with 0 failure(s), 0 warning(s) on 1 files.
```
The file now looks like this:

```yml
---
- hosts: web
  vars:
    config_ini:
      - { section: General, option: force_ssl, value: 1 }
      - { section: General, option: "proxy_client_headers[]", value: HTTP_X_FORWARDED_FOR }

```

Note the extra spaces inside the curly braces.

[...]

PR #2066, opened 2022-04-05
--write: add utils for dumping Jinja from AST

cognifloyd:
TODO:
- [ ] Improve dumping templates that need to be wrapped due to length
- [ ] Add more test cases
- [ ] Reword the PR description

- Add jinja TemplateDumper
- Add some basic tests and dump() func
- Add tests for jinja operators
- Better type hints
- Unused imports
- Ignore pycharm false positive warnings
- Simplify TemplateDumper signature
- reformat with black
- flake8 required fixes
- Add more typehints to satisfy mypy
- simplify elif/else after return or continue for pylint
- disable false positive pylint checks
- Add new lines after blocks in Jinja template dumper
- Consolidate jinja dumper write_block_stmt logic
- better write_block_stmt
- make sphinx happy
- format transform_utils docstrings for sphinx
- tell sphinx to ignore more classes
- Fix newlines in Jinja template dumper
- Correct test for newline feature in TemplateDumper
- Use . instead of [] for getitem dumping in simple cases
- Add TODO about protected name attributes

PR #2077, opened 2022-04-27
Relax the yamllint rules on spaces inside braces for flow mappings

ziegenberg:
This allows zero or one space inside braces for flow mappings.

Fixes: #2076

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -57,6 +57,12 @@ per-file-ignores =
   src/ansiblelint/rules/*.py: D100 D101 D102
   src/ansiblelint/rules/__init__.py: D100 D101 D102
 
+  # FIXME: C901 Function is too complex
+  # FIXME: refactor _defaults_from_yamllint_config using match case
+  #        once python 3.10 is mandatory
+  # Ref: https://github.com/ansible/ansible-lint/pull/2077
+  src/ansiblelint/yaml_utils.py: C901
+
   # FIXME: drop these once they're fixed
   # Ref: https://github.com/ansible-community/ansible-lint/issues/725
   test/*: D100 D101 D102
--- a/src/ansiblelint/yaml_utils.py
+++ b/src/ansiblelint/yaml_utils.py
@@ -59,6 +59,11 @@ rules:
   # you can easily change it or disable in your .yamllint file.
   line-length:
     max: 160
+  # We are adding an extra space inside braces as that's how prettier does it
+  # and we are trying not to fight other linters.
+  braces:
+    min-spaces-inside: 0  # yamllint defaults to 0
+    max-spaces-inside: 1  # yamllint defaults to 0
 """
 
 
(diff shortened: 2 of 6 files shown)


cognifloyd:
Let's see.

This is where we add the extra space:
https://github.com/ansible/ansible-lint/blob/aa305f72e1a56fdf6140d16ae131d136e39fe973/src/ansiblelint/yaml_utils.py#L587-L611

We would need to add a case for `braces` here:
https://github.com/ansible/ansible-lint/blob/aa305f72e1a56fdf6140d16ae131d136e39fe973/src/ansiblelint/yaml_utils.py#L828-L861

And then save the braces setting somewhere here:
https://github.com/ansible/ansible-lint/blob/aa305f72e1a56fdf6140d16ae131d136e39fe973/src/ansiblelint/yaml_utils.py#L780-L787

And pass that setting into the FormattedEmitter here ish:
https://github.com/ansible/ansible-lint/blob/aa305f72e1a56fdf6140d16ae131d136e39fe973/src/ansiblelint/yaml_utils.py#L801-L805

So we can access it in `FormattedEmitter.write_indicator` (where we're adding the space) to determine whether or not to add the space.

Our default config needs to be compatible with prettier's default (add the space in flow-style maps), but I'm happy to respect the (potentially custom) yamllint config in more cases.

cognifloyd:
This is the test for the FormattedEmitter: https://github.com/ansible/ansible-lint/blob/aa305f72e1a56fdf6140d16ae131d136e39fe973/test/test_yaml_utils.py#L171

So we might need additional test cases or an additional test function here to make sure we're respecting the yamllint config.

cognifloyd:
Thank you for working on this!

I've been a little swamped and unable to get back to finishing the `--write` stuff (jinja reformatting is my next on my plate #2066, and then I can finally add rule transforms to automatically address fixing some of the issues identified by our rules).

I will review whatever you do as quickly as I can however, because I hate holding up people's PRs.

ziegenberg:
> That might mean adding something to one or more of the test fixture files (or adding a new fixture file) that has flow maps with too many + too few spaces.

There's already a case here for too few spaces and I will add one more with too many spaces.
https://github.com/ansible/ansible-lint/blob/6516fc9797827adc96dcab52a28a62cbce385cbc/test/fixtures/formatting-before/fmt-1.yml#L25-L26

ziegenberg:
@cognifloyd I got a first version that basically works. Also the tests were extended to test with yamllint after ruaml.yaml did it's thing.

What is missing? Extending `FormattedEmitter.write_indicator`to honor the new settings. I'll probably get that done tomorrow.

Do you see anything completely horrible for now?

ziegenberg:
@cognifloyd I think, this is now ready for review.

ssbarnea (approved):

cognifloyd (approved):

PR #2090, opened 2022-04-30
Fix var-naming rule to show line numbers and apply noqa

notok:
Currently, `noqa` is not working for `vars` section in playbook and vars file.
This fix makes `noqa` working by getting line numbers from `AnsibleUnicode`.


cognifloyd (approved):
Looks sane.

ssbarnea (approved):

Issue #2071, opened 2022-04-11
Write mode rewrites 'poll: 0' as 'poll: 00' in iptables_state


JKAbrams:
##### Summary

--write changes poll: 0 into poll: 00 for iptables_state

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible [core 2.12.3]

ansible-lint --version
ansible-lint 6.0.3.dev10 using ansible 2.12.3

```

##### STEPS TO REPRODUCE

    ansible-lint --write

This task straight from the [documentation](https://docs.ansible.com/ansible/latest/collections/community/general/iptables_state_module.html) triggers the issue. 

Before:

```
---
- hosts: somehost

  tasks:
    # This will load a state from a file, with a rollback in case of access loss
    - name: restore firewall state from a file
      community.general.iptables_state:
        state: restored
        path: /run/iptables.apply
      async: "{{ ansible_timeout }}"
      poll: 0
```

After:

```
---
- hosts: somehost

  tasks:
    # This will load a state from a file, with a rollback in case of access loss
    - name: restore firewall state from a file
      community.general.iptables_state:
        state: restored
        path: /run/iptables.apply
      async: "{{ ansible_timeout }}"
      poll: 00

```

##### Desired Behavior

```
      poll: 0
```

##### Actual Behavior

```
      poll: 00
```

PR #2092, opened 2022-05-01
--write: Do not rewrite zero 0 as octal 00

ziegenberg:
As `ruamel.yaml` only preserves the octal format for YAML 1.2, the effort was made to do the same for YAML 1.1 in `ansible-linter`, as `ruamel.yaml` converts the octal to an int. While the code was partially copied from `ruamel.yaml`, the check for an actual zero got lost, resulting in the behaviour seen in #2071.

Fixes: #2071


cognifloyd (approved):
Thank you for looking into this! Looks like a great fix.

Issue #2036, opened 2022-03-23
Rule request: Disallow prompting in playbooks


cidrblock:
e.g. pause/prompt
e.g. vars_prompt

This better positions playbooks for unattended runs and use in CI pipelines.

ziegenberg:
@cidrblock The [documentation states](https://docs.ansible.com/ansible/latest/user_guide/playbooks_prompts.html): 

> Prompts for individual `vars_prompt` variables will be skipped for any variable that is already defined through the command line `--extra-vars` option, or when running from a non-interactive session (such as cron or Ansible AWX).

Would you still want to discourage the use of`vars_promt`?

And what about `vars_prompt` that defines a `default`?

PR #2093, opened 2022-05-02
Add no-prompting rule as experimental

ziegenberg:
Adds a new experimental rule to disallow the use of `vars_prompt` or `ansible.builtin.pause` to better accommodate unattended playbook runs and use in CI pipelines.

This rule is opt-in for now and needs to be explicitly enabled to be used.

Fixes: #2036


ziegenberg:
I can't get the test to work. It somehow does not include the error from the `matchplay`-call 🤷🏻‍♂️ 
Any help is appreciated.

ssbarnea (approved):

ziegenberg:
I rebased and fixed `PYTEST_REQPASS: 610` in the tox GHA workflow.

Issue #2068, opened 2022-04-06
Command line options to list tags and rules does not include opt-in rules.


zoredache:
##### Summary

I believe someone would be running ` ansible-lint -T` or `ansible-lint -L` to list the all the available rules. But by default neither of these options seem to display the opt-in rules.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible [core 2.12.1]
  config file = None
  configured module search path = ['...']
  ansible python module location = /usr/local/ansible/stable/lib/python3.9/site-packages/ansible
  ansible collection location = ...
  executable location = /usr/local/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 3.0.1
  libyaml = True

ansible-lint --version
ansible-lint 6.0.2 using ansible 2.12.1

```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Recent Debian Bullseye system, ansible, and ansible-lint installed with pip into a venv.

##### STEPS TO REPRODUCE

Run the command `ansible-lint -L`, or `ansible-lint -T`.

##### Desired Behavior

Include the opt-in rules like empty_string_compare, no_log_password, and no_same_owner and any other opt-in rules in the putput of the list rules/tags command. Maybe even with some indication, when they are not enabled.

##### Actual Behavior

None of the opt-in rules are displayed.

PR #2098, opened 2022-05-03
Include opt-in rules when verbosely listing tags and rules

ziegenberg:
With PR #1450 optional rules with the 'opt-in' tag were introduced and according to the docs, listing rules and tags with `-v` should
also list the opt-in rules.

Fixes: #2068


ssbarnea (changes requested):
We do not need to fix listing of opt-in rules and tags only for verbose mode. Can you please remove all the logic around making this dependent on verbosity?

ziegenberg:
Changed as requested. I also adapted the docs and removed the `-v` from the command for listing tags.

ziegenberg:
As PR #2099 is based on and included this PR, this is now closed after rebasing. How do we get this commit into the next release notes?

PR #2107, opened 2022-05-07
Fix support for block tasks in unnamed-task rule

### LINKED ISSUE #2104: Fix support for block tasks in unnamed-task rule
##### Summary

This is a follow-up for #1895 + #2031.

The `unnamed-task` rule does not detect block/rescue/always tasks that are missing `name:`.

In #2031, @nishipy fixed `matchtask()` so that it stops excluding block tasks. But several tests expect `unnamed-task` to skip over block tasks, so #2031 added `not is_nested_task(task)` to preserve the old (wrong) behavior.

At least these tests will need to be adjusted based on https://github.com/ansible/ansible-lint/pull/2031#discussion_r856859577:
- `test_no_same_owner_rule()`
- `test_included_tasks()`
- `test_runner_exclude_globs()`

##### Issue Type

- Bug Report

nishipy:
Fix #2104.

### DIFF
--- a/examples/roles/role_for_no_same_owner/tasks/fail.yml
+++ b/examples/roles/role_for_no_same_owner/tasks/fail.yml
@@ -13,7 +13,8 @@
 
 - name: nested-block
   block:
-    - block:
+    - name: synchronize
+      block:
         - name: synchronize-in-deep-block
           ansible.posix.synchronize:
             src: dummy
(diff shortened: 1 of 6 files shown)


cognifloyd (approved):
LGTM - but the eco changes are extensive, so someone else has to chime in on when we can merge this bugfix.

ssbarnea (approved):

nishipy:
Thanks for updating eco.

Issue #578, opened 2019-09-05
ansible lint should check order of tasks attributes for when and name


ssbarnea:
# Issue Type
- Feature request

# Desired Behaviour

Putting name of the task as first attribute should be verified, even if Ansible would allow mentioning `name:` at any point because it is a dictionary key.

Also `when:` should be after `name:` and clearly before the module itself because putting it at the end introduces two problems:

#1 Likely to introduce indentation related bug with blocks, especially when converting a single task to multiple ones inside a block. If when would be before `block:` this could never happen.

#2 It makes much harder to read tasks, `when` acts like an `if`, and is clearly more natural to "if condition do something" and "run something.... if condition", especially as "something" can easily have a big number of lines, maybe not even fitting a single screen.

```
# good task code:
- name: foo
  when: true
  module: ...

# bad task code:
- module: ...
  name: foo
  when: condition
```

Yes both do execute the same way, but from the maintenance point of view, the first one clearly wins.

MarkusTeufelberger:
I read tasks more like this: "Descriptive title", "the thing that will be done" WHEN "condition is met", so I prefer having the module name as the second member after "name". I'd view your "good" example as bad and would change "module" and "when" around (since "module" will always be there, but "when" is an optional parameter).

I strongly agree that `name` should _always_ be the first member element.

jokajak:
I agree with the original poster that when should be before the module. This aligns more along with programming languages where the conditional is placed before the executed code. I think using name, when, module helps have the best of both worlds. A description of what should be done, WHEN the condition is met, and how it is accomplished.

ssbarnea:
@MarkusTeufelberger I can provide a huge number of examples, even from production. Usually when blocks are one line, or at least nor more than 3-4. On the other hand the module parameters can easily be like 10-20 lines (docker ones especially). This means that the when condition could even go outside the screen.

Also, maybe even more important, there are the blocks. If when is after the block,  user can easily miss-indent the when and make it part of the last task from the block instead of the entire block. Copy/Paste operations are more likely to cause bugs in those cases.

Also there is the literacy aspect. When works similar to if, and in most (programming) languages you do have `if <condition> then <long list of actions>", and not "do <long list of actions> when <condition>".

MarkusTeufelberger:
[...]

PR #2108, opened 2022-05-08
Add rule to check key order

jeefberkey:
Adds a new rule that for the moment checks that name is the first key on tasks. In the future
we will add additional ordering rules to it.

Partial: https://github.com/ansible/ansible-lint/issues/578

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -155,7 +155,7 @@ jobs:
       WSLENV: FORCE_COLOR:PYTEST_REQPASS:TOXENV:TOX_PARALLEL_NO_SPINNER
       # Number of expected test passes, safety measure for accidental skip of
       # tests. Update value if you add/remove tests.
-      PYTEST_REQPASS: 629
+      PYTEST_REQPASS: 631
 
     steps:
       - name: Activate WSL1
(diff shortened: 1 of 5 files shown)


ssbarnea:
@jeefberkey Please update PYTEST_REQPASS and increase its value to match expected number of tests (+2).

jeefberkey:
Thanks for the additional guidance - `rg` was ignoring the directory when I was looking for that number.

ssbarnea:
> Thanks for the additional guidance - `rg` was ignoring the directory when I was looking for that number.

I am an rg user myself, great tool! I have an alias `rg='rg --hidden --no-follow --max-columns 255 --no-heading --column -F'` that you might like.

ssbarnea:
@jeefberkey I pushed some minor changes, like removing opt-int, we will make it experimental from start as it is very useful.

ssbarnea (approved):

jeefberkey:
Thanks for the test fixes, I was working on them but was sick the past few days. The new test name may conflict with other things I'm working on, but we can deal with that later

cidrblock (approved):
love it.

ssbarnea:
> Thanks for the test fixes, I was working on them but was sick the past few days. The new test name may conflict with other things I'm working on, but we can deal with that later

Please mention more details about conflicts we might have? While class name is not important, the `id` of the rule is harder to change once we merge it, so better to get it right from start. I used `key-order` because I want to evolve this rule to take care of ordering of keys in general.

cognifloyd (approved):

jeefberkey:
Nothing, this is actually not a blocker now that I think about it.

PR #2138, opened 2022-05-16
Update JSON Schemas

ssbarnea:


ssbarnea:
eco pipline is known and unrelated, i will sort it in follow-up.

PR #2145, opened 2022-05-16
Bump test dependencies

ssbarnea:


src/ansiblelint/file_utils.py
```diff
@@ -198,7 +198,7 @@ def __getitem__(self, key: Any) -> Any:
     def get(self, key: Any, default: Any = None) -> Any:
         """Provide compatibility subscriptable support."""
         try:
-            return self.__getitem__(key)
+            return self[key]
```
cognifloyd, line 201: Why this change? Why was it the way it was before? does `except NotImplementedError` still do the same as it did before?

ssbarnea: I think it should. The reason for that is that newer pylint uncovers this as a problem. I think that if it passed our testing it should be fine.

cognifloyd (approved):

PR #2167, opened 2022-05-20
Avoid runtime warnings from ansible-playbook --syntax-check

ssbarnea:
This improves the output in case of syntax check failures.

### DIFF
--- a/src/ansiblelint/rules/syntax_check.py
+++ b/src/ansiblelint/rules/syntax_check.py
@@ -47,21 +47,32 @@ class AnsibleSyntaxCheckRule(AnsibleLintRule):
     version_added = "v5.0.0"
 
     @staticmethod
+    # pylint: disable=too-many-locals
     def _get_ansible_syntax_check_matches(lintable: Lintable) -> List[MatchError]:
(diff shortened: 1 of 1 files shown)


src/ansiblelint/rules/syntax_check.py
```diff
@@ -47,21 +47,32 @@ class AnsibleSyntaxCheckRule(AnsibleLintRule):
...
 
         with timed_info("Executing syntax check on %s", lintable.path):
-            extra_vars_cmd = []
+            # To avoid noisy warnings we pass localhost as current inventory:
+            # [WARNING]: No inventory was parsed, only implicit localhost is available
+            # [WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'
+            args = ["-i", "localhost,"]
             if options.extra_vars:
-                extra_vars_cmd = ["--extra-vars", json.dumps(options.extra_vars)]
+                args = ["--extra-vars", json.dumps(options.extra_vars)]
```
cognifloyd, line 62: You need to extend args nope that's it has the inventory bits

cognifloyd (approved):

Issue #2105, opened 2022-05-05
Refactor rules to avoid re-inventing matchtasks()


cognifloyd:
##### Summary

Some tasks avoided using `matchtask()` because, before #2031, block tasks were excluded. Now that block tasks are included, and they are included in a more robust manner than the rule-specific methods, those rules should be refactored to just use `matchtask()`.

Another reason to make sure that rules are using the standard `matchtask()` wherever possible is for the up-coming rule transforms feature (part of `--write`). The `matchtasks()` method (which calls `matchtask()`) enriches any returned match errors with metadata that facilitates finding/fixing the issue in rule transforms.

Here is where the parsed+normalized task gets attached to the MatchError:
https://github.com/ansible/ansible-lint/blob/f9ad8ea39705850b7753e4cb13c3cc7c82f176f8/src/ansiblelint/rules/__init__.py#L177
And here is where `create_matcherror()` adds info about how the match was identified (yaml vs play vs task):
https://github.com/ansible/ansible-lint/blob/f9ad8ea39705850b7753e4cb13c3cc7c82f176f8/src/ansiblelint/rules/__init__.py#L106-L107

These are the rules that need to be adjusted:

- [x] `no-loop-var-prefix`
    - currently uses `matchplay()` with custom match methods: `handle_play()`, `handle_tasks()`, `handle_task()`.
    - unlike the standard `match*()` methodds, the custom `block` detection in this rule is incomplete: it only descends into `block`, not `rescue` or `always`.
    - the `handle_*()` and `matchplay()` methods should be condensed into a single `matchtask()`.

- [x] `no-same-owner`
    - currently uses `matchplay()` with custom match methods: `handle_play()`, `handle_playlist()`, `handle_task()`.
    - like `no-loop-var-prefix` this also misses `rescue` and `always`.
    - the `handle_*()` and `matchplay()` methods should be condensed into a single `matchtask()` (`handle_unarchive()` and `handle_synchronize()` would probably stay as they are)

- [ ] `partial-become`
    - uses `matchplay()` with custom logic to search through the yaml for tasks.
    - actually does handle `block`, `rescue`, and `always`.
    - implementing a rule transform for this will be more difficult because it does not re-use the standard methods that also enrich matches with additional info.
    - `matchplay()` should only look at plays, not tasks.
    - `matchtask()` should handle tasks separately.

- [x] `deprecated-local-action`
    - uses `match()` to look at each line
    - can get false positives in comments or strings
[...]

PR #2216, opened 2022-06-14
Add test for no-loop-var prefix rule

nishipy:
A part of #2105.
Before refactoring, this PR adds some tests for the `no-loop-var-prefix` rule.

### DIFF
--- a/examples/roles/loop_var_prefix/tasks/fail.yml
+++ b/examples/roles/loop_var_prefix/tasks/fail.yml
@@ -6,7 +6,7 @@
   loop:
     - foo
     - bar
-- name: that should fail due to wrong custom
+- name: that should fail due to wrong prefix
   ansible.builtin.debug:
     var: zz_item
   loop:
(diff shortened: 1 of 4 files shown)


examples/roles/loop_var_prefix/tasks/fail.yml
```diff
@@ -0,0 +1,24 @@
+---
+# 3 expected no-loop-var-prefix failures at 3, 9, 19
+- name: that should trigger no-loop-var-prefix
+  ansible.builtin.debug:
+    var: item
+  loop:
+    - foo
+    - bar
+- name: that should fail due to wrong custom
```
cognifloyd, line 9:
```suggestion
- name: that should fail due to wrong prefix
```

cognifloyd (approved):
Nice! The name of one of the test tasks confused me at first. I left a note one possible way to clarify it.

Other than that, it looks good to me.

examples/roles/loop_var_prefix/tasks/fail.yml
nishipy: Thank you!

ssbarnea (approved):

PR #2218, opened 2022-06-15
Refactor no_loop_var_prefix rule to condense methods into matchtask()

nishipy:
A part of #2105.

This PR refactors the `no_loop_var_prefix` rule and resolves the following problem. Please refer to [here](https://github.com/ansible/ansible-lint/issues/2105#issue-1226186054).
> - currently uses matchplay() with custom match methods: handle_play(), handle_tasks(), handle_task().
> - unlike the standard match*() methodds, the custom block detection in this rule is incomplete: it only descends into block, not rescue or always.
> - the handle_*() and matchplay() methods should be condensed into a single matchtask().


nishipy:
I will wait until the schema update gets merged

ssbarnea (approved):

PR #2222, opened 2022-06-23
Update key-order check to check all keys

jeefberkey:
A user can set the key_order setting to one of the following options:

* default - checks only the order of name and action
* enhanced - checks a smaller list of possible keys
* everything - checks a list containing every key


jeefberkey:
Will update this soon - struggling to find time to debug the typing errors recently

ssbarnea:
@jeefberkey Any chance you can resume work on this change? It was a nice addition to make to the linter.

jeefberkey:
@ssbarnea I picked this back up a few days ago - but my tests keep on failing for things I have not modified. Can you help me with that? 

Also, it seems like the eco test have changed. Didn't the results used to be checked in to git? It looks like the test still uses `git diff`, so I'm confused.

jeefberkey:
Everything seems well now - I looked for a bit for a comprehensive list of allowed keys, but never found one. Hopefully the yaml file makes it a little easier to maintain.

@ssbarnea Thanks!

PR #2225, opened 2022-06-28
Allow yaml extension for inventory and requirements

netsandbox:


ssbarnea:
Thanks for raising it, now we need to think a little bit about the implications.

The title is a little bit misleading because the tool does already recognize `.yaml` extension very well, is only the ansible specific **config** files which use only the default/documented extension. That was by design and not a mistake.

I am not sure we want to allow this because using these extensions will likely break LOTS of tools in the ecosystem, as most of them only recognize one filename.

We might require documentation proof that these extensions are allowed. The fact that they might happen to work, may not be enough. I mention this because recently we faced some other request about things that were never documented as being supported/allowed.

Keep in mind that it is summer, so we might need to wait a little bit more when we have to make a decision. Meanwhile you can configure overrides in your ansible-lint config as a workaround.

@cidrblock WDYT?

netsandbox:
For example for inventory, it is documented in the plugin:
https://github.com/ansible/ansible/blob/4bd7e50612d4bff228506ea6f419abe2935bb04d/lib/ansible/plugins/inventory/yaml.py#L21-L25

and also in the schema:
https://github.com/ansible/schemas/blob/a5b163c41ebcfa6f88cfa23c0734671a21ad3bbf/f/ansible-inventory.json#L49-L53

src/ansiblelint/config.py
```diff
@@ -11,14 +11,14 @@
...
-    {"ansible-lint-config": "**/.config/ansible-lint.yml"},
+    {"ansible-lint-config": "**/.config/ansible-lint.{yaml,yml}"},
     {"ansible-navigator-config": "**/ansible-navigator.{yaml,yml}"},
-    {"inventory": "**/inventory/**.yml"},
-    {"requirements": "**/meta/requirements.yml"},  # v1 only
+    {"inventory": "**/inventory/**.{yaml,yml}"},
+    {"requirements": "**/meta/requirements.{yaml,yml}"},  # v1 only
     # https://docs.ansible.com/ansible/latest/dev_guide/collections_galaxy_meta.html
-    {"galaxy": "**/galaxy.yml"},  # Galaxy collection meta
+    {"galaxy": "**/galaxy.{yaml,yml}"},  # Galaxy collection meta
```
cognifloyd, line 21: This is not supported in ansible-core. I prefer `.yaml` but I recently had to switch back to `galaxy.yml` when `ansible-galaxy` couldn't find it.

cognifloyd (commented):
Yeah, I hate how ansible defaults to `.yml` in so many places. But, at least for `galaxy.yml`, that's required.

src/ansiblelint/config.py
netsandbox: Took me some time to find it in the code: https://github.com/ansible/ansible/blob/4c9385dab720ae5d33054d521cecda31464c8857/lib/ansible/galaxy/dependency_resolution/dataclasses.py#L39

I reverted the change for `galaxy.yml` and also for `execution-environment.yml` where also the `.yaml` extension isn't supported.

netsandbox:
I also reverted `.config/ansible-lint.yml` which also doesn't support the `.yaml `extension :cry: :
https://github.com/ansible/ansible-lint/blob/4c4eb53a7854c2a143ccfaf159d034f744528641/src/ansiblelint/cli.py#L100

So there are only `inventory` and `requirements` files left.

cognifloyd:
Both inventory and requirements files are entirely user defined. Ansible only uses them if someone explicitly listed them in a config file, env var, or cli arg. So, we should support `.yaml` at least for those.

I require my team to use `.yaml` for inventory files and anything else that is allowed to avoid the legacy `.yml` extension.

ssbarnea:
I do have some doubts that you will be able to upload collections to galaxy using a `requirements.yaml` file instead of `requirements.yml`. Someone needs to try it.

ssbarnea (approved):

PR #2238, opened 2022-07-14
deprecated-local-action: refactor to use matchtask() 

nishipy:
A part of #2105.

This PR fixes the following points for `deprecated-local-action`:
> - uses match() to look at each line
> - can get false positives in comments or strings
> - does not benefit from any of the additional info that gets added by other standard methods

### DIFF
--- a/src/ansiblelint/rules/deprecated_local_action.py
+++ b/src/ansiblelint/rules/deprecated_local_action.py
@@ -27,7 +27,7 @@ class TaskNoLocalAction(AnsibleLintRule):
     ) -> Union[bool, str]:
         """Return matches for a task."""
         if "old_style" in task.keys():
-            if task["old_style"] == "local_action":
+            if "local_action" in task["old_style"]:
                 return True
 
         return False
--- a/src/ansiblelint/rules/deprecated_local_action.py
+++ b/src/ansiblelint/rules/deprecated_local_action.py
@@ -3,8 +3,14 @@
 # Copyright (c) 2018, Ansible Project
 
 import sys
+from typing import TYPE_CHECKING, Any, Dict, Union
 
 from ansiblelint.rules import AnsibleLintRule
+
+if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


src/ansiblelint/utils.py
```diff
@@ -604,6 +620,11 @@ def normalize_task_v2(task: Dict[str, Any]) -> Dict[str, Any]:
         del arguments["argv"]
 
     result["action"].update(arguments)
+
+    # check if module invocated with old style
+    old_style_keys = ("local_action",)
+    result = _extract_old_style_keys_from_task(result, sanitized_task, old_style_keys)
```
cognifloyd, line 626: This is kind of cool.

But, we might not need to change any of the utils because of another newish feature: `needs_raw_task`.

If the rule class has `needs_raw_task = True` as a class level variable, then `matchtask` gets passed the raw/unparsed task in `task["__raw_task__"]`.

Would that work here?

src/ansiblelint/rules/deprecated_local_action.py
```diff
@@ -16,28 +22,41 @@ class TaskNoLocalAction(AnsibleLintRule):
...
     - name: task example
       local_action:
         module: boto3_facts
     """
 
-    def test_local_action() -> None:
+    SUCCESS_TASK = """
+    - name: task example
+      boto3_facts:
+      delegate_to: localhost # local_action
```
cognifloyd, line 52: Sweet example!

cognifloyd (commented):
Nice! I left a comment about a possible alternative implementation. This is exciting to see all these improvements! Thank you!

src/ansiblelint/utils.py
nishipy: Thank you! Using `task["__raw_task__"]`  looks smart and enough for us. Let me fix that.

cognifloyd (approved):

ssbarnea (approved):

PR #2247, opened 2022-07-25
Add OracleLinux schema

facorazza:


ssbarnea:
This change must be made into https://github.com/ansible/schemas and we will update the vendored schemas here after they get merged there.

Issue #1794, opened 2022-01-06
yamllint config with 'level': 'warning', will trigger fatal violation


konstruktoid:
##### Summary

Even if the custom `.yamllint.yml` is read, the `'line-length': {'level': 'warning', 'max': 120, 'allow-non-breakable-words': True, 'allow-non-breakable-inline-mappings': False}` rule will trigger an fatal violation.

`yamllint` itself triggers only a warning.

See https://github.com/konstruktoid/ansible-role-hardening/runs/4732728069?check_suite_focus=true

`.yamllint.yml`:
```yaml
---
extends: default
ignore: ".tox*\n.git*"
rules:
  line-length:
    max: 120
    level: warning
```

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```
$ ansible-lint --version
[WARNING]: You are running the development version of Ansible. You should only
run Ansible from "devel" if you are modifying the Ansible engine, or trying out
features under development. This is a rapidly changing source of code and can
become unstable at any point.
ansible-lint 5.3.2.dev5 using ansible 2.13.0.dev0
$ ansible-lint -vv
[...]
WARNING  Loading custom .yamllint.yml config file, this extends our internal yamllint config.
[...]

PR #2248, opened 2022-07-25
Skip yamllint warnings

dthomson-triumf:
Hi, I saw that #1794 was marked with a "good first issue" tag, so I thought I'd try doing a small fix for this bug that I just ran up against.

The PR should be fairly straightforward to read, though I'm not sure if this is along the lines of what you might be looking for with a solution.  Comments, suggestions and concerns are all welcome.

Thanks!

### DIFF
--- a/examples/yamllint/yamllint-warning.yaml
+++ b/examples/yamllint/yamllint-warning.yaml
@@ -1,0 +1,8 @@
+---
+rules:
+  line-length:
+    level: warning
+  trailing-spaces:
+    level: warning
+  key-duplicates:
+    level: warning
--- a/src/ansiblelint/rules/__init__.py
+++ b/src/ansiblelint/rules/__init__.py
@@ -85,6 +85,7 @@ class AnsibleLintRule(BaseRule):
         details: str = "",
         filename: Optional[Union[str, Lintable]] = None,
         tag: str = "",
+        ignored: bool = None
     ) -> MatchError:
         """Instantiate a new MatchError."""
         match = MatchError(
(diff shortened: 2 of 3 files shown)


ssbarnea:
@dthomson-triumf I do have the impression that you are trying to fix something that does not need to be fixed.

As far the linter is concerned any message from yamllint is an error. We do not care if yamllint is configured to treat them as errors or warnings, for us all of them are violations.

If you want to make ansible-lint to treat a specific yamllint rule as a warning, you can add these to `warn_list` as `yaml[some-rule]` and it should work.

dthomson-triumf:
I thought that #1794 indicated that this was a bug that needs fixing?

Regardless, using the `yaml[some-rule]` entry in the `warn_list` doesn't seem to work for me with `ansible-lint 6.3.0` and `yamllint 1.27.1`, but maybe I'm misunderstanding something.

The following setup:

```yaml
# playbook.yml
---
- hosts: localhost
  tasks:
    - name: Here is a long task
      ansible.builtin.command: echo "This is the output from a very long command that's not going to make ansible-lint very happy"
      changed_when: false
```

```yaml
# .ansible-lint
---
warn_list:
  - yaml[list-length]
```

```yaml
# .yamllint
---
rules:
  line-length:
    level: warning
    max: 80
```
Still gives me:

```shell
me@myhost:~/ansible-lint-warns$ ansible-lint 
yaml: line too long (130 > 80 characters) (yaml[line-length])
playbook.yml:5

# .config/ansible-lint.yml
  - yaml  # Violations reported by yamllint.

Finished with 1 failure(s), 0 warning(s) on 1 files.
```
Only using:

```yaml
warn_list:
  - yaml
```

Treats the `yamllint` issues as warnings, but that means every single reported result from yamllint, including those that are being listed as `errors` and not just `warnings` will be treated as `warnings` by ansible-lint.

PR #2252, opened 2022-07-28
Upgrade jsonschema to 4.8.0

ssbarnea:
Upgrades required version of jsonschema as newer version changed
some of the validation error messages and we want to be in full
control over the outcome.


ssbarnea:
Eco pipeline will fail but that is only because the output is changed due to new version of the library. That one is safe to ignore.

PR #2259, opened 2022-08-01
Rename var-spacing rule to jinja[spacing]

ssbarnea:
This change prepares for refactoring of jinja2 processing and will
enable us to add other rules related to jinja2 template strings.


src/ansiblelint/rules/jinja.md
```diff
@@ -0,0 +1,20 @@
+## jinja
+
+This rule can report problems related to jinja2 string templates. Ins current
```
cognifloyd, line 3:
```suggestion
This rule can report problems related to jinja2 string templates. The current
```

cognifloyd (approved):

Issue #2171, opened 2022-05-23
Rule request: Name all plays and task blocks


cidrblock:
Starting to pull in possible rule ideas from the community of practice: https://github.com/redhat-cop/automation-good-practices/pull/64

In #2169 it is suggested that all tasks have a name.  

This extends that to cover plays and task blocks.

ssbarnea:
@cidrblock Here are few things I found so far:
- Despite being even mentioned in an example on https://docs.ansible.com/ansible/latest/collections/ansible/builtin/import_playbook_module.html it seems that the "name" alongside `import_playbook` is not displayed by ansible when executing. That works against the case of making it required.
- name on blocks do not appear as listing either
- names on normal plays are visible, so for those we want to make it mandatory

WDYT?

cidrblock:
IMHO, the name for an import playbook and block is equally important, it's unfortunate it doesn't show when the playbook is run, but when looking through the content to better understand what is going on and the original author's intent they can be really useful.

I find import_playbook and a block to be even more complex sometimes than a simple task (e.g. set_fact, debug) , so maybe it's even a little more important to have them here.

Issue #2170, opened 2022-05-23
Rule request: All task names should be written in the imperative


cidrblock:
Starting to pull in possible rule ideas from the community of practice: https://github.com/redhat-cop/automation-good-practices/pull/64

A task in a task file or play performs an action, the name of the task should be written in the imperative.

We might need to pull this one and use some internals rather than maintain an internal list:

https://github.com/PyCQA/pydocstyle/blob/master/src/pydocstyle/data/imperatives.txt

ssbarnea:
Are we ok to add pydocstyle as a dependency for the linter? Do we have any other python library that can help us achieve the same goal?

Issue #2169, opened 2022-05-23
name-capital: All tasks should be named and start with a capital letter


cidrblock:
Starting to pull in possible rule ideas from the community of practice: https://github.com/redhat-cop/automation-good-practices/pull/64

The description of a task (the name key) should start with a capital letter, this will help with readability and help task files and playbook read more like documentation.

ktdreyer:
Anyone have a good `sed` command that will globally fix this problem? It would be great to add this to the ansible-lint docs.

ssbarnea:
@cognifloyd ^ that might be a good opportunity to add rewrite for this one.

I used code (vscode) search and replace in files with regex to fix them.

bolinocroustibat:
I also express strong disagreement with this rule. A name referring to a role should not include capital letters. Capital letter are for complete sentences, not for labels.

PR #2263, opened 2022-08-07
Reimplement unamed-task rule as name[missing]

ssbarnea:
This reimplementation will allow us to implement closely related extra checks such [`name[casing]`](https://english.stackexchange.com/questions/413690/is-casing-a-valid-alternative-to-capitalization#:~:text=Standard%20terminology%20in%20the%20world,it%2C%20so%20that%20the%20typesetting), `name[imperative]`, `name[unique]`, `name[suffix]`.

Related: #2171 #2170 #2169


src/ansiblelint/rules/unnamed_task.py
```diff
@@ -1,48 +0,0 @@
-"""Implementation of unnamed-task rule."""
-
-# Copyright (c) 2016 Will Thames <will@thames.id.au>
```
cognifloyd, line 3: Do we need to include the copyright notice in the new file?

src/ansiblelint/rules/name.py
```diff
@@ -0,0 +1,56 @@
...
+    severity = "MEDIUM"
+    tags = ["idiom"]
+    version_added = "historic"
+
+    def matchtask(
+        self, task: Dict[str, Any], file: "Optional[Lintable]" = None
+    ) -> Union[bool, str, MatchError]:
+        if not task.get("name"):
+            return self.create_matcherror(
+                linenumber=task["__line__"], tag="name[missing]", filename=file
```
cognifloyd, line 31: Let's use the LINE_NUMBER_KEY constant in ansiblelint.utils.LINE_NUMBER_KEY

cognifloyd: Also, this bypasses some of the enrichment that happens in `matchtasks()` like `details` and `match.task`.
https://github.com/ansible/ansible-lint/blob/a5f0d521e404ae8839d5879b20fb44200aeca1f9/src/ansiblelint/rules/__init__.py#L162-L177

I'm not sure where we should add that. Maybe `matchtasks()` should be refactored to update `details` if it is not set, and add `match.task`.

cognifloyd (commented):
Dang. Sorry I didn't review before this was merged. I have a few questions to address in a follow-up.

src/ansiblelint/rules/unnamed_task.py
ssbarnea: Not really needed as it does not have a legal value, also the new one is practically a rewrite, that is why I did not include it.

src/ansiblelint/rules/name.py
ssbarnea: I was aware about missing details but I was not sure what to include in them. The reality is that the generic "name" description is no longer relevant for various violations.

Probably we need to add some longer descriptions but we need to write the text for them.

Issue #2288, opened 2022-08-13
multi-document yaml files are not supported


ssbarnea:
We fail to process at YAML files that are multi-document but:
- pyyaml supports them
- yamllint supports them
- ansible itself does not accept them
- our `parse_yaml_linenumbers` method throws error, while ideally it should not, this being caused by the fact that its implementation relies on Ansible loader and not other one.

PR #2290, opened 2022-08-13
Add ability to process multi-document YAML files

ssbarnea:
Closes: #2288


src/ansiblelint/utils.py
```diff
@@ -709,12 +709,15 @@ def get_action_tasks(data: AnsibleBaseYAMLObject, file: Lintable) -> List[Any]:
     ]
 
 
-@lru_cache(maxsize=128)
-def parse_yaml_linenumbers(lintable: Lintable) -> AnsibleBaseYAMLObject:
+@lru_cache(maxsize=0)
```
cognifloyd: Does 0 make the cache unlimited here?

cognifloyd (commented):
Cool! I hadn't considered multi document yaml files. I wonder if `--write` messes them up.

src/ansiblelint/utils.py
ssbarnea: Fixed it to None to make it unlimited after reading the docs. To my surprise they used None instead of 0.

ssbarnea:
> Cool! I hadn't considered multi document yaml files. I wonder if `--write` messes them up.

Multi-documents are extremely uncommon, but you might want to add some code to avoid messing them if you encounter them. I spotted on inside debops repo, that is why I made this fix.

cognifloyd:
Another example: I frequently see multi-document yaml with kubernetes resources.

PR #2297, opened 2022-08-14
Implement PEP-563

ssbarnea:
As the minimal version of python is py38, we can now avoid having to quote type hint information.

Reference: https://adamj.eu/tech/2021/05/15/python-type-hints-future-annotations/
Partial-Fix: https://github.com/ansible/devtools/issues/65


ssbarnea:
@shatakshiiii Take a look at this, once this merges you can try to do the same change for ansible-navigator.

cognifloyd (approved):

Issue #2271, opened 2022-08-09
yaml[line-length] rule can't be skipped when the triggering line is in a YAML block scalar


brlin-tw:
##### Summary

For readability and line length reasons I prefer folding long lines using YAML's block scalar functionality, however the yaml[line-length] rule can't seem to be skipped using no-qa comment when the long line is in a YAML folded/literal block scalar.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.13.2]
  config file = /stripped/ansible.cfg
  configured module search path = ['/home/brlin/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/brlin/.local/pipx/venvs/ansible/lib/python3.10/site-packages/ansible
  ansible collection location = /home/brlin/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/brlin/.local/bin/ansible
  python version = 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
  jinja version = 3.1.2
  libyaml = True

$ ansible-lint --version
WARNING: PATH altered to include /home/brlin/.local/pipx/venvs/ansible/bin
ansible-lint 6.4.0 using ansible 2.13.2

```

- ansible installation method: pip(via pipx)
- ansible-lint installation method: pip(via pipx)

##### OS / ENVIRONMENT

OS: Ubuntu 22.04 LTS

##### Desired Behavior

The long line should be skippable without changing the task's behavior.

##### Actual Behavior

The following playbook:

```yaml
- name: Example PoC
  hosts: all
  become: true
  tasks:
    - name: Install cronjob
      ansible.builtin.blockinfile:
        path: /etc/crontab
        block: >- # noqa yaml[line-length]
          @reboot root
          curl
          --data key=some.very.looooooooooooooooooooooooooooooooooooooooooooooooong.value
          https://example.com
```

yields:

```
$ ansible-lint playbook.yaml 
WARNING: PATH altered to include /home/brlin/.local/pipx/venvs/ansible/bin
WARNING  Listing 1 violation(s) that are fatal
yaml: line too long (89 > 80 characters) (yaml[line-length])
playbook.yaml:11

You can skip specific rules or tags by adding them to your configuration file:
# .config/ansible-lint.yml
warn_list:  # or 'skip_list' to silence them completely
  - yaml[line-length]  # Violations reported by yamllint.

[...]

PR #2300, opened 2022-08-15
Fix identification of inline noqa skips

ssbarnea:
Fixes: #2271
Fixes: #2320

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -39,7 +39,6 @@ use_default_rules: true
 # This makes linter to fully ignore rules/tags listed below
 skip_list:
   - skip_this_tag
-  - latest[git]
 
 # Any rule that has the 'opt-in' tag will not be loaded unless its 'id' is
 # mentioned in the enable_list:
@@ -57,7 +56,6 @@ enable_list:
 # This makes the linter display but not fail for rules/tags listed below:
 warn_list:
   - skip_this_tag
-  - latest[git]
   - experimental # experimental is included in the implicit list
   # - role-name
   # - yaml[document-start]  # you can also use sub-rule matches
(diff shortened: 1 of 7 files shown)


src/ansiblelint/rules/yaml.py
```diff
@@ -61,17 +60,33 @@ def matchyaml(self, file: Lintable) -> List[MatchError]:
...
+                    break
+            else:
+                filtered_matches.append(match)
         return filtered_matches
 
 
+def _fetch_line_skips(data: Any) -> Generator[Tuple[int, List[str]], None, None]:
+    """Generate (line, skip_list) by looking recursively in."""
+    if hasattr(data, "get") and data.get("__line__") and data.get("skipped_rules", []):
+        yield data.get("__line__"), data.get("skipped_rules", [])
```
cognifloyd, line 78: Use the constants please

src/ansiblelint/rules/yaml.py
```diff
@@ -61,17 +60,33 @@ def matchyaml(self, file: Lintable) -> List[MatchError]:
...
+    if isinstance(data, Iterable) and not isinstance(data, str):
+        if isinstance(data, dict):
+            for entry, value in data.items():
+                for line, rules in _fetch_line_skips(value):
+                    yield line, rules
```
cognifloyd, line 83:
```suggestion
                yield from _fetch_line_skips(value)
```

src/ansiblelint/rules/yaml.py
```diff
@@ -61,17 +60,33 @@ def matchyaml(self, file: Lintable) -> List[MatchError]:
...
+        else:
+            for entry in data:
+                for line, rules in _fetch_line_skips(entry):
+                    yield line, rules
```
cognifloyd, line 87:
```suggestion
                yield from _fetch_line_skips(entry)
```

cognifloyd (changes requested):
Looks good. I left a couple cleanup suggestions earlier.

Issue #2301, opened 2022-08-16
var-spacing rule triggers even if in warn_list


konstruktoid:
##### Summary

`var-spacing` rule triggers even if in `warn_list`, `jinja[spacing]` does not.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible-lint 6.4.1.dev27 using ansible 2.12.7
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

macOS
Python 3.9.13

##### STEPS TO REPRODUCE

```console
$ ansible-lint --version
WARNING: PATH altered to include /usr/local/opt/python@3.9/bin
ansible-lint 6.4.1.dev27 using ansible 2.12.7
$ ansible-lint
[...]
# .config/ansible-lint.yml
warn_list:  # or 'skip_list' to silence them completely
  - jinja[spacing]  # Rule that looks inside jinja2 templates.
  - var-spacing  # Jinja2 variables and filters should have spaces before and after.

Finished with 26 failure(s), 332 warning(s) on 124 files.
$ cat .ansible-lint
---
exclude_paths:
  - .git/
  - .github/
  - tests/
enable_list:
  - fqcn-builtins
warn_list:
  - jinja[spacing]  # Rule that looks inside jinja2 templates.
  - line-length  # Violations reported by yamllint
  - name[casing]  # Rule for checking task names and their content.
  - var-spacing  # Jinja2 variables and filters should have spaces before and after.
  - yaml[line-length]  # Violations reported by yamllint
$ ansible-lint
[...]
You can skip specific rules or tags by adding them to your configuration file:
# .config/ansible-lint.yml
warn_list:  # or 'skip_list' to silence them completely
  - var-spacing  # Jinja2 variables and filters should have spaces before and after.

Finished with 13 failure(s), 345 warning(s) on 124 files.
```

##### Desired Behavior

`var-spacing` should behave as the `jinja[spacing]` rule

konstruktoid:
```sh
---
exclude_paths:
  - .git/
  - .github/
  - tests/
enable_list:
  - fqcn-builtins
warn_list:
  - jinja[spacing]  # Rule that looks inside jinja2 templates.
  - line-length  # Violations reported by yamllint
  - name[casing]  # Rule for checking task names and their content.
  - yaml[line-length]  # Violations reported by yamllint
skip_list:
[...]

Issue #2285, opened 2022-08-12
var-spacing: false-positive in regexp


skwde:
##### Summary

Rather similar to <https://github.com/ansible/ansible-lint/issues/2208> but with variables.

The pipe in the `regexp: '({{ var1 }}|{{ var2 }})'` triggers a false-positive.

Also related

- <https://github.com/ansible/ansible-lint/issues/2241>
- <https://github.com/ansible/ansible-lint/issues/2209>

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
# ansible --version
ansible [core 2.13.2]
  config file = $HOME/ansible-test/ansible.cfg
  configured module search path = ['$HOME/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/miniconda3/envs/ansible/lib/python3.10/site-packages/ansible
  ansible collection location = $HOME/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/miniconda3/envs/ansible/bin/ansible
  python version = 3.10.5 | packaged by conda-forge | (main, Jun 14 2022, 07:04:59) [GCC 10.3.0]
  jinja version = 3.1.2
  libyaml = True

# ansible-lint --version
ansible-lint 6.4.0 using ansible 2.13.2
```

- ansible installation method: conda
- ansible-lint installation method: conda

##### STEPS TO REPRODUCE

```console (paste below)

```

##### Desired Behavior

The regexp should not trigger a var-spacing warning

##### Actual Behavior

A minimal playbook might look like

```yml
---
- name: test
  hosts: localhost
  gather_facts: false
  tasks:
    - name: test
      ansible.builtin.lineinfile:
        path: /some/file.txt
        regexp: '({{ var1 }}|{{ var2 }})'
        line: 'abc'
```

The pipe triggers a var-spacing warning.

Issue #2241, opened 2022-07-20
var-spacing: False-positive on logical OR (||)


imrejonk:
##### Summary

The var-spacing check incorrectly reports a logical OR (||) as not having enough spaces.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.13.2]
  config file = None
  configured module search path = ['/home/imre/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/imre/.local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/imre/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/imre/.local/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 3.1.2
  libyaml = True

$ ansible-lint --version
ansible-lint 6.3.0 using ansible 2.13.2
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Debian 11 GNU/Linux, Python 3.9

##### STEPS TO REPRODUCE

Run ansible-lint against this minimal example playbook:

```
---

- name: Test var-spacing
  hosts: ['localhost']
  gather_facts: False

  tasks:

    - name: Debug
      debug:
        msg: '{{ lookup("pipe", "echo Testing || true") }}'
```

##### Desired Behavior

ansible-lint should not consider a logical OR (||) as a var-spacing issue.

##### Actual Behavior

```
$ ansible-lint test.yml 
WARNING  Overriding detected file kind 'yaml' with 'playbook' for given positional argument: test.yml
WARNING  Listing 1 violation(s) that are fatal
var-spacing: Jinja2 variables and filters should have spaces before and after.
test.yml:9 Task/Handler: Debug

You can skip specific rules or tags by adding them to your configuration file:
# .config/ansible-lint.yml
warn_list:  # or 'skip_list' to silence them completely
  - var-spacing  # Jinja2 variables and filters should have spaces before and after.

Finished with 1 failure(s), 0 warning(s) on 1 files.
```

##### Additional information

Also see these false-positive issues in the var-spacing check: #2208, #2209

ssbarnea:
PR would be more than welcomed, this might also be a duplicate report.

nishipy:
I'd like to work on this issue.

ssbarnea:
[...]

Issue #2209, opened 2022-06-10
var-spacing: False-positive in name


sblask:
##### Summary

I am getting:

```
var-spacing:Jinja2 variables and filters should have spaces before and after.
ansible/includes/zipped_executable.yml:5 Task/Handler: Add {{ product | title }}
```

with ansible-lint 6.3

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.13.0]
  config file = /Users/sebastian/Clones/dotfiles/ansible.cfg
  configured module search path = ['/Users/sebastian/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/sebastian/.opt/ansible/lib/python3.9/site-packages/ansible
  ansible collection location = /Users/sebastian/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/sebastian/.bin/ansible
  python version = 3.9.4 (default, Apr 29 2021, 10:26:40) [Clang 12.0.5 (clang-1205.0.22.9)]
  jinja version = 3.0.3
  libyaml = True

$ ansible-lint --version
WARNING: PATH altered to include /Users/sebastian/.opt/ansible/bin
ansible-lint 6.3.0 using ansible 2.13.0
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

MacOS

##### STEPS TO REPRODUCE

See https://github.com/sblask/dotfiles/runs/6824840729?check_suite_focus=true

You can also just run ansible-lint on https://github.com/sblask/dotfiles/blob/main/ansible/includes/zipped_executable.yml to get the following:

```console (paste below)
$ ansible-lint ansible/includes/zipped_executable.yml
WARNING: PATH altered to include /Users/sebastian/.opt/ansible/bin
WARNING  Listing 3 violation(s) that are fatal
var-spacing: Jinja2 variables and filters should have spaces before and after.
ansible/includes/zipped_executable.yml:5 Task/Handler: Add {{ product | title  }}

risky-file-permissions: File permissions unset or incorrect.
ansible/includes/zipped_executable.yml:8 Task/Handler: Download {{ product | title }}

var-spacing: Jinja2 variables and filters should have spaces before and after.
ansible/includes/zipped_executable.yml:13 Task/Handler: Unpack {{ product | title }}

You can skip specific rules or tags by adding them to your configuration file:
[...]

PR #2306, opened 2022-08-18
Reimplement jinja[spacing] to avoid use of regex

ssbarnea:
This change rewrites jinja2 rule to make it use jinaja lexer and
avoid use of regexes for parsing, as they cause too many false
positives.

Fixes: #2301
Fixes: #2285
Fixes: #2241
Fixes: #2209
Fixes: #2208
Fixes: #2120

### DIFF
--- a/src/ansiblelint/rules/jinja.py
+++ b/src/ansiblelint/rules/jinja.py
@@ -1,27 +1,24 @@
 """Rule for checking content of jinja template strings."""
-# Copyright (c) 2016, Will Thames and contributors
-# Copyright (c) 2018, Ansible Project
 from __future__ import annotations
 
-import re
+import logging
 import sys
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
+import jinja2
 from ansible.parsing.yaml.objects import AnsibleUnicode
 
 from ansiblelint.file_utils import Lintable
 from ansiblelint.rules import AnsibleLintRule
 from ansiblelint.skip_utils import get_rule_skips_from_line
-from ansiblelint.utils import LINE_NUMBER_KEY, parse_yaml_from_file, task_to_str
+from ansiblelint.utils import LINE_NUMBER_KEY, parse_yaml_from_file
 from ansiblelint.yaml_utils import nested_items_path
 
 if TYPE_CHECKING:
     from ansiblelint.errors import MatchError
 
-# Please note that current implementation is only able to report
-# `jinja[spacing]` occurrences, formerly named `var-spacing`. We aim to refactor
-# the implementation to replace currently unreliable regex with use of jinja2
-# parser. That will also allow use to implement additional checks.
+
+_logger = logging.getLogger(__package__)
 
 
 class JinjaRule(AnsibleLintRule):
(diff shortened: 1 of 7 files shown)


cognifloyd (approved):
fascinating use of the jinja lexer.
In the lexer-related work I've been doing, I've been using `TOKEN_*` constants from jinja:
https://github.com/pallets/jinja/blob/main/src/jinja2/lexer.py#L61-L109

Typically I do `from jinja2 import lexer as j2tokens` and then use `j2tokens.TOKEN_OPERATOR` or similar. What do you think of that idiom?

src/ansiblelint/rules/jinja.py
```diff
@@ -94,6 +86,83 @@ def matchyaml(self, file: Lintable) -> List[MatchError]:
...
+                    # breakpoint()
+                    value = " "
+
+                tokens.append((lineno, token_type, value))
+                prev_token_type = token_type
+                prev_value = value
+
+            reformatted = "".join([value for _, _, value in tokens])
+            details = (
+                f"Jinja2 template could better be rewritten as `{reformatted}`."
```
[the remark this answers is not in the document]
ssbarnea: Tx. FYI, I found other issues and I am rewriting it, while keeping the tests and likely adding few more.

ssbarnea:
@drybjed I am not sure how well you will take this but this change seems to uncover at least one genuine jinja2 syntax error in debops repository, but you might find yourself bit overwhelmed by the amount of feedback linter is providing:



Take it easy! And to quote from [thread](https://github.com/ansible/ansible-lint/discussions/2205),... I hope you get paid by the hour... ;)

ssbarnea:
@cognifloyd There are 4 reformatting test that are failing which I do not know how to fix. If you can have a look it would be great. That PR took all my weekend and I might really benefit from another pair of eyes to deal with these last ones.

All these tests were results of different errors I found while testing formatting of jinja2 expression on all eco repositories, so we should be pleased about the results.

Few minutes ago I got another idea,... what if I would try to use black itself to format the snippets, assuming that it might be used programately. Still, doing that would not be a joy as I would have to identify blocks of tokens and recombine them before I feed them to black,.. so maybe not a very good idea.

src/ansiblelint/rules/jinja.py
```diff
@@ -131,58 +344,175 @@ def test_var_spacing(
...
+            pytest.param(
+                # "{% if a|int <= 8 -%} iptables {%- else -%} iptables-nft {%- endif %}",
+                "{% if a|int <= 8 -%} iptables {%- else -%} iptables-nft {%- endif %}",
+                "{% if a | int <= 8 -%} iptables {%- else -%} iptables-nft {%- endif %}",
+                "spacing",
+                id="29",
+            ),
+            pytest.param(
+                # "- 2" -> "-2", minus does not get separated when there is no left side
+                "{{ - 2 }}",
```
cognifloyd, line 485: The AST calls this an Unary node. I looked through the parser, and I don't see an easy way to identify this based only on the lexer tokens.

So maybe track the previous non-whitespace token. See if the token before the `-` or `+` operator is also an operator (one of https://github.com/pallets/jinja/blob/main/src/jinja2/lexer.py#L112-L139 except for right tokens: `]`, `)`, `}`) or if the previous token is a variable start token `{{`.

src/ansiblelint/rules/jinja.py
```diff
@@ -131,58 +344,175 @@ def test_var_spacing(
...
+            pytest.param(
+                # array ranges do not have space added
+                "{{ foo[2:4] }}",
+                "{{ foo[2:4] }}",
+                "spacing",
+                id="32",
+            ),
+            pytest.param(
+                # array ranges have the extra space removed
+                "{{ foo[2: 4] }}",
```
cognifloyd, line 506: In the AST, this is a Slice node.

You will need to track the state to successfully identify this. If within `[` skip the spaces. Maybe have a stack var, and every time you enter a left operator, put it on the stack (a list), then pop it once you hit the closing token. So, if within `{}`, `:` gets a space after. If within `[]` no spaces around `:`.

cognifloyd (commented):
Ew. Yes those are difficult. Here are some ideas to hopefully give you an idea of what to look at next.

drybjed:
@ssbarnea Thanks for the heads up. I'll test the proposed changes and make fixes ASAP.

src/ansiblelint/rules/jinja.py
```diff
@@ -131,58 +357,175 @@ def test_var_spacing(
...
+                "{{ foo[2: 4] }}",
```
cognifloyd, line 520:
```suggestion
                "{{ foo[2:4] }}",
```

src/ansiblelint/rules/jinja.py
ssbarnea: No AST/black for the moment, likely to open another can of worms. Time will tell if we need them or not.

drybjed:
> @drybjed I am not sure how well you will take this but this change seems to uncover at least one genuine jinja2 syntax error in debops repository, but you might find yourself bit overwhelmed by the amount of feedback linter is providing:
> 
> 
> 
> Take it easy! And to quote from [thread](https://github.com/ansible/ansible-lint/discussions/2205),... I hope you get paid by the hour... ;)

@ssbarnea I'm currently working on fixing all these warnings on `ansible-lint` v6.5.0 so that DebOps can be used to fix false-positives in v6.5.1. I wish for one thing - separate Jinja identation into its own rule so that it could be ignored if desired. There are small snippets of Jinja where I had to drop idented `if`/`endif` statements, and that's alright with me, but for larger pieces of Jinja, for example [more extensive templates](https://github.com/debops/debops/blob/master/ansible/roles/journald/templates/etc/systemd/journald.conf.d/ansible.conf.j2), please please please let me keep identation - it makes things much more readable.

drybjed:
@ssbarnea All issues in DebOps have been [fixed](https://github.com/debops/debops/pull/2185) in the `master` branch. There are some false-positives though on the latest `ansible-lint` release.

ssbarnea:
@drybjed Thanks. I got some progress today on them but again failed to pass all tests. That jinja spacing is a bit of a can of worms.

I already introduce a condition to skip the spacing checks when newlines are found inside the jinja blocks, basically because we assume user added them for indentation purposes. The scope of checking spaces was more towards simple blocks.

PR #2315, opened 2022-08-22
Enable flake9-future-annotations

ssbarnea:
Related: https://github.com/ansible/devtools/issues/65

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -125,6 +125,7 @@ repos:
           # - flake8-black>=0.1.1
           - flake8-docstrings>=1.5.0
           - flake8-pytest-style>=1.2.2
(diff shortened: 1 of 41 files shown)


.pre-commit-config.yaml
```diff
@@ -125,6 +125,7 @@ repos:
           # - flake8-black>=0.1.1
           - flake8-docstrings>=1.5.0
           - flake8-pytest-style>=1.2.2
+          - flake8-future-annotations>=0.0.3
```
ssbarnea, line 128: @shatakshiiii Can you please try to make the same change to navigator project? It is enough to modify the `.pre-commit-config.yaml` and run `tox -e lint`, which will tell you which files are missing the imports.

PR #2316, opened 2022-08-22
Enable pyupgrade pre-commit hook

ssbarnea:


cidrblock (approved):
:heart:

cognifloyd (approved):
Woohoo! Type annotations are so much better this way!

PR #2321, opened 2022-08-23
Progressive mode. Return absolute path for Lintable

kostyaplis:
Another issue with progressive mode.
Relative paths are different for files in the current folder and in .cache folder.
Those paths included in `matches`, so `matches` considered different as well.

Added lintable details output to demonstrate the issue:
Latest master:
```
ansible-lint --progressive /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable name: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable self.name: roles/msp.service.legacy_as/defaults/main.yml
Lintable self.path: roles/msp.service.legacy_as/defaults/main.yml
Preparing worktree (checking out 'old-rev')
HEAD is now at 3d62073b0c Automatic merge from XXXXXX -> XXXXXX
Lintable name: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable self.name: ../../../../repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable self.path: ../../../../repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
```

With `normpath` hack:
```
ansible-lint --progressive /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable name: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable self.name: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable self.path: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Preparing worktree (checking out 'old-rev')
HEAD is now at 3d62073b0c Automatic merge from XXXXXX -> XXXXXX
Lintable name: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable self.name: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
Lintable self.path: /Users/konstantinplis/repositories/msp-scm/roles/msp.service.legacy_as/defaults/main.yml
WARNING  Total violations not increased since previous commit, will mark result as success. (1 -> 0)
WARNING  Marked 1 previously known violation(s) as ignored due to progressive mode.
WARNING  Listing 1 violation(s) marked as ignored, likely already known
```

Do let me know if this is a legitimate way to fix the issue, or it needs to be more sophisticated here.

### DIFF
--- a/src/ansiblelint/file_utils.py
+++ b/src/ansiblelint/file_utils.py
@@ -60,7 +60,7 @@ def normpath(path: str | BasePathLike) -> str:
     relpath = os.path.relpath(str(path))
     path_absolute = os.path.abspath(str(path))
     # we avoid returning relative paths that end-up at root level
-    if path_absolute in relpath:
+    if path_absolute in relpath or vars(options).get("progressive"):
         return path_absolute
     return relpath
 


ssbarnea:
That one is indeed a tricky issue to address because, by design, we keep the absolute path of each lintable (and matches too), so we do not endup with cross-contamination.

Still, the way progressive works is that it created a different workspace with with the previous commit and lints its content. That means a different set of matches that point to a different repository on disk. and now we need to diff between these in order to determine if the current commit improved the situation or not.

It would be really useful to have someone that is actively using progressive to do some tests on this fix, just to know that we got it right this time.

ssbarnea:
To be honest I am inclined to close this change unless someone else steps in to verify it and assert that this is indeed desirable. Let me explain why: normally I would expect ansible-lint to report problems only from within current codebase and never to report problems from inside its cache directory or other external place.

If that ever happens it would be up to the user to reconfigure it not to report problems about code that he cannot really fix himself. Basically the fact that cache folder is different between the two runs, should not impact the outcome.

I am afraid to merge this change as it had the danger of introducing a buggy or unexpected behavior for other users, some that might expect relative paths when using progressive.

kostyaplis:
> To be honest I am inclined to close this change unless someone else steps in to verify it and assert that this is indeed desirable. Let me explain why: normally I would expect ansible-lint to report problems only from within current codebase and never to report problems from inside its cache directory or other external place.
> 
> If that ever happens it would be up to the user to reconfigure it not to report problems about code that he cannot really fix himself. Basically the fact that cache folder is different between the two runs, should not impact the outcome.
> 
> I am afraid to merge this change as it had the danger of introducing a buggy or unexpected behavior for other users, some that might expect relative paths when using progressive.

Please give me a time till Monday, I'll open an issue with more details in regards to this PR.

ssbarnea:
@kostyaplis Take your time, I only wanted to state my worries. It can obviously stay open few more days.

kostyaplis:
ok...with further digging i've noticed a misuse on my side.
The issue only exists when passing the absolute file path to ansible-lint.
e.g.

```
cd ~/test-repo
ansible-lint --progressive playbook-test.yaml

result.matches = [[rule_name](rule desc) matched playbook-test.yaml:3]
old_result.matches = [[rule_name](rule desc) matched playbook-test.yaml:3]
``` 
Result: progressive mode works as expected


```
cd ~/test-repo
ansible-lint --progressive ~/test-repo/playbook-test.yaml

result.matches = [[rule_name](rule desc) matched playbook-test.yaml:3]
old_result.matches = [[rule_name](rule desc) matched ../../../test-repo/playbook-test.yaml:3]
```
Result: Relative file paths of current and previous commits doesn't match, so progressive mode doesn't work

Should be closed, I think

PR #2328, opened 2022-08-23
jinja[spacing]: fix negative numbers inside arrays

ssbarnea:
Fixes: #2317

PR #2331, opened 2022-08-24
jinja[spacing] Ensure ~ is correctly spaced

ssbarnea:
Fixes: #2328


felixfontein:
Hmm, felixfontein.acme still has several places where it wants to remove the space after `~`.

PR #2346, opened 2022-08-26
Create template_instead_of_copy.py

GhostLyrics:
I'm sending a PR in response to [a request to upstream a rule ](https://github.com/ansible/ansible-lint/discussions/1754#discussioncomment-3173529)that complains about templating inside of `ansible.builtin.copy`. I'm sorry, I did not know how to decide on severity nor did I know how to find out in which version of Ansible that warning was added to the documentation.


ssbarnea (changes requested):
First, thanks for contributing this new rule. Here are few things missing:
- no tests included
- no markdown documentation included, create a template-instead-of-copy.md file in the same directory as the rule and try to match the same format used from the other rules, with before/after examples.
- rules should trigger only if user tries to use jinja inside content, not on just the presence of content key.

GhostLyrics:
> Here are few things missing:
> 
> * no tests included

Will check.

> * no markdown documentation included, create a template-instead-of-copy.md file in the same directory as the rule and try to match the same format used from the other rules, with before/after examples.

Will check.

> * rules should trigger only if user tries to use jinja inside content, not on just the presence of content key.

IMHO, this is only part of the solution. Thinking about the comment and the linked threads, I probably want to do one of the following is true:
- jinja context
- new lines via escape (`\n`, etc).

What do you think?

cognifloyd:
Please add tests. That will make your target use cases easier to understand.

Please make sure you have a test that shows `content:` with jinja that skips this rule using a comment. I have some self contained playbooks that must not use any other files based on how I distribute/run them (in some kubernetes pods managed by a larger workflow engine). So, in general, this rule sounds good, but I need to make sure I can ignore it in those cases.

ssbarnea (approved):

bendem:
> Items which are templated should use ``template`` instead of ``copy`` with ``content`` to ensure correctness.

I'm kind of confused about this comment. What could be "incorrect" from using copy[content] instead of template?
I can understand it's better taste (though that's an opinion), but I don't understand how it could be incorrect. Am I missing something?

ssbarnea:
@bendem Ping me back if the current version is still unclear, I can ask someone else to produce even more detailed examples and/or explanation. From what I understood, the difference is subtle, but it is. That rule is not about "style" preference, is about the risk of getting an unexpected outcome.

Issue #2340, opened 2022-08-26
name[casing]: False positive when Chinese task/play names are used


brlin-tw:
##### Summary

The following error is triggered in our integration tests:

```
playbooks/ping.yml:6 Task/Handler: 檢查被控端是否可以連接
```

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.13.2]
  config file = /media/veracrypt1/文件/中信安科技/解決方案/自動化 Automation/Ansible 自動化配置工具/Ansible 角色範本/ansible.cfg
  configured module search path = ['/home/brlin/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/brlin/.local/pipx/venvs/ansible/lib/python3.10/site-packages/ansible
  ansible collection location = __REDACTED__/playbooks/collections
  executable location = /home/brlin/.local/bin/ansible
  python version = 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
  jinja version = 3.1.2
  libyaml = True

$ ansible-lint --version
WARNING: PATH altered to include /home/brlin/.local/pipx/venvs/ansible/bin
ansible-lint 6.5.1 using ansible 2.13.2
```

- ansible installation method: pip(pipx)
- ansible-lint installation method: pip(pipx)

##### OS / ENVIRONMENT

OS: Ubuntu 22.04

##### STEPS TO REPRODUCE

```yaml
# playbooks/ping.yml
- name: 測試範例(PoC)
  hosts: all
  tasks:
    - name: 檢查被控端是否可以連接
      ping:
```

##### Desired Behavior

When `name` is start with character that is not in a locale that has the casing concept, skip the rule.

##### Actual Behavior

```paste below
$ ansible-lint playbooks/ping.yml 
WARNING: PATH altered to include /home/brlin/.local/pipx/venvs/ansible/bin
WARNING  Listing 2 violation(s) that are fatal
name: All names should start with an uppercase letter. (name[casing])
playbooks/ping.yml:2

name: All names should start with an uppercase letter. (name[casing])
playbooks/ping.yml:5 Task/Handler: 檢查被控端是否可以連接


Finished with 0 failure(s), 2 warning(s) on 1 files.

```

ssbarnea:
[...]

PR #2347, opened 2022-08-27
name[casing]: now applies only to scripts supporting uppercase

ssbarnea:
Fixes: #2340

### DIFF
--- a/src/ansiblelint/rules/name.py
+++ b/src/ansiblelint/rules/name.py
@@ -69,7 +69,10 @@ class NameRule(AnsibleLintRule):
     def _check_name(
         self, name: str, lintable: Lintable | None, linenumber: int
     ) -> MatchError | None:
-        if not name[0].isupper():
+        # This rules applies only to scripts that do have uppercase and
+        # lowercase letter, so we ignore anything else. On Unicode islower()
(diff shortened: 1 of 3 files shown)


src/ansiblelint/rules/name.py
```diff
@@ -69,7 +69,10 @@ def matchtask(
     def _check_name(
         self, name: str, lintable: Lintable | None, linenumber: int
     ) -> MatchError | None:
-        if not name[0].isupper():
+        # This rules applies only to scripts that do have uppercase and
+        # lowercase letter, so we ignore anything else. On Unicode islower()
+        # is not the opposite of islower()
```
cognifloyd, line 74:
```suggestion
        # This rules applies only to languages that do have uppercase and
        # lowercase letter, so we ignore anything else. On Unicode isupper()
        # is not necessarily the opposite of islower()
```

cognifloyd (commented):
Sorry I missed reviewing earlier

src/ansiblelint/rules/name.py
ssbarnea: Thanks, I included the fix into https://github.com/ansible/ansible-lint/pull/2351

PR #2359, opened 2022-08-29
name[play]: now also correctly works with import_playbook blocks

apatard:
When running ansible-lint on a playbook like:

```
---
- ansible.builtin.import_playbook: "test.yml"
```

ansible-lint is now "complaining":
```
WARNING  Ignored exception from NameRule.<bound method AnsibleLintRule.matchyaml of name: Rule for checking task and play names.>: 'name'
```

The test in src/ansiblelint/rules/name.py:matchplay() is trying to ignore
the import_playbook case but with current test leads to the line:

```
data["name"], lintable=file, linenumber=data[LINE_NUMBER_KEY]
```

and in this example, the key "name" doesn't exist, leading to the warning.

The proposed solution here is to split the test into two and return
no error if "name" is not set for import_playbook.

[ Not sure what's currently best practice with import_playbook:
should a name be set or not ? ]

### DIFF
--- a/src/ansiblelint/rules/name.py
+++ b/src/ansiblelint/rules/name.py
@@ -31,10 +31,7 @@ class NameRule(AnsibleLintRule):
         """Return matches found for a specific play (entry in playbook)."""
         if file.kind != "playbook":
             return []
-        if "name" not in data and not any(
-            key in data
-            for key in ["import_playbook", "ansible.builtin.import_playbook"]
-        ):
+        if "name" not in data:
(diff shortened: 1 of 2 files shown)


ssbarnea:
Older versions did not check for it but based on my discussions with @cidrblock i think it would be better to require it here too, for documentation purposes. At this moment ansible does not display the name for these in the UI but this may also change in the future.

ssbarnea (approved):

PR #2364, opened 2022-08-30
Handle FileNotFoundError caused by processing new file in progressive mode

### LINKED ISSUE #2363: Progressive mode is failing for new files
##### Summary

When scanning a newly added file with `progressive: true`, `FileNotFoundError` exception is raised because the file doesn't exist in the previous commit
Note: only reproducible when passing the file path as an argument to ansible-lint.

##### Issue Type

- Bug Report

```console
$ ansible-lint --version                                                                    
ansible-lint 6.5.1 using ansible 2.13.3
```

##### STEPS TO REPRODUCE

1. Add `playbook-test1.yml` to the existing git repository
2. Commit
3. Run ansible-lint in progressive mode 

```console (paste below)
$ ansible-lint --progressive playbook-test1.yml

Preparing worktree (checking out 'old-rev')
HEAD is now at f17ba41 Second commit. Fix 1 violation
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/__main__.py", line 281, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/__main__.py", line 211, in main
    old_result = _get_matches(rules, options)
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/runner.py", line 221, in _get_matches
    matches.extend(runner.run())
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/runner.py", line 120, in run
    lintable.data
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/file_utils.py", line 294, in data
    self._data = parse_yaml_linenumbers(self)
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/utils.py", line 739, in parse_yaml_linenumbers
    loader = AnsibleLoader(lintable.content, **kwargs)
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/file_utils.py", line 220, in content
    self._populate_content_cache_from_disk()
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/site-packages/ansiblelint/file_utils.py", line 212, in _populate_content_cache_from_disk
    self._content = self.path.resolve().read_text(encoding="utf-8")
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/pathlib.py", line 1132, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
  File "/Users/konstantinplis/.pyenv/versions/3.10.1/lib/python3.10/pathlib.py", line 1117, in open
    return self._accessor.open(self, mode, buffering, encoding, errors,
FileNotFoundError: [Errno 2] No such file or directory: '/Users/konstantinplis/.cache/ansible-compat/f23de3/old-rev/playbook-test1.yml'
```

##### Desired Behavior

Should previous versions of newly added files be treated as an empty string?

##### Actual Behavior

Unhandled exception

kostyaplis:
Fixes #2363

### DIFF
--- a/src/ansiblelint/file_utils.py
+++ b/src/ansiblelint/file_utils.py
@@ -209,7 +209,13 @@ class Lintable:
 
     def _populate_content_cache_from_disk(self) -> None:
         # Can raise UnicodeDecodeError
-        self._content = self.path.resolve().read_text(encoding="utf-8")
+        try:
+            self._content = self.path.resolve().read_text(encoding="utf-8")
+        except FileNotFoundError as ex:
+            if vars(options).get("progressive"):
+                self._content = ""
+            else:
+                raise ex
         if self._original_content is None:
             self._original_content = self._content
 


kostyaplis:
Apparently, this breaks desired behaviour when passing non-existent file to ansible-lint.
Maybe we should be more specific that we are checking old revision right now, so only in this case we can swallow `FileNotFoundError`.
E.g.
1. Add `in_old_revision=False` to `options` Namespace
2. Set it to True when entering [with _previous_revision()](https://github.com/ansible/ansible-lint/blob/f2d9471417df1c0ec4bc4f1cea6a044cae162236/src/ansiblelint/__main__.py#L214) and set back to `False` before exiting
3. check `if vars(options).get("in_old_revision"):` in [_populate_content_cache_from_disk()](https://github.com/ansible/ansible-lint/blob/f2d9471417df1c0ec4bc4f1cea6a044cae162236/src/ansiblelint/file_utils.py#L210), then swallow `FileNotFoundError`

Does this sound reasonable?

ssbarnea:
If I understood correctly, you were using progressive mode while still passing filenames as arguments to the linter. TBH, I never considered this use-case as most of the time we call it without even passing any file arguments.

ssbarnea:
I personally do not like this hack because it is inside load content, I wonder if we could catch this error in a different place and ignore it.

Do you happen to have a repository where running linter in progressive can reproduce the bug? If so, I may be able to find a better place to catch this error and to ignore it.

kostyaplis:
Yes, you can found test repository [here](https://github.com/kostyaplis/ansible-lint-test)
Run `ansible-lint --progressive playbook-test1.yml` to reproduce an error.

> If I understood correctly, you were using progressive mode while still passing filenames as arguments to the linter. TBH, I never considered this use-case as most of the time we call it without even passing any file arguments.

Our repository contains >400 playbooks and >300 roles, so our idea is to only lint changed/added files and only report regressions(at least for now). That's why all the recent activity around progressive mode, as I'm actively testing it with our use-cases.

ssbarnea (approved):
It seems to be working but because no test was added, we might see it again in the future. Thanks!

PR #2377, opened 2022-09-01
[Docs] md doc for deprecated_module

oraNod:
@ssbarnea Hi, I've finally got a start on adding docs for rules. It's a small step but I'm planning to add more soon. Could you please take a quick look and let me know if this seems OK before I tackle more? Thanks.


ssbarnea (approved):
Tx!

ssbarnea:
@oraNod BTW, I replaced your original comment from the description because it was not suitable for a commit message. Because we merge with squash, bug title/description become commit message, so we often need to alter them.

That first contribution was flawless! Good example and explanation, clearly will help few users. I will merge it as soon GHA is passing.

oraNod:
Thanks @ssbarnea ! I'll keep the description in mind for my next PR.

ssbarnea:
In fact I was wrong, only the title is used. The merge message is combined from the commits, so forget about my comment.

Issue #2103, opened 2022-05-04
meta-version: confirm collection version is >= 1.0.0


thedoubl3j:
confirm via galaxy.yml file if key `version:` is greater than or equal to 1.0.0. 

rule should be optional and only enabled at users discretion. Error level should be error/fatal (1) as Automation Hub cannot house content that is not fully released via semantic versioning standards. 

based off of code found [here](https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/schema.py#L149)

KB-perByte:
This looks interesting, I would take a look at this enhancement.

CC, @ssbarnea

felixfontein:
Seriously, but this rule makes zero sense and should be removed. There is no single reason I can think of why a collection version should be 1.0.0 or larger.

felixfontein:
(If the intention was that this is optional and needs to be explicitly enabled, then the current implementation does not take this into account.)

ssbarnea:
@felixfontein That rule was included in [shared](https://ansible-lint.readthedocs.io/profiles/#shared) profile, so that is why is enabled by default (no profile mentioned equals with everything being enabled). As @thedoubl3j showed, there is code inside galaxy importer that requests a version like this. I guess the reason is related to semver, as nothing before `1.0.0` can be trusted for keeping backwards compatibility. The rule can be disabled, so I do not find this as too bad.

@cidrblock WDYT?

felixfontein:
@ssbarnea collection releases with version < 1.0.0 have always been possible on Ansible Galaxy and still are. See for example https://galaxy.ansible.com/community/internal_test_tools and https://galaxy.ansible.com/felixfontein/acme, which both still have 0.x.y versions, and most community and cloud collections, which started with 0.1.x (https://galaxy.ansible.com/community/general, https://galaxy.ansible.com/amazon/aws) or even 0.0.1 (can't find an example now, but I'm pretty sure I've seen that in the past).

felixfontein:
Right now the 1.0.0 restriction is only there for collections that are included in the Ansible package, and apparently (according to @thedoubl3j, I've never used or published collections to AH) for collections that are published on Automation Hub.

PR #2383, opened 2022-09-02
Add rule to verify that galaxy.yml collection version is present and >= 1.0.0

KB-perByte:
Rule to check collection version in galaxy.yml is >= 1.0.0

```yaml
---
name: foo
namespace: bar
version: 0.2.3 # <-- collection version should be <= 1.0.0
authors:
  - John
readme: ../README.md
description: "..."
```

Fixes: #2103


src/ansiblelint/rules/galaxy_collection_version.py
```diff
@@ -0,0 +1,103 @@
...
+                    message="collection version should be greater than or equal to 1.0.0",
+                    linenumber=data[LINE_NUMBER_KEY],
+                    tag="collection-version[galaxy]",
+                    filename=file,
+                )
+            ]
+
+
+@total_ordering
+class Version:
```
KB-perByte, line 54: @ssbarnea I wanna move this to somewhere generic, like a utility part for it to be reused among the whole collection. would it be the right candidate for `ansiblelint.utils` ?

ssbarnea:
Do not bother to put the code in a shared lib yet, we can do it as soon we will need it.

ssbarnea (approved):

Issue #457, opened 2019-01-04
[E602] Don't compare to empty string - what's a better way?


geerlingguy:
# Issue Type

- Bug report

# Ansible and Ansible Lint details

```
$ ansible --version
ansible 2.7.5
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/Users/jgeerling/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python2.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 2.7.15 (default, Jul 17 2018, 19:41:20) [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)]

$ ansible-lint --version
ansible-lint 4.0.0
```

- ansible installation method: pip
- ansible-lint installation method: pip

# Desired Behaviour

Using a string comparison to check for a string being empty should be A-OK:

```yaml
- name: Do something with my_var.
  command: "do_something_with {{ my_var }}"
  when: my_var != ''
```

ansible-lint should throw no errors for this code.

# Actual Behaviour (Bug report only)

ansible-lint throws the following:

```
[602] Don't compare to empty string
```

The problem is, if I set it to:

```yaml
- name: Do something with my_var.
  command: "do_something_with {{ my_var }}"
  when: my_var
```

Then I get the error:

```
fatal: [localhost]: FAILED! =>
The conditional check 'my_var' failed. The error was: error while evaluating conditional (my_var): 'testing' is undefined
```

So I'm wondering—if we're not supposed to compare to empty strings in a duck-typed language like Python, what _are_ we supposed to do? There are a ton of use cases for comparison to an empty string.

## Workarounds

Currently there are two ways to get past this:

  1. Head-in-the-sand:
     1. Add `.ansible-lint` file in your role/playbook.
     1. Add the following to skip the rule entirely:

            skip_list:
              - '602'

  2. Bypass the rule's regex:

         when: my_var | length > 0

Number 1 is just ignoring the issue. Number 2 is more obtuse and makes for more unreadable code (at least IMO)... but is that preferred if we truly need to check if a string has a value?

geerlingguy:
For reference, original discussion was from: https://github.com/ansible/ansible-lint/issues/450

geerlingguy:
[...]

PR #2400, opened 2022-09-08
Docs: empty string compare rule

oraNod:
Add MD documentation for the empty_string_compare rule.

### DIFF
--- a/src/ansiblelint/rules/empty_string_compare.md
+++ b/src/ansiblelint/rules/empty_string_compare.md
@@ -1,0 +1,44 @@
+# empty-string-compare
+
+This rule checks for empty string comparison in playbooks.
+To ensure code clarity you should avoid using empty strings in conditional statements with the `when` clause.
+
+- Use `when: var|length > 0` instead of `when: var != ""`.
+- Use `when: var|length == 0` instead of `when: var == ""`.
+
+This is an opt-in rule.
+You must enable it in your Ansible-lint configuration as follows:
+
+```yaml
+enable_list:
+  - empty-string-compare
+```
+
+## Problematic Code
+
+```yaml
+---
+- name: Example playbook
+  hosts: all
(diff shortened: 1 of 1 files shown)


src/ansiblelint/rules/empty_string_compare.md
```diff
@@ -0,0 +1,44 @@
...
+```yaml
+---
+- name: Example playbook
+  hosts: all
+  tasks:
+    - name: Shut down
+      ansible.builtin.shell: |
+        /sbin/shutdown -t now
+        echo $var ==
+      when: ansible_os_family
```
oraNod, line 43: @ssbarnea I hope you don't read this on your PTO but I wanted to add a comment in this PR. I'm not sure what this is doing and it even seems to violate the rule. I asked about this in slack and have read through https://github.com/ansible/ansible-lint/issues/457 

IMO it would be good to clarify this example. Please let me know what you think (when you're back and ready to deal with it ofc). Thanks!

ssbarnea:
@oraNod it is much better to make the change atomic (per rule), even if that means opening 10 pull requests. While I made few fixes on your change I observed that the no-handler docs and change seems wrong and I would like to remove it from this change, but that is not so easy if everything is in a bundle.

oraNod:
@ssbarnea Do you want me to break this one into multiple PRs too? Or was it just moving the no_handler rule to a separate PR?

ssbarnea:
@oraNod If you can split it it would be awesome as it would make much easier to review, update and merge them one by one.

Some of them would be harder to review and we want to merge the easy ones ASAP.

oraNod:
@ssbarnea Sounds good. I'll do that today, just wasn't sure as you mentioned in chat you might just merge this one. Thanks!

ssbarnea (approved):

PR #2401, opened 2022-09-09
Docs: no jinja nesting rule

oraNod:
Add MD documentation for the no-jinja-nesting rule.

### DIFF
--- a/src/ansiblelint/rules/no_jinja_nesting.md
+++ b/src/ansiblelint/rules/no_jinja_nesting.md
@@ -1,0 +1,35 @@
+# no-jinja-nesting
+
+This rule checks for nested Jinja expressions.
+Avoid nesting Jinja expressions with curly brackets `{{ }}` inside other Jinja expressions.
+
+An Ansible rule is that "moustaches don't stack".
+Nested expressions are an Ansible anti-pattern and do not produce expected results.
+
+## Problematic Code
+
+```yaml
+---
+{{ list_one + {{ list_two | max }} }}
+```
+
+```yaml
+---
+{{ somevar_{{other_var}} }}
+```
+
+## Correct Code
+
+```yaml
+---
+{ { list_one + max(list_two) } }
+```
+
+```yaml
+---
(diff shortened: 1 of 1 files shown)


src/ansiblelint/rules/no_log_password.md
```diff
@@ -0,0 +1,38 @@
...
+  hosts: localhost
+  tasks:
+    - name: Log user passwords
+      ansible.builtin.user:
+        name: johnd
+        comment: John Doe
+        uid: 1040
+        group: admin
+        password: password
+      no_log: False # <- Sets the no_log attribute to false.
```
oraNod, line 20: @ssbarnea Again I hope you don't read this while on PTO. But I wasn't able to trigger this rule, even using the tests. Am I missing something here? Thanks.

src/ansiblelint/rules/no_jinja_nesting.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+```yaml
+---
+# Uses the hostvars variable.
+{{ hostvars[inventory_hostname]['somevar_' ~ other_var] }}
+
+# Uses the ansible.builtin.vars lookup plugin.
+{{ lookup('vars', 'somevar_' ~ other_var) }}
+```
```
oraNod, line 35: @ssbarnea One other question I have in this PR is about contextualizing these examples into playbooks. I found this one a little difficult and worry any playbook example might be nonsense. I'm still fairly new to Ansible and Jinja. Do we have "advanced" examples anywhere?

ssbarnea: While looking at this rule I realised that I will probably remove this rule by encapsulating the feature it provides into `jinja` one.

I will keep this PR open until I have the proof that I can bring that functionality into the jinja rule, especially as that would make it much more trustable (avoiding the regex, which brings false positivies)

ssbarnea:
Closing as #2436 removed this rule.

PR #2406, opened 2022-09-12
Docs: no handler rule

oraNod:
Add MD docs for the no_handler rule.


ssbarnea:
@oraNod On that one you did not get the explanation right. Please read my fix and let me know if it was clear or if it can be improved.

ssbarnea (approved):

PR #2412, opened 2022-09-13
fqcn-builtins: add documentation

oraNod:
Add MD docs for the fqcn-builtins rule.

### DIFF
--- a/src/ansiblelint/rules/fqcn_builtins.md
+++ b/src/ansiblelint/rules/fqcn_builtins.md
@@ -1,0 +1,25 @@
+# fqcn-builtins
+
+This rule checks that playbooks use the fully qualified collection name (FQCN) for modules in the `ansible.builtin` collection.
+
+## Problematic Code
+
+```yaml
+---
+- name: Example playbook
+  hosts: all
+  tasks:
+    - name: Shell
+      shell: echo # <- This does not use the FQCN for the shell module.
+```
+
+## Correct Code
+
+```yaml
+---
+- name: Example playbook
+  hosts: all
+  tasks:
+    - name: Shell
+      ansible.builtin.shell: echo # <- This uses the FQCN for the shell module.
+```


src/ansiblelint/rules/fqcn_builtins.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Correct Code
+
+```yaml
+---
+- name: Example playbook
+  hosts: all
+  tasks:
+    - name: Shell
+      ansible.builtin.shell: echo # <- This uses the FQCN for the shell module.
```
bcoca, line 24: this changes the meaning, to preserve the same meaning it would have to be `ansible.legacy.shell`

oraNod: Thanks for spotting that @bcoca 

I'm going to update the examples here to use one of the `ansible.builtin` modules. I could be wrong in my thinking but it seems like we wouldn't want users to go with the `ansible.legacy.shell` module so putting it in the "Correct Code" example might not be the right thing to do.

@ssbarnea I actually got this from the test at: https://github.com/ansible/ansible-lint/blob/e79e65c4c9619791944432329c3689969553982e/src/ansiblelint/rules/fqcn_builtins.py#L109

Do you think this might need an update based on Brian's comment as well? Thanks.

bcoca: the difference between legacy and builtin is subtle and most users won't care, but the ones that do ... will find themselves with an unexpected change in behavior. legacy allows local overrides while builtin does not, most people don't have local overrides, but those that do, depend on them.

ssbarnea: @bcoca Any documentation links pointing to difference between the two that we can include in this markdown file?

In the end we just give an example of "modernizing" the code, that does not mean that the right answer is always appending "ansible.builtins." prefix. Still, adding an extra note/link regarding the nuances could only be seen as beneficial, convincing user to make a more informed decision.

bcoca: :-( no, cannot find it ever making the docs

ssbarnea (approved):

PR #2416, opened 2022-09-13
Document metadata rules

oraNod:
Add MD docs for the following rules:

- meta-incorrect
- meta-no-info
- meta-no-tags
- meta-video-links

### DIFF
--- a/src/ansiblelint/rules/meta_incorrect.md
+++ b/src/ansiblelint/rules/meta_incorrect.md
@@ -1,0 +1,35 @@
+# meta-incorrect
+
+This rule checks role metadata for fields with undefined or default values.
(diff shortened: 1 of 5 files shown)


oraNod:
@ssbarnea I globbed the metadata rule docs into a single PR. Let me know if you want me to break this into multiple PRs though. Cheers.

ssbarnea (approved):

ssbarnea (approved):
Thanks. The sad part is that we are likely to remove most of these rule as they are covered or can be covered by schema meta schema. Still, even when we remove the rules, we will just move the documentation inside schema.md file, so the effort is not lost.

PS. rule removal because test would be redundant, not because they are not good.

PR #2419, opened 2022-09-13
schema: meta/main.yml now requires a version field

ssbarnea:
As `meta/main.yml` files can have different format if they are from old standalone roles (v1) and those from inside collections (v2), our schema will require user to explicitly specify which version they want to use.

This counts as a breaking change because will force users to add either `version: 1` or `version: 2` to all their meta files. Still this would allow us to use the proper sub-schema for each of them.

### DIFF
--- a/examples/roles/dependency_in_meta/meta/main.yml
+++ b/examples/roles/dependency_in_meta/meta/main.yml
@@ -1,5 +1,6 @@
 ---
 # meta file, determined by ending in meta/main.yml
+version: 1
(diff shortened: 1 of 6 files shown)


cidrblock:
Ty for this one. Although adding a version field will be a little extra step for everyone, this will certainly help improve the accuracy of the file structure and prevent surprises later.

cidrblock (approved):

ssbarnea:
Added skip-changelog as this change was reverted before the release.

PR #2424, opened 2022-09-15
Change ansible-lint logo to bracket-dot from codicons

ssbarnea:
In order to make linter logo easier to recognize and also possible to display in vscode status-bar, we adopt the bracket-dot icon.

This should also sort the lack of favicon on multiple browsers by making use of the most compatible format for these (.ico).


cidrblock:
+1

PR #2433, opened 2022-09-17
Respect environment variable ANSIBLE_VERBOSITY

davedittrich:
This simple change enables `ansible-lint` to respect the environment variable `ANSIBLE_VERBOSITY` like other Ansible CLIs. It simplifies debugging `molecule` as you don't have to both export `ANSIBLE_VERBOSITY` for `ansible-playbook` and also add `-vv`s to `ansible-lint` command lines. You can control both with a simple `ANSIBLE_VERBOSITY=2 molecule test` command line.


ssbarnea:
@cidrblock @ganeshrn WDYT about this behavior change? My personal take is that I kinda dislike linking it to `ANSIBLE_VERBOSITY` as we lose some control over enabling just one of them.

There are also other implications, as we call ansible-playbook --syntax-check and currently setting ANSIBLE_VERBOSITY=1 does increase the verbosity of these calls without increasing verbosity of the linter.

I am softly against, probably @cidrblock will want to say something about navigator, as the same kind of question could be asked about navigator.

davedittrich:
Thank you for considering my PR.

My TL;DR response can be summarized as follows:

- I don’t believe you lose any control of setting verbosity for only one program in the Ansible suite, since all this does is take advantage of the `argparse` keyword argument `default` to make the default match that value defined in the environment rather than it being hard-coded to `0` (and thus requiring the user edit a file to change the command line argument). Whatever the user specifies on the command line takes precedence.

- I would respectfully say that from a UX perspective, making things easier for users combining multiple open source tools to accomplish high-level tasks should take precedence over making things easier for developers making internal subprocess calls (and thus the place to control the `ansible-playbook --syntax-check` call would be when `exec`ing the subprocess.) The programmer can formulate whatever command line arguments or make whatever per-call environment variable changes they need to control internal behavior, while leaving the higher-level UX matching user expectations.

- Many old-timer Unix programmers like me are used to taking advantage of the environment inheritance feature of the Unix `fork`/`exec` model to consistently control the behavior of all programs down the process tree. I made the assumption that all Ansible programs would set their verbosity accordingly and was confused when it seemed this was not working in this specific situation.

That said, here is a longer exposition of similar end-user requests, and other Ansible-related programs using the environment variable (or close to it), that I found while debugging why my expectations of program behavior motivating this PR were not manifesting.

----

Quoting from [this blog post](https://sbarnea.com/molecule/molecule-101/), “the [favored] way of passing custom configuration [is by] defining `ANSIBLE_` prefixed environment variables [that will] be recognized by Ansible.” I was following that guidance and found it did not work as I expected.

I tried setting the documented `log: True` option in `molecule.yml` and found that it had no effect, either  set to `True` or `False`.

Searching for solutions I found that many others were making the same assumption I was and having the same user experience.

- "I was surprised to find out that there is no way to control ansible verbosity level via an environment variable. There are environment variables for even the weirdest things like `ANSIBLE_COLOR_DEBUG` but somehow none for verbosity."

  "It would have been nice if the difference and use case between `ANSIBLE_DEBUG` and `ANSIBLE_VERBOSITY` was documented? Now both of these variables control the ansible log verbosity somehow, but the user has no idea why there are two different configuration variables... if he by chance happens to even notice that there are two different variables…"

  https://github.com/ansible/ansible/issues/22087

- "Apologies if this is working as intended. I've tried to search for it but didn't see anywhere that this was a deprecation change. Specifying the `vvv: true` option on the provider section of a `molecule.yml` has no effect even when `molecule` is ran with or without the `--debug` flag."

  "I think this feature is very useful. Not sure why it has been removed, but I don't think it should be. When you use the `ANSIBLE_VERBOSITY` environment variable, you actually end up getting all molecule phases outputting at that verbosity level rather than just the provisioner phases.”

  https://github.com/ansible-community/molecule/issues/3159

- “I want to control the verbosity of ansible playbooks using an environment variable or a global configuration item. This is because ansible is called from multiple places in multiple ways and I want to change the logging level for all further executions from the same shell session."

  https://stackoverflow.com/questions/42510032/how-to-change-ansible-verbosity-level-without-changing-the-command-line-argument

I spent quite a while searching through source code. I found that `ansible/config/base.yml` defines `ANSIBLE_VERBOSITY` as the environment variable controlling the default verbosity. I didn’t take the time to run down (or interactively debug) the core programs like `ansible-console`, but it looks like all the CLI apps–`ansible`, `ansible-console`, `ansible-doc`, `ansible-galaxy`, `ansible-inventory`, `ansible-pull`, and `ansible-vault`--use the same options processing logic (which would make sense) and thus do the same thing and all respect `ANSIBLE_VERBOSITY`. (Please correct me if this is not the case).

```
$ grep -r verbos cli/*.py
cli/__init__.py:            if not have_cli_options or have_cli_options and context.CLIARGS['verbosity'] > 2:
cli/adhoc.py:        display.verbosity = options.verbosity
cli/config.py:        opt_help.add_verbosity_options(common)
cli/config.py:        display.verbosity = options.verbosity
cli/console.py:       - `verbosity [num]`: set the verbosity level
cli/console.py:        display.verbosity = options.verbosity
cli/console.py:            # FIXME: add traceback in very very verbose mode
cli/console.py:    def do_verbosity(self, arg):
cli/console.py:        """Set verbosity level"""
cli/console.py:            display.display('Usage: verbosity <number>')
cli/console.py:                display.verbosity = int(arg)
cli/console.py:                display.v('verbosity level set to %s' % arg)
cli/console.py:                display.error('The verbosity must be a valid integer: %s' % to_text(e))
cli/doc.py:        display.verbosity = options.verbosity
cli/doc.py:                if display.verbosity >= 3:
cli/doc.py:                    display.verbose(traceback.format_exc())
cli/doc.py:            filename, fragment_loader, verbose=(context.CLIARGS['verbosity'] > 0),
cli/galaxy.py:        opt_help.add_verbosity_options(common)
cli/galaxy.py:        display.verbosity = options.verbosity
cli/inventory.py:                           'ansible_verbosity',
cli/inventory.py:        display.verbosity = options.verbosity
cli/playbook.py:        display.verbosity = options.verbosity
cli/pull.py:        display.verbosity = options.verbosity
cli/pull.py:        if context.CLIARGS['verbosity'] > 0:
cli/pull.py:            base_opts += ' -%s' % ''.join(["v" for x in range(0, context.CLIARGS['verbosity'])])
cli/vault.py:        opt_help.add_verbosity_options(common)
cli/vault.py:        display.verbosity = options.verbosity
```

I don’t think that core logic is intended to be used in other community programs like `ansible-lint`, though it does add confusion that only some programs that start with `ansible-` behave one way and others behave another way (without clear documentation as to which way and how to control the programs). That level of opacity results in a frustrating user experience.

I found a couple other programs that go even farther in terms of granularity of the environment variable control method:

- [Ansible Operator](https://sdk.operatorframework.io/docs/building-operators/ansible/reference/advanced_options/) provides support for fine-grained control of verbosity on a per-kind basis using environment variables of the form `ANSIBLE_VERBOSITY_<kind>_<group>` (e.g., `ANSIBLE_VERBOSITY_MONGODB_DB_EXAMPLE_COM` for the `MongoDB` kind in the `db.example.com` group.)

- [The `ovirt-engine` (as of v4.2.13)](https://gerrit.ovirt.org/c/ovirt-engine/+/86429/)  uses a similar method to pass along verbosity settings to `ansible-playbook` using the environment variable `ANSIBLE_PLAYBOOK_VERBOSITY_LEVEL`.

I don’t think that level of complexity and fine-grained control is necessary, but it does take advantage of the environment variable injection method of controlling subprocess behavior down through the process hierarchy without having to edit any files.

Thanks again for taking the time to consider this. I may look into the

cidrblock (commented):
I recall a similar conversation early on about ansible-navigator, using ANSIBLE_VERBOSITY as an indicator for navigator's log level. 

We decided against it for a couple reasons, which may or may not apply here.

1) navigator uses python log levels so it uses the string variant of the level
2) Since navigator has an interface, "verbosity" didn't translate well, so navigator calls it LOG_LEVEL
3) The need to debug ansible-playbook is more common (should be?) than the need to debug navigator, so they are separately configured
4) the navigator log level can be configured in it's settings file, with an environment variable and at the command line. So if desired, it can just be left in the setting file and isn't required at the command line for multiple invocations (this is true for every navigator settings option)
5) Care was taken to keep ansible's configuration separate from navigator's settings in several places to avoid confusion, so it could completely recreate the ansible-playbook experience without a detailed understanding of all navigator's knobs.

All this said, navigator is a completely different use case than lint.

I'm curious if introducing lint specific environment variables for some of it's settings over time might be a middle ground here (e.g ANSIBLE_LINT_VERBOSITY)

I'm leaning toward keeping them separate but wondering if there is a need to make configuring lint easier?

ssbarnea:
As main contributor of the linter, I often need to adjust verbosity in order to debug different behaviors and I do not remember any case where I needed to enable both, ansible own one and linter one. Connecting one to another would prevent people from increasing just one of them.

Due to this, I will close this PR. Still, please note that that does mean that I am against exposing the ability to control linter options via environment variables, just like ansible. We already have one example: `ANSIBLE_LINT_CUSTOM_RULESDIR`.

davedittrich:
I'm a bit at a loss for a proposal for how to use environment variables to control `molecule` subprocess calling verbosity. Is there a good place to have such discussions besides in a PR thread?

The issue I raised that I'm not seeing addressed completely in @ssbarnea 's response is that it is difficult for a user to guess how to increase verbosity in the context of `molecule` because setting `ANSIBLE_VERBOSITY` before running `molecule` *will* increase verbosity for `ansible`, `ansible-galaxy`, `ansible-vault`, `ansible-playbook`, but not for `ansible-lint` or `ansible-compat` (both of the latter being invoked from `molecule` directly, or via `pytest` from `molecule`. That is really  confusing and makes debugging quite difficult.

I guess one could say that the name `ANSIBLE_VERBOSITY` has grown over time to be overly-general and perhaps should have been renamed `ANSIBLE_CORE_VERBOSITY` to make it clear it only applies to the first set of programs listed above and then have more specific `ANSIBLE_LINT_VERBOSITY`, `ANSIBLE_COMPAT_VERBOSITY`, etc. for programs that are not part of `ansible-core`.

The other issue is the fact that `molecule` is essentially a runtime framework that controls many different programs behind the scenes, with the `molecule.yml` file as the driver. I've been spending hours trying to wade through source code and single-step through VSCode to wrap my head around the complexities of the implicit process trees to learn how to control them and debug failures. I see the following call hierarchy:

```
molecule  -run_command->  ansible-playbook
molecule  -run_command->  ansible-galaxy
molecule  -run_command->  ansible
molecule  -run_command->  pytest
molecule  -run_commmand->  linting
molecule  -imports->  ansible-compat
ansible-compat  -runtime.execs->  ansible
ansible-compat  -runtime-execs->  whatever is exec()ed
pytest  -runs->  testinfra
pytest  -runs->  ansible
testinfra  -runs->  docker
```

It would seem to me that having a generic verbosity level indicator (e.g., `verbosity`) for every stage as part of the schema (that is then translated to whatever mechanism--`-vvv` flags, an environment variable in the `exec` call--is required when baking the command) would be a better user experience.  Then environment variables for each stage (for times when you need to debug just one stage at a time) and one over-arching variable that could control them all at once (similar to `ANSIBLE_VERBOSITY` for the subset that respect it) that could be injected via the process environment would be easier to understand and use:

```
MOLECULE_VERBOSITY
MOLECULE_CREATE_VERBOSITY
MOLECULE_LINT_VERBOSITY
MOLECULE_PREPARE_VERBOSITY
MOLECULE_CONVERGE_VERBOSITY
MOLECULE_IDEMPOTENCE_VERBOSITY
MOLECULE_VERIFY_VERBOSITY
```

Thanks for considering these ideas.

Issue #2426, opened 2022-09-16
ansible-lint fails when linting a collection installed within ansible-compat cache


stenh0use:
##### Summary

When running `molecule==4.0.1` within a collection that has a `galaxy.yml` file, it will install the collection in to `~/.cache/ansible-compat`.

Before being installed into that directory everything works well.

After the collection is installed into `~/.cache/ansible-compat` directory if you run ansible-lint it will error out after hitting a permission issue in `_mockings.py` library.

Issue was introduced from `ansible-lint 6.0.0` onwards.

I know where the code that causes the issue is, but havent had the time to get to understand the library yet. Happy to try and make the fix if you can point me in the right direction.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console
ansible --version
ansible [core 2.13.3]
  config file = /Users/jstenhouse/.ansible.cfg
  configured module search path = ['/Users/jstenhouse/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/jstenhouse/alpha/infra/ansible/collections/ansible_collections/nvidia/cuda/.venv/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/jstenhouse/alpha/infra/ansible/collections
  executable location = /Users/jstenhouse/alpha/infra/ansible/collections/ansible_collections/nvidia/cuda/.venv/bin/ansible
  python version = 3.10.6 (main, Aug 11 2022, 13:47:18) [Clang 12.0.0 (clang-1200.0.32.29)]
  jinja version = 3.1.2
  libyaml = True

ansible-lint --version
ansible-lint 6.5.2 using ansible 2.13.3
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

MacOS 10.15.7

##### STEPS TO REPRODUCE

`cd` to a collection with a galaxy.yml file and then run the following

```console (paste below)
ansible-galaxy collection install -v --force -p ~/.cache/ansible-compat/e636c9/collections .
ansible-lint playbooks/image.yml
```

Where playbook has something similar to this
```yaml
---
- hosts: all
  become: true
  gather_facts: false
  roles:
    - namespace.collection.role_name
```

##### Desired Behavior

yamllint to execute successfully

##### Actual Behavior

when running with `ansible-lint 5.4.0` we see success
```shell
$ ansible-lint --version
ansible-lint 5.4.0 using ansible 2.13.3

[...]

Issue #2265, opened 2022-08-07
Local collection install conflict between ansible-lint and molecule


aldehir:
##### Summary

It appears there is a conflict with how molecule and ansible-lint install a project if it is a collection.

Molecule uses `ansible-galaxy` to install the collection in the `ansible-compat` cache directory, while ansible-lint attempts to create a symlink to the collection in the same cache directory. If you run molecule first, ansible-lint will fail and conversely if you run ansible-lint first, molecule will fail.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console
$ ansible --version
ansible [core 2.13.2]
  config file = None
  configured module search path = ['/home/alde/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/alde/example/venv/lib/python3.10/site-packages/ansible
  ansible collection location = /home/alde/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/alde/example/venv/bin/ansible
  python version = 3.10.0rc2 (default, May  1 2022, 08:45:48) [GCC 10.3.1 20210422 (Red Hat 10.3.1-1)]
  jinja version = 3.1.2
  libyaml = True

$ ansible-lint --version
ansible-lint 6.4.1.dev5 using ansible 2.13.2

$ molecule --version
molecule 4.0.1 using python 3.10
    ansible:2.13.2
    delegated:4.0.1 from molecule
    podman:2.0.2 from molecule_podman requiring collections: containers.podman>=1.7.0 ansible.posix>=1.3.0
```

- ansible installation method: pip
- ansible-lint installation method: source, main branch

##### OS / ENVIRONMENT

* Fedora 33
* Python 3.10

##### STEPS TO REPRODUCE

```console (paste below)
$ ansible-galaxy collection init dummy.example
$ cd dummy/example
$ molecule init scenario
$ molecule create
$ ansible-lint
Traceback (most recent call last):
  File "/home/alde/example/venv/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
  File "/home/alde/example/venv/lib/python3.10/site-packages/ansiblelint/__main__.py", line 266, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
  File "/home/alde/example/venv/lib/python3.10/site-packages/ansiblelint/__main__.py", line 166, in main
    app = get_app(offline=options.offline)
[...]

PR #2434, opened 2022-09-17
Clean up link handling logic

davedittrich:
This change cleans up the logic surrounding handling linked directories.

I was getting confusing exceptions when running `molecule` (which in turn runs `ansible-lint`) because `ansible-compat` created a directory, not a link. This more clearly expresses the logical flow of this section of code and hopefully makes it easier for others to debug similar indirect exceptions raised during `molecule` runs.


ssbarnea (changes requested):
I did not check why the jobs failed but I can see a clear regression, new code is not able to repair an existing symlink that is either broken or pointing to a wrong location. We also want to avoid changing things unless needed. Be aware that python can behave unexpected when it comes to performing exist tests on symlinks.

If you want to fix a bug on this area, how about just adding a test that reproduce the bug? I could do the fix if I have the test.

heindsight:
@davedittrich the error you describe sounds like the same issue that was reported in #2265. I created [heindsight/ansible-lint-vs-molecule](https://github.com/heindsight/ansible-lint-vs-molecule) as a demonstration/reproducer of that bug.

@ssbarnea I've taken a stab at writing a test to reproduce this here: https://github.com/ansible/ansible-lint/compare/main...heindsight:ansible-lint:cache-link-conflict

davedittrich:
I just saw [this fix](https://github.com/ansible/ansible-lint/pull/2466#issuecomment-1255208471) and had a deja-vu ("I've seen that code somewhere before!"). It resonates with [this comment](https://github.com/ansible/ansible-lint/issues/2426#issuecomment-1251588548) and might apply here as well..

I think the issue is a combination of non-DRY coding (same code, different places) combined with the logic issue I raised, which is use of `try`/`except`/`pass` instead of clear `if`/`else` logic.

What do y'all think?

P.S. I actually was using a much simpler (but similar) mechanism of symbolic links, but I was explicit about adding another path to `ANSIBLE_COLLECTIONS_PATH`. It is [described here](https://github.com/davedittrich/utils#testing). My method was working fine over a year ago, but broke horribly when I updated all the components to start working on it again. I had to significantly change my code to use modern `molecule` and keep my code (which used `testinfra`) but the symbolic link bug related to the code in this PR had me spinning my wheels for days.

ssbarnea:
This was already fixed on main branch.

PR #2444, opened 2022-09-19
Add name[template] violation for tasks not using jinja at the end of the string

vineethreddykaturu:


ssbarnea:
Closed in favour of #2483

PR #2450, opened 2022-09-19
Docs: yamllint rule

oraNod:
This PR updates documentation for the yaml rule.


ssbarnea:
You need to keep the `yaml.md` filename. We had to avoid using `yaml.py` on this rule, but the documentation must remain `yaml.md`. Is the only exception.

oraNod:
> You need to keep the `yaml.md` filename. We had to avoid using `yaml.py` on this rule, but the documentation must remain `yaml.md`. Is the only exception.

Thanks, Sorin. I wasn't sure about this one so took a punt and renamed the md file. I've changed it back.

ssbarnea (approved):

PR #2470, opened 2022-09-20
Refactor no-loop-var-prefix rule

oraNod:
- rename rule to loop-var-prefix and keep alias for old name
- make rule produce more detailed messages
- add documentation to the rule


oraNod:
@ssbarnea Hi. Not sure about this one at all and am kind of reaching. I couldn't trigger this rule and am unsure about the description. I think I'm also missing some nuances about roles in the docs. If you could please take a look and let me know where I'm going wrong I'd appreciate it. Cheers.

ssbarnea (approved):

Issue #2357, opened 2022-08-29
ansible-lint print WARNING (PATH altered to include)


hardcore:
##### Summary

ansible-lint --version print:

WARNING: PATH altered to include /opt/homebrew/Cellar/ansible-lint/6.5.1/libexec/bin
ansible-lint 6.5.1 using ansible 2.13.3

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```
❯ ansible-lint --version
WARNING: PATH altered to include /opt/homebrew/Cellar/ansible-lint/6.5.1/libexec/bin
ansible-lint 6.5.1 using ansible 2.13.3
```

```console (paste below)
ansible --version
ansible [core 2.13.3]
  config file = None
  configured module search path = ['/Users/sash/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/lib/python3.9/site-packages/ansible
  ansible collection location = /Users/sash/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.9.13 (main, Aug 11 2022, 13:40:36) [Clang 13.1.6 (clang-1316.0.21.2.5)]
  jinja version = 3.1.2
  libyaml = True

ansible-lint --version
WARNING: PATH altered to include /opt/homebrew/Cellar/ansible-lint/6.5.1/libexec/bin
ansible-lint 6.5.1 using ansible 2.13.3

brew --version
Homebrew 3.5.10
Homebrew/homebrew-core (git revision c85c8882a61; last commit 2022-08-29)
Homebrew/homebrew-cask (git revision b422fa8bf7; last commit 2022-08-29)
```

- ansible installation method: homebrew
- ansible-lint installation method: homebrew

##### OS / ENVIRONMENT

macOS Monterey Version 12.5.1, MacBook Pro (14", 2021)

##### STEPS TO REPRODUCE

Run ansible-lint within a homebrew installation. 

##### Desired Behavior

This behavior should be fixed in https://github.com/ansible/ansible-lint/pull/2129

##### Actual Behavior

ansible-lint thorws an error.

ssbarnea:
I use homebrew too. The real question is how did you install ansible-lint?

dawidd6:
> ansible installation method: homebrew
ansible-lint installation method: homebrew

Can confirm on Ubuntu with Homebrew-installed ansible and ansible-lint:
```
[...]

PR #2477, opened 2022-09-21
Use site.getuserbase() to detect ansible executable

ssbarnea:
As it is quite common for many users to miss adding python bin scripts
folder to their PATH, even if pip does warn them about the issue, we
now add it dynamically.

Fixes: https://github.com/ansible/ansible-lint/issues/2357


hardcore:
> Installed on macOs 12.6 w/o venv, version reported 0.0.0

did you update ansible-lint with "brew install ansible-lint"?

ssbarnea:
@hardcore No. We are not updating brew recipes ourselves, mainly because brew requires manual intervention in order to make any new release. We release ansible-lint quite often, sometimes with even two releases in a single week. If you convince brew folks to automate their process, it would be cool. That is one one the reasons `brew` is not listed on our [installation](https://ansible-lint.readthedocs.io/installing/) documentation. I love brew and use it a LOT, but not for python tools/packages, as it is always lagging.

I suspect that the brew version bad error is because they installed the package from a shallow clone. To get the proper version number, you need a full clone, or the python wheel (pip).

hardcore:
Great! Thank you for the necessary info about brew and python things :-)

Issue #2110, opened 2022-05-09
Add support to display score on running lint for a given profile


ganeshrn:
##### Summary

Add support to display score (similar to pylint) on running lint for a given profile. The score can be considered as one of the metric for code quality and adherence to rule set for a given profile.

##### Issue Type

- Feature

PR #2495, opened 2022-09-25
Adds summary stats about found violations

ssbarnea:
Fixes: #2110

sample-screenshot

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -40,6 +40,7 @@ max-line-length = 100
 extend-ignore =
   E203,
   E501,
+  C901, # complexity is also measured by pylint: too-many-branches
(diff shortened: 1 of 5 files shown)


briantist (approved):
A warning that I'm not too familiar with this codebase, but the results look nice to me and nothing stands out to me on the implementation side. Thanks!

ssbarnea:
> A warning that I'm not too familiar with this codebase, but the results look nice to me and nothing stands out to me on the implementation side. Thanks!

Thanks! I was interested in opinions regarding the output in particular. I wanted to avoid making it too verbose, but also keep it usable. The rules are ordered by importance (profile), not count. If you address them from the top the scoring can increase.

Issue #2065, opened 2022-04-04
var_naming_pattern not working for set_fact


SpikePy:
##### Summary
var_naming_pattern is buggy for set_fact variables

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

ansible-lint --version
```
ansible-lint 6.0.2 using ansible 2.12.4
```

##### STEPS TO REPRODUCE

.ansible-lint file:
```
var_naming_pattern: "^[a-z][a-z0-9_]*[a-z0-9]__[a-z][a-z0-9_]*[a-z0-9]$"
```

in a role set a variable in defaults/main.yml to:
```
k8s_grafana_users__namespace: monitoring
```

in tasks/main.yml set:
```
- name: foobar
  ansible.builtin.set_fact:
    k8s_grafana_users__namespace: "foo"
```

##### Desired Behavior

no error regarding variable pattern neither in defaults nor in the tasks file

##### Actual Behavior

```
var-naming: Task uses 'set_fact' to define variables that violates variable naming standards
tasks/main.yml:2 Task/Handler: foobar
```

defaults/main.yml is ok but set_fact variable with the same name is not ok!

nishipy:
Hi team, if no one hasn't started to work on this, I'd like to take a look.

PR #2496, opened 2022-09-26
var-naming: also recognize vars with set_fact module

nishipy:
Fixes: #2065


ssbarnea (changes requested):
I am not sure if I fully understand the bug, but if that is fixing something that was not working correctly before, why I am not seeing any change being made to the tests? ... a change that would ensure
that we not see a regression in the future.

Look at the bottom of the file for few tests, you might be able to just
alter them to test the new outcome. I would be able to check manually
the tests without including you change and see that they fail with current implementation.

nishipy:
Got it. I will add tests.

ssbarnea (approved):

Issue #2501, opened 2022-09-27
docs: copy/template rule needs better reasoning


greg-hellings:
##### Summary

template-instead-of-copy rule finds false positive when `content` argument is used that has template values in it.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

Linux/NixOS

##### STEPS TO REPRODUCE

Sample play
```yaml
- name: Write file contents from variable
  ansible.builtin.copy:
    dest: /some/dest/path
    content: "{{ some_var.subfield }}"
```

##### Desired Behavior

The above code should be perfectly acceptable, as all templating will be handled locally within the file.

##### Actual Behavior

Error `template-instead-of-copy` is thrown.

MarkusTeufelberger:
I thought this was exactly the reason this rule was created? It now tries to make you use `ansible.builtin.template` on a template file containing only `{{ some_var.subfield }}` instead.

From looking through the code, a false positive might happen if you try to write two literal `{{` characters as file content, but the case you describe seems like a true positive.

Maybe @GhostLyrics can give a better explanation since he authored this rule.

ssbarnea:
I also think that this is not a false positive but I will keep the bug open because we need to fix the documentation for this rule to explain exactly why this needs to be done, current docs page is not enough. I do remember reading a blog post about the risks but I am unable to find it now.

GhostLyrics:
This works as (I) intended it to. It is even a little bit more lenient than I wrote it. The original version disallowed the content key entirely due to formatting also not always being applied perfectly. Maintainers changed it to only trigger on Jinja in the content key.

MarkusTeufelberger:
I'm not a big fan of that rule personally by the way, since it is useful if a "template" is just a single line of Jinja to have it right there in the task, but maybe that's something that IDEs should solve instead (e.g. by allowing to view a referenced template directly from the tasks file).

The upstream docs state however "For advanced formatting or if content contains a variable, use the [ansible.builtin.template](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/template_module.html#ansible-collections-ansible-builtin-template-module) module." (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html) so I guess the current way this rule is written reflects this recommendation. I don't see how the documentation change in #2512 explains this though, since no variable is actually used.

ssbarnea:
[...]

PR #2512, opened 2022-09-28
Replace template-instead-of-copy with avoid-implicit[copy-content]

ssbarnea:
- Replace to aggressive and indiscriminate rule template-instead-of-copy with new one that only discourages the implicit stringification on copy content.
- New rule will be just a warning

Fixes: #2501


GhostLyrics (commented):
Is the change from double brackets to single brackets intentional?

ssbarnea:
> Is the change from double brackets to single brackets intentional?

That is the expected outcome from the code point of view, but I doubt that many users would discover the difference, even those that know why it happens.

I would like to be able to detect which input is safe or not in order to avoid triggering this violation on safe strings, but at this moment I do not know what would be safe or not.

Issue #1070, opened 2020-10-06
Rule 208 (Wrong/Missing file permission) ignores configured module_defaults


fleaz:
##### Summary

Hey everyone,
I used [module_defaults](https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html) in my playbook to save me some writing by configuring defaults for owner,group and mode of the *copy* and *template* module. This works as expected, but ansible-lint still complains with rule 208 because my copy and template tasks (allegedly) don't have correct file permissions set.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

ansible --version
```
ansible 2.9.12
  config file = /home/fleaz/codemonauts/ffni-monitoring/ansible.cfg
  configured module search path = ['/home/fleaz/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.5 (default, Jul 27 2020, 08:42:51) [GCC 10.1.0]
```

ansible-lint --version
```
ansible-lint 4.3.6.dev5+gcff352f

(also tested with 4.3.5 from pip)
```

- ansible installation method: OS package 
- ansible-lint installation method: pip and git

##### OS / ENVIRONMENT

Arch Linux

##### STEPS TO REPRODUCE

Define owner,group and mode with `module_defaults` for e.g. the copy module and then use the copy module in your playbook.

test.yml:

```console (paste below)
---
- hosts: all
  module_defaults:
    copy:
      owner: root
      group: root
      mode: 0644
  tasks:
    - name: copy a file
      copy:
        src: files/foo
        dest: /tmp/foo

```

##### Desired Behaviour

Ansible-lint should interpret the module_defaults and see that the copy module has implicit values for owner,mode,group via the module_defaults

##### Actual Behaviour

Ansible-lint complains that the copy task has no configured file permissions.

```paste below
# ansible-lint test.yml                       
[208] File permissions unset or incorrect
foo.yml:10
Task/Handler: copy a file
```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

pratheekrebala:
[...]

PR #2519, opened 2022-09-29
Document lack of module_defaults support

ssbarnea:
Fixes: #1070

### DIFF
--- a/src/ansiblelint/rules/risky_file_permissions.md
+++ b/src/ansiblelint/rules/risky_file_permissions.md
@@ -23,9 +23,10 @@ Modules that are checked:
 - [`community.general.archive`](https://docs.ansible.com/ansible/latest/collections/community/general/archive_module.html)
 - [`community.general.ini_file`](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html)
 
-````{warning}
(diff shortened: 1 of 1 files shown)


src/ansiblelint/rules/risky_file_permissions.md
```diff
@@ -23,6 +23,13 @@ Modules that are checked:
 - [`community.general.archive`](https://docs.ansible.com/ansible/latest/collections/community/general/archive_module.html)
 - [`community.general.ini_file`](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html)
 
+```{warning}
+This rule does not take [module_defaults](https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html)
+configuration into account, and there are no plans to implement such feature,
+mainly because when a task is moved from one location to another, its behavior
+might change without noticing.
+```
```
oraNod, line 31:
```suggestion
```{warning}
This rule does not take [module_defaults](https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html) configuration into account.
There are currently no plans to implement this feature because changing task location can
also change task behavior.
```
```

Issue #2117, opened 2022-05-10
no-shorthand: Disallow `=` style argument defintion


cidrblock:
As an opt in rule until profile support is available we should disallow yaml ignorant style argument definition for tasks:

e.g.

```
- debug: msg="END connection={{ ansible_connection }} nxos_linkagg sanity test"
```

Related #2116

IPvSean:
I like this idea, you can always turn it off.... but I am a fan of being more prescriptive / opinionated on this, and this confuses people new to Ansible

MarkusTeufelberger:
Duplicate of https://github.com/ansible/ansible-lint/discussions/1609

PR #2527, opened 2022-09-30
Add no-shorthand rule as experimental

ssbarnea:
Fixes: #2117


hswong3i:
@ssbarnea refer to #2537, shall we exclude ```raw``` from this short hand rule?

Issue #2502, opened 2022-09-27
role-name doesn't recognize roles in subpaths


MarkusTeufelberger:
##### Summary

The role-name rule (and the schema rule too which also validates role names these days!) don't trigger on a role that's in a subfolder.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible [core 2.13.4]
  config file = /home/[snip]/ansible.cfg
  configured module search path = ['/home/[snip]/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/[snip]/collections
  executable location = /home/[snip]/.local/bin/ansible
  python version = 3.10.6 (main, Aug 10 2022, 11:40:04) [GCC 11.3.0]
  jinja version = 3.0.3
  libyaml = True

ansible-lint --version
ansible-lint --version
Command 'ansible-lint' not found, but can be installed with:
sudo apt install ansible-lint
```

(`ansible-lint` 6.7.0 is run via pre-commit, not installed on the host...)

- ansible installation method: OS package
- ansible-lint installation method: pip (in pre-commit venv)

##### OS / ENVIRONMENT

Xubuntu 22.04

##### STEPS TO REPRODUCE

Consider the following playbook:

```console (paste below)
---
- name: Testcase
  hosts: localhost
  roles:
    - valid_prefix/this-should-trigger  # <-- no error here!
    - invalid-prefix/this-should-trigger-too  # <-- only complains about "invalid-prefix"
```

##### Desired Behavior

At least `role-name` and probably also the `schema` rule should trigger on this.

##### Actual Behavior

Please give some details of what is actually happening.
Include a [minimum complete verifiable example] with:

- playbook
- output of running ansible-lint
- if you're getting a stack trace, output of
  `ansible-playbook --syntax-check playbook`

```paste below

```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

PR #2532, opened 2022-10-01
role-name[path]: Now detects use of paths when including roles

ssbarnea:
Fixes: #2502


MarkusTeufelberger:
Hehe, that's one way of fixing this! ;-)

Issue #2537, opened 2022-10-02
no-shorthand + ansible.builtin.raw: The raw module takes a free form command to run.


hswong3i:
##### Summary

https://github.com/ansible/ansible-lint/pull/2527 introduce ```no-shorthand``` rule as experimental, but https://docs.ansible.com/ansible/latest/collections/ansible/builtin/raw_module.html only support free form arguments.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible [core 2.13.4]
  config file = None
  configured module search path = ['/home/hswong3i/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/hswong3i/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.6 (main, Aug 10 2022, 11:40:04) [GCC 11.3.0]
  jinja version = 3.1.2
  libyaml = True

ansible-lint --version
ansible-lint 6.8.0b1 using ansible 2.13.4
```

- ansible installation method: OS package (https://build.opensuse.org/package/show/home:alvistack/ansible-community-ansible-build-data-6.4.0)
- ansible-lint installation method: OS package (https://build.opensuse.org/package/show/home:alvistack/ansible-community-ansible-lint-6.8.0-beta1)

##### OS / ENVIRONMENT

```
$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

##### STEPS TO REPRODUCE

Consider follow task with raw module for bootstrapping python3 (https://github.com/alvistack/vagrant-ubuntu/blob/597240a8b75db0d286f58861e398f104d431143b/playbooks/prepare.yml#L35-L43):
```
    - name: debian | apt-get install python3
[...]

PR #2541, opened 2022-10-03
no-shorthand: Avoid false positive with raw

ssbarnea:
- Add special case for `raw`, where we would look for `executable=` string
- Add warning if we see `raw` being called with anything else than a string

Fixes: #2537


hswong3i:
Do we need examples and test cases for this, too?

ssbarnea:
> Do we need examples and test cases for this, too?

 I just pushed a new revision, with tests and updated examples. Good idea!

PR #390, opened 2018-10-31
Fix running with role path containing single or multiple dirs

awcrosby:
Fixes #210 - this PR only adds the trailing slash when multiple dirs are in the role path, and allows these 5 role path scenarios to function:

`code$ ansible-lint -p my_role/`  3.5.1 no output, 3.4.23 no output
`code$ ansible-lint -p my_role`   3.5.1 no output
`~$ ansible-lint -p code/my_role/`
`~$ ansible-lint -p code/my_role`    3.4.23 no output
`my_role$ ansible-lint -p .`

Test cases were added for each scenario.

PR #2557, opened 2022-10-05
Remove enrich as a dependency

ssbarnea:


src/ansiblelint/__main__.py
```diff
@@ -350,5 +349,50 @@ def path_inject() -> None:
...
+    if term == "dumb":
+        return False
+
+    # Use tty detection logic as last resort because there are numerous
+    # factors that can make isatty return a misleading value, including:
+    # - stdin.isatty() is the only one returning true, even on a real terminal
+    # - stderr returning false if user user uses a error stream coloring solution
+    return stream.isatty()
+
+
```
davedittrich, line 396: The comment on this PR and commit message suggest removal of `enrich` as a dependency is the only change, so the added lines above caught my eye. I recognized those lines of code as being the same as that involved in a `stdout` processing bug in [PR #3699](https://github.com/ansible-community/molecule/pull/3669).

I also noted that there were four places that the `enrich` package was referenced (two with just the package name and no version specification, but from different sources, one with a minimum version `>=1.2.6` specified, and the other with explicit version `==1.2.7`.) That means *four* different versions could potentially be involved in different aspects of testing or runtime installation. Removing `enrich` reduces the potential it would have this problem, but is that the only package with that potential? I'm not sure if a more DRY way to specify packages in all of those places is possible, but you may want to look into it for more consistency.

What might be more problematic here (but also related to DRYness) is that the lines added here to `ansible-lint` are [the same lines of code](https://github.com/ansible-community/molecule/blob/d0cb7b7193652084909892bca6ce6e3129abab38/src/molecule/console.py#L29-L71) that exist in `molecule`. I wanted to make sure that was known.

The potential impact here is that should changes be made to the now-common lines of code in `molecule` as part of merging [PR #3699](https://github.com/ansible-community/molecule/pull/3669), those changes would need to be replicated here in `ansible-lint` as well, else the runtime behavior be different between one case and the other.  Because `molecule` runs `ansible-lint`, this could result in a bug that would be very difficult to find and fix as one could assume it was resolved by changes to `molecule` and not even realize the same code existed in `ansible-lint` running in a subprocess. Since both contain the comment "Based on Ansible implementation" this might indicate the same lines of code also exist in a third project, increasing the random bug problem.

Might it be more robust (more DRY) to have this code exist in someplace like `ansible-core` and have both `molecule` and `ansible-lint` import the functions, as `ansible-core` is already a dependency of both?

Lastly, I notice a couple issues with the above code related to the way color-capability is being determined. By using the `TERM` environment variable to infer based only on 'xterm' being found in the string, this leaves out many other terminal types that similarly are color capable. For example, if [`byobu`](https://www.byobu.org) is used for terminal multiplexing, or either [`screen`](https://www.gnu.org/software/screen/) or [`tmux`](https://github.com/tmux/tmux) (which `byobu` front-ends), the `TERM` environment variable could be something like `screen-256color`, or `tmux-256color`, both of which are as much color-capable as `xterm`, `xterm-256color`, etc. Even if the test was to look for `color` in the `TERM` setting rather than just `xterm`, I suspect that won't properly identify all terminal types supporting color. The comment in lines #390-#393 reflects the difficulty in detection (though it says "stdin.isatty() is the only one returning true", yet the test being called is `stream.isatty()` and when `stream` is not passed to the function that would default to testing `stdout` instead of `stdin`: either the comment is wrong, or the test result may be the opposite of what is intended).

Users expect to have control of terminal behavior by explicitly setting the `TERM` environment variable appropriate to the terminal they are using. If a child process of a command line program the user runs in their terminal changes that behavior incorrectly due to an mistaken inference (as opposed to an explicit setting), that can lead to a very frustrating user experience and be extremely difficult to debug. Wouldn't a more opaque method of getting colored output via configuration settings or command line options that is well-documented in both `molecule` and `ansible-lint` (for those who use `ansible-lint` apart from `molecule`) be more consistent and produce a better user experience?

ssbarnea: PS. Sorry for not going deeper into this, now is probably the busiest time of the year due to AnsibleFest 2022, but I will more than happy to work on this right after it (Oct 22)

PR #2572, opened 2022-10-09
Add internal_names rule which flags usage of internal names in collections that must not be used

felixfontein:
For example, the FQCNs `community.general.system.sudoers` and `community.network.network.edgeos.edgeos_facts` should never be used for actions, since they are internal to the collection and can even change in bugfix releases.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -55,6 +55,7 @@ docstrings
 doctest
 doctrees
 docutils
+edgeos
 ematcher
 ematchtestfile
 execnet
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -166,7 +166,7 @@ jobs:
       WSLENV: FORCE_COLOR:PYTEST_REQPASS:TOXENV:TOX_PARALLEL_NO_SPINNER
       # Number of expected test passes, safety measure for accidental skip of
       # tests. Update value if you add/remove tests.
-      PYTEST_REQPASS: 705
+      PYTEST_REQPASS: 707
 
     steps:
       - name: Activate WSL1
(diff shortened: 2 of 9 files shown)


felixfontein:
The cisco.nxos `eco` failures are unrelated, they have been failing for a couple of days now (at least since 6.8.0 has been released).

ssbarnea:
That is really interesting because we had a chat last week with others related the same question: which form should be considered the normalized one. In our context it was more about which auto-complete value should be advertised by our language server, as we clearly wanted to avoid showing duplicates.

1. value exposed by collection metadata
2. location on disk (can be a symlink)
3. resolved/effective location from disk (not a disk location), which is the deeper.
4. targeted location from ansible internal redirects

My impression was that "3" was the winning option and that is exactly the opposite of this rule.

I need to ping @bcoca @sivel @cidrblock @ganeshrn for input on this as we want to be sure we give only one recommendation everywhere, at least one recommendation for specific environment obviously, as collection version and ansible own version might impact that.

sivel:
The fact that a few collections have decided that their use of subdirectories is not a feature, and are not subject to backwards compat sounds like the exception to the rule, and not a general recommendation.  I'd actually state that the decision by these collections to not consider the real FQCN a feature, and not subject to backwards compat is the real bug here.  If a collection were to move a plugin from a subdir, it should be adding redirections with deprecations.

There are use cases where this could be preferred even, and seeing as though this is an actual feature of ansible-core, we should not discourage its use.

bcoca:
To clarify 'internal names' is not a feature in ansible-core, we allow 'aliases', for many reasons (rename/moves, deprecation, etc) and this is a generic feature for all collections.  The 'canonical' name for a plugin should be the name that it resolves to when loading, while the rest of names are either intermediate or parallel loading paths.

PR #2593, opened 2022-10-11
Added rule warning about use of run_once with strategy: free

ajinkyau:
Fixes [1026](https://github.com/ansible/ansible-lint/issues/1026)


ssbarnea:
@ajinkyau You forgot to update the rule counter and when you created the PR you forgot to give permissions for maintainers to fix your PR, so I cannot fix it for you.

ssbarnea:
In fact the mistake was not to create a branch, as we cannot push to default branch.

PR #2594, opened 2022-10-11
Docs: Add MD for var-naming

oraNod:
This PR adds docu for the var-naming rule.

### DIFF
--- a/src/ansiblelint/rules/var_naming.md
+++ b/src/ansiblelint/rules/var_naming.md
@@ -1,0 +1,34 @@
+# var-naming
+
+This rule checks variable names to ensure they conform with requirements.
+
+Variable names must contain only lowercase alphanumeric characters and the underscore `_` character.
+Variable names must also start with either an alphabetic or underscore `_` character.
+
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

src/ansiblelint/rules/var_naming.md
```diff
@@ -0,0 +1,34 @@
...
+```yaml
+---
+- name: Example playbook
+  hosts: localhost
+  vars:
+    CamelCase: true # <- Contains a mix of lowercase and uppercase characters.
+    this_is_valid: # <- This is a dict, not a variable
+      ALL_CAPS: bar # <- Contains only uppercase characters.
+      v@r!able: baz # <- Contains special characters.
+```
```
ssbarnea, line 21: That is not correct because variables are only the keys at the first level under vars. They can have any kind of value and if they are dictionaries the keys of these dictionaries are not forced to follow the same rules. Basically only the CameCase example is correct.

ssbarnea (changes requested):
see comment

src/ansiblelint/rules/var_naming.md
oraNod: How about now? I just remove the dictionary to avoid confusion.

ssbarnea (approved):

PR #2603, opened 2022-10-14
Allow toggling strict in the config file

hemberger:
This commit adds `strict` to the config file schema, equivalent to passing `--strict` on the command line.

Since the config file is the only way to customize behavior in the ansible-lint action, it is useful to have parity between the command line and the config file.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -10,6 +10,7 @@ exclude_paths:
   - test/fixtures/formatting-prettier/
 # parseable: true
 # quiet: true
+# strict: true
 # verbosity: 1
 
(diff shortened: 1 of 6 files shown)


ssbarnea (approved):

ssbarnea (approved):
May I say that is a real pleasure to see a new contribution that gets everything right from start? Most people do not run tests locally, so they miss lots of details, like updating PYTEST_REQPASS number.

I hope to see more contributions from you and I will try to fast track this, so you can make use of it for the action.

hemberger:
That's very kind of you to say, I really appreciate the encouraging words!

Thank you for all the hard work you put in to make this invaluable tool available for everyone.

PR #2615, opened 2022-10-24
Add warning message about outdated linter version

ssbarnea:
Checks every 24h if current version is latest and prints warning message if so. If HTTP requests fails, no error is displayed.

### DIFF
--- a/src/ansiblelint/app.py
+++ b/src/ansiblelint/app.py
@@ -13,7 +13,7 @@ from rich.table import Table
 from ansiblelint import formatters
 from ansiblelint._mockings import _perform_mockings
 from ansiblelint.color import console, console_stderr, render_yaml
-from ansiblelint.config import PROFILES
+from ansiblelint.config import PROFILES, get_version_warning
 from ansiblelint.config import options as default_options
 from ansiblelint.constants import RULE_DOC_URL, SUCCESS_RC, VIOLATIONS_FOUND_RC
 from ansiblelint.errors import MatchError
@@ -204,8 +204,8 @@ warn_list:  # or 'skip_list' to silence them completely
 
         return SUCCESS_RC if mark_as_success else VIOLATIONS_FOUND_RC
 
-    @staticmethod
     def report_summary(  # pylint: disable=too-many-branches,too-many-locals
+        self,
         summary: SummarizedResults,
         changed_files_count: int,
         files_count: int,
@@ -289,6 +289,11 @@ warn_list:  # or 'skip_list' to silence them completely
         if summary.fixed:
             msg += f", and fixed {summary.fixed} issue(s)"
         msg += f" on {files_count} files."
+
+        if not self.options.offline:
+            version_warning = get_version_warning()
+            if version_warning:
+                msg += f"\n{version_warning}"
 
         console_stderr.print(msg)
 
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -1,14 +1,31 @@
 """Store configuration options as a singleton."""
 from __future__ import annotations
 
+import json
+import logging
 import os
 import re
+import sys
+import time
+import urllib.request
+import warnings
 from argparse import Namespace
 from functools import lru_cache
 from pathlib import Path
 from typing import Any
+from urllib.error import HTTPError, URLError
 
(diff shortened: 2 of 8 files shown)


src/ansiblelint/app.py
```diff
@@ -290,6 +290,8 @@ def report_summary(  # pylint: disable=too-many-branches,too-many-locals
             msg += f", and fixed {summary.fixed} issue(s)"
         msg += f" on {files_count} files."
 
+        msg += get_version_warning()
```
apatard, line 293: Do you plan to make this optional ? Having something pinging github every day doesn't seem a good idea [I'm thinking of cases like air-gap users] but it's my personal opinion.

src/ansiblelint/config.py
```diff
@@ -171,3 +182,75 @@ def parse_ansible_version(stdout: str) -> tuple[str, str | None]:
...
+            return ""
+
+    html_url = data["html_url"]
+    tag_name = data["tag_name"]  # name
+
+    if current_version < new_version:
+        msg = f"""\n\n[warning]A new release of ansible-lint is available: [red]{current_version}[/] → [green][link={html_url}]{tag_name}[/][/][/]
+Do not report bugs without trying the latest version."""
+
+    pip = guess_install_method()
```
apatard, line 252: Is this really needed ? Can't we just say that upgrading is a good idea ?

src/ansiblelint/config.py
```diff
@@ -171,3 +182,75 @@ def parse_ansible_version(stdout: str) -> tuple[str, str | None]:
...
+    return pip
+
+
+def get_version_warning() -> str:
+    """Display warning if current version is outdated."""
+    msg = ""
+    new_version = Version("9999.99.9")
+    data = {}
+    current_version = Version(__version__)
+    cache_dir = os.path.expanduser("~/.cache/ansible-lint")
```
apatard, line 218: Can't we use ``XDG_`` stuff ? It's already used in ``yaml_utils.py`` after all

src/ansiblelint/app.py
ssbarnea: In air-gap environments the warning will not appear, by design. But on all other cases it will tell user to update and not raise bugs. We had more than 2/3 reports on old version about stuff that was already fixed... so that drive this feature.

src/ansiblelint/config.py
ssbarnea: Indeed, already fixed.

src/ansiblelint/app.py
apatard: Still, it would be nice to be able to disable it imho. I totally understand your aim but if I don't want to update for whatever reason / regression, it would be nice to be able to not get the message. Of course, for the exact reasons you're giving, it obviously should be enabled by default. Not the opposite.

src/ansiblelint/config.py
```diff
@@ -171,3 +187,76 @@ def parse_ansible_version(stdout: str) -> tuple[str, str | None]:
...
+    """Display warning if current version is outdated."""
+    msg = ""
+    data = {}
+    current_version = Version(__version__)
+    if not os.path.exists(CACHE_DIR):
+        os.makedirs(CACHE_DIR)
+    cache_file = f"{CACHE_DIR}/latest.json"
+    refresh = True
+    if os.path.exists(cache_file):
+        age = time.time() - os.path.getmtime(cache_file)
```
apatard, line 227: btw, since the cache file is a json file, what about putting the check time in it ?

src/ansiblelint/config.py
```diff
@@ -171,3 +187,76 @@ def parse_ansible_version(stdout: str) -> tuple[str, str | None]:
...
+        if age < 24 * 60 * 60:
```
apatard, line 228: You're going to hate me for that but why not make it a configuration option  ? :)

src/ansiblelint/config.py
ssbarnea: I don't want to modify it, i only cache it on disk.

src/ansiblelint/config.py
ssbarnea: Not initially, maybe in the future.

FloMiau:
How do I disable this update-check?

Issue #2622, opened 2022-10-27
erroneous "outdated linter version" message


Sergio-IME:
##### Summary

It appears that the package disagrees with its name (`ansiblelint` / `ansible-lint`) and/or version
and asks to be upgraded with a non-existent package while using the latest version

probably introduced here
https://github.com/ansible/ansible-lint/pull/2615/files#diff-638e063b189a4b2b5effe4845ba4d615b63bfb3b2e5c81cb0f80280ad0c9aa0aR206

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console
ansible --version
ansible [core 2.13.5]
  config file = /home/user/project/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/project/venv/lib/python3.10/site-packages/ansible
  ansible collection location = /home/ansbl/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/project/venv/bin/ansible
  python version = 3.10.8 (main, Oct 13 2022, 21:13:48) [GCC 12.2.0]
  jinja version = 3.1.2
  libyaml = True

ansible-lint --version
ansible-lint 6.8.3 using ansible 2.13.5
 Upgrade by running: pip install --upgrade ansiblelint
```

- ansible installation method: pip venv
- ansible-lint installation method: pip venv

##### OS / ENVIRONMENT

Arch / x86_64

##### STEPS TO REPRODUCE

- get latest version with `venv`
```console
pip install --upgrade -r requirements.txt
...
Requirement already satisfied: ansible in ./venv/lib/python3.10/site-packages (from -r requirements.txt (line 2)) (6.5.0)
Requirement already satisfied: ansible-lint in ./venv/lib/python3.10/site-packages (from -r requirements.txt (line 3)) (6.8.3)
...
```
- run `ansible-lint --version`
```console
(venv) $ ansible-lint --version
ansible-lint 6.8.3 using ansible 2.13.5
 Upgrade by running: pip install --upgrade ansiblelint
```

##### Desired Behavior

ask to upgrade the right version

##### Actual Behavior

```console
(venv) $ ansible-lint --version
ansible-lint 6.8.3 using ansible 2.13.5
 Upgrade by running: pip install --upgrade ansiblelint
```

```console
(venv) $ ansible-lint deploy.yml
Passed with production profile: 0 failure(s), 0 warning(s) on 1 files.
[...]

PR #2625, opened 2022-10-27
Fix package name in pip upgrade instructions

davedittrich:
Fixes https://github.com/ansible/ansible-lint/issues/2622 and adds support for Conda virtual environments.

### DIFF
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -192,27 +192,31 @@ def parse_ansible_version(stdout: str) -> tuple[str, str | None]:
 
 def in_venv() -> bool:
     """Determine whether Python is running from a venv."""
-    if hasattr(sys, "real_prefix"):
+    if (
+        hasattr(sys, "real_prefix")
+        or os.environ.get('CONDA_EXE', None) is not None
+    ):
         return True
+
     pfx = getattr(sys, "base_prefix", sys.prefix)
     return pfx != sys.prefix
 
 
 def guess_install_method() -> str:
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

src/ansiblelint/config.py
```diff
@@ -192,27 +192,31 @@ def parse_ansible_version(stdout: str) -> tuple[str, str | None]:
 
 def in_venv() -> bool:
     """Determine whether Python is running from a venv."""
-    if hasattr(sys, "real_prefix"):
+    if (
+        hasattr(sys, "real_prefix")
+        or os.environ.get('CONDA_EXE', None) is not None
+    ):
```
apatard: If the commit is about package name, why is this change here ?

apatard:
While you're at it, could you fix also the issue of always printing the " Upgrade by running: " message please ? (see my comment in #2622 )

src/ansiblelint/config.py
davedittrich: Because it looked from the path comparison logic like the code didn't take into consideration use of Conda as a virtual environment (though it is). I wasn't running in VScode, so maybe "real_prefix" is in `sys` with Conda?

apatard: Oh, I was not commenting the change by itself (since I'm not sure to understand it), just that imho, it was deserving its own commit.

isuftin:
@apatard  - Same https://github.com/ansible/ansible-lint/issues/2622#issuecomment-1294068923

PR #2626, opened 2022-10-27
Add experimental rule for run_once

ssbarnea:
Fixes: #1026
Closes: #2593


bluikko:
What - a warning will be printed if strategy `free` is used? That is not what 1026 is about at all, it is about the combined use of `free` and `run_once` (which is not supported by Ansible).

MicLennartz:
The logic in the rule is giving a warning, whenever `run_once` is used. No just in combination with a 'free' strategy.

I think, there are very legitimate use-cases for this keyword and it can be done in a predictable way. (we actually do this in our infrastructure). Also there's no hint in the Ansible reference docs, that the use of `run_once` is seen as problematic.

Please reconsider this PR and check only for the risky combination of `run_once` and `strategy: free`.

PR #2630, opened 2022-10-28
Avoid upgrade recommendation for bleeding edge users

ssbarnea:


apatard:
aren't you supposed to patch main() to not print an empty line in case of equality ?

ssbarnea:
> aren't you supposed to patch main() to not print an empty line in case of equality ?

Done! Thanks point pointing it.

cidrblock (approved):

Issue #2629, opened 2022-10-28
no-changed-when behavior is inconsistent


nre-ableton:
##### Summary

This issue is the same as https://github.com/ansible/ansible-lint/issues/1894, but since I don't have permission to re-open issues, I'm filing a new one (with the sample code updated for new rules such as `key-order`). :slightly_smiling_face: I made two comments on the closed issue, but since it's a closed issue it is possible that they were not seen. For bookkeeping purposes, if the project admins want to close this issue as a dupe and reopen #1894, I'd be fine with that.

Anyways, the original issue was [claimed](https://github.com/ansible/ansible-lint/issues/1894#issuecomment-1264473362) to have been fixed in 6.8.0b1, but is still easily reproducible in 6.8.4 with the below codeblock. 

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.13.5]
  config file = None
  configured module search path = ['/home/nre/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/nre/.cache/virtualenv/ansible-role-amtool/lib/python3.10/site-packages/ansible
  ansible collection location = /home/nre/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/nre/.cache/virtualenv/ansible-role-amtool/bin/ansible
  python version = 3.10.6 (main, Sep 14 2022, 13:37:48) [GCC 11.2.0]
  jinja version = 3.1.2
  libyaml = True

$ ansible-lint --version
ansible-lint 6.8.4 using ansible 2.13.5
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Ubuntu Linux 22.04

##### STEPS TO REPRODUCE

For an example, see the `build-sources.yml` task file of the `ableton.ccache` role [here](https://github.com/Ableton/ansible-role-ccache/blob/25cd3ae4b9c0174c7ce5de3e2dc17b117768630d/tasks/unix/build-sources.yml#L10-L57). An abbreviated version of this file follows:

```yaml
---
- name: Check ccache version
  ansible.builtin.command: "ccache --version"
  environment:
    PATH: "{{ ccache_install_prefix }}/bin:{{ ansible_pkg_mgr_path }}"
  changed_when: false
  failed_when: false
  register: ccache_version_cmd

- name: Build ccache from sources
  when: >
    ccache_version_cmd.rc != 0 or
[...]

PR #2653, opened 2022-11-04
Consider conditions at the block level for the tasks inside a block

ajinkyau:
Now conditions at the block level are considered for the tasks inside a block
Fixes [2629](https://github.com/ansible/ansible-lint/issues/2629)


src/ansiblelint/_internal/rules.py
```diff
@@ -104,6 +104,28 @@ def matchtask(
         the original task (before normalization) will be made available under
         ``task["__raw_task__"]``.
         """
+
```
ssbarnea, line 107: You do not want to modify implementation matchtask, instead you want to modify the implementation of `iter_tasks_in_file` function, which is called by matchtask (the function calling matchtask for each task).

Basically adding the inherited tasks should be done automatically by `iter_tasks_in_file`.

ssbarnea (changes requested):
See prev comment related to iter_tasks_in_file

PR #2656, opened 2022-11-04
Refactor get_app

ssbarnea:
Change implementation of get_app to make it avoid initializing the app more than once. This should sort some weird behaviours with parallel testing.

### DIFF
--- a/src/ansiblelint/app.py
+++ b/src/ansiblelint/app.py
@@ -325,9 +325,10 @@ def _sanitize_list_options(tag_list: list[str]) -> list[str]:
     return sorted(set(tags))
 
 
-@lru_cache(maxsize=1)
-def get_app(offline: bool = False) -> App:
-    """Return the application instance."""
+@lru_cache
+def get_app() -> App:
+    """Return the application instance, caching the return value."""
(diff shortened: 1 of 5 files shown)


src/ansiblelint/app.py
```diff
@@ -324,13 +325,15 @@ def _sanitize_list_options(tag_list: list[str]) -> list[str]:
     return sorted(set(tags))
 
 
-@lru_cache(maxsize=1)
```
cognifloyd, line 327: Maybe we should bump this to `@lru_cache(maxsize=2)` which would cover both offline=True and offline=False. This would mean there are at most two app instances. It would also mean that syntax_check and _role_path can continue to request offline mode, and the expense of setting up the app for that only happens one extra time. Would that resolve the issue you're experiencing?

With this PR, only the first invocation of `get_app` will use the `offline` arg, otherwise the arg will be ignored.

ssbarnea: I am still playing with it. I think that I will remove the parameter as I am not aware of any real "production" use case where you would want to instantiate more than one and with different types. When you run it, you will either want online or offline mode.

Issue #2672, opened 2022-11-11
Docs: Installation and configuration


oraNod:
##### Summary

Edits and structural changes needed to improve general quality and completeness for user stories around installation, configuration, and usage.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

- using latest pip install of ansible-lint

##### OS / ENVIRONMENT

F36

##### STEPS TO REPRODUCE

N/A

##### Desired Behavior
N/A

##### Actual Behavior

N/A

PR #2685, opened 2022-11-15
Improve documentation on custom rules

oraNod:
Partially fixes https://github.com/ansible/ansible-lint/issues/2672 with documentation for custom lint rules


ssbarnea (changes requested):
Please avoid adding ansible-lint to any documentation titles as this will make the sidebar menu harder to read.

There is no need to repeat the name of the product in titles, it is already part of the URL and page structure.

ssbarnea (approved):

PR #2703, opened 2022-11-19
Enable dynamic schema refresh

ssbarnea:


felixfontein:
I'm not sure this is a good idea. It basically means that two folks two run ansible-lint for a specific x.y.z version end up with different results, especially when one is using `--offline` (like you should do in CI to make results repoducable) and the other one not (developers usually locally don't add that flag).

ssbarnea:
> I'm not sure this is a good idea. It basically means that two folks two run ansible-lint for a specific x.y.z version end up with different results, especially when one is using `--offline` (like you should do in CI to make results repoducable) and the other one not (developers usually locally don't add that flag).

The difference is quite subtle because schemas changes are of minor nature. Look a https://github.com/ansible/schemas/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Amerged to get an idea about what kinds of changes are applied to them. I do not want to have to make a new release just to document existence of a new OS release in galaxy schema, it is to much work for both upstream/downstream use.

The reality is that we will slowly see more "online" feature being introduced in the future as we want to speedup the development.

The reproducibility, especially on CI is not something that is 0 or 1, it is always a gray agrea, ranging from very-good to very-poor. Different results can be produced even when running in offline mode or the extreamly of it the locked-down environment. One example is that SSL certificates expire, so running the same thing would produce different results in isolation, only if the clock is different.

The real question about this feature is if the current implementation is resilient enough to fallback to current copy when network is unavailable. We don't want to see any failure related to not being able to contact github.

felixfontein:
> The reality is that we will slowly see more "online" feature being introduced in the future as we want to speedup the development.

IMO that's a really dangerous slippy slope that a linter should really, **really** avoid.

ganeshrn (approved):

Issue #2664, opened 2022-11-09
role-name in a skip_list not loading


y0rune:
##### Summary
After added the file `.ansible-lint`, the file has been loaded to the settings but still I got the error with it. 
Error:
```txt
...
As an alternative, you can add 'role-name' to either skip_list or warn_list.
```

As you can see below (from -vv) the skip list and all settings has been loaded/

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.14.0]
  config file = None
  configured module search path = ['/home/yorune/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/yorune/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/yorune/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/yorune/.local/bin/ansible
  python version = 3.10.4 (main, Mar 31 2022, 13:15:24) [GCC 9.3.0] (/usr/lib/python-exec/python3.10/python3)
  jinja version = 3.1.2
  libyaml = True

$ ansible-lint --version
ansible-lint 6.8.6 using ansible 2.14.0
```

- ansible installation method: pip
- ansible-lint installation method: pipe

##### OS / ENVIRONMENT
OS: Gentoo
```txt
yorune@Gentoo-d9yc132 ~ master $ emerge --info                                                                                                                                                                                                                  


Portage 3.0.30 (python 3.9.12-final-0, default/linux/amd64/17.1, gcc-10.3.0, glibc-2.35-r7, 5.4.66-gentoo-x86_64 x86_64)                                                                                                                                        

=================================================================                                                                                                                                                                                               

System uname: Linux-5.4.66-gentoo-x86_64-x86_64-Intel-R-_Core-TM-_i7-4790_CPU_@_3.60GHz-with-glibc2.35                                                                                                                                                          

KiB Mem:    16303496 total,   1488888 free                                                                                                                                                                                                                      

KiB Swap:          0 total,         0 free                                                                                                                                                                                                                      

Timestamp of repository gentoo: Wed, 19 Oct 2022 12:02:06 +0000 
```

##### Behaviour

```console (paste below)
$ ansible-lint --force-color -vv
DEBUG    Logging initialized to level 10
[...]

PR #2706, opened 2022-11-21
Respect warn_list and skip_list

rekup:
Respect `warn_list` and `skip_list` while preparing environment

Fixes: #2664

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -163,7 +163,7 @@ jobs:
       WSLENV: FORCE_COLOR:PYTEST_REQPASS:TOXENV:TOX_PARALLEL_NO_SPINNER
       # Number of expected test passes, safety measure for accidental skip of
       # tests. Update value if you add/remove tests.
-      PYTEST_REQPASS: 713
+      PYTEST_REQPASS: 714
 
     steps:
       - name: Activate WSL1
--- a/examples/roles/invalid-name/tasks/main.yaml
+++ b/examples/roles/invalid-name/tasks/main.yaml
@@ -1,4 +1,4 @@
 ---
 - name: Foo
-  debug:
+  ansible.builtin.debug:
     msg: foo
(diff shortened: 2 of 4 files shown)


ssbarnea:
@rekup By any chance, could you also add a test for it so we do not break it again while refactoring the code?

ssbarnea (approved):

bviktor:
After hours of messing around with this, I finally fixed it on localhost.

... then I saw this commit in Git.

Unfortunately this logic makes `role-name` get into the `warn_list` regardless of what you set in your config. You can add `role-name` to `skip_list`, or `enable_list`, or even _both_, it will pass with a warning.

Why? Relax, it's not this commit's fault. It's because `role-name` is in [DEFAULT_WARN_LIST](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/config.py#L29) for whatever reason. So even though the role-name rule is

- not labeled as `opt-in`
- part of the `basic` profile

it always passes role-name checks with a warning. So this PR as it is, is still not fixing the issue. Before, it always threw an error because of ansible-compat's `prepare_environment` defaulting to `role_name_check = 0`, this commit at least fixed that, now we pass the variable as we should.

But now it always throws a warning because of `DEFAULT_WARN_LIST`.

We could mess around with an if-else rework::

```python
    role_name_check = 1
    if "role-name" in app.options.enable_list:
        role_name_check = 0
    elif "role-name" in app.options.warn_list:
        role_name_check = 2
```

So we start with the (default) `warning_list`. If the user explicitly puts it on the `enable_list`, it will error out. If the user explicitly puts it on the `skip_list`, it will pass without warnings.

So it's **almost** good, and would definitely improve the logic, **but** there's caveat: it wouldn't be enabled **by default**, even though the docs and the labels indicate so.

So (at least from a skip/warn/enable_list perspective) a 100% correct solution is to remove `role-name` from `DEFAULT_WARN_LIST`.

In that case:

- by default, ansible-lint throws an error, as we would expect it
- if you add it to `warn_list`, it throws a warning and passes, as we would expect it
- if you add it to `skip_list`, it passes without warnings, , as we would expect it

The only thing left to be figured out is, why was role-name added to the default warning list in the first place. I can't really come up with any explanations, as it was **always** erroring out previously, so it made absolutely no difference to put it in there.

PR #2708, opened 2022-11-21
Clarify progressive mode documentation

MarcinWieczorek:
Hello.

In this PR I have fixed two issues regarding the progressive mode.

1. The default config lacks progressive mode, it is only mentioned in the cli help. I have added `progressive: false` with a comment (which is modified from the original, see 2.)

2. The logic behind progressive marking a success is a comparison of sets. The help says that the number of violations must be reduced, which is not true, because keeping the same number of violations is still a success: `len({1,2,3} - {1,2,3})` evaluates to `0` (`src/ansiblelint/__main__.py#233`). Therefore ansible-lint returns success if the number of violations has not increased. This takes no change in number into account. Once again, reducing means that every commit needs to remove at least 1 violation. Note, that the warning in `__main__.py` states correctly that `Total violations not increased since previous commit`, so does the `docs/usage.md`: `as long the total number of violations did not increase since the previous commit.`

I have read `test_progressive.py` file and came to the conclusion that these are issues of purely documentation nature, there's no need to additionally test, but I will continue to verify that once resolvelib dependency is fixed in my distro...

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -72,6 +72,11 @@ warn_list:
 # Offline mode disables installation of requirements.yml
 offline: false
 
+# Return success if number of violations compared with previous git
+# commit has not increased. This feature works only in git
+# repositories.
+progressive: false
+
 # Define required Ansible's variables to satisfy syntax check
 extra_vars:
   foo: bar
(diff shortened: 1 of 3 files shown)


ssbarnea (approved):

MarcinWieczorek:
Thanks for merging!

PR #2710, opened 2022-11-22
allow additional collections in only-builtins

evgeni:


evgeni:
I do not exactly understand why it fails on MacOS.

ssbarnea (approved):

evgeni:
@ssbarnea can/will you investigate the macOS errors, or should I try to? (I don't have macOS here, so will have to fall back to abusing GH actions :crying_cat_face: )

ssbarnea:
> @ssbarnea can/will you investigate the macOS errors, or should I try to? (I don't have macOS here, so will have to fall back to abusing GH actions 😿 )

The bad news is that I do have MacOS and I am not able to reproduce the error locally. Feel free to abuse them as much as you need.

evgeni:
Okay, so the error on macOS is:
```
syntax-check[specific]: couldn't resolve module/action 'fake_namespace.fake_collection.fake_module'. This often indicates a misspelling, missing collection, or incorrect module path.
```
I don't exactly understand why it resolves fine on Linux but not on macOS, but I will try to explicitly add it to the mocked list in the config.

PR #2715, opened 2022-11-22
Update packit

ssbarnea:


.packit.yaml
```diff
@@ -18,8 +18,8 @@ jobs:
     metadata:
       targets:
         # - fedora-34  # has no https://src.fedoraproject.org/rpms/python-build
```
evgeni, line 20: I guess this comment can be removed, now that f34 is EOL?

evgeni (approved):

PR #2716, opened 2022-11-22
Reduce test dependency on galaxy server

ssbarnea:


examples/reqs_v1/requirements.yml
```diff
@@ -1,3 +1,10 @@
 ---
 # v1 requirements test file
-- src: geerlingguy.mysql
+
+# As Jeff never created releases on hit repo, we are unable to download tar.gz
+# archives, so we need to use the slower git cloning instead.
+# - src: https://github.com/geerlingguy/mysql/archive/refs/tags/4.2.0.tar.gz
```
felixfontein, line 6: Why not just use https://github.com/geerlingguy/ansible-role-mysql/archive/4.2.0.tar.gz instead of cloning https://github.com/geerlingguy/ansible-role-mysql at tag 4.2.0?

felixfontein: I don't see where you updated this PR.

PR #2727, opened 2022-11-24
Treat ansible.legacy the same way ansible.builtin

evgeni:


evgeni:
"Coverage failure: total of 87 is less than fail-under=88" -- it's still one statement, why would the coverage be lower than before?

ssbarnea (approved):

ssbarnea:
> "Coverage failure: total of 87 is less than fail-under=88" -- it's still one statement, why would the coverage be lower than before?

It can happen but is quite rare, safe to ignore it.

PR #2732, opened 2022-11-24
Allow additional modules in only-builtins

evgeni:
This is most useful for old-style roles that ship own modules in a `library` folder.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -101,3 +101,11 @@ kinds:
   # - vars: "**/vars/*.yml"
   # - meta: "**/meta/main.yml"
   - yaml: "**/*.yaml-too"
+
+# List of additional collections to allow in only-builtins rule.
+# only_builtins_allow_collections:
+#   - example_ns.example_collection
+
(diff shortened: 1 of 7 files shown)


ssbarnea (approved):

ssbarnea (changes requested):
Please add both new options for the sample config file from root of repository.

Also modify the ansible-lint-config.json file to allow them. That is the JSON Schema for the config file. Let me know if you do not know how to add them as "list of strings", I can help.

evgeni:
> Please add both new options for the sample config file from root of repository.

Will do

> Also modify the ansible-lint-config.json file to allow them. That is the JSON Schema for the config file. Let me know if you do not know how to add them as "list of strings", I can help.

Both are (collections was before, this PR adds modules) part of the schema:
https://github.com/ansible/ansible-lint/blob/5c13da155d1a9225028806334bc183c9034df638/src/ansiblelint/schemas/ansible-lint-config.json#L59-L72

ssbarnea (approved):

Issue #2690, opened 2022-11-16
Rule name[casing] conflicts with COP 3.1.16


alice-rc:
##### Summary
The current rule <i>name[casing]: All names should start with an uppercase letter</i> directly conflicts with 
COP reference [3.1.16 Prefix names in sub-tasks files of roles with the subtask's filename (Example: subtask filename | subtask description)](https://redhat-cop.github.io/automation-good-practices/#_prefix_task_names_in_sub_tasks_files_of_roles) 

##### Issue Type
- Bug Report

##### Ansible and Ansible Lint details

```
ansible --version
ansible [core 2.13.4]
  config file = /home/USER/venv/ee-rcstandard-rhel8-183/meta/ansible.cfg
  configured module search path = ['/home/USER/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/USER/venv/ee-rcstandard-rhel8-183/lib64/python3.9/site-packages/ansible
  ansible collection location = /home/USER/venv/ee-rcstandard-rhel8-183/collections
  executable location = /home/USER/venv/ee-rcstandard-rhel8-183/bin/ansible
  python version = 3.9.7 (default, Apr 11 2022, 06:30:15) [GCC 8.5.0 20210514 (Red Hat 8.5.0-10.0.1)]
  jinja version = 3.1.2
  libyaml = True

ansible-lint --version
ansible-lint 6.8.6 using ansible 2.13.4
```
- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT

Not relevant, but Oracle Linux 8

##### STEPS TO REPRODUCE
Line in the ansible subtask file:
```
- name: type_server | Setting facts from CMDB info
```
Output from ansible-lint:
```
roles/cmdb/tasks/type_server.yml:1: name[casing]: All names should start with an uppercase letter.
```

##### Desired Behavior

The rule should NOT apply to the subtask portion of the task name, just the part that falls after the "|"

##### Actual Behavior

The correctly implemented COP 3.1.16 name is triggering the ansible-lint name[casing] rule

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

ssbarnea:
@cidrblock WDYT about this, should we update the rule to document and support that? I looks a little bit weird to name a task `foo | Do something` but I do also understand why they wanted to add it.

alice-rc:
[...]

PR #2740, opened 2022-11-25
Allow a prefix in subtasks names

shatakshiiii:
- Adds `name[prefix]` in the NameRule.
- Fixes: #2690

### DIFF
--- a/src/ansiblelint/rules/name.py
+++ b/src/ansiblelint/rules/name.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import re
 import sys
+from copy import deepcopy
 from typing import TYPE_CHECKING, Any
 
 from ansiblelint.constants import LINE_NUMBER_KEY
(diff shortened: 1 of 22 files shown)


src/ansiblelint/rules/name.py
```diff
@@ -74,6 +74,19 @@ def _check_name(
         # lowercase letter, so we ignore anything else. On Unicode isupper()
         # is not necessarily the opposite of islower()
         results = []
+        task_components = name.split("|")
+        if len(task_components) > 1:
+            prefix_name = task_components[0].strip()
+            if prefix_name == "main":
```
ssbarnea, line 80: You also need to check if the prefix is matching allowed pattern. I would suggest to take a different approach and build the expected prefix based on prefix template string. If it does not match we can raise an error with tag `name[prefix]` and a clear message.

This also means that you can easily check if filename base name (w/o extension) is main and make the expected prefix = "".

You will be able to generate an error message like "Task name should use `<foo>` prefix in order to make it easy to identify.

ssbarnea (changes requested):
see comment

ssbarnea (approved):

cidrblock (approved):
+1 for _collection and the test coverage looks good

PR #2745, opened 2022-11-26
Improve RPM Packaging

gotmax23:
(See the detailed commit bodies for more information :). I hope this repository doesn't use squash merges; I spent extra time splitting this up into logical changes... )

This improves ansible-lint.spec to follow best practices. It also addresses the glaring uninstallability problem. The pythondist runtime dependency generator preserves the version constraints you set, so if those aren't available in the Fedora repositories, the package won't install. %pyproject_buildrequires generates buildtime dependencies based on the upstream metadata after `%prep`, so the build will fail if those aren't installable, instead of silently producing broken packages.

Along with this, the version constraints were relaxed to versions that are actually available in the Fedora 37 and Rawhide repositories. Fedora 36's jsonschema and ansible-compat are a major version behind, so I left that out.

Side note: While aggressive version pins might seem beneficial from the upstream perspective, they make it difficult for distribution package maintainers to package your software. I urge you to reconsider this practice and only use pin versions when it's absolutely necessary.

### DIFF
--- a/.config/ansible-lint.spec
+++ b/.config/ansible-lint.spec
@@ -45,6 +45,7 @@ potentially be improved.
 
 %install
 %pyproject_install
+%pyproject_save_files ansiblelint
 
 
 %if %{with check}
@@ -62,9 +63,7 @@ PYTHONPATH=%{buildroot}%{python3_sitelib} \
 %endif
 
 
-%files
-%{python3_sitelib}/ansiblelint/
-%{python3_sitelib}/ansible_lint-*.dist-info/
+%files -f %{pyproject_files}
 %{_bindir}/ansible-lint
 %license COPYING
 %doc docs/* README.md
--- a/.config/ansible-lint.spec
+++ b/.config/ansible-lint.spec
@@ -23,7 +23,7 @@ BuildRequires:  python%{python3_pkgversion}-devel
 BuildRequires:  python%{python3_pkgversion}-pytest
 BuildRequires:  python%{python3_pkgversion}-pytest-xdist
 BuildRequires:  python%{python3_pkgversion}-libselinux
-BuildRequires:  git
+BuildRequires:  git-core
 %endif
 
 
--- a/.config/ansible-lint.spec
+++ b/.config/ansible-lint.spec
@@ -17,33 +17,15 @@ Source0:        %{pypi_source}
 
 BuildArch:      noarch
 
-BuildRequires:  pyproject-rpm-macros
-BuildRequires:  python%{python3_pkgversion}-build
 BuildRequires:  python%{python3_pkgversion}-devel
-BuildRequires:  python%{python3_pkgversion}-pip
-BuildRequires:  python%{python3_pkgversion}-setuptools
-BuildRequires:  python%{python3_pkgversion}-setuptools_scm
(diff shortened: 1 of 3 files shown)


.config/ansible-lint.spec
```diff
@@ -11,39 +11,21 @@ Version:        VERSION_PLACEHOLDER
 Release:        1%{?dist}
 Summary:        Ansible-lint checks ansible content for common mistakes
 
-License:        MIT
+License:        GPL-3.0-or-later AND MIT
```
ssbarnea, line 14: Drop MIT, keep GPL-3.0-or-later as that is the distribution licensing.

.config/ansible-lint.spec
```diff
@@ -76,12 +64,9 @@ PYTHONPATH=%{buildroot}%{python3_sitelib} \
...
 
 
-%files
-%{python3_sitelib}/ansiblelint/
-%{python3_sitelib}/ansible_lint-*.dist-info/
+%files -f %{pyproject_files}
 %{_bindir}/ansible-lint
-%license COPYING
-%doc docs/* README.md
+%license COPYING docs/licenses/LICENSE.mit.txt
```
ssbarnea, line 69: GPL

gotmax23:
I'm thinking we should run the builds with networking and `--with check`. Downstream distributions build without networking, so even if the build environment wouldn't be as representative, the package would have better coverage overall. What do you think?

.config/ansible-lint.spec
gotmax23: This is the correct License tag per the new Fedora licensing guidelines. Both MIT and GPL-3.0-or-later apply to the distribution. Distributing the package under the GPL-3.0-or-later doesn't erase the MIT licensed code.

ssbarnea:
@gotmax23 Just to give you an early warning, I already started implementation of PEP-621, see https://github.com/tox-dev/tox-extra/pull/61 - I expect to migrate all projects before the end of the year. If the RPM process is using "python3 -m build", I do not see reasons to worry, is just that `setup.cfg` and `setup.py` files will no longer exist.

gotmax23:
> @gotmax23 Just to give you an early warning, I already started implementation of PEP-621, see [tox-dev/tox-extra#61](https://github.com/tox-dev/tox-extra/pull/61) - I expect to migrate all projects before the end of the year. If the RPM process is using "python3 -m build", I do not see reasons to worry, is just that `setup.cfg` and `setup.py` files will no longer exist.

I would suggest against that. RHEL 9 and Fedora 36's setuptools version is too old for that.  I believe ansible-lint is part of AAP, so using setuptools with PEP 621 presents a problem there. FWIW, EPEL 7, 8, 9, and all Fedoras have flit-core which does support this. It's not part of RHEL itself, however, so I guess flit-core would also present a problem for AAP.

gotmax23:
> If the RPM process is using "python3 -m build", I do not see reasons to worry, is just that setup.cfg and setup.py files will no longer exist.

To be clear, our packaging tools do support PEP 517 build backends, and this does work if the setuptools version is new enough to understand PEP 621 metadata.

ssbarnea (approved):

ssbarnea:
It is up to you, but versions in requirements.txt are more likely to be better because these are the tested versions.

Issue #2758, opened 2022-11-30
risky-octal validation partially broken for non-quoted permissions


tomsiewert:
##### Summary

If the file permission is set to `0646`, ansible-lint throws a `risky-octal` error. When the file permission is quoted or set to something else than `0646`, the issue does not occur.

##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
ansible --version
ansible [core 2.14.0]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.11/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.11.0 (main, Nov 12 2022, 07:10:51) [GCC 11.2.1 20220219] (/usr/local/bin/python)
  jinja version = 3.1.2
  libyaml = False

ansible-lint --version
ansible-lint 6.9.0 using ansible 2.14.0

```

- ansible installation method: pip3
- ansible-lint installation method: pip3

##### OS / ENVIRONMENT

uname: `6.0.9 #1-NixOS SMP PREEMPT_DYNAMIC Wed Nov 16 09:04:15 UTC 2022 x86_64 GNU/Linux`

ansible-lint runs in a `python:alpine` Docker image

##### STEPS TO REPRODUCE

ansible-lint config:
```
--- .ansible-lint ---
# merge-with-defaults
skip_list:
  - package-latest
  - no-changed-when
  - schema
  - experimental
  - name[play]
  - name[casing]
  - name[template]
  - jinja[spacing]
  - template-instead-of-copy
warn_list:
  - jinja[invalid]
exclude_paths:
  - .git/
  - .cache/
  - roles/
  - files/
enable_list:
  - yaml
extra_vars:
  servers: localhost
  hostvars:
    DUMMY_HOST:
      hosts:
        - localhost
kinds:
  - playbook: '**/play*.{yml,yaml}'
---------------

--- .yamllint ---
# merge-with-defaults
rules:
  brackets:
    min-spaces-inside: 1
    max-spaces-inside: 1
    min-spaces-inside-empty: 0
    max-spaces-inside-empty: 0
    level: error
  braces:
    max-spaces-inside: 1
    min-spaces-inside: 1
    min-spaces-inside-empty: 0
    max-spaces-inside-empty: 0
    level: error
[...]

PR #2759, opened 2022-11-30
Refactor risky-octal and improve its messaging

ssbarnea:
Fixes: #2758

New errors looks like:
```
risky-octal: `mode: 600` is risky and possible wrong, rewrite it as a string such `mode: "01130" or use symbolic mode.
risky-octal: `mode: 710` is risky and possible wrong, rewrite it as a string such `mode: "01306" or use symbolic mode.
risky-octal: `mode: 123` is risky and possible wrong, rewrite it as a string such `mode: "0173" or use symbolic mode.
risky-octal: `mode: 2000` is risky and possible wrong, rewrite it as a string such `mode: "03720" or use symbolic mode.
```


src/ansiblelint/rules/risky_octal.md
```diff
@@ -1,6 +1,11 @@
 # risky-octal
 
-This rule checks that octal file permissions either contain a leading zero or are a string value.
+This rule checks that octal file permissions are strings that contain a leading
+zero or are written in symbolic mode (such as `u+rwx` or `u=rw,g=r,o=r`).
```
oraNod, line 4:
```suggestion
This rule checks that octal file permissions are strings that contain a leading
zero or are written in symbolic mode, such as `u+rwx` or `u=rw,g=r,o=r`.
```

src/ansiblelint/rules/risky_octal.md
```diff
@@ -1,6 +1,11 @@
...
+
+This is because using a numeric decimal or octal in YAML can easily lead to
+costly unexpected mistakes. For example, `0644` is interpreted as `420` as
+a decimal.
```
oraNod, line 8: I'd like to clarify this bit some more. Is this intended to recommend symbolic mode instead of octal number?

When would the YAML parser interpret 0644 as the 420 decimal number? Does that happen during serialization or something?

src/ansiblelint/rules/risky_octal.py
```diff
@@ -98,6 +100,93 @@ def matchtask(
             if isinstance(mode, str):
                 return False
 
-            if isinstance(mode, int):
-                return self.is_invalid_permission(mode)
+            if isinstance(mode, int) and self.is_invalid_permission(mode):
+                return f'`mode: {mode}` is risky and possible wrong, rewrite it as a string such `mode: "0{mode:o}" or use symbolic mode.`.'
```
oraNod, line 104:
```suggestion
                return f'`mode: {mode}` should have a string value with leading zero `mode: "0{mode:o}"` or use symbolic mode.'
```

src/ansiblelint/rules/risky_octal.md
ssbarnea: It is happening during the loading, so is outside our control.

oraNod: So is it recommended to use symbolic mode instead of something like 0644 to avoid the wrong interpretation?

ssbarnea: From our point of view using a string with `0644` is equally safe as symbolic mode. User can decide. In the future we might even suggest how symbolic representation for the octal, so they can switch from one to another. Still, for now we only want to avoid integers/octals inside the YAML, as they are tricky.

src/ansiblelint/rules/risky_octal.md
```diff
@@ -1,6 +1,11 @@
 # risky-octal
 
-This rule checks that octal file permissions either contain a leading zero or are a string value.
+This rule checks that octal file permissions are strings that contain a leading
+zero or are written in symbolic mode, such as `u+rwx` or `u=rw,g=r,o=r`.
+
+This is because using a numeric decimal or octal in YAML can easily lead to
+costly unexpected mistakes. For example, `0644` is interpreted as `420` as
+a decimal.
```
oraNod, line 8:
```suggestion
Using an octal string in YAML can result in unexpected error. For example, the YAML parser can interpret `0644` as the decimal number `420`.
```

oraNod: @ssbarnea Not sure if this changes meaning too much but I think it needs to be clarified a little more.

ssbarnea: I used your version to produce something in between as you got a small mistake.

examples/playbooks/rule-risky-octal-pass.yml
```diff
@@ -0,0 +1,42 @@
...
+      ansible.builtin.file: path=baz mode=0711 # noqa: no-free-form
+
+    - name: Permissions test success (0777)
+      ansible.builtin.file: path=baz mode=u+rwx # noqa: no-free-form
+
+    - name: Octal permissions test success (777)
+      ansible.builtin.file: path=baz mode=777 # noqa: no-free-form
+
+    - name: Octal permissions test success (733)
+      ansible.builtin.file: path=baz mode=733 # noqa: no-free-form
```
oraNod, line 42: Why do these last two pass? Is it because they are free form?

If I do this it fails:

```
      ansible.builtin.file:
        path: foo
        mode: 733
 ```

ssbarnea: Yes, free from is even more painful as YAML loader does happen so I suppose (but I am not sure) that 733 might be loaded as anything, string, integer or octal, based on how the module author decided to parse the output. That one is messed at multiple levels.

PR #2770, opened 2022-12-02
Add meta-unsupported-ansible rule for checking requires_ansible values

### LINKED ISSUE #2102: meta-unsupported-ansible: requires_ansible key
affirm that the `requires_ansible:` key in the meta/runtime.yml file has a value within current supported versions of ansible-core

rule should be optional and only enabled at users discretion. Error level should be fatal (1) as galaxy-importer will disallow the import if not present currently. 

based off of code found [here](https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/loaders/runtime_file.py#L46)

similar context to #2101

alisonlhart:
Adding rule to check that the requires_ansible variable in meta/runtime.yml is set to a supported version of ansible-core. 

Example error:
```yml
---
requires_ansible: ">=2.9.0"  <---- not a supported version of ansible

```

Example pass:
```yml
---
requires_ansible: ">=2.9.10"  <---- a supported version of ansible

```

Fixes #2102


ssbarnea (approved):

SRv6d:
This rule breaks version numbers that omit the patch version. 
Omitting the patch version is common practice and even [suggested in the docs](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html#meta-directory-and-runtime-yml). 
If this was an intentional decision, it should at least be documented.

alisonlhart:
@ssbarnea RE: comment above - There's an argument to be had for omitting the patch version requirement. However, as far as I can see, the only instance needed for omitting patch version is for compatibility with prerelease versions of Ansible. As stated in the docs: `Ansible deviates from PEP440 behavior by truncating prerelease segments from the Ansible version. Ansible 2.11.0b1 is compatible with something that requires_ansible: ">=2.11".`

So, since the intention of the `shared` profile is best practices for building useful content to be shared on Galaxy, Automation Hub, and private instances, I would argue there's not really an instance where using a patch version specifier would be an issue, unless you think the above would be a valid reason to change it. 

I would say it's ideal to always specify patch versions, but again, I could see the benefit with removing it.

PR #2782, opened 2022-12-05
Fix error on empty config list

julien-lecomte:
Fixes the error when a ansible-lint config file has an empty list such as

~~~yaml
enable_list:

~~~

### DIFF
--- a/src/ansiblelint/cli.py
+++ b/src/ansiblelint/cli.py
@@ -495,7 +495,7 @@ def merge_config(file_config: dict[Any, Any], cli_config: Namespace) -> Namespac
     for entry, default in lists_map.items():
         if getattr(cli_config, entry, None) or entry in file_config.keys():
             value = getattr(cli_config, entry, [])
-            value.extend(file_config.pop(entry, []))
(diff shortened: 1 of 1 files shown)


ssbarnea:
I am not sure we want to allow this, `enable_list` is a list by definition, passing it anything else than a list should still be an error IMHO.

Even `enable_list: null` or `enable_list: {}` should be errors, the only way to empty it should be `enable_list: []`.

While ansible might be quite open about what you feed it, it does not mean we need to make the same kind of decisions, especially as this likely caused by an incomplete commenting in linter config.

@cidrblock @ganeshrn WDYT?

cidrblock:
I tend to agree, if it is present, it should be a list or an error is presented (not a stack tho :) )

ssbarnea (changes requested):
I see two ways to fix this bug properly: either use jsonschema validation before loading the file (ideal), or just testing that the loaded entry is of type list before doing anything with it.

The second approach is easier to implement but the first one is much better, helping with all the other options, likely sorting similar bugs with other options.

Issue #2729, opened 2022-11-24
--offline (or offline: true) does not do, what is said in the docs


dbk-rabel:
# Issue Type

- Bug report

# Ansible and Ansible Lint details

```
ansible-lint --version
ansible-lint 6.8.7 using ansible 2.14.0
```

- ansible-lint installation method: pip

# Desired Behavior

Behaviour of parameter --offline (oder equivalently option offline: true) is coherant with documentation:

--offline             Disable installation of requirements.yml

# Actual Behavior (Bug report only)

What --offline really does, are apparently these two things:

1) Do not connect to Github. (That's very useful btw.)
2) Pass --offline to 'ansible-galaxy collection install -r requirements.yml', which seems to have no effect at all.

```
1:42:12   WARNING  Retrying execution failure 250 of: ansible-galaxy collection install -v --offline -r requirements.yml -p /opt/app-root/src/.cache/ansible-compat/68c7a5/collections
11:42:12   ERROR    Using /<censored>/ansible.cfg as config file
11:42:12   to see the full traceback, use -vvv
11:42:12   
11:42:12   ERROR    ERROR! Unexpected Exception, this is probably a bug: [Errno 2] No such file or directory: b'https://<censored>.tar.gz
11:42:12   
11:42:12   Got 250 exit code while running: ansible-galaxy collection install -v --offline -r requirements.yml -p /opt/app-root/src/.cache/ansible-compat/68c7a5/collections
```

# Proposed solution

1) Add the "do not connect to github" behaviour to the docs.
2) Fix --offline parameter, so that there is not ansible-galaxy collection install instead of just forwarding the --ofline parameter.

Thanks.
David

MarkusTeufelberger:
I have a similar issue (also using ansible-core 2.14.0 and the newer ansible-lint 6.9.0), I have downloaded collections to their proper place already AND set the collections_paths in ansible-cfg correctly, there is absolutely no need for ansible-lint to run `ansible-galaxy collection install`, with an `--offline` flag or without. Just don't run galaxy at all if `--offline` is specified and don't mess with my collections path settings please. Now I need to waste traffic and time again for a while as well as deal with flaky availability of galaxy by turning offline mode off until this is fixed... :-(

MarkusTeufelberger:
Likely related to #2683 btw

bluikko:
Same here. I wonder why is `ansible-galaxy` called to download collections in the first place when all the collections are already installed in the venv?

[...]

PR #2785, opened 2022-12-06
Correct schema refresh logic

ssbarnea:
- Ignore ConnectionResetError, reported on https://github.com/ansible/ansible-lint/issues/2729#issuecomment-1339099602
- Stop refresh process on first found error, to avoid delays when
  networking is not available
- Include etags for current schemas to avoid refreshing them
- Update schemas from linter repo too if they changed

Please note that offline mode does disable schema completely.


erikgb:
Answering https://github.com/ansible/ansible-lint/issues/2729#issuecomment-1339176974, since the issue is now locked. 😡 

Thanks for the update @ssbarnea. Sorry for "raising my voice", but there have been numerous issues with this "thing" in the latest patch! releases of ansible-lint. I find it very annoying.

MarkusTeufelberger:
> Please note that offline mode does disable schema completely.

What? Can't they be vendored into this repository for example?

ssbarnea:
> Answering [#2729 (comment)](https://github.com/ansible/ansible-lint/issues/2729#issuecomment-1339176974), since the issue is now locked. 😡
> 
> Thanks for the update @ssbarnea. Sorry for "raising my voice", but there have been numerous issues with this "thing" in the latest patch! releases of ansible-lint. I find it very annoying.

It is ok. We all make mistakes and try to do better after. I will release this today. If you want to keep an eye on open pull requests, you might help us get them right from first attempt.

ssbarnea:
> > Please note that offline mode does disable schema completely.
> 
> What? Can't they be vendored into this repository for example?

We considered it but did not make sense, basically because inside a container, like https://github.com/ansible/creator-ee, we never update the packages. We wanted to ensure that someone that uses the linter on CI, github action or another pipeline, can benefit from using the latest schemas, without having to update the linter itself, especially because schemas changes are unlikely to cause any regressions.

MarkusTeufelberger:
Ah, I answered before you added the word "updates" there. This makes more sense of course.

PR #2814, opened 2022-12-13
Add schema for Antsibull changelogs

ssbarnea:


felixfontein:
I hope you are aware that this schema only covers a small subset of things that are allowed in changelogs/changelog.yaml. The full format is specified in https://github.com/ansible-community/antsibull-changelog/blob/main/docs/changelog.yaml-format.md.

ssbarnea:
OMG, not over-engineered at all. That will keep me busy for today.

PR #2832, opened 2022-12-15
Add no-changelog check to galaxy rule

### LINKED ISSUE #2101: no-changelog: check_collection_changelog
check the collection structure to confirm if there is a changelog file present at the root of the directory. 

rule should be optional and only enabled at users discretion. Error level should be Warn/Info (0)

based off of code found [here ](https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/loaders/collection.py#L119)

alisonlhart:
Adding a changelog check to the existing `galaxy` rule. 

Checks for a changelog in the following locations:

- <collection_root>/CHANGELOG.rst
- <collection_root>/CHANGELOG.md
- <collection_root>/changelogs/changelog.yml

If no changelog is found, then the following error is thrown:

```
galaxy[no-changelog]: "No changelog found. Please add a changelog file. Refer to the galaxy.md file for more info."
```

Rule information added to the `galaxy.md` file. 

Resolves #2101


ssbarnea (changes requested):
The correct check is for `changelog.yaml` not `changelog.yml`. That is the only documented filename.

alisonlhart:
@ssbarnea Whoops, fixed!

ssbarnea (approved):

PR #2834, opened 2022-12-16
Complete schema for Antsibull changelogs

shatakshiiii:
Related: #2814


ssbarnea:
@felixfontein Any chance you can check this one? Is the new scheme working as expected?

ssbarnea (changes requested):
Few observations remarked by @felixfontein 👍 
- missing known_issues
- missing trivial category
- I see various markdownDescription fields containing a leading space
- "Value of `short_description from plugin `DOCUMENTATION`." - is missing one backquote
- a lot of duplication, fields that are the same should never be duplicated inside the file and references should be used instead.

src/ansiblelint/schemas/changelog.json
```diff
@@ -18,31 +46,69 @@
...
+              "type": "array"
+            },
             "minor_changes": {
               "items": {
                 "type": "string"
               },
               "type": "array"
             },
             "release_summary": {
+              "markdownDescription": "This must be valid [reStructuredText](https://en.wikipedia.org/wiki/ReStructuredText).",
```
felixfontein, line 68: This also applies to every list entry in the other sections.

src/ansiblelint/schemas/changelog.json
```diff
@@ -57,6 +123,100 @@
...
           "type": "array"
         },
+        "objects": {
+          "additionalProperties": false,
+          "properties": {
+            "playbook": {
+              "items": {
+                "properties": {
+                  "description": {
+                    "markdownDescription": "Value of `short_description` from plugin `DOCUMENTATION`.",
```
felixfontein, line 133: Playbooks are not documentable yet.
```suggestion
                    "markdownDescription": "A short description of what the playbook does.",
```

src/ansiblelint/schemas/changelog.json
```diff
@@ -57,6 +123,100 @@
...
+                },
+                "type": "object"
+              },
+              "type": "array"
+            },
+            "role": {
+              "items": {
+                "properties": {
+                  "description": {
+                    "markdownDescription": "Value of `short_description` from plugin `DOCUMENTATION`.",
```
felixfontein, line 155:
```suggestion
                    "markdownDescription": "Value of `short_description` from role's argument spec.",
```

src/ansiblelint/schemas/changelog.json
```diff
@@ -57,6 +123,100 @@
...
+            "callback": {
+              "$ref": "#/$defs/plugin-descriptions"
+            },
+            "cliconf": {
+              "$ref": "#/$defs/plugin-descriptions"
+            },
+            "connections": {
+              "$ref": "#/$defs/plugin-descriptions"
+            },
+            "filter": {},
```
felixfontein, line 194: Why is this empty?

src/ansiblelint/schemas/changelog.json
```diff
@@ -57,6 +123,100 @@
...
+            "netconf": {
+              "$ref": "#/$defs/plugin-descriptions"
+            },
+            "shell": {
+              "$ref": "#/$defs/plugin-descriptions"
+            },
+            "strategy": {
+              "$ref": "#/$defs/plugin-descriptions"
+            },
+            "test": {},
```
felixfontein, line 213: Same question here.

src/ansiblelint/schemas/changelog.json
```diff
@@ -70,15 +230,14 @@
     "semver": {
       "pattern": "\\d+.\\d+.\\d+.*",
```
felixfontein, line 231: See "Is there a suggested regular expression (RegEx) to check a SemVer string?" on https://semver.org/

felixfontein (commented):
Line 119 should be
```
                "type": "string"
```
`null` isn't allowed for modules. (I cannot add a comment to that line, so writing it here.)

src/ansiblelint/schemas/changelog.json
ssbarnea: The doc did not give any details regarding what filter allows.

src/ansiblelint/schemas/changelog.json
ssbarnea: Same here.

src/ansiblelint/schemas/changelog.json
ssbarnea: JSON specification does only allow very limited set of regex. https://json-schema.org/understanding-json-schema/reference/regular_expressions.html -- very limited!

src/ansiblelint/schemas/changelog.json
ssbarnea: I am not going to copy the text in each section. There is no way to inherit it.

src/ansiblelint/schemas/changelog.json
felixfontein: Exactly the same as every other plugin type.

src/ansiblelint/schemas/changelog.json
felixfontein: I would have assumed you could do something like `"$ref": "#/$defs/changelog-section"` (for a generic 'list' section).

src/ansiblelint/schemas/changelog.json
```diff
@@ -18,45 +46,182 @@
...
           "type": "object"
         },
         "codename": {
           "type": "string"
         },
+        "fragments": {
+          "items": {
+            "type": "string"
+          },
+          "markdownDescription": "List of strings representing filenames with framents.",
```
felixfontein, line 100:
```suggestion
          "markdownDescription": "List of strings representing filenames of changelog fragments.",
```

src/ansiblelint/schemas/changelog.json
```diff
@@ -18,45 +46,182 @@
...
               "name": {
-                "markdownDescription": " It must not be the FQCN, but the name inside the collection.",
+                "markdownDescription": "It must not be the FQCN, but the name inside the collection.",
                 "pattern": "[a-zA-Z0-9_]+",
                 "title": "Short module name",
                 "type": "string"
               },
               "namespace": {
-                "markdownDescription": "Must be `null` for plugin and objects. For modules it must be either empty string. The namespace is used to group new modules by their namespace inside the collection.",
+                "markdownDescription": "Must be `''` for modules directly in `plugins/modules/`, or the dot-separated list of directories the module is in inside the `plugins/modules/` directory. This is mostly relevant for large collections such as `community.general` and `community.network`. For example, the `community.general.docker_container` module is in the directory `plugins/modules/cloud/docker/`, hence its namespace must be `cloud.docker`. The namespace is used to group new modules by their namespace inside the collection.",
```
felixfontein, line 118: docker_container hasn't been in community.general for years, and community.general no longer has hat mentioned directory structure for two months. (I see where you copied this from, I guess I have to adjust that description :) )

src/ansiblelint/schemas/changelog.json
```diff
@@ -18,45 +46,182 @@
...
                 "title": "Namespace",
-                "type": ["string", "null"]
+                "type": ["string"]
```
felixfontein, line 120: Shouldn't this be
```suggestion
                "type": "string"
```
? (I guess both work, but this is the syntax you use in other cases.)

src/ansiblelint/schemas/changelog.json
felixfontein: Ref: https://github.com/ansible-community/antsibull-changelog/pull/95

felixfontein (approved):

felixfontein:
@shatakshiiii @ssbarnea thanks for improving the schema!

PR #2842, opened 2022-12-21
src/ansiblelint/schemas/meta.json: add openSUSE 15.5

johanneskastl:


johanneskastl:
Looks like someone has to add a label for the checks to succeed. I seem to have no permission to do so.

PR #2860, opened 2023-01-01
Make git a runtime dependency

ssbarnea:
As we received numerous bug reports which proved to be caused by
the missing git, we make git a permanent runtime dependency. This will
make the tool behavior more consistent as before that behavior would
have being different if git was missing.

### DIFF
--- a/.config/ansible-lint.spec
+++ b/.config/ansible-lint.spec
@@ -25,6 +25,7 @@ BuildRequires:  python%{python3_pkgversion}-pytest-xdist
 BuildRequires:  python%{python3_pkgversion}-libselinux
 BuildRequires:  git-core
 %endif
+Requires:       git-core
 
 
 %description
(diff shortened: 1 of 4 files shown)


fclaerho:
Hi @ssbarnea !

I just hit this weird case of installing innocently ansible-lint in a minimal python docker image to lint a playbook, and having it fail flat with this:
```
# pipenv run ansible-lint -vv --parseable --skip-list fqcn-builtins,unnamed-task,key-order playbook.yml
…
Failed to find runtime dependency 'git' in PATH
```

I've traced back this error to this issue, which make "git" a required dependency of ansible-lint, a change very strange in itself (ansible-lint being a Python tool initially, now we have a non-python dependency -- the above ticket description refers to git-related issues, could this be detailed?) — and as importantly it's simply not documented anywhere in the installation page (and indeed the change list above only covers the code):
https://ansible-lint.readthedocs.io/installing/

Could this new requirement be added to the doc?

ssbarnea:
Hi! Shotly, there is no need to add it to the docs because we have an incoming change that will remove the need for `git` executable.

fclaerho:
Good to know, thanks for the info 👍🏻

PR #2876, opened 2023-01-09
Avoid caching on role_name regex (#2876)Co-authored-by: Sorin Sbarnea <ssbarnea@redhat.com>Co-authored-by: Sorin Sbarnea <ssbarnea@redhat.com>

DanInProgress:
Previously, when the `role_name` rule matched a given role, it would fail to match during future runs from within the same python process.

This change avoids this by caching the results of the regex `match` call, but does not suppress the output for subsequent matches.

This will make use of the `--progressive` option and `ansiblelint`'s python api more predictable, atleast in this scenario.


DanInProgress:
Sorry about the flurry of notifications, I spend more of my time lately on gitlab (where the prefix `Draft: ` is the signifier for draft status), and didn't realize that I hadn't *properly* marked this as draft.

This now should be ready (pending a `label` from a maintainer)

--also, loving the pre-commit.ci bot, I'll have to start using it for my projects

ssbarnea:
@DanInProgress The convert to draft button is very well hidden but as hint for you, use "find in page with drafter" and you will find the link. I will do the conversion for you, but I mentioned this for the future. It is very easy to switch between draft and non-draft and that has no side-effects, it does not even trigger a rebuild.

ssbarnea (approved):

Issue #2869, opened 2023-01-04
ansible-lint hangs indefinitely on "DEBUG    Refreshing ansible-lint-config schema ..."


Andeye:
If there are network issues, ansible-lint hangs indefinitely with no output. The reason can only be found out in debug mode.

The issue is here:
https://github.com/ansible/ansible-lint/blob/d74ef7d6d76b081f07ff50fa23a75e90eb446b6b/src/ansiblelint/schemas/main.py#L102

And here is the relevant documentation:
> You can tell Requests to stop waiting for a response after a given number of seconds with the `timeout` parameter. Nearly all production code should use this parameter in nearly all requests. Failure to do so can cause your program to hang indefinitely:
- https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts

p.s. Adding the `--offline` flag can circumvent this issue.

ssbarnea:
Default DNS timeout is 30s so, I would not call that "indefinitely". Also we stop processing all schemas on first error, so the final delay is not multiplied by the number of files to be checked.

Andeye:
Hello and thank you for your quick response!

I left it running for 18 minutes and nothing happened, and while 18 minutes is still quite far from "indefinitely", I think we can both agree that that is not ideal either.
"Indefinitely" was also the word used by the authors of the `requests` package, but I now noticed that `Request` that you are using is from the `urllib` package and not from the `requests` package. My bad!

However, I did some additional testing and noticed that if I just unplug my computer, the default timeout you mentioned works as expected, and even faster than 30s. My problems arise when I run ansible-lint in a Docker container **and** while connected to VPN. I've had problems with Docker + VPN before on my computer, e.g. `apt` is not working either in Debian based images, and I haven't been able to figure out what causes this and how to fix it. Might be something with the configuration of the corporate VPN, but that is outside of my control.

But I did a quick test and replaced:
https://github.com/ansible/ansible-lint/blob/d74ef7d6d76b081f07ff50fa23a75e90eb446b6b/src/ansiblelint/schemas/main.py#L107
with:
```python
            with urllib.request.urlopen(request, timeout=3) as response:
```
and then it worked fine.

ssbarnea:
@Andeye please make a PR to fix this bug, I am sure others will be happy. I think that 10s should be a safe default.

Andeye:
[...]

PR #2895, opened 2023-01-16
Fix infinite hang on schema refresh by adding timeout parameter

Andeye:
Under some special network conditions, such as running inside a Docker container and over a VPN, the schema refresh can hang indefinitely. This is fixed by adding a timeout parameter.

https://github.com/ansible/ansible-lint/issues/2869


ssbarnea:
@Andeye Looks good very few fixed needed to get it full free.

ssbarnea (approved):

PR #2897, opened 2023-01-16
Fix 'location' attribute for CodeclimateJSONFormatter

4ch1m:
Hi there,

I believe the current implementation of the [CodeclimateJSONFormatter](https://github.com/ansible/ansible-lint/blob/dc42168472d0c32a23df8eef0f3c17583e3ce5ee/src/ansiblelint/formatters/__init__.py#L130) is not quite correct.

Please have a look at this block ...

https://github.com/ansible/ansible-lint/blob/dc42168472d0c32a23df8eef0f3c17583e3ce5ee/src/ansiblelint/formatters/__init__.py#L163-L171

... which **always** generates a `lines`-object.

Furthermore the `begin`-attribute **changes type** (to either `int` or `object`)  depending of the presence of a column-value. (Which makes automatic object-binding extremely complicated if not impossible when parsing the JSON-string with other languages/frameworks.)

Aside from that, the [codeclimate docs](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#user-content-data-types) state:

> Locations refer to ranges of a source code file. A Location contains a path and a source range (expressed as lines or positions).

So it should **either** be a `lines`-object ...
```
{
  "path": "path/to/file.css",
  "lines": {
    "begin": 13,
    "end": 14
  }
}
```
... with a begin/end range, **or** a `positions`-object ...
```
{
  "path": "path/to/file.css",
  "positions": {
    "begin": {
      "line": 3,
      "column": 10
    },
    "end": {
      "line": 4,
      "column": 12
    }
  }
}
```
... with line/column values.

This PR should fix that.

Thanks for reading. :smile:


ssbarnea:
@4ch1m Update PYTEST_REQPASS to include the test you added. Other than this it looks good to me. Thanks!

4ch1m:
> @4ch1m Update PYTEST_REQPASS to include the test you added. Other than this it looks good to me. Thanks!

Done. Thank you. :thumbsup:

ssbarnea (approved):

PR #2902, opened 2023-01-17
Added tests for text.py

ajinkyau:
Signed-off-by: Ajinkya Udgirkar <audgirka@redhat.com>

### DIFF
--- a/test/test_text.py
+++ b/test/test_text.py
@@ -3,15 +3,15 @@ from typing import Any
 
 import pytest
 
-from ansiblelint.text import strip_ansi_escape, toidentifier, has_glob, has_jinja
+from ansiblelint.text import has_glob, has_jinja, strip_ansi_escape, toidentifier
 
 
 @pytest.mark.parametrize(
     ("value", "expected"),
     (
         pytest.param("\x1b[1;31mHello", "Hello", id="0"),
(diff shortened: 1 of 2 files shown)


ssbarnea (changes requested):
While in good direction, it shill does not increase coverage of the two lines we needed to cover.

You need to pass a bytes object to do this, not a string.

For the second one, you want to pass some data that would force it to raise and Error, as that was the execution path that was not covered.

test/test_text.py
```diff
@@ -3,7 +3,36 @@
...
+        pytest.param("foo--bar", "foo_bar", id="1"),
+        pytest.param("example_test.zip", "example_test.zip", id="2"),
+    ),
+)
+def test_toidentifier(value: Any, expected: str) -> None:
+    """Tests for toidentifier()."""
+    try:
+        assert toidentifier(value) == expected
+    except RuntimeError as err:
+        assert isinstance(err, RuntimeError)
```
ssbarnea, line 35: Please use pytest.raises instead

ssbarnea (changes requested):

test/test_text.py
ajinkyau: done

ssbarnea (approved):

Issue #2908, opened 2023-01-17
meta-unsupported-ansible rule produces extraneous output


kpfleming:
https://github.com/ansible/ansible-lint/blob/4db702e9b50f0e7b8b7ce70a2e8f8dfd92920a57/src/ansiblelint/rules/meta_unsupported_ansible.py#L46

This `print()` should not be present :-)

kpfleming:
It's also not clear that the logic in this rule is correct; it will pass if at least one supported version is listed in `requires_ansible`, even if it also includes unsupported versions. It will only fail if *all* of the versions in `requires_ansible` are unsupported.

PR #2913, opened 2023-01-18
Fix print leak with meta-unsupported-ansible

ssbarnea:
Fixes: #2908


cidrblock (approved):
+10000000000000 for the flake-print add

kpfleming:
Wow, `print` was all over the place!

PR #2917, opened 2023-01-19
Makes empty-playbook warning non fatal

nkakouros:
This closes #2868.

Issue #2868, opened 2023-01-04
Warning is fatal and stops further rule checks


nkakouros:
##### Summary

A `warning[empty-playbook]` warning is fatal, when it should be just that, a warning.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.10.2 using ansible 2.14.1
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

I have a simple role, with 2 files:

```
# myrole/tasks/main.yml
- debug: ~
```

```
myrole/molecule/default/molecule.yml
---
```

The molecule file is empty because I use a global one for all my roles.

##### Desired Behavior

The empty playbook warning should just be listed as such and I should receive an error about the unnamed tasks in tasks/main.yml and other errors.

##### Actual Behavior

```paste below
(.venv) [nikos@instance-2 myrole]$ ansible-lint molecule/default/moelcule.yml tasks/main.yml
WARNING  Listing 1 violation(s) that are fatal
warning[empty-playbook]: Empty playbook, nothing to do (warning)
molecule/default/molecule.yml:1:1

You can skip specific rules or tags by adding them to your configuration file:
# .config/ansible-lint.yml
warn_list:  # or 'skip_list' to silence them completely
  - experimental  # all rules tagged as experimental

                       Rule Violation Summary
 count tag                     profile rule associated tags
     1 warning[empty-playbook]         core, experimental (warning)

Passed with production profile, 5/5 star rating: 0 failure(s), 1 warning(s) on 2 files.
```

The empty playbook is a warning. I don't get why it is fatal! 

<del>The only way to make ansible-lint go further is to add `warning[empty-playbook]` to `skip_list`. </del>
There is not way to ignore this fatal error.

I don't know why molecule.yml is treated as a playbook and I would be happy to learn how to instruct ansible-lint to not do that, but regardless of this, it seems wrong that a warning could be fatal.

isuftin:
.ansible-lint

```yaml
exclude_paths:
  - ./molecule
  - ./whatever_molecule_playbooks_to_ignore.yml
```

nkakouros:
But I still want molecule.yml to be included in any further ansible-lint checks (as well as yamllint through ansible-lint). The important part of my issue, though, is why warning is fatal.

nkakouros:
[...]

PR #2925, opened 2023-01-20
Allow warnings to be skipped

ssbarnea:
Fixes: #2868
Closes: #2917

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -69,7 +69,7 @@ jobs:
       WSLENV: FORCE_COLOR:PYTEST_REQPASS:TOXENV:GITHUB_STEP_SUMMARY
       # Number of expected test passes, safety measure for accidental skip of
       # tests. Update value if you add/remove tests.
-      PYTEST_REQPASS: 769
+      PYTEST_REQPASS: 770
 
     steps:
       - name: Activate WSL1
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -281,6 +281,9 @@ def main(argv: list[str] | None = None) -> int:  # noqa: C901
     if options.strict and result.matches:
         mark_as_success = False
 
+    # Remove skipped list items from the result
+    result.matches = [m for m in result.matches if m.tag not in app.options.skip_list]
+
     app.render_matches(result.matches)
 
     _perform_mockings_cleanup()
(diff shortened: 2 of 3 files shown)


cidrblock (approved):

nkakouros:
@ssbarnea I don't think this is the right fix. What this PR does is not output the error about the empty-playbook if the rule is excluded. Still, `warning[emtpy-playbook]` remains fatal and no further checks are done, which should not be the case given that an empty playbook really is not fatal. For example, given the files below:

```
# vars/main.yml
q: 1
```

```
# playbooks/main.yml
---
```

if I run `ansible-lint -x 'warning[empty-playbook]' vars playbooks`, I get:

```

Passed with production profile: 0 failure(s), 0 warning(s) on 6 files.
You are using a pre-release version of ansible-lint.
```

`ansible-lint` never reaches the point where it checks that vars file (to complain about a missing doc start).

And there is also the question if a `warning` should be considered as `fatal error` by default (I argue it should not).

ssbarnea:
You are confusing the special rune named "warning" with the warning level. They are two different things, and the warning meta-rule can act as a warning or error (levels).

nkakouros:
There are 3 issues here and I think you would agree that there is sth at least inconsistent if not wrong. It would be great to hear your view on these 3 points.

1. It is not me that confuses the two. The [docs](https://ansible-lint.readthedocs.io/rules/warning/) say:

> warning is a special type of internal rule that is used to report generic runtime warnings found during execution. As stated by its name, they are not counted as errors, so they do not influence the final outcome.

So, `warning` rules are about warnings according to the docs

2. Regardless of naming, do you think it is fine for this rule to be fatal and stop execution of further rules? 

`warning[empty-playbook]` breaks further execution because it relies on ansible-playbook's `--syntax-check` behavior where an empty playbook is fatal. But even ansible-playbook can run an empty playbook without issue (`ansible-playbook path/to/empty/playbook` works fine and further playbooks in the command line are run without issue).

There is nothing really other than the above that prevents `ansible-lint` from continuing to check files, there is no for example bad yaml where ansible-lint will not know how to parse it. This is apparent in [my PR](https://github.com/ansible/ansible-lint/pull/2917) where I modified the code to not follow `ansible-playbook --check` by making that warning non-breaking, but still reporting it.

3. Especially with this PR here, if the user skips `warning[empty-playbook]`, they may see:

```
Passed with production profile: 0 failure(s), 0 warning(s) on 6 files.
```

which is misleading. Ansible-lint has not actually checked 6 files and found them ok, it just stopped after the empty playbook and did not applied most of the rest of the rules.

ssbarnea:
Please create a new issue for this and we can work next week to address it. It might need multiple changes but that is not a problem, we should fix it and probably do it will also involve avoiding using "warning" as a rule name.

nkakouros:
I opened #2927.

Issue #2824, opened 2022-12-15
args[module] is confused when there is a module and an action plugin


evgeni:
##### Summary

When there is an action plugin and a module by the same name, Ansible calls the action plugin which then will call the module internally.

These two can, however, have different parameters, like we see in the `async_status` combo. The action plugin has two: `jid` and `mode`, while the module has a third, hidden one: `_async_dir`.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console 
% ansible-lint --version     
ansible-lint 6.10.0 using ansible 2.14.1
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

This is the example from `ansible-doc async_status`:

```yaml
---
- name: Lol
  hosts: localhost
  tasks:
    - name: Asynchronous yum task
      ansible.builtin.yum:
        name: docker-io
        state: present
      async: 1000
      poll: 0
      register: yum_sleeper

    - name: Wait for asynchronous job to end
      ansible.builtin.async_status:
        jid: '{{ yum_sleeper.ansible_job_id }}'
      register: job_result
      until: job_result.finished
      retries: 100
      delay: 10
```

```console
% ansible-lint l2.yml   
WARNING  Overriding detected file kind 'yaml' with 'playbook' for given positional argument: l2.yml
WARNING  Listing 1 violation(s) that are fatal
args[module]: missing required arguments: _async_dir (warning)
l2.yml:13 Task/Handler: Wait for asynchronous job to end

You can skip specific rules or tags by adding them to your configuration file:
# .config/ansible-lint.yml
warn_list:  # or 'skip_list' to silence them completely
  - experimental  # all rules tagged as experimental

                  Rule Violation Summary                   
 count tag          profile rule associated tags           
     1 args[module]         syntax, experimental (warning) 

Passed with production profile, 5/5 star rating: 0 failure(s), 1 warning(s) on 1 files.
```

##### Desired Behavior

`args[module]` should not be raised

##### Actual Behavior

```
args[module]: missing required arguments: _async_dir (warning)
```

evgeni:
A similar issue happens with `ansible.builtin.service`.
The doc of `ansible.builtin.service` says: "This module acts as a proxy to the underlying service manager module. While all arguments will be passed to the underlying module, not all modules support the same arguments."

The following invocation works just fine (`daemon_reload` is accepted and passed to `systemd`):
```yaml
- name: Restart rhcd
[...]

PR #2931, opened 2023-01-23
bug: Ignore non-existing missing required argument _async_dir

klaus-tux:
This fixes the initial problem reported in #2824 .


ssbarnea:
Sorry but that is not the correct way of fixing that bug (special ignore case).

PR #2948, opened 2023-01-28
Remove zuul-jobs and tripleo-ansible from eco pipelines

ssbarnea:
Removing zuul-jobs and tripleo-ansible as the maintainers of these
repositories did not show active interest in updating their codebase
based on ansible-lint updates.


cidrblock:
+1, ty

Issue #2977, opened 2023-02-03
version 6.12.0 ignores noqa in playbook?


psvmcc:
##### Summary

Failure with noqa set in playbook. Works on 6.11.0, and failed on 6.12.0.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 6.12.0 using ansible 2.12.10
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

```console (paste below)
---
- name: Upgrade host
  hosts: all
  become: true
  gather_facts: false
  tasks:
    - name: Install updates
      ansible.builtin.dnf:
        name: '*'
        state: latest  # noqa package-latest
        update_cache: true
```

##### Desired Behavior

With ansible-lint==6.11.0
```console
$ ansible-lint test.yaml
WARNING  Overriding detected file kind 'yaml' with 'playbook' for given positional argument: test.yaml

Passed with production profile: 0 failure(s), 0 warning(s) on 1 files.
```

##### Actual Behavior
With ansible-lint==6.12.0
```console
$ ansible-lint test.yaml
WARNING  Ignore loading rule from /Users/ps/a/.venv/lib/python3.11/site-packages/ansiblelint/rules/galaxy.py due to No module named 'pytest'
WARNING  Listing 1 violation(s) that are fatal
package-latest: Package installs should not use latest.
test.yaml:7 Task/Handler: Install updates

You can skip specific rules or tags by adding them to your configuration file:
# .config/ansible-lint.yml
warn_list:  # or 'skip_list' to silence them completely
  - package-latest  # Package installs should not use latest.

              Rule Violation Summary
 count tag            profile rule associated tags
     1 package-latest safety  idempotency

Failed after moderate profile, 2/5 star rating: 1 failure(s), 0 warning(s) on 1 files.
```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

apatard:
Please update to 6.12.1 and close the bug if it's solved.
See https://github.com/ansible/ansible-lint/releases/tag/v6.12.1:
```
Allow noqa comments without colons (https://github.com/ansible/ansible-lint/pull/2971) @ssbarnea
```

psvmcc:
No, I see same error with 6.12.1 version :(

```
$ ansible-lint --version
ansible-lint 6.12.1 using ansible 2.12.10
You are using a pre-release version of ansible-lint.


[...]

PR #2984, opened 2023-02-05
src/ansiblelint/file_utils.py: Fix playbook/rulebook detection

apatard:
Don't try detecting the file type unless file already loaded in Lintable._data, otherwise the line numbers for inline rule skipping will be set to the line where's it's used since self.kind is 'yaml'.
In case of a playbook, after _guess_kind() call, the kind will be set to playbook and the error line will be the one of the task, which is possibly different that the one with the 'noqa:' comment.

Fixes: #2977

### DIFF
--- a/src/ansiblelint/file_utils.py
+++ b/src/ansiblelint/file_utils.py
@@ -242,7 +242,7 @@ class Lintable:
         self.abspath = self.path.expanduser().absolute()
 
         if self.kind == "yaml":
-            self._guess_kind()
+            self.data
 
     def _guess_kind(self) -> None:
         if self.kind == "yaml":
@@ -371,6 +371,11 @@ class Lintable:
                     )
 
                     self._data = parse_yaml_linenumbers(self)
+                    # now that _data is not empty, we can try guessing if playbook or rulebook
+                    # it has to be done before append_skipped_rules() call as it's relying
+                    # on self.kind.
+                    if self.kind == "yaml":
+                        self._guess_kind()
                     # Lazy import to avoid delays and cyclic-imports
                     if "append_skipped_rules" not in globals():
                         # pylint: disable=import-outside-toplevel


src/ansiblelint/file_utils.py
```diff
@@ -242,7 +242,7 @@ def __init__(
         self.abspath = self.path.expanduser().absolute()
 
         if self.kind == "yaml":
-            self._guess_kind()
+            self.data
```
CarstenGrohmann, line 245: I'm only here by chance and don't know much about the internals of ansible-lint, but this statement is meaningless from my point of view.

apatard: But it is not. The self.data function is "filling" self._data at the first call and is returning self._data the next calls.

CarstenGrohmann: Is this a function and need `()` or does it use `__getattr__()`?

ssbarnea: The only reason why I did not merge this fix yet is because it does not update the tests. We need a reproducer for original bug as a test in order to ensure that next refactoring does not bring the bug back.

If you do not have time to write it, I will probably try to push the test myself.

apatard: Oops. Sorry for that. I'll try to write it before the end of the week.

ssbarnea:
Closing this because the original reported bug was already fixed and this change did not refer to any other bug.

apatard:
@ssbarnea afaik, no, the bug is not solved. The issue 2977 is not about noqa vs noqa:. It's about playbook being first detected as yaml and then detected as playbook. Leading to wrong file lines used to determine if the rule should be fixed or not.  Just test the reproducer given in 2977 and you'll see it should still be valid.

ssbarnea:
@apatard I tried the mentioned example and it worked fine for me. That example was affected by two bugs that were fixed on main. We can clearly reopen it once we find a way to reproduce it on main branch.

apatard:
Not sure how you managed to not reproduce it:

```
$ pip3 freeze|grep ansible_lint
-e git+https://github.com/apatard/ansible-lint@26a12aa70f949a02287ac56530d04f045ec07a81#egg=ansible_lint
$ cat > test.yml <<EOF
---
- name: Upgrade host
  hosts: all
  become: true
  gather_facts: false
  tasks:
    - name: Install updates
      ansible.builtin.dnf:
        name: '*'
        state: latest  # noqa: package-latest
        update_cache: true
EOF
$ ansible-lint test.yml 
package-latest: Package installs should not use latest.
test2.yml:7 Task/Handler: Install updates

# .config/ansible-lint.yml
  - package-latest  # Package installs should not use latest.

              Rule Violation Summary               
 count tag            profile rule associated tags 
     1 package-latest safety  idempotency          

Failed after moderate profile, 2/5 star rating: 1 failure(s), 0 warning(s) on 1 files.
```

apatard:
fwiw, I've just updated my branch (https://github.com/apatard/ansible-lint/tree/bug2977)  with a test. Can you please have a look so that I know if opening a new PR is a waste of time or not.

ssbarnea:
Open a PR as draft and I will look at it tomorrow. I cannot reopen this one due to the forced push.

apatard:
ok. Done.

Issue #1584, opened 2021-05-24
Configurable violation ignore list file


ssbarnea:
##### Summary

I would like to implement a configurable option for silencing specific errors similar to the one used by flake8, where you can acknowledge and ignore specific errors inside specific files, even mentioning the line numbers of you want.

If we do this we may even avoid the need to add YAML comments with noqa, which can be really spammy and only record exceptions in ignore.txt files, very similar with what ansible-test does now.

##### Details

While initially this can be implemented as a feature, it does open the door of removing ruyaml loading completely and rely on pyyaml alone, simplifying the corebase considerably.

Still, decision about if we should remove noqa comments or not should require a wider community consultation.

Example from flake8 config:
```
per-file-ignores =
  docs/conf.py: D
  src/molecule/test/*: D100,D103,D104
```
As one can see this allows to disable rules on files and patterns, being very easy to maintain.

Please use thumbs up down or comments to provide feedback on this proposal.

ssbarnea:
I was looking at this proposal and trying to find a good way to represent these ignores inside YAML as we don't want to parse the strings to implement it like pylint. I came with 3 ideas:

```
# new feature file rule skips (V1)
skips:
  git-latest:
    - "*.*"

# V2
skips:
  "*.*": git-latest

# V3 not the exclude_paths already exists at root level and has the same format, this one has
# the advantage of allowing use to add other rule specific configuration in the future.
rules:
  git-latest:
    exclude_paths:
      - "*.*"
```

WDYT?

cidrblock:
seems sane to me, does this leave the door open for subrule disable support?

```rules:
  name:
    exclude_paths:
      - "*.*"
    subrules:
      - block: false
      - import_play: true #default
      - tasks: true
      - imperative: false
 ```

thaumos:
I like v3, that seems the cleanest and easily legible as to what is being accomplished.

nogweii:
I've run into needing this a couple of times in my own infrastructure. I've been able to work around it by extracting the relevant bits of YAML into a separate file and excluding it using `exclude_paths`. However, I feel like that disables too much.

The V3 proposal is very nice and would work for my use case, being able to disable only certain violations on a file rather than all of them.

ssbarnea:
For the moment we will not deprecate any inline noqa.

PR #3004, opened 2023-02-09
Load rule ignores from external text file

ssbarnea:
Ansible-lint will load skip rules from `.ansible-lint-ignore` file that is adjacent to its config file. The file format is very simple, containing the filename and the rule to be ignored. It also supports comments starting with `#`.

- Adds ability to load ignores from file
- Implements `--generate-ignore` for dumping ignores for current violations
- Adds documentation about ignores
- Change summary message to point to documentation for ignores
- This feature also deprecates progressive mode due to its performance and testing problems, like testing of pull-requests with multiple commits.

Fixes: #1584
Fixes: #3012
Closes: #2934
Related: https://github.com/ansible/ansible-lint/issues/2880

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -44,7 +44,13 @@ use_default_rules: true
 # rulesdir:
 #   - ./rule/directory/
 
-# Ansible-lint completely ignores rules or tags listed below
+# Ansible-lint is able to recognize and load skip rules stored inside
+# `.ansible-lint-ignore` files. To skip a rule just enter filename and tag,
+# like "playbook.yml package-latest" on a new line. Optionally you can add
+# comments after the tag, prefixed by "#". We discourage the use of skip_list
+# below because that will hide violations from the output. When putting ignores
+# inside the ignore file, they are marked as ignored, but still visible, making
+# it easier to address later.
 skip_list:
   - skip_this_tag
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -38,6 +38,22 @@ counterparts:
 --8<-- ".ansible-lint"
 ```
 
+## Ignoring rules for entire files
+
+Ansible-lint will load skip rules from `.ansible-lint-ignore` file that is
+adjacent to its config file. The file format is very simple, containing the
+filename and the rule to be ignored. It also supports comments starting with
+`#`.
+
+```yaml title=".ansible-lint-ignore"
+# this is just a comment
+playbook.yml package-latest # disable package-latest rule for playbook.yml
+playbook.yml deprecated-module
+```
+
+The file can also be created by adding `--generate-ignore` to the command line.
+Keep in mind that this will override any existing file content.
+
 ## Pre-commit setup
 
 To use Ansible-lint with [pre-commit], add the following to the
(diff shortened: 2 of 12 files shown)


apatard:
Typo in the commit message title ? :)

.ansible-lint
```diff
@@ -44,7 +44,10 @@ use_default_rules: true
 # rulesdir:
 #   - ./rule/directory/
 
-# Ansible-lint completely ignores rules or tags listed below
+# Ansible-lint is able to recognize and load skip rules stored inside
+# `tests/sanity/ignore.txt` files (same as ansible-test). To skip
```
mattclay, line 48: Since this is unrelated to the `ansible-test` `sanity` test, I don't think it makes sense to put it in the `tests/sanity/` directory.

For the same reason, naming the file `ignore.txt` (especially in the directory used by `ansible-test`) could be confusing to users, since the name is so close to the one used by `ansible-test`. There's also the potential for further confusion since the file formats are different.

docs/configuring.md
```diff
@@ -38,6 +38,18 @@ counterparts:
 --8<-- ".ansible-lint"
 ```
 
+## Ignoring rules for entire files
+
+Ansible-lint will load skip rules from `tests/sanity/ignore.txt` file, if one is
+detected, the same as
+[ansible-test](https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/ignores.html).
```
mattclay, line 45: Since this feature is unrelated to `ansible-test`, I don't think it makes sense to reference it here. Not only is the behavior different, but as mentioned above, the file format is different also.

mattclay:
@ssbarnea Can you clarify what you mean by this part of the summary?

> This will also make it easier to integrate results
> from `ansible-test sanity` directly into ansible-lint, as we will store the ignores into a single location.

mattclay (changes requested):

cidrblock:
Related: https://github.com/ansible/ansible-lint/issues/3012

.ansible-lint
mattclay: The references to `tests/sanity/ignore.txt` and `ansible-test` should be removed.

mattclay (changes requested):

src/ansiblelint/loaders.py
```diff
@@ -24,4 +29,29 @@ def yaml_from_file(filepath: str | Path) -> Any:
...
+    """Return a list of rules to ignore."""
+    result = defaultdict(set)
+    if os.path.isfile(filepath):
+        with open(str(filepath), encoding="utf-8") as content:
+            _logger.debug("Loading ignores from %s", filepath)
+            for line in content:
+                entry = line.split("#")[0].rstrip()
+                if entry:
+                    try:
+                        path, rule = entry.split()
```
cidrblock, line 42: if path.endwith(':'), path = path[0:-1] ?

cidrblock (approved):

mattclay (commented):
Now that ansible-test references have been removed, I'll leave the reviewing to the regular ansible-lint contributors.

src/ansiblelint/loaders.py
ssbarnea: Nope, only spaces supported for now. No plan to add unneeded complexity.

Issue #3008, opened 2023-02-10
XDG_CONFIG_CACHE should be XDG_CACHE_HOME


gangefors:
##### Summary

In the current source XDG_CONFIG_CACHE is used as env.var. for the cache directory. This is not a valid XDG environment variable. Instead XDG_CACHE_HOME should be used. See https://wiki.archlinux.org/title/XDG_Base_Directory

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 6.12.1 using ansible 2.14.2
```

Introduced in #2615
https://github.com/ansible/ansible-lint/blob/2df8d95748ef1f17d8719da72d30799e9736ed3b/src/ansiblelint/config.py#L25-L27

PR #3009, opened 2023-02-10
Use XDG_CACHE_HOME instead of XDG_CONFIG_CACHE

gangefors:
XDG_CONFIG_CACHE does not exist in the XDG Base directory listing.
XDG_CACHE_HOME is the correct env.var. to use for cache directories.

Reference: https://wiki.archlinux.org/title/XDG_Base_Directory
Fixes: #3008

### DIFF
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -23,7 +23,7 @@ _logger = logging.getLogger(__name__)
 
 
 CACHE_DIR = (
-    os.path.expanduser(os.environ.get("XDG_CONFIG_HOME", "~/.cache")) + "/ansible-lint"
+    os.path.expanduser(os.environ.get("XDG_CACHE_HOME", "~/.cache")) + "/ansible-lint"
 )
 
 DEFAULT_WARN_LIST = [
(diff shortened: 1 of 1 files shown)


src/ansiblelint/config.py
```diff
@@ -23,7 +23,7 @@
 
 
 CACHE_DIR = (
-    os.path.expanduser(os.environ.get("XDG_CONFIG_CACHE", "~/.cache")) + "/ansible-lint"
+    os.path.expanduser(os.environ.get("XDG_CONFIG_HOME", "~/.cache")) + "/ansible-lint"
```
apatard, line 26: What about $XDG_CACHE_HOME ?
Ref: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html:
"$XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used."

gangefors: Yeah. I had it right in the first commit subject but managed to mess up all the rest.
Should be XDG_CACHE_HOME everywhere now.

ssbarnea (approved):

ssbarnea:
Thanks for spotting this one!

apatard (approved):
thanks for noticing and fixing

PR #3014, opened 2023-02-11
Increase the test coverage

ssbarnea:


ssbarnea:
@cidrblock look at this for the two variables that you need to pass for the coverage to work with subprocesses.

Issue #2927, opened 2023-01-21
`warning[empty-playbook]` is fatal and stops further rule checks


nkakouros:
##### Summary

The [docs](https://ansible-lint.readthedocs.io/rules/warning/) say:

> warning is a special type of internal rule that is used to report generic runtime warnings found during execution. As stated by its name, they are not counted as errors, so they do not influence the final outcome.

So, `warning` rules are about warnings according to the docs and they should not be stopping further execution of other rules.

However, `warning[empty-playbook]` forces ansible-lint to exit before most other rules are processed.

This is because `ansible-lint` relies on ansible-playbook's `--syntax-check` behavior where an empty playbook is fatal. But even ansible-playbook can run an empty playbook without issue (`ansible-playbook path/to/empty/playbook` works fine and further playbooks in the command line are run without issue).

There is nothing really other than the above that prevents `ansible-lint` from continuing to check files, there is no for example bad yaml where ansible-lint will not know how to parse it. This is apparent in [this PR](https://github.com/ansible/ansible-lint/pull/2917) where I modified the code to not follow `ansible-playbook --check` by making that warning non-breaking, but still reporting it and still processing further rules.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
nsible-lint 6.11.1.dev12 using ansible 2.14.1
You are using a pre-release version of ansible-lint.
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Consider the following files:

```
# vars/main.yml
q: 1
```

```
# playbooks/main.yml
---
```

Run: `ansible-lint  -x 'warning[empty-playbook]' playbooks vars`.

##### Desired Behavior

All errors are displayed.

##### Actual Behavior

The doc start error is not emitted. Instead, ansible-lint displays just:

```
Passed with production profile: 0 failure(s), 0 warning(s) on 6 files.
You are using a pre-release version of ansible-lint.
```

See also: #2868 #2917 #2925

nkakouros:
@ssbarnea How should we proceed with this? Do you need me to do anything?

PR #3019, opened 2023-02-12
Move empty-playbook to syntax-check

ssbarnea:
Closes: #2927

### DIFF
--- a/src/ansiblelint/_internal/warning.md
+++ b/src/ansiblelint/_internal/warning.md
@@ -4,7 +4,6 @@
 runtime warnings found during execution. As stated by its name, they are not
 counted as errors, so they do not influence the final outcome.
 
-- `warning[empty-playbook]` is raised when a playbook file has no content.
 - `warning[raw-non-string]` indicates that you are using
   `[raw](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/raw_module.html#ansible-collections-ansible-builtin-raw-module)`
   module with non-string arguments, which is not supported by Ansible.
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -36,7 +36,6 @@ DEFAULT_WARN_LIST = [
     "name[play]",
     "name[prefix]",
     "role-name",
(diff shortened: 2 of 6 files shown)


nkakouros:
If I understand correctly, with this PR:

- it is not possible to skip the checking of the `empty-playbook` rule (despite the test):
  ```
  > ansible-lint -x 'syntax-check[empty-playbook]' plays/
  Rule 'syntax-check' is unskippable, you cannot use it in 'skip_list' or 'warn_list'. Still, you could exclude the file.
  ```

- ansible-lint will fail to process further files if an empty playbook is encountered (following the behavior of `ansible-playbook --syntax-check` but not `ansible-playbook`)

The only thing that has improved for the user is the communication that this is not a warning anymore, but syntax check. The user still has to either remove the empty playbook or somehow exclude it from ansible-lint.

To me, the proper solution is to report but not fail on empty playbooks as there is **absolutely no reason** to do so other than following what `--syntax-check` does. But in any case, either of the two bullets should be addressed.

ssbarnea:
@nkakouros Ansible reports fatal error (return code 4) on these "empty playbooks", so linter will be have the same behavior.

The warning approach we use was more of a hack because until recently the linter assumed that any given argument is a playbook, which no longer does since https://github.com/ansible/ansible-lint/releases/tag/v6.12.0 -- now you can pass it an empty yml file and it will not complain (unless is named "playbook"), as it is perfectly fine to have an empty YAML file.

In fact "empty-playbook" is not really a real rule, it is an error message reported by ansible.

If you could provide a real life example we might be able to check if we can tune the rule to make it less painful.

Still, it is linter goal to prevent accidental inclusion of non-playbook inside a folder named "playbooks/", so if you do `touch playbooks/foo.yml`, you will get an error from the linter complaining about an empty playbook. That is by design, to prevent runtime accidents. Remember that various ansible tools might inspect the filesystem in order to discover existing playbooks (including playbooks part of a collection) and we know that specific folders are not expected to have random mixed content.

nkakouros:
OK, I see, it makes sense. I will open an issue in case an empty yml gets picked up as a playbook. Thank you for looking into this.

PR #3022, opened 2023-02-13
Test coverage for meta_incorrect, no_jinja_when rules

ajinkyau:
- Test Coverage improvement for meta_incorrect, no_jinja_when rules


src/ansiblelint/rules/no_jinja_when.py
```diff
@@ -52,11 +52,6 @@ def matchplay(self, file: Lintable, data: dict[str, Any]) -> list[MatchError]:
                             linenumber=role[LINE_NUMBER_KEY],
                         )
                     )
-        if isinstance(data, list):
```
ajinkyau, line 55: Removed this check and related code below it, tried passing multiple plays still this part of code is never invoked

Issue #2965, opened 2023-02-01
Risky octal shouldn't recommend `0` prefixing the permissions.


jcgruenhage:
##### Summary

`yaml` is a notoriously complicated format, with different pitfalls depending on which version you're looking at. What Ansible Lint recommends for permissions only works because the underlying `yaml` parser does not implement the latest version (which has been out for over a decade). Octal numbers in yaml specification >= 1.2.2 work by prefixing `0o` and not `0`.

My suggestion here would be to mandate strings and completely disallow int literals. This avoids the ambiguity between encodings of different yaml versions. Mandating `0o` can't work, because that's not supported by ansible.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

not applicable

##### STEPS TO REPRODUCE

not applicable

##### Desired Behavior

Don't recommend setting values that will break when the yaml parser is eventually upgraded to the new version.

##### Actual Behavior

It recommends prefixing the int with an `0`, which will break eventually.

ssbarnea:
I checked and while Ansible supports the new `0o` prefixes documented by [YAML 1.2.2 specification](https://yaml.org/spec/1.2.2/ext/changes/), that is apparently not documented anywhere.


- [Ansible YAML Syntax page](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html) does not contain any documentation about octals and it should be updated to include that ASAP
- Core modules accepting octals (usually `mode` argument), still advertise the old octals and we need to fix them.
- We will update the linter to recommend the modern octals, so people will have plenty of time to migrate to more future proof values.

ssbarnea:
@jcgruenhage I am quite curious about your statement related to "0o can't work, because that's not supported by ansible." -- which based on my own tests prove to be false. Yes, it is not documented, but when trying to use `0o` prefix with file module, I got perfect results. Maybe some 3rd party modules might fail to use the correct ansible methods for processing the octals, but Ansible itself seems to work fine with new ones.

jcgruenhage:
https://github.com/ansible/ansible/blob/fa382670dc35d89652c03c3d7f7911533a36b29d/lib/ansible/module_utils/basic.py#L897 is the line that parses the strings, for which apparently both `0o755` and `0755` turn into `493` in decimal, so that solves that bit of mystery.

sivel:
I should also note that jinja2 native resolves the issues as well, and we hope to switch to that being the default soon.

[...]

PR #3030, opened 2023-02-14
Allow only use of quoted octals

ssbarnea:
In order to prevent errors caused by different versions of YAML loaders we now forbid use of implicit and explicit octals and ask user to add quotes, so the behavior will be consistent.

```yaml
foo: 0777 # <-- yaml[octal-values]
bar: 0o777 # <-- yaml[octal-values]
---
# Fixed:
foo: "0777"
bar: "0o777"
```
Fixes: #2965

### DIFF
--- a/src/ansiblelint/rules/risky_octal.md
+++ b/src/ansiblelint/rules/risky_octal.md
@@ -1,11 +1,12 @@
 # risky-octal
 
 This rule checks that octal file permissions are strings that contain a leading
-zero or are written in [symbolic modes](https://www.gnu.org/software/findutils/manual/html_node/find_html/Symbolic-Modes.html),
+zero or are written in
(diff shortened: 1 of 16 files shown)


src/ansiblelint/rules/risky_octal.md
```diff
@@ -43,11 +44,6 @@ Modules that are checked:
...
         group: foo
-        mode: 0644 # <- Leading zero will prevent Numeric file permissions to behave in unexpected ways.
-    - name: Safe example of declaring Numeric file permissions (2nd solution)
-      ansible.builtin.file:
-        path: /etc/foo.conf
-        owner: foo
-        group: foo
-        mode: "0644" # <- Being in a string will prevent Numeric file permissions to behave in unexpected ways.
+        mode: "0644" # <- quoting and the leading zero will prevent surprises
+        # "0o644" is also a valid alternative.
```
ajinkyau, line 48: +1

ajinkyau (approved):

Issue #2870, opened 2023-01-05
`var-naming` is enforcing personal preference


The-Judge:
##### Summary

[var-naming](https://ansible-lint.readthedocs.io/rules/var-naming/) seems to enforce personal preferences on variable naming, that are neither required nor recommended by Ansible, nor is a rationale provided in the docs; not even in the doc referenced in that RTD-Page ([creating valid variable names](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#creating-valid-variable-names)). Also, the check seems not to be able to be configured by changing this behavior.

What makes absolute sense is that it enforces stuff that is plain wrong in Ansible, like using `-` instead of `_` or using special characters.
But no Ansible Doc I know (or is referenced by Ansible-lint) declares Uppercase variable names wrong. The only indicator that this may be considered best practice is that the examples in [creating valid variable names](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#creating-valid-variable-names) do it like that. But: `Examples != Rules`.

CamelCase may not look very nice to some people, but nowhere it is stated that you may not use that in Ansible variable names. That's why I state this is personal preference being enforced here.

Other Ansible-lint rules offer sub-rules that can be enabled/disabled individually; such as [name](https://ansible-lint.readthedocs.io/rules/name/), which has `name[casing]`, `name[prefix]`, `name[template]`, ...

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
6.10.2
```

##### STEPS TO REPRODUCE

1. Activate `var-naming` rule in ansible-lint
2. Create an Ansible file that uses uppercase letters in a variable name, like this:

```console (paste below)
---
- name: demo
  set_fact:
    thisIsSparta: yes
...
```

3. Check the file using ansible-lint.

##### Desired Behavior

The module should do one of the following things:

1. Check only for things that are officially (by Ansible upstream) considered bad practice. In this case, this reference should be in the ansible-lint docs as a rationale.
2. Offer the casing as a sub-rule that users can either activate or deactivate (like `var-naming[casing]`).
3. Provide an objective rationale at [var-naming](https://ansible-lint.readthedocs.io/rules/var-naming/) why this is a must-have.

My favorited solution by far would be option 2.

##### Actual Behavior

See `STEPS TO REPRODUCE`

briantist:
In some sense, I would say all linting is subjective (otherwise it's a syntax check).

[...]

PR #3042, opened 2023-02-15
Improve documentation on var-naming

ssbarnea:
Fixes: #2870


cidrblock (approved):
Th@nk$--for,~~d0ing=this!

PR #3043, opened 2023-02-15
profiles: include all rules when listing all rules or tags

ziegenberg:
When using a profile some rules are unloaded. This also happens when listing all rules or tags with `--list-tags` or `--list-rules`. This commit changes the behaviour and prevents the unloading of rules when listing all rules or tags.


ziegenberg:
I rebased.

ssbarnea (approved):

PR #3046, opened 2023-02-15
Fix ignore file generation

jcgruenhage:
Previously, generated ignore files didn't put newlines between separate
ignores. We're now appending `\n` to each line before writing them out.


jcgruenhage:
There should probably be a better test case here (one with multiple lines), but I'm currently on sick leave and only doing very little due to that..

ssbarnea (approved):

PR #3050, opened 2023-02-17
Improve no-changed-when rule

ssbarnea:
- avoid looking for `when:` when evaluating rule
- clarify documentation
- ensure rule works w/ and w/o fqcn actions
- make test file standalone for easy debugging

Fixes: #2629

### DIFF
--- a/src/ansiblelint/rules/no_changed_when.md
+++ b/src/ansiblelint/rules/no_changed_when.md
@@ -12,8 +12,9 @@ detect if a change has occurred or not. Some of the most common examples are
 [shell] and [command] modules, which run arbitrary commands.
 
 One very common workaround is to use a boolean value like `changed_when: false`
-but you can also use any expressions, including ones that use the registered
-result of a task, like in our example below.
+if the task never changes anything or `changed_when: true` if it always
+changes something, but you can also use any expressions, including ones that
+use the registered result of a task, like in our example below.
 
 ## Problematic Code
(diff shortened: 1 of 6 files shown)


src/ansiblelint/rules/no_changed_when.md
```diff
@@ -1,12 +1,19 @@
...
 - Use the `creates` or `removes` argument.
 
-You should use the `when` clause to run tasks only if a check returns a particular result.
+You should always use the `changed_when` clause on tasks that do not naturally
+detect if a change has occurred or not. Some of the most common examples are
+[shell] and [command] modules, which run arbitrary commands.
+
+One very common workaround is to use a boolean value like `changed_when: false`
+but you can also use any expressions, including ones that use the registered
+result of a task, like in our example below.
```
bluikko, line 16: My minor suggestion would be to expand on the use of boolean values, something like:

```suggestion
if the task never changes anything or `changed_when: true` if it always
changes something, but you can also use any expressions, including ones that
use the registered result of a task, like in our example below.
```
I added the `true` case for completeness but my point was to explain *when* to use the boolean case - it could be condensed to just explaining "if the task never changes anything, but you can also ...". Unfortunately after this addition it is quite a long single sentence - I am sure a better than my wording can be found or possibly split the sentence before the "but".

priyamsahoo (approved):

Issue #3066, opened 2023-02-19
incomplete/non-intuitive handling of '.ansible-lint-ignore' file


4ch1m:
##### Summary

The `.ansible-lint-ignore` file is not being considered/loaded when ...

* the config file is located under `.config/ansible-lint.yml`
or 
* the config file is being referenced via `--config-file` command line parameter.

The official [documentation](https://ansible-lint.readthedocs.io/configuring/#ignoring-rules-for-entire-files) currently states:

> Ansible-lint will load skip rules from .ansible-lint-ignore file that is adjacent to its config file.

The keyword being **adjacent** here.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

Tested on Linux.

```
ansible-lint --version

ansible-lint 6.13.1 using ansible 2.14.1
```

##### STEPS TO REPRODUCE

```
.
├── .ansible-lint
├── .ansible-lint-ignore
└── tasks
    └── test.yml

ansible-lint tasks/test.yml
```
:heavy_check_mark: This will work as expected. Rules listed in `.ansible-lint-ignore` actually do get ignored.

---

```
.
├── .config
│   ├── .ansible-lint-ignore
│   └── ansible-lint.yml
└── tasks
    └── test.yml

ansible-lint tasks/test.yml
```
:x: This does **not** work as expected/documented. Rules listed in `.config/.ansible-lint-ignore` are not being considered. (Whereas the config file itself will be used as expected.)

---

```
.
├── my_config
│   ├── .ansible-lint
│   └── .ansible-lint-ignore
└── tasks
    └── test.yml

ansible-lint -c my_config/.ansible-lint tasks/test.yml
```
:x: This also does **not** work as expected/documented. Rules listed in `my_config/.ansible-lint-ignore` are not being considered. (Whereas the config file itself will be used as expected.)

##### Desired Behavior

The ignore file should be loaded in all listed cases above.

It's still debatable if the ignore file should be named `ansible-lint-ignore.txt` (instead of `.ansible-lint-ignore`) if it resides in a `.config` subdir.
(Which should be documented accordingly then.)

In general I'd suggest to provide an additional command-line parameter (which complements `-c / --config-file`):

```
 -i IGNORE_FILE, --ignore-file IGNORE_FILE
 
Specify ignore file to use. By default it will look for '.ansible-lint-ignore' or '.config/ansible-lint-ignore.txt'
```

I hope something like this could be implemented in the near future.

[...]

PR #3075, opened 2023-02-21
Improved ignore file handling

4ch1m:
Hey there,

this will fix and/or improve #3066.
Let me know what you think. :smile:
Thanks!

Depends on #3071


4ch1m:
Hmmm... seems like this PR needs a _label_ to pass the missing pre-check.
Unfortunately I can't set ony myself (due to missing rights?!).
Help/assistance would be appreciated.

ajinkyau (approved):

PR #3080, opened 2023-02-22
Fix: Loosen the schema for .ansible-lint to allow custom rules specific configuration

ssato:
The schema for .ansible-lint is too strict and only allows rules specific configurations of the default rules listed in it.

The changes remove that too strict limitation of the schema and allows custom rules specific configuration.


src/ansiblelint/schemas/ansible-lint-config.json
```diff
@@ -122,60 +122,7 @@
...
-          "risky-file-permissions",
-          "risky-octal",
-          "risky-shell-pipe",
-          "role-name",
-          "schema",
-          "syntax-check",
-          "var-naming",
-          "yaml"
-        ]
+        "pattern": "^[a-z0-9-]+$"
```
MarkusTeufelberger, line 125: `[` and `]` are also part of rule names, e.g. `fqcn[action-core]`, there are none with numbers though.

ssato: Right, I updated the pattern. Thanks!

src/ansiblelint/schemas/ansible-lint-config.json
```diff
@@ -122,60 +122,7 @@
...
-          "risky-file-permissions",
-          "risky-octal",
-          "risky-shell-pipe",
-          "role-name",
-          "schema",
-          "syntax-check",
-          "var-naming",
-          "yaml"
-        ]
+        "pattern": "^[a-z0-9-\\[\\]]+$"
```
MarkusTeufelberger, line 125: No current rule name that I'm aware of contains numbers, so why include these. On the other hand, underscores and capital letters might be something custom rule authors want to use?

ssato: I think rule names (IDs) and file names should be standardized, but I don't have a strong opinion on how these should be.  Also, IMHO, it's not a point of discussion for this issue. We can update the pattern after the standard is fixed.

For now, the pattern is tentatively specified based on the IDs of the standard rules and the IDs of some custom rules under test/custom_rules/.

ssbarnea (approved):

Issue #3026, opened 2023-02-14
Ensure rules have full code coverage


ssbarnea:
Coverage report with currently uncovered code from within `rules/`.

- [ ] src/ansiblelint/rules/args.py                          136     10    93%   164, 175, 181, 198-199, 206-215
- [ ] src/ansiblelint/rules/command_instead_of_module.py      82      7    91%   251-262
- [ ] src/ansiblelint/rules/deprecated_bare_vars.py           34      1    97%   66
- [ ] src/ansiblelint/rules/deprecated_command_syntax.py      23      1    96%   63
- [ ] src/ansiblelint/rules/empty_string_compare.py           40      1    98%   47
- [ ] src/ansiblelint/rules/inline_env_var.py                 19      1    95%   68
- [ ] src/ansiblelint/rules/jinja.py                         227     16    93%   91, 96-105, 108, 162-166, 615-616
- [ ] src/ansiblelint/rules/loop_var_prefix.py                41      1    98%   45
- [ ] src/ansiblelint/rules/meta_no_info.py                   42      1    98%   75
- [ ] src/ansiblelint/rules/meta_runtime.py                   38      1    97%   74
- [ ] src/ansiblelint/rules/meta_video_links.py               35      1    97%   41
- [ ] src/ansiblelint/rules/no_free_form.py                   45      1    98%   51
- [ ] src/ansiblelint/rules/no_log_password.py                72      1    99%   47
- [ ] src/ansiblelint/rules/no_prompting.py                   32      2    94%   32, 36
- [ ] src/ansiblelint/rules/no_relative_paths.py              20      1    95%   45
- [ ] src/ansiblelint/rules/no_same_owner.py                  53      2    96%   34, 68
- [ ] src/ansiblelint/rules/no_tabs.py                        28      1    96%   45
- [ ] src/ansiblelint/rules/syntax_check.py                   93      4    96%   77, 121-123, 152
- [ ] src/ansiblelint/rules/var_naming.py                    110      5    95%   37, 54, 58-59, 62
- [ ] src/ansiblelint/rules/yaml_rule.py                      87      3    97%   50, 112, 179

trishnaguha:
@ajinkyau Could you share the ETA for this task?

ajinkyau:
@trishnaguha I think it can be completed until the month end

trishnaguha:
@ssbarnea @ajinkyau Can we close this issue and open a new issue with the postponed list? 
[...]

PR #3088, opened 2023-02-27
Adding coverage for yaml rule

ajinkyau:
Related: #3026

### DIFF
--- a/src/ansiblelint/rules/yaml_rule.py
+++ b/src/ansiblelint/rules/yaml_rule.py
@@ -48,8 +48,6 @@ class YamllintRule(AnsibleLintRule):
             self.severity = "VERY_LOW"
             if problem.level == "error":
                 self.severity = "MEDIUM"
-            if problem.desc.endswith("(syntax)"):
-                self.severity = "VERY_HIGH"
             matches.append(
                 self.create_matcherror(
                     # yamllint does return lower-case sentences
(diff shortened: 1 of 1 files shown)


src/ansiblelint/rules/yaml_rule.py
```diff
@@ -48,8 +48,6 @@ def matchyaml(self, file: Lintable) -> list[MatchError]:
             self.severity = "VERY_LOW"
             if problem.level == "error":
                 self.severity = "MEDIUM"
-            if problem.desc.endswith("(syntax)"):
```
ajinkyau, line 51: Removed these two lines as the tests don't run if we have a syntax error in the yaml file

src/ansiblelint/rules/yaml_rule.py
```diff
@@ -177,5 +175,5 @@ def test_yamllint_has_help(default_rules_collection: RulesCollection) -> None:
                 assert collection.help is not None
```
ssbarnea, line 175: Can you also please rename `collection` variable to `rule`, I don't know how this code went in because we are talking about a rule. The bug is unrelated to your change but is in the same area, so better to address it.

ssbarnea (changes requested):

src/ansiblelint/rules/yaml_rule.py
ajinkyau: ack

ssbarnea (approved):

ssbarnea (approved):

PR #3097, opened 2023-02-28
Fix for skipped rules key in yaml rule

ajinkyau:
Bug fix for yaml rule when we pass a list and try to skip a rule.
Also added tests to improve coverage.

Signed-off-by: Ajinkya Udgirkar <audgirka@redhat.com>


ajinkyau:
@ssbarnea please could you review and merge #3096 first and then this one as the total number of tests that will pass is one more than #3096

ssbarnea (approved):

Issue #3110, opened 2023-03-01
Ansible Lint reports the `decrypt` parameter in the `ansible.builtin.copy` module is not supported


alessfg:
##### Summary

Ansible Lint reports the `decrypt` parameter in the `ansible.builtin.copy` module is not supported.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
$ ansible --version
ansible [core 2.14.3]
  config file = /Users/a.faelgarcia/.ansible.cfg
  configured module search path = ['/Users/a.faelgarcia/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/a.faelgarcia/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.11.2 (main, Feb 16 2023, 02:55:59) [Clang 14.0.0 (clang-1400.0.29.202)] (/opt/homebrew/opt/python@3.11/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True

$ ansible-lint --version
ansible-lint 6.13.1 using ansible 2.14.3

$ python --version
Python 3.11.2
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

1. Create a task/playbook containing the `ansible.builtin.copy` module and using the `decrypt` parameter:

```yaml
- name: (Alpine Linux) Copy NGINX Plus certificate
  ansible.builtin.copy:
    src: "{{ nginx_license['certificate'] }}"
    dest: /etc/apk/cert.pem
    decrypt: true
    mode: "0444"
```

2. Run Ansible Lint

```console
$ ansible-lint
...
args[module]: Unsupported parameters for ansible.builtin.copy module: decrypt. Supported parameters include: _original_basename, attributes, backup, checksum, content, dest, directory_mode, follow, force, group, local_follow, mode, owner, remote_src, selevel, serole, setype, seuser, src, unsafe_writes, validate (attr). (warning)
tasks/plus/setup-license.yml:15 Task/Handler: (Alpine Linux) Copy NGINX Plus certificate
...
```

##### Desired Behavior

Per the `ansible.builtin.copy` module docs, the `decrypt` parameter should be supported https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html#parameter-decrypt

```console
[...]

PR #3113, opened 2023-03-02
Generalize args rule exception handling

ajinkyau:
Fixes: #3110

### DIFF
--- a/src/ansiblelint/rules/args.py
+++ b/src/ansiblelint/rules/args.py
@@ -38,6 +38,19 @@ ignored_re = re.compile(
     flags=re.MULTILINE | re.DOTALL,
 )
 
+workarounds_drop_map = {
+    # https://github.com/ansible/ansible-lint/issues/3110
+    "ansible.builtin.copy": ["decrypt"],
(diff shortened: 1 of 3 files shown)


MarkusTeufelberger:
This means no parameter in this (rather common) module will be validated... would it be possible to manually inject the missing parameter(s) instead?

src/ansiblelint/rules/args.py
```diff
@@ -88,10 +88,14 @@ def matchtask(
...
-        if loaded_module.resolved_fqcn == "ansible.builtin.service":
-            _logger.debug(
-                "Skipped service module validation as not being supported yet."
-            )
+        if loaded_module.resolved_fqcn in (
+            "ansible.builtin.service",
+            "ansible.builtin.copy",
+        ):
+            module_name_list = module_name.split(".")
+            name = module_name_list[len(module_name_list) - 1]
```
MarkusTeufelberger, line 96:
```suggestion
            name = module_name_list[-1]
```

ajinkyau:
> This means no parameter in this (rather common) module will be validated... would it be possible to manually inject the missing parameter(s) instead?

It would not be possible to inject missing params, per my understanding this is something that needs to be updated in core

PR #3145, opened 2023-03-06
Implementing the count no of tasks rule

Ruchip16:


priyamsahoo (changes requested):
This is not correct. As I mentioned, you should not commit (and push) your local venv. This is the reason why no. of files changed is almost 3K!

ssbarnea:
@Ruchip16 Please cleanup you PR and avoid adding files what we do not track. You might want to tune your user level gitignore file to exclude some folders/files you might have added inside the repository.

Ruchip16:
> This is not correct. As I mentioned, you should not commit (and push) your local venv. This is the reason why no. of files changed is almost 3K!

yes, i realized it after committing it and since then i am working on my venv 
:cry:

PR #3206, opened 2023-03-22
Do not trigger no_changed_when when "when" is used

sathieu:
I have this:

```yaml
  - name: Fact
    set_fact:
      foo: bar

  - name: Test
    command:
      argv:
      - echo
      - Hellow
    when:
    - foo == 'bar'
```


ssbarnea:
This rule has nothing to do with `when`, you still need to add a `changed_when` to inform ansible if a change was made when the task was executed or not.

sathieu:
Back in https://github.com/ansible/ansible-lint/pull/3050. Thanks :pray: !

PR #3207, opened 2023-03-22
Document fix for literal_compare `== False`

jemand771:
While this should be obvious to experienced ansible users, I didn't know you could just use the `not` keyword like that. Documenting this doesn't take up much extra space but can potentially save time wihle fixing linter warnings.


4ch1m:
> I've spent quite some time to understand what's going on; trying to fix it. No luck so far. cry
> 
> Maybe @ssbarnea can help us out.

I found the root cause of the issue.

The [__store__.json](https://github.com/ansible/ansible-lint/blob/42329279496085a30946c1c93fc2fa8e0dbb2d5b/src/ansiblelint/schemas/__store__.json) is under version control and part of the unit-tests.

The file holds several ETags of GitHub-server-files/-resources.

IMHO this is not suitable for unit-tests, since ETags are - per definition - non-deterministic (check: [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)).

Meaning: It's up to the server(-implementation) how ETags are being generated; and how/if they change over time.

@ssbarnea : Maybe we should find a workaround for this?

jemand771:
ah, thanks for looking into that! I couldn't really figure out what was frong at first glance.
I just saw something about failed schemas and thought that maybe help texts are only allowed to one paragraph long or something like that, but apparently that's not it

ssbarnea (approved):

PR #3221, opened 2023-03-28
Extend the var_naming rule to also check for the role name

aehliglucas:
## Summary
Hi ansible-lint team!

The project team that I am currently part of wishes to include the role name in all variables that are defined in a role scope.
In order to enforce this using ansible-lint, one of my colleagues created a discussion for it (https://github.com/ansible/ansible-lint/discussions/3197). 
Out of curiosity, I started working on it myself and this small modification is the result. 

It allows users to add "{role}" into the var_naming_pattern field in the .ansible-lint config file.
Example:
`var_naming_pattern: "^{role}[a-z_][a-z0-9_]*$"`

The rule then obtains the role name based on the absolute path of the vars or tasks file if it is part of a role and adjusts the Regex pattern accordingly.

## Example
In this example, we are checking if the role name is provided in front of the variable name itself. (like a prefix)
For the sake of testing, the role is called "example_role".

**.ansible-lint:**
```
[...]
# Enforce variable names to follow pattern below, in addition to Ansible own
# requirements, like avoiding python identifiers. To disable add `var-naming`
# to skip_list.
var_naming_pattern: "^{role}[a-z_][a-z0-9_]*$"
[...]
```
### This will fail:
**roles/example_role/vars/main.yml:**
```
[...]
action: lookup
[...]
```
**ansible-lint output:**
```
var-naming: File defines variable 'action' that violates variable naming standards
roles/example_role/vars/main.yml:1

Read documentation for instructions on how to ignore specific rule violations.

            Rule Violation Summary             
 count tag        profile rule associated tags 
     1 var-naming basic   idiom                

Failed after min profile: 1 failure(s), 0 warning(s) on 11 files.
```

### This will succeed:
**roles/example_role/vars/main.yml:**
```
[...]
example_role_action: lookup
[...]
```
**ansible-lint output:**
```
Passed with production profile: 0 failure(s), 0 warning(s) on 11 files.
```


aehliglucas:
> Single curly braces look similar to Jinja2 syntax, but not the same. Is there a way maybe that you could use existing template capabilities of either Python (f-strings) or Ansible (Jinja2) instead of inventing yet another way of templating?
> 
> In general I like the idea of having a little bit of context available while evaluating files. I wonder how this is supposed to work within playbook files for example, if configured globally in the `.ansible-lint` config file.

Thanks for your remarks.
If I got you right, you proposed to use f-strings (or Jinja2) for templating. My solution actually uses formatted strings, as it reads the value of the "var_naming_pattern" field in .ansible-lint and formats it using the role name, if there is one. Using an f-string would force us to set it permanently in the rule script, which would be impractical in case a user doesn't want to check if their variable names contain the corresponding role. 
I took the "task_name_prefix" field as an inspiration, as it also uses single curly braces.

On your question how this would work in playbooks: The role name check will only consider files inside the roles/ folder. Any other files outside of the roles folder will be linted as usual, following the "^[a-z_][a-z0-9_]*$" pattern.

ssbarnea (changes requested):
I can understand that feature but this change is missing any tests.

ssbarnea:
#3422 will fix this one, we refactored the code.

PR #3227, opened 2023-03-29
name[casing]: Add transform to automatically fix this during --write

MarkusTeufelberger:
One of the easier rules for adding transforms I guess, but it is probably time to actually start doing it.

I didn't find any transformations being used (or tested) in the repo so far, but it seems to work and makes this rather annoying rewrite _much_ easier.

### DIFF
--- a/src/ansiblelint/rules/name.py
+++ b/src/ansiblelint/rules/name.py
@@ -8,13 +8,13 @@ from typing import TYPE_CHECKING, Any
 
 from ansiblelint.constants import LINE_NUMBER_KEY
 from ansiblelint.errors import MatchError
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
We need a test for it.

MarkusTeufelberger:
Yeah, about that...

How are transforms going/supposed to be tested? I suspect that I should re-do a few existing tests, but I'm unsure for example how to assert on the results.

ssbarnea:
@MarkusTeufelberger #3268 implement your change. I was not able to fix this PR, probably because i did not have permission to push to your fork.

PR #3230, opened 2023-03-30
Update no-handler rule to accept 'not changed'

jpopelka:
Hello,

I don't know how to convert the following to a handler so I think the `no-handler` rule should accept it:

```yaml
      tasks:
        - name: Registering task
          command: echo Hello
          register: r
          changed_when: true

        - name: When task not changed
          command: echo Not changed
          when: not r.changed
```


ssbarnea:
@jpopelka Thanks for the fix. I only had to tune to testing but it makes sense as handlers do not have conditions.

Issue #3245, opened 2023-04-06
Missing "kind" in message for several files/


bluikko:
##### Summary

Running at verbosity 10 reveals that several files stored in `rolename/files/` are missing "kind", which is probably fine internally but the output could be improved. They are actually printed with the field that usually lists the "kind" missing in the output which produces a strange looking message where there are 2 spaces where the missing "kind" should be:
```console
data set to None for roles/rolename/files/php.conf due to being of  kind.
                                                                   ^ here
```
Example of file where "kind" is detected:
```console
data set to None for roles/rolename/files/php.ini due to being of text/ini kind.
```

If this "empty kind" is supposed to mean "file kind was not detected or was unknown" then it would be better to print that instead of just leaving the field empty in the output. Examples:
* `due to being of unrecognized kind.`
* `due to being of undetected kind.`
* `due to being of unknown kind.`
* or even `due to being of (skipped) kind.`

I would also like to note that "type" is used more universally for this meaning instead of "kind"...

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.14.4 using ansible 2.14.4
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

1. Create a role.
2. Put some files in `files/`, for example 1 file with extension `.ini` and other file with extension `.conf`.
3. Run `ansible-lint -vv`.
4. Check output for missing "kind" for the `.conf` file.

```console (paste below)

```

##### Desired Behavior

Improve the message so that the "kind" is not left to empty which leaves 2 consecutive spaces in the output, instead use a suitable word such as "undetected", "unknown", "unrecognized", "(skipped)", etc in the field.

Possible security bugs should be reported via email to `security@ansible.com`

##### Actual Behavior

```paste below
data set to None for roles/rolename/files/php.conf due to being of  kind.
```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

PR #3247, opened 2023-04-06
Updated 'kind' message if file kind is unknown

ajinkyau:
Fixes: #3245


src/ansiblelint/file_utils.py
```diff
@@ -388,7 +388,7 @@ def data(self) -> Any:
                     logging.debug(
                         "data set to None for %s due to being of %s kind.",
                         self.path,
-                        self.base_kind,
+                        self.get("base_kind", "unknown"),
```
ssbarnea, line 391: I think that the right way to do it should be `self.base_kind or "unknown"`.

ssbarnea (approved):

Issue #3261, opened 2023-04-11
ansible-lint schema check fails for doc_changes with network collections


KB-perByte:
lint fails for _doc_changes_
valid section across Ansible network modules 
https://github.com/ansible-collections/cisco.ios/blob/main/changelogs/config.yaml#L29
##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
```

- ansible installation method: one of source, pip, OS package
- ansible-lint installation method: one of source, pip, OS package

##### STEPS TO REPRODUCE

```console (paste below)

```

##### Desired Behavior

Possible security bugs should be reported via email to `security@ansible.com`

##### Actual Behavior

Please give some details of what is happening.
Include a [minimum complete verifiable example] with:

- minimized playbook to reproduce the error
- the output of running ansible-lint including the command line used
- if you're getting a stack trace, also the output of
  `ansible-playbook --syntax-check playbook`

```paste below
WARNING  Listing 1 violation(s) that are fatal
schema[changelog]: Additional properties are not allowed ('doc_changes' was unexpected)
changelogs/changelog.yaml:1  Returned errors will not include exact line numbers, but they will mention
the schema name being used as a tag, like ``schema[playbook]``,
``schema[tasks]``.

This rule is not skippable and stops further processing of the file.

If incorrect schema was picked, you might want to either:

* move the file to standard location, so its file is detected correctly.
* use ``kinds:`` option in linter config to help it pick correct file type.


Error: Additional properties are not allowed ('doc_changes' was unexpected)
Read documentation for instructions on how to ignore specific rule violations.

                Rule Violation Summary                
 count tag               profile rule associated tags 
     1 schema[changelog] basic   core  
```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

PR #3262, opened 2023-04-11
Ignore unknown sections under changes key on changelogs

ssbarnea:
Fixes: #3261

### DIFF
--- a/test/schemas/test/changelogs/maximal/changelog.yaml
+++ b/test/schemas/test/changelogs/maximal/changelog.yaml
@@ -32,6 +32,8 @@ releases:
         - something that is not included in release notes
       known_issues:
         - som other
+      xxx:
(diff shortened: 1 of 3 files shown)


test/schemas/test/changelogs/maximal/changelog.yaml
```diff
@@ -32,6 +32,8 @@ releases:
         - something that is not included in release notes
       known_issues:
         - som other
+      xxx:
+        - we should ignore unknown keys because used can define custom section in changelogs/config.yaml file
```
ajinkyau, line 36: we should ignore unknown keys because `user` can define custom section in changelogs/config.yaml file

ajinkyau (changes requested):

ajinkyau (approved):

PR #3267, opened 2023-04-11
Remove meta-no-info rule due to being covered by schema[meta]

ssbarnea:
As our schema validation covers all the tests from meta-no-info,
it makes no sense to keep the rule.


cidrblock (approved):
nice, less code

Issue #3177, opened 2023-03-10
Schema validation fails for variables files in tasks/vars/


CarstenGrohmann:
##### Summary

The schema detection for variables files in `roles/<role name>/tasks/vars` is failing. 

```
schema[tasks]: {'city': 'Dresden'} is not of type 'array', 'null'
roles/myrole/tasks/vars/config.yml:1  Returned errors will not include exact line numbers, but they will mention
the schema name being used as a tag, like ``schema[playbook]``,
``schema[tasks]``.
```

```
# cat ./roles/myrole/tasks/vars/config.yml
---
city: Dresden
...
```

As I understand the error message, Ansible-lint expect files with tasks schema in `tasks/vars`. But we placed a file with variables inside this directory.

We collect tasks in a helper role and import that tasks on demand in other roles. One of the helper task needs additional variables. The variables will be included with `include_vars`. As we import the task from another role, we've to place the variables file next to the task, otherwise Ansible didn't find it. 

According to the [Ansible search order](https://docs.ansible.com/ansible/latest/playbook_guide/playbook_pathing.html#resolving-local-relative-paths) we placed the variables file in `tasks/vars`. That's in the search path and `include_vars` works well.

Now we are introducing Ansible-lint and schema check fails.

According to the [documentation of the schema rule](https://github.com/ansible/ansible-lint/blob/36ef08b965cb63a8a7d1efafc45cd569e1a4f87f/src/ansiblelint/rules/schema.md?plain=1#L53-L56) Ansible-lint expects tasks file in `tasks/**/*.yml` (directly in `tasks` or in next level of subdirectories) and variables files in `vars/*.yml` and `defaults/*.yml`.

Based on my understanding from the Ansible documentation, I expect that YAML files in `tasks/vars/` are thread as variables file with a corresponding scheme.

What do you think about it?

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
$ ansible-lint --version
ansible-lint 6.14.2 using ansible 2.13.8
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

```console (paste below)
# find .
.
./roles
./roles/myrole
./roles/myrole/vars
./roles/myrole/tasks
./roles/myrole/tasks/vars
./roles/myrole/tasks/vars/config.yml

# cat ./roles/myrole/tasks/vars/config.yml
---
city: Dresden
...

#  ansible-lint 
[...]

PR #3289, opened 2023-04-14
Fix Schema validation for variables in tasks/vars directory

ajinkyau:
Fixes: #3177


ssbarnea (changes requested):
This is missing a test, be sure that you also add a test that would have failed without this change as I showed you yesterday. Enough to create a file inside `examples/playbooks/tasks/vars/bug-3289.yml` that is a vars file and include it in testing.

You need to add this test to `test_kinds`

PR #3291, opened 2023-04-14
Extend ruff checks

ssbarnea:
- enable all checks while adding the non-compliant ones to ignore
- remove isort to avoid conflicts


cidrblock (approved):
:dog:

PR #3297, opened 2023-04-17
Change logging to use rich output

ssbarnea:
- use dimmed coloring for stderr output to increase visibility


ajinkyau:
@ssbarnea the tests are failing for this PR

PR #3300, opened 2023-04-17
Refactor exit codes to use an enum

ssbarnea:


src/ansiblelint/constants.py
```diff
@@ -7,11 +7,18 @@
...
-SUCCESS_RC = 0
-VIOLATIONS_FOUND_RC = 2
-INVALID_CONFIG_RC = 3
-LOCK_TIMEOUT_RC = 4
-EXIT_CONTROL_C_RC = 130
+
+# Not using an IntEnum because only starting with py3.11 it will evaluate it
+# as int.
+class RC:  # pylint: disable=too-few-public-methods
+    """All exist codes used by ansible-lint."""
```
ajinkyau, line 14: `exit` instead of `exist`

shatakshiiii (approved):

PR #3303, opened 2023-04-18
Require at least one known subfolder for roles

ssbarnea:
Identified folders as roles only if they contain one of the known subfolders used by roles, such "tasks", "meta", "vars", "defaults", "handlers".

This behaviour change should prevent undesired identification of other folders as roles, like we seen on debops repository, where documentation folders were wrongly identified as roles.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -330,6 +330,7 @@ ssbarnea
 stylesheet
 subdir
 subelements
+subfolders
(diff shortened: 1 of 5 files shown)


ssbarnea:
Please note that "eco" pipeline failure seems expected in this case, as it does change the output for debops repository.

ajinkyau (approved):

Issue #3200, opened 2023-03-17
Since 6.10.0: New args rule broken for most windows modules. (Unable to load module <most windows modules> for options validation)


gone-for-coding:
##### Summary

When linting playbooks with newer versions of ansible lint this message is printed for windows modules:

`Unable to load module <most windows modules> for options validation`

The last working version is 6.9.1 - The issue shows up since 6.10.0!

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.14.2 using ansible 2.14.2
```

- ansible installation method: one of source, pip, OS package -> pip
- ansible-lint installation method: one of source, pip, OS package -> pip

##### STEPS TO REPRODUCE

Install collection ansible.windows then lint the following playbook:

```yaml
---
- name: Configure Windows Systems
  hosts: all
  tasks:
    - name: Create software directory
      ansible.windows.win_file:
        path: "c:\\testdir"
        state: directory
```

```console (paste below)
ansible-lint playbook.yml
root@60f15304ae34:/tmp/poc# ansible-galaxy collection list

# /usr/share/ansible/collections/ansible_collections
Collection                      Version
------------------------------- -------
ansible.windows                 1.13.0
community.crypto                2.10.0
community.general               6.3.0
community.mysql                 3.6.0
community.vmware                2.10.3
community.windows               1.12.0
```

##### Desired Behavior

The linter should not produce a warning.

##### Actual Behavior

```console paste below
root@60f15304ae34:/tmp/poc# ansible-lint playbook.yml
WARNING  Unable to load module ansible.windows.win_file at playbook.yml:5 for options validation

Passed with production profile: 0 failure(s), 0 warning(s) on 1 files.

root@60f15304ae34:/tmp/poc# ansible-lint -vvv playbook.yml
DEBUG    Logging initialized to level 10
[...]

PR #3305, opened 2023-04-19
Avoid running args rule if we encounter a windows module

ajinkyau:
Fixes: #3200

### DIFF
--- a/src/ansiblelint/rules/args.py
+++ b/src/ansiblelint/rules/args.py
@@ -18,7 +18,7 @@ from unittest.mock import patch
 # pylint: disable=reimported
 import ansible.module_utils.basic as mock_ansible_module
 from ansible.module_utils import basic
-from ansible.plugins import loader
+from ansible.plugins.loader import PluginLoadContext, module_loader
 
 from ansiblelint.constants import LINE_NUMBER_KEY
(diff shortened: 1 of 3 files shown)


src/ansiblelint/rules/args.py
```diff
@@ -105,6 +105,15 @@ def matchtask(
             return []
 
         loaded_module = load_module(module_name)
+        if loaded_module.plugin_resolved_path.endswith(".ps1"):
+            _logger.warning(
```
ssbarnea, line 109: No need to log anything about it, it will spam output and confuse users. You can keep a code comment stating that `ps1` modules are unlikely to be executable on our platform.

Still I am quite curious if we could install pwsh on linux to be able to execute these, it would be an interesting exercise.

examples/playbooks/rule-args-module-fail.yml
```diff
@@ -34,3 +34,8 @@
       ansible.builtin.file:
         path: /opt/software/deployment
         state: away
+
+    - name: Create software directory (Windows module - Bug 3200) # we display a warning
```
ssbarnea, line 38: Move this into `-pass` file as we expect it to pass w/o any other side-effects, not to fail.

Issue #2506, opened 2022-09-27
Issue since v6.6.0 with a warning "Attempting to decrypt but no vault secrets found" for all group_vars files with vault variables


alice-rc:
##### Summary
All versions since 6.6.0 (6.6.1, 6.7.0) now report the following for all group_vars files with vaulted variables.
NOTE: the environment variable ANSIBLE_VAULT_PASSWORD_FILE is set.
```
WARNING  Ignored exception from YamllintRule.<bound method YamllintRule.matchyaml of yaml: Violations reported by yamllint.> while processing group_vars/all.yml (vars): Attempting to decrypt but no vault secrets found
```
##### Issue Type

- Bug Report

##### Ansible and Ansible Lint details

```console (paste below)
$ ansible --version
ansible [core 2.13.4]
  config file = /home/julie.blake/venv/ee-183-lint/meta/ansible.cfg
  configured module search path = ['/home/julie.blake/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/julie.blake/venv/ee-183-lint/lib64/python3.9/site-packages/ansible
  ansible collection location = /home/julie.blake/venv/ee-183-lint/collections
  executable location = /home/julie.blake/venv/ee-183-lint/bin/ansible
  python version = 3.9.7 (default, Apr 11 2022, 06:30:15) [GCC 8.5.0 20210514 (Red Hat 8.5.0-10.0.1)]
  jinja version = 3.1.2
  libyaml = True
$ ansible-lint --version
ansible-lint 6.7.0 using ansible 2.13.4
$ ansible-lint --version
ansible-lint 6.6.1 using ansible 2.13.4
$ ansible-lint --version
ansible-lint 6.6.0 using ansible 2.13.4
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### OS / ENVIRONMENT
```
$ cat /etc/os-release
NAME="Oracle Linux Server"
VERSION="8.6"
ID="ol"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="8.6"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Oracle Linux Server 8.6"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:oracle:linux:8:6:server"
HOME_URL="https://linux.oracle.com/"
BUG_REPORT_URL="https://bugzilla.oracle.com/"

[...]

PR #3306, opened 2023-04-19
Improve tests related to vault processing

ssbarnea:
As we were not able to reproduce the original bug #2504, we add a test for it and close it, so
we would prevent it from resurfacing later.

Fixes: #2506


ajinkyau (approved):

ktomk:
Thank you!

examples/playbooks/vars/vault_full.yml
```diff
@@ -1,3 +1,4 @@
+---
```
pascal-hofmann, line 1: Adding the document start breaks ansible-vault for me:
```
ERROR! input is not vault encrypted data. …
```

Issue #2881, opened 2023-01-11
(H/C) ansible-lint lint docstrings/examples in plugins


ssbarnea:

cidrblock:
https://github.com/ansible-collections/ansible.utils/blob/main/plugins/modules/cli_parse.py#L86

example of examples

Qalthos:
I'm looking into making this work.

PR #3309, opened 2023-04-19
Allow linting plugin EXAMPLES as playbooks

Qalthos:
Add a means of reading ansible plugins and linting their EXAMPLES string.

Fixes: https://github.com/ansible/ansible-lint/issues/2881

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -259,6 +259,7 @@ pipx
 pkgcache  # linux
 pkgs
 placefolder
+plainexamples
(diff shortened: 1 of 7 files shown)


cidrblock:
Thinking about this more...............

What if we insisted that sidecar docs were used rather than inline (we could probably write a little utility to do the conversion in masse)

Would that make this easier?

ssbarnea:
@Qalthos Any chance you can rebase this so we can merge it?

PR #3320, opened 2023-04-21
Removed TRY from ruff ignores

ajinkyau:


ssbarnea (changes requested):
This does not look ok to me at all. TypeError is a very different thing.

I think that what we might need/want to do to address TRY is to subclass python own exceptions like RuntimeError or ValueError and use our classes for that. In some cases we might even need to include the error message inside the subclassed exception, so we do not need to pass it as an argument.

@cidrblock might also want to check this, maybe I am missing something.

ajinkyau:
Closing it as of now coz we have other things to push

PR #3350, opened 2023-04-25
Remove progressive mode

ssbarnea:


rndmh3ro:
Shouldn't removals of features be done in a major release according to semver?

> Deprecating existing functionality is a normal part of software development and is often required to make forward progress. When you deprecate part of your public API, you should do two things: (1) update your documentation to let users know about the change, (2) issue a new minor release with the deprecation in place. Before you completely remove the functionality in a new major release there should be at least one minor release that contains the deprecation so that users can smoothly transition to the new API.

https://semver.org/#how-should-i-handle-deprecating-functionality

ssbarnea:
At this moment we use versioning similar to ansible-core, meaning that the 1st number will be bumped only on really-big changes. Removal of a corner-case deprecated feature, is considered minor (basically that is relative to the number of users).

Think that even a bugfix can break someone's pipeline, if they rely on it.

PR #3356, opened 2023-04-26
Remove `EM` from ruff ignores

shatakshiiii:
Exception must not use an string or f-string literal, assign to variable first
As per: https://beta.ruff.rs/docs/rules/#flake8-errmsg-em


src/ansiblelint/cli.py
```diff
@@ -88,7 +88,8 @@ def load_config(config_file: str | None) -> tuple[dict[Any, Any], str | None]:
 
     config = clean_json(config_lintable.data)
     if not isinstance(config, dict):
-        raise RuntimeError("Schema failed to properly validate the config file.")
+        msg = "Schema failed to properly validate the config file."
+        raise RuntimeError(msg)
```
ssbarnea, line 92: This does not make sense, is a static string. We should use a variable only if we have expansion.

ssbarnea (approved):

Issue #1632, opened 2021-06-17
no-handler: should not react on when-conditions containing "and" or "or"


phgogo:
##### Summary
Right now the rule `Tasks that run when changed should likely be handlers` (which BTW, i am a big fan of) would produce findings for all of this lines:

`when: mytask.changed`
`when: mytask is changed`
...
`when: mytask is changed and wartherIsNice|bool`

While i totally agree that the first two examples are bad practices and should produce a linter warning, i would not agree, that the last example should.

##### Proposed solution

As mentioned in #419 i could imagine of splitting up E503 into two rules, one of which reacts to single conditions and one for more complex conditions involving `and` or `or` - that way both could be skipped/disabled seperately.

As @ssbarnea pointed out, it might also be a solution to disable the check completeley for complex conditons.

##### Issue Type

- Bug Report

- ansible installation method: OS package
- ansible-lint installation method: pip

ssbarnea:
The bug report form states to report bugs only on latest version, 2.6.1 is FEW YEARS OLD, "503" rule id no longer exists, python2.7 is no longer supported either. Please cleanup the report to avoid confusing the reader. I happen to have background knowledge otherwise I would have just closed it ;)

And yes, I would support changing the rule only match the two specific simple syntaxes and ignore more complex ones.

lilatomic:
It's possible to add a `when` clause to the handlers themselves. so:
```yaml
- name: mytask
  ...
- name: example task
  ...
  when: mytask is changed and wartherIsNice|bool
```
becomes
```yaml
- name: mytask
  ...
  notify: example handler
---
- name: example handler
  when: wartherIsNice|bool
```
should we really not match on complex conditions?

ssbarnea:
We talked in our weekly meeting and I think we would support a PR that makes it ignore complex conditions.

StopMotionCuber:
As I wrote in the PR making a bugfix to complex condition detection:

Is the information still current that a PR ignoring complex conditions would be supported?

klaus-tux:
@StopMotionCuber Do you want to make a PR or should I take over?

StopMotionCuber:
@klaus-tux Feel free to go ahead with that. You can ping me to additionally review that PR then

PR #3358, opened 2023-04-26
Find "changed" not only in first "when" condition

klaus-tux:
Prior to this change, the rule `no_handler` did not find the string "changed" if it was in the second `when` condition or any later `when` condition:
```
- name: This should be a handler 1
  ansible.builtin.debug:
    msg: why isn't this a handler
  when:
    - result.changed # ansible-lint finds this
    - second_condition

- name: This should be a handler 2
  ansible.builtin.debug:
    msg: why isn't this a handler
  when:
    - first_condition
    - result.changed # ansible-lint misses this
```
This change fixes the implementation so that it finds the second case as well.


ssbarnea (approved):

StopMotionCuber:
I just stumbled upon that as this PR broke our CI on `ansible-lint` and raise that something into this direction was discussed on https://github.com/ansible/ansible-lint/issues/1632

While I agree that it should be consistent (either match on _any_ when condition or none at all if there is more than 1 line available), I'm not sure whether these cases with multiple conditions should be linted.

PR #3359, opened 2023-04-26
Remove `RET` from ruff ignores

shatakshiiii:
Unnecessary variable assignment before `return` statement.
As per: https://beta.ruff.rs/docs/rules/#flake8-return-ret


shatakshiiii:
@cidrblock @ssbarnea I totally agree, wish they had something which actually enforces variable assignment before return. Closing it then :)

Issue #3351, opened 2023-04-25
[Docs] Suggestion for developer guide


cidrblock:
https://ansible-lint.readthedocs.io/contributing/#contributing-to-ansible-lint

Might be an improvement if we mention the need to 

`git submodule init`

and 

`git submodule update`

before running tox if pulling down a PR for testing

PR #3362, opened 2023-04-26
Add initial handling of submodules to contribution docs

klaus-tux:
Fixes: #3351

I submitted my last PR to ansible-lint before the submodule was introduced. If this was mentioned in the documentation it would have saved a lot of time for me.


ssbarnea (changes requested):
Better than adding extra commands is to add `--recursive` to git clone one.

klaus-tux:
The man page does not document `--recursive` but `--recurse-submodules`. I prefer the latter, because it makes more explicit that submodules are used. Drop me a comment if you prefer `--recursive` nevertheless.

ssbarnea:
Please use recursive, as we might drop the submodule in the future and we would not need to update the docs anymore.

ssbarnea:
BTW, thanks for contributing, even minor changes like this can help others.

klaus-tux:
Changed as requested :-)

ssbarnea (approved):

PR #3378, opened 2023-04-27
ruff: Address TCH

ssbarnea:


cidrblock (approved):
I really like this rule :)

Issue #3310, opened 2023-04-20
Wildcards / globbing not working for --exclude anymore (since 6.12)


frittentheke:
##### Summary

We use the globbing pattern `ansible-lint --exclude '**/secrets.yml` to exclude binary files (which are binary due to applied git-crypt encryption). This worked fine until `creator-ee:v0.13.0 (ansible-lint 6.11) ` and broke with `creator-ee:v0.14.0 (ansible-lint 6.12.2)`

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
```

```
root@creator-ee:v0.16.0: /builds (main)
# ansible-lint --version
ansible-lint 6.14.6 using ansible 2.14.4
```
We are using the `quay.io/ansible/creator-ee:v0.16.0` container within a GitlabCI job

##### STEPS TO REPRODUCE

1. 
```console (paste below)
# Create a binary file which causes ansible-lint to fail
dd if=/dev/urandom of=secrets.yml bs=1 count=1000
```

2. Add that file to your group_vars

3. Run ansible-lint

4. You see a stack trace with a UnicodeDecodeError exception:

```
Traceback (most recent call last):                                                        
  File "/usr/local/bin/ansible-lint", line 8, in <module>                                 
    sys.exit(_run_cli_entrypoint())                                                       
             ^^^^^^^^^^^^^^^^^^^^^                                                        
  File "/usr/local/lib/python3.11/site-packages/ansiblelint/__main__.py", line 354, in _run_cli_entrypoint                                                                           
    sys.exit(main(sys.argv))                                                              
             ^^^^^^^^^^^^^^                                                               
  File "/usr/local/lib/python3.11/site-packages/ansiblelint/__main__.py", line 244, in main                                                                                          
    result = _get_matches(rules, options)                                                 
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                 
  File "/usr/local/lib/python3.11/site-packages/ansiblelint/runner.py", line 216, in _get_matches                                                                                    
    lintables = ansiblelint.utils.get_lintables(opts=options, args=options.lintables)                                                                                                
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                
  File "/usr/local/lib/python3.11/site-packages/ansiblelint/utils.py", line 965, in get_lintables                                                                                    
[...]

PR #3401, opened 2023-05-05
Fix exclude patterns logic

ssbarnea:
Fixes: #3310

### DIFF
--- a/src/ansiblelint/cli.py
+++ b/src/ansiblelint/cli.py
@@ -34,7 +34,6 @@ if TYPE_CHECKING:
 
 _logger = logging.getLogger(__name__)
 _PATH_VARS = [
-    "exclude_paths",
     "rulesdir",
 ]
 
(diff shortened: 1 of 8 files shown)


src/ansiblelint/cli.py
```diff
@@ -594,6 +594,11 @@ def get_config(arguments: list[str]) -> Options:
         msg = f"Failed to determine a valid project_dir: {options.project_dir}"
         raise RuntimeError(msg)
 
+    # expand user home dir in exclude_paths
+    options.exclude_paths = [
+        os.path.expandvars(os.path.expanduser(p)) for p in options.exclude_paths
```
ajinkyau, line 599: @ssbarnea I think we can use `Path.expanduser(p)`

ssbarnea: Not in this case, there is no expandvars inside pathlib and for these we need to stick to of paths as they are glob patterns, not real paths.

ajinkyau (approved):

PR #3405, opened 2023-05-05
Add Amazon Linux 2023 as a supported platform

artis3n:
Amazon Linux 2023 is a supported platform on Ansible Galaxy now - 
image

https://galaxy.ansible.com/api/v1/platforms/?search=Amazon+Linux+2023

So add the platform to ansible-lint's meta checks. And fix these failing linting errors that should pass :) 

https://github.com/artis3n/ansible-role-tailscale/actions/runs/4873967123/jobs/8694284112?pr=329#step:6:16

Fixes: #3411

### DIFF
--- a/src/ansiblelint/schemas/meta.json
+++ b/src/ansiblelint/schemas/meta.json
@@ -91,6 +91,25 @@
       "title": "Amazon Linux 2PlatformModel",
       "type": "object"
     },
+    "Amazon_Linux_2023PlatformModel": {
+      "properties": {
+        "name": {
+          "const": "Amazon Linux 2023",
+          "title": "Name",
+          "type": "string"
+        },
+        "versions": {
+          "default": "all",
+          "items": {
+            "enum": ["all"],
+            "type": "string"
+          },
+          "type": "array"
+        }
+      },
+      "title": "Amazon Linux 2023PlatformModel",
+      "type": "object"
+    },
     "ArchLinuxPlatformModel": {
       "properties": {
         "name": {
--- a/src/ansiblelint/schemas/meta.json
+++ b/src/ansiblelint/schemas/meta.json
@@ -72,36 +72,17 @@
       "title": "AmazonPlatformModel",
       "type": "object"
     },
-    "Amazon_Linux_2023PlatformModel": {
+    "AmazonLinuxPlatformModel": {
       "properties": {
         "name": {
-          "const": "Amazon Linux 2023",
+          "const": "Amazon Linux",
           "title": "Name",
           "type": "string"
         },
         "versions": {
           "default": "all",
           "items": {
-            "enum": ["all"],
-            "type": "string"
-          },
-          "type": "array"
-        }
-      },
-      "title": "Amazon Linux 2023PlatformModel",
-      "type": "object"
-    },
-    "Amazon_Linux_2PlatformModel": {
-      "properties": {
-        "name": {
-          "const": "Amazon Linux 2",
-          "title": "Name",
-          "type": "string"
-        },
-        "versions": {
-          "default": "all",
-          "items": {
-            "enum": ["all"],
+            "enum": ["all", "1", "2", "2013"],
             "type": "string"
           },
           "type": "array"
@@ -1225,7 +1206,7 @@
             "$ref": "#/$defs/AmazonPlatformModel"
           },
           {
-            "$ref": "#/$defs/Amazon_Linux_2PlatformModel"
+            "$ref": "#/$defs/AmazonLinuxPlatformModel"
           },
           {
             "$ref": "#/$defs/aosPlatformModel"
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
2023 is a version, not part of the distro name.

To add this correctly, we need a generic Amazon Linux Platform which would have these versions: 1, 2, 2023 like we see at https://hub.docker.com/_/amazonlinux

artis3n:
@ssbarnea Sure, I was following the previously established pattern in the meta scheme - `AmazonPlatformModel` for Amazon Linux, `Amazon_Linux_2PlatformModel` for Amazon Linux 2. As Amazon Linux 2023 is a new distro based on fedora vs. Amazon Linux 2 based on CentOS, I figured following that pattern made sense.

Also, now that I think of it, a generic Amazon Linux Platform doesn't follow Ansible Galaxy's schema model at https://galaxy.ansible.com/api/v1/platforms/ - does `ansible-lint` need to match Ansible Galaxy's schema expectations exactly? Let me know. I would assume ansible-lint is merely exposing the schema expectations of Galaxy, in which case I believe this PR is correct.

e.g. "Amazon Linux" shows 2 different top-level distros each with an "any" version - 2 and 2023. https://galaxy.ansible.com/api/v1/platforms/?search=Amazon+Linux

"Amazon" shows the original Amazon Linux - https://galaxy.ansible.com/api/v1/platforms/?search=Amazon , this is a distro that has many versions underneath it.

Setting a different model for Amazon Linux 2023 appears to match the behavior for distros like Ubuntu, in which new versions are entries - https://galaxy.ansible.com/api/v1/platforms/?search=Ubuntu . 

Amazon Linux 2 and 2023 don't follow these patterns, which is on Amazon, but leaves me confused how to ensure the correct schema behavior is modeled through ansible-lint.

ssbarnea:
Ansible Galaxy HUB currently allows any kind of data to uploaded for platform and version, but the linter does not. We aim to ensure people use consistent naming, avoiding confusing and increasing the overall quality of the content uploaded.

In time the upload will likely become more restrive too, so people will be able fix their metadata.

I will contact the hub team for some recommendation on this, but I am almost sure we want a single platform for `Amazon Linux` with `1`, `2`, and `2023` as known versions.

artis3n:
Understood! Let me know what you hear and I'll make the appropriate changes.

ssbarnea (approved):

ssbarnea:
I made some modification but let's not merge them before we get some feedback from galaxy/hub team as the side effect of fixing the mistake of including "Amazon Linux 2" (with version) as a platform was unfortunate and the fix will force people using the linter to remove the version from the platform name to pass linting.

src/ansiblelint/schemas/meta.json
```diff
@@ -38,6 +38,25 @@
...
+      "properties": {
+        "name": {
+          "const": "Amazon Linux",
+          "title": "Name",
+          "type": "string"
+        },
+        "versions": {
+          "default": "all",
+          "items": {
+            "enum": ["all", "1", "2", "2013"],
```
ajinkyau, line 51: Need `2023` instead of `2013`

alisonlhart:
I have fixed this on Galaxy:
Now the scheme is `Amazon Linux` with versions `1`, `2`, `2023`.

artis3n:
The `AmazonPlatformModel` remains as the versioned Amazon Linux 1 model, which a number of version strings. If `1` is being included in the `AmazonLinuxPlatformModel`, do we need to restructure that as well?

src/ansiblelint/schemas/meta.json
```diff
@@ -38,6 +38,25 @@
...
+        "versions": {
+          "default": "all",
+          "items": {
+            "enum": ["all", "1", "2", "2023"],
+            "type": "string"
+          },
+          "type": "array"
+        }
+      },
+      "title": "Amazon Linux 2PlatformModel",
```
artis3n, line 57:
```suggestion
      "title": "Amazon Linux PlatformModel",
```

awcrosby:
Some galaxy background on this... Platforms and cloud platforms will be going away from the galaxy api and ui. Note these are not “supported” platforms, but just a list of existing platforms. Platforms are used in legacy [galaxy.ansible.com](url) for standalone roles not collections (since they are not part of the collection spec). The addition of legacy roles into `ansible/galaxy_ng` are not fully featured beyond keeping automation working, and do not include platforms. Platforms are not parsed by `ansible/galaxy-importer` and are not displayed on [beta-galaxy.ansible.com](url).

ssbarnea:
@awcrosby Thanks for the reply. This means that from now on, DevTools team will take care of maintaining the platforms and their versions updated. We do think that this metadata is quite useful for other parts of the ecosystem.

src/ansiblelint/schemas/meta.json
artis3n: @ssbarnea is this a typo or intended title?

Issue #2784, opened 2022-12-06
Add prefix checking for facts and variables


ssbarnea:
##### Summary

As one of our clients suggested, they would like to enforce the presence of a particular prefix for all facts they use, something like:

* All facts within a role should match `{role_name_*}` pattern
* All inventory variables should match `inv_*` pattern
* All credentials from tower/app should match `credential_*` pattern
* All survey variables should match `survey_*` pattern

It should be noted that this rule is closely related to one that we already have for `loop_var` prefix, where we require users to use unique variables, especially within roles.

##### Notes

* facts seems relatively easy to implement, by checking `set_fact`. Still we need to add some context information, so a task would know if is part of a role (and its name) or a playbook.
* inventory could be implemented at source but likely only for static YAML inventory files, as we do not have plans to start linting `ini` files. Also, it will likely not work for dynamic inventories, at least initially. Still, we could use ansible-inventory output and check only this, making it work for any kind of inventory, but we will not be able to tell exactly where the non-conformance comes from. Still, would help with enforcement.
* No idea about how to check for credential and survey. We need to investigate that.
* We could check that any used variable is matching one of these patterns, making much harder for a user to accidentally introduce non-conforming names.
* What should we do about locally used vars? (contained in one task), like below:

```yaml
- name: Print something
  vars:
    foo: "some text"
  ansible.builtin.debug:
    msg: "{{ foo }}"
```

In my view the example above is legit as this variable can be considered "local", so there is no real need to enforce prefixes on it.

Still, if that `vars:` would be at playbook level or even a block level, I could see more reasons to be careful with its naming.

It is interesting to mention that I did not see any proposal from the customer for adding prefixes to role `vars/` and `defaults/`. I will wait for feedback and update the ticket.

As usual, any feedback is welcomed.

MarkusTeufelberger:
Another case to consider is `register: foo` in tasks where other subsequent tasks then use the output of this action.

ssbarnea:
@MarkusTeufelberger You are right and I was already thinking about it. For example, I would not want to discourage uses of `register: result` if the result is used right away in the next task. Still, if that variable is used later, it would clearly need a better name.

That would an interesting rule to write, to scan the task tree for bad usages while ignoring basic ones. Still, it would clearly help some people improve their playbook authoring skills.

MarkusTeufelberger:
[...]

PR #3422, opened 2023-05-09
Upgrade var-naming rule to include role name prefix

ssbarnea:
Fixes: #3221
Fixes: #3197
Fixes: #2784


cidrblock (approved):

ssbarnea:
@gardar Please raise a ticket for this request. I can understand your case, especially with an well establish code base. I already contacted @cidrblock so we can consult with others and see if we should relax this rule or not. For the moment you made the right call of temporary disabling it for your repo.

PR #3437, opened 2023-05-12
Do not trigger no_handler upon complex conditions

### LINKED ISSUE #1632: no-handler: should not react on when-conditions containing "and" or "or"

klaus-tux:
Fixes #1632

This change suppresses the no_handler rule in the following cases:
* if the condition contains "or"
* if the list of conditions has more than one entry

With the existing implementation this leads to the situation that the rule does not trigger if there are multiple condition or if the condition contains "and" or "or".


klaus-tux:
@StopMotionCuber Feel free to have a look at this PR for #1632

ssbarnea (approved):
LGTM but I want extra pair of eyes from others before merging this change, especially to estimate if the new behavior is the optimal one.

ajinkyau (approved):

StopMotionCuber:
LGTM

I have thought a bit about this and it doesn't trigger when doing something like `x.changed or x.changed`. But I came to the conclusion that probably no sane person would do that and instead just add `noqa` to ignore the linting instead of finding a loophole.

Issue #2924, opened 2023-01-20
ansible-lint warns when about FQCN when linting collection with dependencies


AKorezin:
##### Summary

When I lint my collection with dependencies, `ansible-lint` prints warnings about FQCN related to dependencies.

```
WARNING  Unable to load module ansible.posix.sysctl at roles/test/tasks/main.yml:2 for options validation
WARNING  Unable to resolve FQCN for module ansible.posix.sysctl
```

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 6.11.0 using ansible 2.14.1
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

1. Install `ansible-lint==6.11.0` and `ansible-core=2.14.0`.
1. Install git.
1. Create a new collection.
1. Add dependencies to `galaxy.yml`.
   ```yaml
   dependencies:
     ansible.posix: "*"
   ```
1. Create a role in `roles` dir.
1. Create a task with `ansible.posix.sysctl`.
1. Run `ansible-lint`.

##### Desired Behavior

Properly find installed dependencies.

##### Actual Behavior

You can check the repository with an example https://gitlab.com/AKorezin/ansible-lint-issue-2924.
I use gitlab-ci to lint the collection. You can check the script [here](https://gitlab.com/AKorezin/ansible-lint-issue-2924/-/blob/main/.gitlab-ci.yml).
Job logs
- [simple](https://gitlab.com/AKorezin/ansible-lint-issue-2924/-/jobs/3636944451)
- [verbose](https://gitlab.com/AKorezin/ansible-lint-issue-2924/-/jobs/3637013087)

```console
# ansible-lint -vvvvvv
DEBUG    Logging initialized to level 10
[...]

PR #3446, opened 2023-05-15
Install requirements from galaxy.yml and tests/requirements.yml

ssbarnea:
Fixes: #2924


ajinkyau (approved):

isuftin:
Don't know if anyone's watching this, but at what point would we be putting a requirements.yml in tests/ ?

ssbarnea:
These are test requirements, not runtime ones. Still, for linting we need both.

Issue #3453, opened 2023-05-16
Update to var-naming[no-role-prefix] breaks on include_tasks (ansible-lint 6.16.1 using ansible-core:2.14.5)


bnogaro:
##### Summary

On 'ansible.builtin.include_tasks', two infrigements to the var-naming[no-role-prefix] rule are systematically infriged, for vars:

- \_\_file\_\_
- \_\_line\_\_

These variables are not defined in my playbook, I guess they are internal include_tasks variables;

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
ansible-lint 6.16.1 using ansible-core:2.14.5 ruamel-yaml:None ruamel-yaml-clib:None
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Run ansible-lint on a playbook with an include_tasks, either in free-form or with file: filename.yaml, usings vars:

```console
- name: Import task
      ansible.builtin.include_tasks: mytask.yml
      vars:
        role_name_my_var: 'toto'
```

##### Desired Behavior

No issue should be raised, as all variables follow the role_prefix rule

##### Actual Behavior

The following issues are raised:
```console
roles/myrole/tasks/main.yml:XX: var-naming[no-role-prefix]: Variables names from within roles should use role_name_ as a prefix. (vars: __file__)
roles/myrole/tasks/main.yml:XX: var-naming[no-role-prefix]: Variables names from within roles should use role_name_ as a prefix.  (vars: __line__)
```

tomaxsas:
same here

nodiscc:
This is not specific to the `include_task` module, I am getting this issue/false positive on a task that uses the `command` module:

```bash
var-naming[no-role-prefix]: Variables names from within roles should use role_name_ as a prefix. (vars: __file__)
roles/tt_rss/tasks/tt-rss.yml:120 Task/Handler: load/upgrade tt-rss database schema

var-naming[no-role-prefix]: Variables names from within roles should use role_name_ as a prefix. (vars: __line__)
roles/tt_rss/tasks/tt-rss.yml:120 Task/Handler: load/upgrade tt-rss database schema
```

```yaml
- name: load/upgrade tt-rss database schema
  become: yes
  become_user: tt-rss
  command:
    cmd: /usr/bin/php {{ tt_rss_install_dir }}/update.php --update-schema=force-yes
  register: tt_rss_db_upgrade
[...]

PR #3458, opened 2023-05-17
Fix for var-naming rule to not break on include_tasks and vars

ajinkyau:
Fixes: #3453

### DIFF
--- a/src/ansiblelint/rules/var_naming.py
+++ b/src/ansiblelint/rules/var_naming.py
@@ -10,7 +10,7 @@ from ansible.parsing.yaml.objects import AnsibleUnicode
 from ansible.vars.reserved import get_reserved_names
 
 from ansiblelint.config import options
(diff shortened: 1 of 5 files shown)


ssbarnea (changes requested):
-1 for missing any tests. We need to modify tests to ensure that we do not further regressions.

src/ansiblelint/rules/var_naming.py
```diff
@@ -46,44 +46,44 @@ def get_var_naming_matcherror(
                 message="Variables names must be strings.",
                 rule=self,
             )
+        if ident not in ANNOTATION_KEYS:
```
ssbarnea, line 49: If you reverse the condition you can just return None fast and avoid adding a new level of indentation to this function. Basically we already know that we don't need to do anything if the key (ident) is one of these.

ssbarnea (approved):

Issue #3456, opened 2023-05-17
var-naming[no-role-prefix] should not trigger on variables starting with ansible_ (builtins)


fanderchan:
##### Summary 
ansible-lint reports an error requiring my role variable name to add the prefix "role_name_", but modifying the built-in variable ansible_python_interpreter causes the program to crash.
##### Issue Type 
- Bug Report 
##### OS / ENVIRONMENT
```
ansible-lint --version
ansible-lint 6.16.1
```
- ansible installation method: pip
- ansible-lint installation method: pip
##### STEPS TO REPRODUCE 
```yaml
- name: Set Python interpreter
  set_fact: 
    ansible_python_interpreter: /usr/bin/python3
```  
When running ansible-lint check, the error reported: 
```  
WARNING  Listing 1 violation(s) that are fatal  
var-naming[no-role-prefix]: Variables names from within roles should use role_name_ as a prefix. (set_fact: ansible_python_interpreter)
roles/pre_check_and_set/tasks/main.yml:15 Task/Handler: Set Python interpreter
```
##### Desired Behavior
ansible-lint should skip checking ansible built-in variables to avoid requiring prefixes and causing program errors. 
##### Actual Behavior
ansible-lint requires role variable names to be prefixed with "role_name_", but modifying the built-in variable ansible_python_interpreter causes the program to crash.
  
"role_name_ansible_python_interpreter": /usr/bin/python3  
  
Please let ansible-lint skip checking ansible built-in variable names to avoid similar issues.

ajinkyau:
@fanderchan When I modifying the built-in variable, and run ansible-lint on it the program doesn't crash for me.
And I think it would not be possible to skip the built-in variable check.
@ssbarnea thoughts?

ssbarnea:
That is indeed a bug, we should just ignore any name starting with `ansible_` as maintaining a list would be probably too much.

MarkusTeufelberger:
There is also https://docs.ansible.com/ansible/latest/reference_appendices/config.html#inject-facts-as-vars (see discussion at https://old.reddit.com/r/ansible/comments/13giahm/is_it_preferable_to_utilize_ansible/) - probably there should even be a rule that actually only allows `ansible_facts` and disallows anything that depends on `INJECT_FACTS_AS_VARS=True`?

Especially considering the comment by @sivel there:

[...]

PR #3462, opened 2023-05-18
Enable var-name rule to detect read-only variables

ssbarnea:
Related: #3456

### DIFF
--- a/src/ansiblelint/rules/var_naming.md
+++ b/src/ansiblelint/rules/var_naming.md
@@ -9,6 +9,11 @@ alphabetic or underscore `_` character.
 For more information see the [creating valid variable names][var-names] topic in
 Ansible documentation and [Naming things (Good Practices for Ansible)][cop].
 
+You should also be fully aware of [special variables][magic-vars], also known as
+magic variables, especially as most of them can only be read. While Ansible will
+just ignore any attempt to set them, the linter will notify the user, so they
+would not be confused about a line that does not effectively do anything.
+
 Possible errors messages:
 
 - `var-naming[non-string]`: Variables names must be strings.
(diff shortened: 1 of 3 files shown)


src/ansiblelint/rules/var_naming.py
```diff
@@ -33,6 +33,53 @@ class VariableNamingRule(AnsibleLintRule):
...
+    # does not include connection variables, which we expect users to tune in
+    # specific cases.
+    # https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html
+    read_only_names = {
+        "ansible_check_mode",
+        "ansible_collection_name",
+        "ansible_config_file",
+        "ansible_dependent_role_names",
+        "ansible_diff_mode",
+        "ansible_facts",  # special fact
```
bcoca, line 46: this is actually updatable by user on purpose .. so they can 'fake facts' if needed

ssbarnea: Thanks, I moved it to the allowed list.

src/ansiblelint/rules/var_naming.md
```diff
@@ -7,7 +7,9 @@ underscore `_` character. Variable names must also start with either an
 alphabetic or underscore `_` character.
 
 For more information see the [creating valid variable names][var-names] topic in
-Ansible documentation and [Naming things (Good Practices for Ansible)][cop].
+Ansible documentation and [Naming things (Good Practices for Ansible)][cop]. You
+should also be fully aware of [special variables][magic-vars], also known as
+magic variables, especially as most of them can only be read.
```
bcoca, line 12: maybe inform that 'while it is possible to assign to them w/o an error, this will be ignored'

cidrblock (approved):

Issue #3466, opened 2023-05-19
ansible-lint lints successfully non-existing directories


vlotorev:
##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 6.16.1 using ansible-core:2.14.5 ruamel-yaml:None ruamel-yaml-clib:None
```
- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

```console (paste below)
mkdir tmp
cd tmp
ansible-lint asdf
Passed with production profile: 0 failure(s), 0 warning(s) on 1 files.
```

##### Desired Behavior
ansible-lint fails (exits with non-zero status code) and reports what items (directories) are not found.

##### Actual Behavior

ansible-lint finishes successfully and exits with zero status code

ssbarnea:
I am quite curious how you install the linter considering that it reports version none for ruamel-yaml, which indicates that it will either not work or that we have another bug related to reading its version.

What is the output of `pip freeze|grep ruamel` ?

PR #3467, opened 2023-05-20
Ensure we return error if argument gives are not present

ssbarnea:
Fixes: #3466

### DIFF
--- a/src/ansiblelint/runner.py
+++ b/src/ansiblelint/runner.py
@@ -188,6 +188,16 @@ class Runner:
                     ),
                 )
                 lintable.stop_processing = True
+            # identify missing files/folders
+            if not lintable.path.exists():
+                matches.append(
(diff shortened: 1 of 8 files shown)


src/ansiblelint/runner.py
```diff
@@ -188,6 +188,16 @@ def _run(self) -> list[MatchError]:  # noqa: C901
                     ),
                 )
                 lintable.stop_processing = True
+            # identify missing files/folders
+            if not lintable.path.exists():
+                matches.append(
+                    MatchError(
+                        lintable=lintable,
+                        message="File or found not found.",
```
vlotorev, line 196: File or directory not found?

ssbarnea: Thanks you for reviewing it. I included a fix in next PR.

Issue #3421, opened 2023-05-09
`ansible-lint` does not contain the v3 `ansible-builder` execution environment schema


cidrblock:
https://github.com/ansible/ansible-lint/blob/34735422c1ab09cf2dff387e350a75aa454bec34/src/ansiblelint/schemas/execution-environment.json#L1

builder is at 3.0rc2, when 3.0 is release we should be able to switch to the new schema

https://github.com/ansible/ansible-builder/blob/5a748e08bf2f56ea72ec15cc378690f8e1adc0b2/src/ansible_builder/ee_schema.py#L197

PR #3480, opened 2023-05-22
Add support for ansible-builder v3 schema

ajinkyau:
Fixes: #3421
- Added support for ansible-builder v3 schema


ssbarnea (approved):

ssbarnea (changes requested):
Please also add one sample test file using v3 schema that makes used of all allowed fields.

ssbarnea (changes requested):
Please also add one sample test file using v3 schema that makes used of all allowed fields.

Issue #3481, opened 2023-05-23
Allowed become_methods in ansible-lint don't match with what ansible provides


romankuchin:
##### Summary

List of become_methods in ansible:
```
$ ansible-doc -t become -l
ansible.netcommon.enable     Switch to elevated permissions on a network device
community.general.doas       Do As user
community.general.dzdo       Centrify's Direct Authorize
community.general.ksu        Kerberos substitute user
community.general.machinectl Systemd's machinectl privilege escalation
community.general.pbrun      PowerBroker run
community.general.pfexec     profile based execution
community.general.pmrun      Privilege Manager run
community.general.sesu       CA Privileged Access Manager
community.general.sudosu     Run tasks using sudo su -
runas                        Run As user
su                           Substitute User
sudo                         Substitute User DO
```

What ansible-lint allows: https://github.com/ansible/ansible-lint/blob/42329279496085a30946c1c93fc2fa8e0dbb2d5b/src/ansiblelint/schemas/ansible.json#L43-L65

```
"oneOf": [
        {
          "enum": [
            "sudo",
            "su",
            "pbrun",
            "pfexec",
            "runas",
            "dzdo",
            "ksu",
            "doas",
            "machinectl"
          ],
          "type": "string"
        },
        {
          "$ref": "#/$defs/full-jinja"
        }
      ],
```
##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
$ ansible-lint --version
ansible-lint 6.16.2 using ansible-core:2.13.5 ruamel-yaml:0.17.26 ruamel-yaml-clib:0.2.7
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Valid ansible tasks:

```console (paste below)
  - command: whoami
    become: true
    become_method: community.general.sudosu
    become_user: nessus
    register: output
  
  - debug:
      var: output
```

##### Desired Behavior

Ansible-lint doesn't generate error on above code.

##### Actual Behavior

```paste below
$ ansible-lint playbook.yaml
schema[playbook]: {'command': 'whoami', 'become': True, 'become_method': 'community.general.sudosu', 'become_user': 'nessus', 'register': 'output'} is not valid under any of the given schemas
playbook.yaml:1  Returned errors will not include exact line numbers, but they will mention
the schema name being used as a tag, like ``schema[playbook]``,
``schema[tasks]``.

This rule is not skippable and stops further processing of the file.

[...]

PR #3499, opened 2023-05-26
Check validity of 'become_method' values from code

ajinkyau:
Related: #3481


src/ansiblelint/rules/schema.py
```diff
@@ -75,6 +77,49 @@ class ValidateSchemaRule(AnsibleLintRule):
...
+                    message=f"'become_method' must be one of the currently installed plugins: {', '.join(available_become_plugins_keys)}",
+                    lintable=file or Lintable(""),
+                    rule=ValidateSchemaRule(),
+                    details=ValidateSchemaRule.description,
+                    tag=f"schema[{file.kind}]",
+                ),
+            )
+
+        # check at task level
+        tasks = data.get("tasks", [])
```
ssbarnea, line 106: This is not how you look inside tasks, we have `matchtask` method to implement which is called for each task within the rule. Look at parent implementation to see how is called.

Basically matchplay should never look inside `tasks` as we have a specific method that takes care of task level checks.

src/ansiblelint/rules/schema.py
```diff
@@ -120,6 +165,9 @@ def matchyaml(self, file: Lintable) -> list[MatchError]:
                     tag=f"schema[{file.kind}]",
                 ),
             )
+
+        if not result:
+            result = super().matchyaml(file)
         return result
```
ssbarnea, line 173: That is missing tests, it needs at least two tests, one with playbook and one with a task file.

ssbarnea: I would also recommend making this easier to extend for other similar checks, where we want to verify that a specific field is one of list of generated values, maybe similar to:

```python
FIELD_CHECKS = {
"become_method": get_app().runtime.plugins.become.keys()
}
```


Also be sure that this is defined at module or class level and not inside the method being called.

The effective error message can be built from this dictionary, so we would not need to create new message for each field check that we add later.

ssbarnea (changes requested):

src/ansiblelint/rules/schema.py
ajinkyau: updated

ssbarnea (approved):

PR #3513, opened 2023-05-31
Allow use as a reusable github action

ssbarnea:
- add self testing
- fixes current problems preventing use as an action
- make use of pip caching for linter requirements

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -39,11 +39,21 @@ jobs:
             eco
             py-devel
           platforms: linux,macos
-
+  test-action:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


ssbarnea:
This is not ready yet because it uncovers concurrency errors during testing when calling ansible-galaxy. We will need to improve ansible-compat in order to make it work better when called in parallel.

evgeni:
> This is not ready yet because it uncovers concurrency errors during testing when calling ansible-galaxy. We will need to improve ansible-compat in order to make it work better when called in parallel.

Ah, yes, I've just reported it as a dedicated issue (https://github.com/ansible/ansible-lint/issues/3560) as it happened to me in a different context.

PR #3527, opened 2023-06-01
Handling of exclude paths with trailing slashs

### LINKED ISSUE #3524: Directories with trailing slash in .gitignore will not be exluded in version 6.17
##### Summary

directories with trailing slashes in `.gitignore` won't be exluded in ansible-lint 6.17.0.
I guess this is related to the reimplementation of the file exclusion logic (https://github.com/ansible/ansible-lint/pull/3507).

This statement https://github.com/ansible/ansible-lint/blob/98063af1d3b95475dc60e3eecbd9bec5c1683862/src/ansiblelint/file_utils.py#L541 will return `false` for a comparison of Path(`foo`) with `foo/` in the pathspec object. It will even return `false` for `Path('foo/')` as pathlib will strip trailing slashes (see https://github.com/python/cpython/issues/65238)

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.17.0 using ansible-core:2.14.6 ruamel-yaml:None ruamel-yaml-clib:None
```

##### STEPS TO REPRODUCE

```console (paste below)
echo "foo/" > .gitignore
mkdir foo
echo 'foobar' > foo/bar.yml
ansible-lint -v
```

This will result in following error:

```
INFO     Loading ignores from .gitignore
load-failure: Failed to load or parse file.
foo/bar.yml:1


             Rule Violation Summary              
 count tag          profile rule associated tags 
     1 load-failure min     core, unskippable    

Failed after : 1 failure(s), 0 warning(s) on 2 files.
```

As you can see, it won't be excluded.

##### Desired Behavior

It should work the same as if the directory is specified without trailing slash:

```console (paste below)
echo "foo" > .gitignore
mkdir foo
echo 'foobar' > foo/bar.yml
ansible-lint -v
```

```
INFO     Loading ignores from .gitignore
INFO     Excluded: foo

Passed with production profile: 0 failure(s), 0 warning(s) on 1 files.
```

hille721:
Fixes: #3524

### DIFF
--- a/.config/requirements.in
+++ b/.config/requirements.in
@@ -7,7 +7,7 @@ black>=22.8.0 # MIT
 filelock>=3.3.0 # The Unlicense
 jsonschema>=4.10.0 # MIT, version needed for improved errors
 packaging>=21.3 # Apache-2.0,BSD-2-Clause
-pathspec>=0.9.0 # Mozilla Public License 2.0 (MPL 2.0)
+pathspec>=0.10.3 # Mozilla Public License 2.0 (MPL 2.0)
 pyyaml>=5.4.1 # MIT (centos 9 has 5.3.1)
 rich>=12.0.0 # MIT
 ruamel.yaml>=0.17.0,<0.18,!=0.17.29,!=0.17.30 # MIT, next version is planned to have breaking changes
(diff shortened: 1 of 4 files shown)


ssbarnea (approved):

ssbarnea:
@hille721 I also added some tests as I was not convinced it worked and also because I don't want to accidentally lose this change when we refactor something els.e

Issue #3541, opened 2023-06-10
ansible-lint's version check makes incorrect assumptions


gotmax23:
##### Summary

ansible-lint's version check breaks completely or gives incorrect instructions in certain cases.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

Fedora Rawhide container running ansible-lint 6.16.2

With pipx:

```console (paste below)
$ pipx install ansible-lint==6.16.2
$ ansible-lint --version
WARNING: PATH altered to include /home/ME/.local/pipx/venvs/ansible-lint/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.
ansible-lint 6.16.2 using ansible-core:2.15.0 ruamel-yaml:0.17.31 ruamel-yaml-clib:0.2.7
A new release of ansible-lint is available: 6.16.2 → 6.17.0 Upgrade by running: pip install --upgrade ansible-lint
```

With RPM:

```
# dnf install ansible-lint
$ /usr/bin/ansible-lint --version
ansible-lint 6.16.2 using ansible-core:2.15.0 ruamel-yaml:0.17.24 ruamel-yaml-clib:0.2.7
Traceback (most recent call last):
  File "/usr/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/ansiblelint/__main__.py", line 310, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/ansiblelint/__main__.py", line 223, in main
    msg = get_version_warning()
          ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/ansiblelint/config.py", line 302, in get_version_warning
    pip = guess_install_method()
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/ansiblelint/config.py", line 229, in guess_install_method
    for _ in uninstallation_paths(dist):
  File "/usr/lib/python3.11/site-packages/pip/_internal/req/req_uninstall.py", line 45, in unique
    for item in fn(*args, **kw):
  File "/usr/lib/python3.11/site-packages/pip/_internal/req/req_uninstall.py", line 83, in uninstallation_paths
    raise UninstallationError(msg)
[...]

PR #3551, opened 2023-06-12
Avoid adding pip related messages if installation method is different

ajinkyau:
Fixes: #3541

### DIFF
--- a/src/ansiblelint/config.py
+++ b/src/ansiblelint/config.py
@@ -10,7 +10,7 @@ import urllib.request
 import warnings
 from dataclasses import dataclass, field
 from functools import lru_cache
-from importlib.metadata import PackageNotFoundError, version
+from importlib.metadata import PackageNotFoundError, distribution, version
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 from urllib.error import HTTPError, URLError
@@ -152,7 +152,7 @@ class Options:  # pylint: disable=too-many-instance-attributes,too-few-public-me
     config_file: str | None = None
     generate_ignore: bool = False
     rulesdir: list[Path] = field(default_factory=list)
-    cache_dir_lock: FileLock | None = None
+    cache_dir_lock: FileLock | None = None  # type: ignore[valid-type]
     use_default_rules: bool = False
     version: bool = False  # display version command
     list_profiles: bool = False  # display profiles command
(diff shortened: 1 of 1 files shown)


src/ansiblelint/config.py
```diff
@@ -204,7 +204,18 @@ def in_venv() -> bool:
 
 def guess_install_method() -> str:
     """Guess if pip upgrade command should be used."""
+    # pylint: disable=import-outside-toplevel
+    from importlib.metadata import distribution
```
ssbarnea, line 208: Put it outside, no need to use this approach in this case. Only with ansible core we need these.

ajinkyau: ack

src/ansiblelint/config.py
```diff
@@ -152,7 +152,7 @@ class Options:  # pylint: disable=too-many-instance-attributes,too-few-public-me
     config_file: str | None = None
     generate_ignore: bool = False
     rulesdir: list[Path] = field(default_factory=list)
-    cache_dir_lock: FileLock | None = None
+    cache_dir_lock: FileLock | None = None  # type: ignore[valid-type]
```
ssbarnea, line 155: I kinda doubt this is needed, probably filelock was not included as a dependency in mypy, which would be the correct way to address the issue.

ssbarnea (approved):

src/ansiblelint/config.py
```diff
@@ -205,6 +205,14 @@ def in_venv() -> bool:
 def guess_install_method() -> str:
     """Guess if pip upgrade command should be used."""
     package_name = "ansible-lint"
+
+    try:
+        if distribution(package_name).read_text("INSTALLER").strip() != "pip":  # type: ignore[union-attr]
```
gotmax23, line 210:
```
         if distribution(package_name).read_text("INSTALLER").strip() != "pip":  # type: ignore[union-attr]
```

This returns None if the file doesn't exist which can happen, as I don't think INSTALLER is a mandatory part of the dist-info directory, so calling .strip()` directly is incorrect. Mypy warns about this for a reason...

You can probably change this to

```
        if (distribution(package_name).read_text("INSTALLER") or "").strip() != "pip":
```

ajinkyau: Updated in PR https://github.com/ansible/ansible-lint/pull/3553

PR #3556, opened 2023-06-13
Change RuleCollection to use offline mode if options are not given

ssbarnea:
Improves testability. Hopefully, this should sort concurrency issues seen in related PR.
Related: #3554

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -244,7 +244,11 @@ def main(argv: list[str] | None = None) -> int:
     from ansiblelint.rules import RulesCollection
     from ansiblelint.runner import _get_matches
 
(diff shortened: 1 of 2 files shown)


src/ansiblelint/__main__.py
```diff
@@ -244,7 +244,11 @@ def main(argv: list[str] | None = None) -> int:
     from ansiblelint.rules import RulesCollection
     from ansiblelint.runner import _get_matches
 
-    rules = RulesCollection(options.rulesdirs, profile_name=options.profile)
+    rules = RulesCollection(
```
ajinkyau, line 247: @ssbarnea Instead of sending three params we can just send one param `options` as the value for other params is just part of `options` object / dict

Issue #2173, opened 2022-05-23
size[block]: Limit the maximum block depth


cidrblock:
Starting to pull in possible rule ideas from the community of practice: https://github.com/redhat-cop/automation-good-practices/pull/64

Ofter, using an include within a block leads to more readable playbooks and task file vs. a deep block depth.

It might not be possible to check across files and I'm not sure lint tracks block state, but within a single file we should be able to limit it.

2 was suggested only as a starting point

Ruchip16:
@ssbarnea I can try fixing this

ssbarnea:
@Ruchip16 Any updates on it? Asking as I might do it myself if you don't have time.

ajinkyau:
Closing since PR #3156 is merged

PR #3602, opened 2023-07-04
Limit the maximum block depth

Ruchip16:
Since inside tasks there can be nested blocks so this PR limits the depth of the block by a certain limit which helps to maintain the readability of the playbooks.

fixes: #2173


examples/playbooks/rule-complexity-fail.yml
```diff
@@ -26,3 +26,85 @@
...
+
+        - name: Nested Task 18
+          ansible.builtin.debug:
+            msg: "This is nested task 18"
+
+        - name: Nested Task 19
+          ansible.builtin.debug:
+            msg: "This is nested task 19"
+
+        - name: Nested Task 20
```
ajinkyau, line 108: @Ruchip16 we need one more task added here and it should fail as the limit is 20 tasks

ajinkyau: This change should result in 2 failures for this playbook on line:
https://github.com/ansible/ansible-lint/pull/3602/files#diff-8e317bdeb6fb7d81373aa20796059598fecdb9c8c50255d1edbc5d1d946d7a3aR91

ajinkyau (changes requested):

src/ansiblelint/rules/complexity.py
```diff
@@ -42,6 +45,32 @@ def matchplay(self, file: Lintable, data: dict[str, Any]) -> list[MatchError]:
             )
         return results
 
+    def matchtask(self, task: Task, file: Lintable | None = None) -> list[MatchError]:
+        """Check if the task is a block and count the number of items inside it."""
+        results: list[MatchError] = []
+
+        if task.get("action") == "block":
```
ajinkyau, line 52: @Ruchip16 This condition will not return true as `task.get("action")` returns a dict instead of a string

ssbarnea:
> hey @ssbarnea @ajinkyau can you review this please?

Please check the last commit and see all the extra changes I had to do there. The original code was not really working and the tests were failing to check for the behavior. Still, it was not too hard to fix.

cidrblock (approved):

Ruchip16:
> > hey @ssbarnea @ajinkyau can you review this please?
> 
> Please check the last commit and see all the extra changes I had to do there. The original code was not really working and the tests were failing to check for the behavior. Still, it was not too hard to fix.

the main thing i got wrong and really confused was with the example of the block, i although discussed that once with ajinkya, but nevertheless, apologies for the extra work and would contribute more to get a good hang of it!

bluikko:
Does this PR really limit maximum block depth, or tasks inside a block?

There seems to be a conflict between the PR name and the help files contained in the PR diff:
- PR name talks about "block depth", i.e. nested blocks.
- PR diff contains 2 new rule description items which only talk about number of tasks inside block.

Please clarify which one it is, something is wrong here.

Ruchip16:
`limit maximum block` depth means that it limits the depth of any block (nested or not) up to a certain number which is currently 20 by default, if the tasks inside a block are > 20 it would raise a warning message, so maximum block depth refers to the depth of the block and nesting is a scenario inside that, hope that clears your confusion

bluikko:
> `limit maximum block` depth means that it limits the depth of any block

It does not, at all. It strengthens my opinion that the words are misused here.  
"Depth" refers to how many nested layers there are in my opinion.

PR #3605, opened 2023-07-06
Allow to set gather_facts as templated boolean

noonedeadpunk:
Right now schema does not allow gather_facts to be a templated value, even though it would resolve to a proper boolean. This is quite a valid usecase for operators that have a set of playbooks that run each after another and want facts to be gathered only in some of them, while disable facts gathering conditionally for others.

This should enable users to use variables for facts gathering


noonedeadpunk:
Hm, how to setup required labels?

noonedeadpunk:
Realized I had to rename branch, so closed this one in favor of https://github.com/ansible/ansible-lint/pull/3606

PR #3610, opened 2023-07-11
command-instead-of-module: allow `git rev-parse`

JohnVillalovos:
If a user wants to call `git rev-parse`, allow this without raising a `command-instead-of-module` error.

One use would be to do: `git rev-parse HEAD` which will get the SHA value of the current HEAD of the git repository.

The `ansible.builtin.git` module does not have support for doing this.

### DIFF
--- a/src/ansiblelint/rules/command_instead_of_module.py
+++ b/src/ansiblelint/rules/command_instead_of_module.py
@@ -68,7 +68,7 @@ class CommandsInsteadOfModulesRule(AnsibleLintRule):
     }
 
     _executable_options = {
-        "git": ["branch", "log", "lfs"],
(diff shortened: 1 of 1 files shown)


ajinkyau:
@ssbarnea any thought on this, do we want to allow this change?

JohnVillalovos:
I looked at the three failing checks and don't believe this PR is the cause of them. Well I guess someone needs to add a label to this PR to fix the `ack / ack / ack (pull_request_target)` to fix that check. But the other two don't seem to be caused by this PR.

JohnVillalovos:
Please let me know if there is anything needed from me on this PR. Thanks.

Issue #3583, opened 2023-06-21
--write mode creates problematic yaml output with multiline strings


bnogaro:
##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
ansible-lint 6.17.1 using ansible-core:2.15.1 ruamel-yaml:0.17.26 ruamel-yaml-clib:0.2.7
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

As a mean of complying with the yaml[line-length] rule with long strings without spaces (URL, for instance), one could do the following:
```yaml
---
my_var: "
  Lorem_ipsum_dolor_sit_amet__consectetur_adipiscing_elit__Mauris_vitae_consectetur_lacus__ac_sagittis_metus__Curabitur_\
  nibh_diam__molestie_et_odio_a__volutpat_lacinia_nisi__Mauris_consequat_mi_felis__rutrum_sodales_sapien_varius_placerat
  "
```

This file passes all ansible-lint tests. However, running `ansible-lint --write` reassembles the two lines into one
This exceeds ansible-lint yaml[line-length] criterium, litterally meaning that ansible-lint --write created a ansible-lint violation.

```console
                Rule Violation Summary                
 count tag               profile rule associated tags 
     1 yaml[line-length] basic   formatting, yaml    
```

##### Desired Behavior

Hopefully, ansible-lint --write would preserve the line-length rule, not removing the escaped newlines

##### Actual Behavior

The variable is rewritten on a single line, failing the test.

```yaml
---
my_var: "Lorem_ipsum_dolor_sit_amet__consectetur_adipiscing_elit__Mauris_vitae_consectetur_lacus__ac_sagittis_metus__Curabitur_nibh_diam__molestie_et_odio_a__volutpat_lacinia_nisi__Mauris_consequat_mi_felis__rutrum_sodales_sapien_varius_placerat"
```

ajinkyau:
@bnogaro we will be happy to accept a PR for it.

bnogaro:
Before any contribution, I may need the input from a more experienced developer on this project.
As far as I understant the problem, this seems to come from the 'round-trip' functionality of ruamel.yaml, which doesn't seem to customisable in terms of max length of line in the output.
In your opinion, should it be a contribution to ruamel.yaml, or should we process the ruamel.yaml output with another tool to break long lines into multiline strings, using the yamllint configured max line length?

shatakshiiii:
[...]

PR #3653, opened 2023-08-10
Document `yaml[line-length]` rule

shatakshiiii:
Closes: https://github.com/ansible/ansible-lint/issues/3583
Related Ruamel issue: https://sourceforge.net/p/ruamel-yaml/tickets/474/

### DIFF
--- a/src/ansiblelint/rules/yaml.md
+++ b/src/ansiblelint/rules/yaml.md
@@ -53,6 +53,7 @@ Some of the detailed error codes that you might see are:
 - `yaml[empty-lines]` - _too many blank lines (...> ...)_
 - `yaml[indentation]` - _Wrong indentation: expected ... but found ..._
 - `yaml[key-duplicates]` - _Duplication of key "..." in mapping_
+- `yaml[line-length]` - _Line too long (... > ... characters)_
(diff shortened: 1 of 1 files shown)


src/ansiblelint/rules/yaml.md
```diff
@@ -72,6 +73,11 @@ for it does check these. If for some reason, you do not want to follow our
 defaults, you can create a `.yamllint` file in your project and this will take
 precedence over our defaults.
 
+## Additional Information
```
ajinkyau, line 76: should we mention the title as `Additional Information For Multiline Strings`.
That way user would be able to directly reference it instead of searching for multiline related info

shatakshiiii: I made the title as `Additional Information` so that if in future we need to add in other yaml rule specific info, it can directly go inside this title. But for now, we can name it as `Additional Information For Multiline Strings`.

shatakshiiii: @ajinkyau Done with the changes!

ajinkyau (approved):

cidrblock:
Good stuff @shatakshiiii

Issue #3226, opened 2023-03-29
--write mode creates problematic yaml output with list-of-lists


MarkusTeufelberger:
##### Summary

Running `ansible-lint --write` on https://github.com/ansible-collections/collection_template rewrites some files (sadly expected) but creates additional `yaml[hyphens]: Too many spaces after hyphen` errors along the way (unexpected).

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
WARNING: PATH altered to include /home/__removed___/.local/pipx/venvs/ansible-lint/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.
ansible-lint 6.14.3 using ansible 2.14.1
```

- ansible installation method: pipx
- ansible-lint installation method: pipx

##### STEPS TO REPRODUCE

```console (paste below)
git clone https://github.com/ansible-collections/collection_template collection_template
cd collection_template
ansible-lint --write
# some errors/warnings...
ansible-lint --write
# different errors/warnings, including new yaml ones
```

##### Desired Behavior

Do not introduce additional issues after `--write`

##### Actual Behavior

```paste below
sections:
- - major_changes
  - Major Changes
- - minor_changes
  - Minor Changes
- - breaking_changes
  - Breaking Changes / Porting Guide
- - deprecated_features
  - Deprecated Features
- - removed_features
  - Removed Features (previously deprecated)
- - security_fixes
  - Security Fixes
- - bugfixes
  - Bugfixes
- - known_issues
  - Known Issues

gets rewritten to:

sections:
  -   - major_changes
      - Major Changes
  -   - minor_changes
      - Minor Changes
  -   - breaking_changes
      - Breaking Changes / Porting Guide
  -   - deprecated_features
      - Deprecated Features
  -   - removed_features
      - Removed Features (previously deprecated)
  -   - security_fixes
      - Security Fixes
  -   - bugfixes
      - Bugfixes
  -   - known_issues
      - Known Issues

this apparently looks even worse than before to yamllint
```

klaus-tux:
I had a look into this and found the following:
Setting `self.compact_seq_seq` to `False` in `FormattedYAML` produces the extra linebreaks, but the nested sequences are indented too far:
```
sections:
  -
      - major_changes
      - Major Changes
  -
      - minor_changes
      - Minor Changes
...
```
[...]

PR #3671, opened 2023-08-17
Limit sequence-of-sequence indents to a single space

Qalthos:
This isn't ideal, but it does appropriately align the sequence items. Further work is probably desired to compact them into nested lists.

Fixes: https://github.com/ansible/ansible-lint/issues/3226

Before:

```
---
sequence:
  -    -    - 111
            - 112
       - 12
  -    - 21
       -    - 221
```

After:

```
---
sequence:
  -
    -
      - 111
      - 112
    - 12
  -
    - 21
    -
      - 221
```

Ideal:

```
---
sequence:
  - - - 111
      - 112
    - 12
  - - 21
    - - 221
```


src/ansiblelint/yaml_utils.py
```diff
@@ -598,6 +598,36 @@ def choose_scalar_style(self) -> Any:
...
+                #              self.best_map_indent
+                # self.indent = self.best_sequence_indent
+                self.indent = self.requested_indent
+            else:
+                self.indent = 0
+        elif not indentless:
+            if self.indents.last_seq():
+                self.indent += self.best_sequence_indent  # - self.sequence_dash_offset
+                if sequence:
+                    self.indent -= self.sequence_dash_offset
```
Qalthos, line 620: This is what's doing the squeezing. If the _last_ indentation it desired was for a sequence, and this one is too, then ignore the dash offset.

This is likely also the reason for the extra newlines. ruamel is getting confused about where in the file we are, possibly as a consequence of `self.indents` getting out of sync? `self.column` ends up referring to a place far beyond the column we are at in the file and inserts a newline to try to catch up.

This can all probably get smartened up, this is just the first version I produced after playing around with it that made valid YAML instead of a huge mess.

Issue #3670, opened 2023-08-17
Running --write on a project having Ansible files results in stack trace


ganeshrn:
##### Summary
When ansible-lint (version 6.17.2) is executed with --write option against the repository https://github.com/geerlingguy/ansible-for-devops, it fails with:

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
6.17.2
```

- ansible installation method: one of source, pip, OS package
- ansible-lint installation method: one of source, pip, OS package

##### STEPS TO REPRODUCE

```console (paste below)
$ ansible-lint --write ansible-for-devops 
Traceback (most recent call last):
  File "/home/ttakamiy/.local/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ansiblelint/__main__.py", line 317, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ansiblelint/__main__.py", line 281, in main
    _do_transform(result, options)
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ansiblelint/__main__.py", line 192, in _do_transform
    transformer.run()
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ansiblelint/transformer.py", line 98, in run
    ruamel_data = yaml.loads(data)
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ansiblelint/yaml_utils.py", line 931, in loads
    data = self.load(stream=text)
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ruamel/yaml/main.py", line 426, in load
    return constructor.get_single_data()
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ruamel/yaml/constructor.py", line 111, in get_single_data
    node = self.composer.get_single_node()
  File "/home/ttakamiy/.local/lib/python3.9/site-packages/ruamel/yaml/composer.py", line 75, in get_single_node
    raise ComposerError(
ruamel.yaml.composer.ComposerError: expected a single document in the stream
  in "<unicode string>", line 5, column 1:
    kind: ClusterRole
    ^ (line: 5)
but found another document
  in "<unicode string>", line 29, column 1:
    ---
    ^ (line: 29)
```

##### Desired Behavior

[...]

PR #3679, opened 2023-08-22
Use yaml.load_all for multiple yaml document

shatakshiiii:
Fixes: https://github.com/ansible/ansible-lint/issues/3670


ajinkyau (approved):

ssbarnea (changes requested):
That change has not tests, please adds at least one test that proves that the new code is able to deal with multi-document YAML files.

ssbarnea (approved):

Issue #3634, opened 2023-07-27
[RFE] Implement transform (--write) functionality for applicable rules


ganeshrn:
##### Summary

Implement write functionality for the rules in the below list as applicable.

 Must have:
- [x] [command-instead-of-shell](https://ansible.readthedocs.io/projects/lint/rules/command-instead-of-shell/#correct-code) https://github.com/ansible/ansible-lint/pull/3675
- [x] [deprecated-local-action](https://ansible.readthedocs.io/projects/lint/rules/deprecated-local-action/#correct-code) https://github.com/ansible/ansible-lint/pull/3689
- [x] [jinja[spacing]](https://ansible.readthedocs.io/projects/lint/rules/jinja/) https://github.com/ansible/ansible-lint/pull/3691
- [x] [key-order](https://ansible.readthedocs.io/projects/lint/rules/key-order/) https://github.com/ansible/ansible-lint/pull/3705
- [x] [no-free-form](https://ansible.readthedocs.io/projects/lint/rules/no-free-form/) https://github.com/ansible/ansible-lint/pull/3725
- [x] [no-jinja-when](https://ansible.readthedocs.io/projects/lint/rules/no-jinja-when/) https://github.com/ansible/ansible-lint/pull/3721
- [x] [no-log-password](https://ansible.readthedocs.io/projects/lint/rules/no-log-password) https://github.com/ansible/ansible-lint/pull/3708
- [x] [partial-become](https://ansible.readthedocs.io/projects/lint/rules/partial-become/) https://github.com/ansible/ansible-lint/pull/3692

Good to have:

- [ ] [deprecated-module](https://ansible.readthedocs.io/projects/lint/rules/deprecated-module/)
- [ ] [empty-string-compare ](https://ansible.readthedocs.io/projects/lint/rules/empty-string-compare/)
- [x] [fqcn](https://ansible.readthedocs.io/projects/lint/rules/fqcn/)
- [ ] [inline-env-var/](https://ansible.readthedocs.io/projects/lint/rules/inline-env-var/)
- [ ] [literal-compare](https://ansible.readthedocs.io/projects/lint/rules/literal-compare/)
- [ ] [no-tabs](https://ansible.readthedocs.io/projects/lint/rules/no-tabs)
- [ ] [risky-file-permissions](https://ansible.readthedocs.io/projects/lint/rules/risky-file-permissions/)
[...]

PR #3691, opened 2023-08-28
Add auto-fixing implementation for `jinja[spacing]` rule

ajinkyau:
Related: #3634
Depends on #3693 

When finding jina2 spacing errors a recommendation is produced, the key, value and path within the task is known.  This information is added to the MatchError to be used later.

When transforming for a specific MatchError, use the metadata captured during the identification of the spacing issue to iterate the task and when a matching key and value are found, use the previously generated recommendation.

Because the MatchError was generated using a "normalized task" and the transform is operating on the original task, the path cannot be used to locate the key. This introduces the possibility or 2 identical key/value combinations being corrected out of order and is noted in the doc string of the respective function.

### DIFF
--- a/examples/playbooks/transform-jinja.transformed.yml
+++ b/examples/playbooks/transform-jinja.transformed.yml
@@ -1,0 +1,24 @@
+---
+- name: Fixture
+  hosts: localhost
+  vars:
+    my_list:
+      - foo
+      - bar
+  tasks:
+    - name: A block used to check that we do not identify error at correct level
+      block:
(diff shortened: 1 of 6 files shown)


cidrblock (changes requested):
I think we can persist the recommendation in the class instance rather than parse are own message

cidrblock (approved):
I'll approve, but @ssbarnea will need to as well since I've got commits in here.

ssbarnea (approved):

PR #3692, opened 2023-08-28
Transform for `partial-become` rule

shatakshiiii:
Related: https://github.com/ansible/ansible-lint/issues/3634


cidrblock (approved):
I'll approve, but will need additional approvals since I have commits.

TY @shatakshiiii

Issue #3645, opened 2023-08-04
ansible-lint fails validating github workflows due their use of "on:" keyword


johanneskastl:
This is  the same issue as [reported in the old ansible-lint-action](https://github.com/ansible/ansible-lint-action/issues/178)

##### Summary

ansible-lint fails while validating the `.github/workflows/ansible-lint.yml` file.

Example: https://github.com/johanneskastl/ansible-role-obs_server/actions/runs/5759297496/job/15613149480

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

GitHub action installed as [documented in the marketplace readme](https://github.com/marketplace/actions/run-ansible-lint)

##### STEPS TO REPRODUCE

Add `.github/workflows/ansible-lint.yml` file to repo, push, let the action run.

##### Desired Behavior

The action should either ignore the workflows files or it should spit out a useful warning, that is actually related to the file's contents. There is no "truthy value" in that file at all, and not in line 2...

##### Actual Behavior

The action files due to an error in the workflows file:
https://github.com/johanneskastl/ansible-role-obs_server/actions/runs/5759297496/job/15613149480

ssbarnea:
This is not really a bug in ansible-lint and is caused by Github not following the YAML specification. Read https://github.com/adrienverge/yamllint/issues/430 for details.

We currently do not plan to do anything about it. If you want, you can add `.github` folder to exclude to avoid being annoyed by this.

johanneskastl:
Thanks for the hints and explanation, @ssbarnea !

PR #3698, opened 2023-08-30
Avoid yamllint truthy violations with github workflows

ssbarnea:
Fixes: #3645


Qalthos (approved):

cidrblock (commented):
what else could we do right? :)

Issue #2941, opened 2023-01-25
--write mode has issues with block indentation indicators


MarkusTeufelberger:
##### Summary

If blocks are indented using indentation indicators (https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator), the `--write` mode seems to have a hard time and constantly re-formats the file until it outputs invalid YAML

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.11.0 using ansible 2.14.1
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

When trying to automatically prettify a larger set of Ansible roles, I detected some strange diffs in files with multiline blocks. I fixed some other stuff in the meantime and after 2-3 ansible-lint runs, it suddenly started crashing.

Here is a reproducing case:

```console (paste below)
foo.yml:
---
- name: Demo
  hosts: all
  tasks:
    - name: Demo
      ansible.builtin.debug:
        msg: |3
               multi
             line
             message

```

##### Desired Behavior

Either leave this as-is or at least don't break it over time.

##### Actual Behavior

Please give some details of what is happening.
Include a [minimum complete verifiable example] with:

- minimized playbook to reproduce the error
- the output of running ansible-lint including the command line used
- if you're getting a stack trace, also the output of
  `ansible-playbook --syntax-check playbook`

```paste below
# First run:
ansible-lint --write foo.yml

---
- name: Demo
  hosts: all
  tasks:
    - name: Demo
      ansible.builtin.debug:
        msg: |4
              multi
            line
            message

# Running again:
ansible-lint --write foo.yml

---
- name: Demo
  hosts: all
  tasks:
    - name: Demo
      ansible.builtin.debug:
        msg: |4
            multi
          line
          message

# Running _again_:
ansible-lint --write foo.yml
# Now it crashes:
ruamel.yaml.parser.ParserError: while parsing a block mapping
  in "<unicode string>", line 7, column 9:
            msg: |4
            ^ (line: 7)
expected <block end>, but found '<scalar>'
  in "<unicode string>", line 9, column 11:
              line
              ^ (line: 9)

```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

ssbarnea:
To be honest, I am not surprised it chokes on this feature. I am aware of reformatting bugs with far more popular YAML options, that one being one that I never used myself.

[...]

PR #3700, opened 2023-08-31
Test transformer with block indentation indicators

shatakshiiii:
Fixes: https://github.com/ansible/ansible-lint/issues/2941

With the current version of ruamel.yaml, ansible-lint transform is not producing the stack trace mentioned in the linked issue. And the --write mode behavior is as expected, which we would be testing as part of this PR.


ssbarnea (approved):
Thanks!

Issue #3696, opened 2023-08-30
WARNING  Unable to resolve FQCN for module ansible.builtin.include


lisenet:
##### Summary

Ansible-lint cannot resolve FQCN for module `ansible.builtin.include`.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.18.0 using ansible-core:2.15.3 ansible-compat:4.1.6 ruamel-yaml:None ruamel-yaml-clib:None
```

- ansible installation method: pip
- ansible-lint installation method: pip
- OS: Rocky 9.2
- Python: 3.9

##### STEPS TO REPRODUCE

Running `ansible-lint` against the following piece of code results in a violation of the rule `fqcn[action-core]`.

```console (paste below)
---
- name: Include playbook
  include: Linux.yml
  when: ansible_system == 'Linux'
```

Resolving the issue by using an FQCN (see documentation [here](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_module.html)) below results in the warning reported in this issue:

```console (paste below)
---
- name: Include playbook
  ansible.builtin.include: Linux.yml
  when: ansible_system == 'Linux'
```

Ansible-lint output:

```
WARNING  Unable to resolve FQCN for module ansible.builtin.include
```

Ansible-lint configuration below.

```
---
# .ansible-lint
profile: production

exclude_paths:
  - collections

parseable: true
quiet: true
strict: true
verbosity: 1

var_naming_pattern: "^[a-z_][a-z0-9_]*$"

use_default_rules: true

skip_list:
  - role-name
  - deprecated-module
  - yaml[line-length]

enable_list:
  - args
  - empty-string-compare # opt-in
  - no-log-password # opt-in
  - no-same-owner # opt-in
  - name[prefix] # opt-in
  - yaml

offline: true
```

##### Desired Behavior

No warning is printed when FQCN is used.

##### Actual Behavior

- minimised playbook to reproduce the error:

```yaml
---
- name: Include playbook
  include: Linux.yml
  when: ansible_system == 'Linux'
```

- the output of running ansible-lint including the command line used:

```bash
ansible-lint roles/hl.timezone
WARNING  Unable to resolve FQCN for module ansible.builtin.include
```

PR #3715, opened 2023-09-05
Avoid warnings from loading of deprecated modules

ssbarnea:
- move module loading code in utils module
- mimic core behavior and fallback on legacy modules

Fixes: #3696


cognifloyd (approved):
Makes sense

PR #3732, opened 2023-09-09
Add auto-fixing implementation for `no-handler` rule

ajinkyau:
Related: #3634


examples/playbooks/transform-no-handler.transformed.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Example of no-handler rule
+  hosts: localhost
+  tasks:
+    - name: Register result of a task
+      ansible.builtin.copy:
+        dest: /tmp/placeholder
+        content: Ansible made this!
+        mode: "0600"
+      register: result # <-- we register the result of the task
```
cidrblock, line 10: I think this needs a notify...

cidrblock:
This is a tricky one, we need to watch for varialbe reuse, overlapping handler names and playbooks that have an include
```
- hosts: localhost
  gather_facts: no
  tasks:
  - name: Create a file
    ansible.builtin.file:
      path: /tmp/ansible_1.txt
      state: touch
    register: result
  - name: Print the result
    ansible.builtin.debug:
      msg: "/tmp/ansible_1.txt changed"
    when: result.changed 

  - name: Create a file
    ansible.builtin.file:
      path: /tmp/ansible_2.txt
      state: touch
    register: result
  - name: Print the result
    ansible.builtin.debug:
      msg: "/tmp/ansible_2.txt changed"
    when: result.changed 
```

Issue #3742, opened 2023-09-18
Replace `--write` with `--fix`


cidrblock:
We'll refer to this as `autofix` now like ruff does as well.

(we can add an alias for 6mo or so)

PR #3748, opened 2023-09-19
Rename --write option to --fix

ssbarnea:
This change will make ansible-lint auto-fixing feature in line with other tools that can perform similar transformations (ruff, black, ...).

For compatibility, we automatically convert old option into new one and display a warning.

Fixes: #3742


cidrblock (approved):
:heart:

Issue #3624, opened 2023-07-20
RFE: Expose which rules have a transform capabilitiy in docs and via CLI


cidrblock:
The per rule docs should show if a given rule has a transform ability.

We should also make the list of rules that have transform available from the CLI in aggregate.
```[tasklist]
### Tasks
- [x] Expose list of auto-fix rules in documentation
- [x] Make the auto-fix rule list dynamic in documentation using an include and generate_docs.py script
- [x] Fail generate_docs.py if a rule with autofix tag does not contain `autofix` inside its own markdownfile
- [x] Add documentation for all autofix rules. Required to be able to merge previous one.
```

shatakshiiii:
Initial idea to expose auto-fix in upstream docs:

- Add a separate section for auto-fix over [here](https://ansible.readthedocs.io/projects/lint/), to explain what it does and to mention the list of rules covered by it 
(just like Profiles). Also link it to the per rule docs.
- The per rule doc should show the before and after behavior of using auto-fix (via examples)

More inputs are welcomed!

PR #3788, opened 2023-09-29
Add documentation for autofixing `key-order` rule

shatakshiiii:
Related: https://github.com/ansible/ansible-lint/issues/3624

### DIFF
--- a/docs/autofix.md
+++ b/docs/autofix.md
@@ -12,7 +12,7 @@ Following is the list of supported rules covered under autofix functionality.
 - [deprecated-local-action](rules/deprecated-local-action.md)
 - [fqcn](rules/fqcn.md)
 - [jinja](rules/jinja.md)
-- [key-order](rules/key-order.md)
+- [key-order](rules/key-order.md#auto-fixing-capability)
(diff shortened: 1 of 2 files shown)


ssbarnea (changes requested):
My view on this is that we do not need to flood the user with extra examples for the auto-fix when the change is obvious and matches our already existing examples in rule documentation.

I looked at the produced documentation page and I seen it becoming long enough to detect many people from reading it.

I personally think that it was enough to mention that this rule can be automatically fixed in a paragraph inside the correct example section.

@ganeshrn WDYT about this?

ganeshrn:
> My view on this is that we do not need to flood the user with extra examples for the auto-fix when the change is obvious and matches our already existing examples in rule documentation.
> 
> I looked at the produced documentation page and I seen it becoming long enough to detect many people from reading it.
> 
> I personally think that it was enough to mention that this rule can be automatically fixed in a paragraph inside the correct example section.
> 
> @ganeshrn WDYT about this?

Yup a visible mention that the rule supports autofix will be good enough.

cidrblock (approved):

PR #3798, opened 2023-09-29
Do not use github action path for running ansible lint

### LINKED ISSUE #3703: The vault password file was not found
##### Summary

Ansible lint gives warnings/errors that it cannot find the vault password file. I create my vault password using the following step before the ansible lint step.

```yml
run: echo ${{ secrets.MYSECRET || '1234' }} > .vault-password
```

I run ansible lint with this step:

```yml
uses: ansible/ansible-lint@v6.18.0
with:
  args: "playbooks/"
```

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
```

6.18.0 (using ansible-lint as github action)

- ansible installation method: pypi
- ansible-lint installation method: using github action

##### STEPS TO REPRODUCE

Set password file location in config.

**ansible.cfg**
```ini
vault_password_file=.vault-password
```

Create vault passowrd file in workflow, as shown above.

Run ansible lint action, as shown above.

##### Desired Behavior

Find the file, that was created.

Possible security bugs should be reported via email to `security@ansible.com`

##### Actual Behavior

File was not found.

```txt
Warning: /some_playbook.yml:1 [WARNING]: Error getting vault password file (default): The vault password file /runner/_work/ansible-repo/ansible-repo/.vault-password was not found
```

**Additional Information**

I use the same method of creating the .vault-password file to actually run ansible playbooks inside of GitHub actions, and that works fine. I also did not have this problem on the now depreciated `ansible-lint-action` repository.

ReenigneArcher:
- Allow user to specify the working directory of the `ansible-lint` command
- Default working directory will be the `github.workspace` instead of `github.action_path`
- Allow user to skip installing python if they already have a specific version of python setup
- Removes the checkout step... a normal expectation is that the user will checkout their own repository... and when this action does so it overwrites any customizations/edits the user has done to their directory (such as creating a vault password file).

Fixes the following issues:
- Fixes #3703 
- Fixes #3713

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -7,17 +7,20 @@ branding:
   color: red
 inputs:
   args:
-    description: Arguments to be passed to ansible-lint command
+    description: Arguments to be passed to ansible-lint command.
+    required: false
+    default: ""
+  setup_python:
+    description: If false, this action will not setup python and will instead rely on the already installed python.
+    required: false
+    default: true
+  working_directory:
+    description: The directory where to run ansible-lint from. Default is `github.workspace`.
     required: false
     default: ""
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@v4
-      with:
-        fetch-depth: 0 # needed by setuptools-scm
-        submodules: true
-
     # Due to GHA limitation, caching works only for files within GITHUB_WORKSPACE
     # folder, so we are forced to stick this temporary file inside .git, so it
     # will not affect the linted repository.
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -7,17 +7,20 @@ branding:
...
-    description: Arguments to be passed to ansible-lint command
+    description: Arguments to be passed to ansible-lint command.
+    required: false
+    default: ""
+  setup_python:
+    description: If false, this action will not setup python and will instead rely on the already installed python.
+    required: false
+    default: true
+  working_directory:
+    description: The directory where to run ansible-lint from. Default is `${{ github.workspace }}`.
```
ReenigneArcher, line 18:
```suggestion
    description: The directory where to run ansible-lint from. Default is `github.workspace`.
```

ssbarnea (changes requested):
If you remove the checkout from it, you will need to add it to the usage example, or the users will be surprised that the linter is not linting their code.

ssbarnea (approved):

PR #3809, opened 2023-10-03
Enable YAML 1.2 support for non-ansible files

ssbarnea:
Fixes: #3790

### DIFF
--- a/src/ansiblelint/transformer.py
+++ b/src/ansiblelint/transformer.py
@@ -8,7 +8,11 @@ from ruamel.yaml.comments import CommentedMap, CommentedSeq
 
 from ansiblelint.file_utils import Lintable
 from ansiblelint.rules import AnsibleLintRule, TransformMixin
-from ansiblelint.yaml_utils import FormattedYAML, get_path_to_play, get_path_to_task
+from ansiblelint.yaml_utils import (
+    FormattedYAML,
+    get_path_to_play,
+    get_path_to_task,
+)
 
 if TYPE_CHECKING:
     from ansiblelint.config import Options
@@ -92,7 +96,12 @@ class Transformer:
                 # We need a fresh YAML() instance for each load because ruamel.yaml
                 # stores intermediate state during load which could affect loading
                 # any other files. (Based on suggestion from ruamel.yaml author)
-                yaml = FormattedYAML()
+                yaml = FormattedYAML(
+                    # Ansible only uses YAML 1.1, but others files should use newer 1.2 (ruamel.yaml defaults to 1.2)
+                    version=(1, 1)
+                    if file.is_owned_by_ansible()
+                    else None,
+                )
 
                 ruamel_data = yaml.load(data)
                 if not isinstance(ruamel_data, (CommentedMap, CommentedSeq)):
@@ -107,11 +116,6 @@ class Transformer:
 
             if self.write_set != {"none"}:
                 self._do_transforms(file, ruamel_data or data, file_is_yaml, matches)
-
-            if not file.is_owned_by_ansible():
-                # Do nothing until we enable support for YAML 1.2
-                # https://github.com/ansible/ansible-lint/issues/3811
-                continue
 
             if file_is_yaml:
                 _logger.debug("Dumping %s using YAML (%s)", file, yaml.version)
--- a/src/ansiblelint/yaml_utils.py
+++ b/src/ansiblelint/yaml_utils.py
@@ -35,7 +35,7 @@ from ansiblelint.utils import Task
 if TYPE_CHECKING:
     # noinspection PyProtectedMember
     from ruamel.yaml.comments import LineCol
-    from ruamel.yaml.compat import StreamTextType
+    from ruamel.yaml.compat import StreamTextType, VersionType
     from ruamel.yaml.nodes import ScalarNode
     from ruamel.yaml.representer import RoundTripRepresenter
     from ruamel.yaml.tokens import CommentToken
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
     from ansiblelint.file_utils import Lintable
 
 _logger = logging.getLogger(__name__)
+
 
 YAMLLINT_CONFIG = """
 extends: default
@@ -750,13 +751,14 @@ class FormattedEmitter(Emitter):
 class FormattedYAML(YAML):
     """A YAML loader/dumper that handles ansible content better by default."""
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         *,
         typ: str | None = None,
         pure: bool = False,
         output: Any = None,
         plug_ins: list[str] | None = None,
+        version: VersionType | None = None,
     ):
         """Return a configured ``ruamel.yaml.YAML`` instance.
 
@@ -816,10 +818,12 @@ class FormattedYAML(YAML):
               tasks:
                 - name: Task
         """
-        # Default to reading/dumping YAML 1.1 (ruamel.yaml defaults to 1.2)
-        self._yaml_version_default: tuple[int, int] = (1, 1)
-        self._yaml_version: str | tuple[int, int] = self._yaml_version_default
-
+        if version:
+            if isinstance(version, str):
+                x, y = version.split(".", maxsplit=1)
+                version = (int(x), int(y))
+            self._yaml_version_default: VersionType = version
+            self._yaml_version: VersionType = self._yaml_version_default
         super().__init__(typ=typ, pure=pure, output=output, plug_ins=plug_ins)
 
         # NB: We ignore some mypy issues because ruamel.yaml typehints are not great.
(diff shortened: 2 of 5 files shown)


test/test_yaml_utils.py
```diff
@@ -202,8 +203,7 @@ def test_custom_ruamel_yaml_emitter(
     assert output == expected_output
 
 
-@pytest.fixture(name="yaml_formatting_fixtures")
-def fixture_yaml_formatting_fixtures(fixture_filename: str) -> tuple[str, str, str]:
```
cognifloyd, line 206: I made this a fixture so that any issues here would show up as a test error instead of a test failure. Why did you change it to a helper function?

src/ansiblelint/yaml_utils.py
```diff
@@ -817,8 +819,12 @@ def __init__(
                 - name: Task
         """
         # Default to reading/dumping YAML 1.1 (ruamel.yaml defaults to 1.2)
```
cognifloyd, line 821: This comment is out of date now, because this uses ruamel.yaml's default of 1.2. It should probably be moved to `transformer.py`.

src/ansiblelint/yaml_utils.py
```diff
@@ -939,7 +950,11 @@ def version(self, value: str | tuple[int, int] | None) -> None:
         So, if a file does not include the directive, it sets this to None.
         But, None effectively resets the parsing version to YAML 1.2 (ruamel's default).
         """
-        self._yaml_version = value if value is not None else self._yaml_version_default
+        if value is not None:
+            self._yaml_version = value
+        elif hasattr(self, "_yaml_version_default"):
+            self._yaml_version = self._yaml_version_default
+        # We do nothing if the object did not had a previous default version defined
```
cognifloyd, line 957:
```suggestion
        # We do nothing if the object did not have a previous default version defined
```

src/ansiblelint/yaml_utils.py
```diff
@@ -874,6 +880,9 @@ def __init__(
         # (see https://stackoverflow.com/a/44314840/1134951)
         # self.Representer.add_representer(
 
+        if version:
+            self.version = version
```
cognifloyd, line 884: Why did you add this? Setting `self.version` actually sets `self._yaml_version` which is already configured above. Why do we need to set it again?

src/ansiblelint/yaml_utils.py
```diff
@@ -977,7 +992,11 @@ def dumps(self, data: Any) -> str:
                 stream.write(preamble_comment)
             self.dump(data, stream)
             text = stream.getvalue()
-        return self._post_process_yaml(text)
+        strip_version_directive = hasattr(self, "_yaml_version_default")
```
cognifloyd, line 995: Is this ever false? I thought ruamel.yaml had a default for `self._yaml_version_default`.

src/ansiblelint/transformer.py
```diff
@@ -92,7 +96,9 @@ def run(self) -> None:
                 # We need a fresh YAML() instance for each load because ruamel.yaml
                 # stores intermediate state during load which could affect loading
                 # any other files. (Based on suggestion from ruamel.yaml author)
-                yaml = FormattedYAML()
+                yaml = FormattedYAML(
+                    version=(1, 1) if file.is_owned_by_ansible() else None,
```
cognifloyd, line 100: The comment about the default version being 1.1 should probably move here.

cognifloyd (dismissed):

src/ansiblelint/yaml_utils.py
ssbarnea: That is there for a good reason, that attribute is not always defined!

test/test_yaml_utils.py
ssbarnea: It made the testing harder and also maintenance, vscode got bit confused by it. I also think that having a test failure is fine.

cidrblock (approved):

src/ansiblelint/transformer.py
```diff
@@ -92,7 +96,12 @@ def run(self) -> None:
                 # We need a fresh YAML() instance for each load because ruamel.yaml
                 # stores intermediate state during load which could affect loading
                 # any other files. (Based on suggestion from ruamel.yaml author)
-                yaml = FormattedYAML()
+                yaml = FormattedYAML(
+                    # Ansible only uses YAML 1.1, but others files should use newer 1.2 (ruamel.yaml defaults to 1.2)
+                    version=(1, 1)
```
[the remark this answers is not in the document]
ssbarnea: There reason why I wanted to fallback to null was because that would make it make it future-proof for when ruamel.yaml will switch to versions such 1.3 or more.

What we want is to override the default for ansible file, and not touch it for others.

src/ansiblelint/yaml_utils.py
cognifloyd: Doh! Of course. I added `self._yaml_version` and `self._yaml_version_default` to work around ruamel.yaml's quirky use of the `version` attribute. This PR changes that so those are only present sometimes. Hmm...

src/ansiblelint/yaml_utils.py
```diff
@@ -816,10 +818,12 @@ def __init__(
...
-        # Default to reading/dumping YAML 1.1 (ruamel.yaml defaults to 1.2)
-        self._yaml_version_default: tuple[int, int] = (1, 1)
-        self._yaml_version: str | tuple[int, int] = self._yaml_version_default
-
+        if version:
+            if isinstance(version, str):
+                x, y = version.split(".", maxsplit=1)
+                version = (int(x), int(y))
+            self._yaml_version_default: VersionType = version
+            self._yaml_version: VersionType = self._yaml_version_default
```
cognifloyd, line 826:
```suggestion
        if isinstance(version, str):
            x, y = version.split(".", maxsplit=1)
            version = (int(x), int(y))
        self._yaml_version_default: VersionType | None = version
        self._yaml_version: VersionType | None = self._yaml_version_default
        self._strip_yaml_version_directive = version is not None
```

src/ansiblelint/yaml_utils.py
```diff
@@ -920,16 +924,18 @@ def _defaults_from_yamllint_config() -> dict[str, bool | int | str]:
...
 
         Ansible uses PyYAML which only supports YAML 1.1. ruamel.yaml defaults to 1.2.
         So, we have to make sure we dump yaml files using YAML 1.1.
         We can relax the version requirement once ansible uses a version of PyYAML
         that includes this PR: https://github.com/yaml/pyyaml/pull/555
         """
-        return self._yaml_version
+        if hasattr(self, "_yaml_version"):
+            return self._yaml_version
+        return None
```
cognifloyd, line 938:
```suggestion
        return self._yaml_version
```

src/ansiblelint/yaml_utils.py
```diff
@@ -939,7 +945,11 @@ def version(self, value: str | tuple[int, int] | None) -> None:
         So, if a file does not include the directive, it sets this to None.
         But, None effectively resets the parsing version to YAML 1.2 (ruamel's default).
         """
-        self._yaml_version = value if value is not None else self._yaml_version_default
+        if value is not None:
+            self._yaml_version = value
+        elif hasattr(self, "_yaml_version_default"):
+            self._yaml_version = self._yaml_version_default
+        # We do nothing if the object did not have a previous default version defined
```
cognifloyd, line 952: I like the ternary expression that was here before. But this would work too:

```suggestion
        if value is not None:
            self._yaml_version = value
        else:
            self._yaml_version = self._yaml_version_default
```

src/ansiblelint/yaml_utils.py
```diff
@@ -977,7 +987,11 @@ def dumps(self, data: Any) -> str:
                 stream.write(preamble_comment)
             self.dump(data, stream)
             text = stream.getvalue()
-        return self._post_process_yaml(text)
+        strip_version_directive = hasattr(self, "_yaml_version_default")
+        return self._post_process_yaml(
+            text,
+            strip_version_directive=strip_version_directive,
+        )
```
cognifloyd, line 994:
```suggestion
        return self._post_process_yaml(
            text,
            strip_version_directive=self._strip_yaml_version_directive,
        )
```

cognifloyd (changes requested):
The purpose of relying on the presence/absence of the `_yaml_version*` attributes was not clear to me. What about using a separate `_strip_yaml_version_directive` attribute to define that explicitly instead of implying it using `hasattr`?

src/ansiblelint/transformer.py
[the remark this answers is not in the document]
cognifloyd: Agreed. We should stick with `None` as the default value.

src/ansiblelint/yaml_utils.py
cognifloyd: So, this would be my preference for this block:
```suggestion
        self._yaml_version = value if value is not None else self._yaml_version_default
```

src/ansiblelint/yaml_utils.py
ssbarnea: I do not want to touch this right now, especially as I know that a newer patch of the library introduces some data that we might want to use, look at 0.18.4 changelog entries https://pypi.org/project/ruamel.yaml/

We can do extra polishing in follow-up, but I prefer to use whatever they have instead of setting new properties that have meaning only for our subclass. I need to play with it bit more too see what data I have available.

PR #3812, opened 2023-10-04
Remove bump-homebrew-formula configuration

mislav:
Moving the conversation from https://github.com/Homebrew/homebrew-core/pull/149334#issuecomment-1746766176 /cc @ssbarnea 

This updates the `download-url` input to the `bump-homebrew-formula-action` to be a URL in the format of `https://pypi.org/packages/source/a/ansible-lint/ansible-lint-<VERSION>.tar.gz`.

That resource ultimately resolves to `https://files.pythonhosted.org/packages/*/*/*/ansible-lint-<VERSION>.tar.gz`, which is the URL format currently present in the homebrew-core formula. I find the latter form unwieldy, and I think that the `pypi.org`-hosted resource looks much nicer and should be the same trustworthiness. I'm _not_ familiar with python-hosted anything, so you should decide which canonical form works best for you (and for the homebrew-core team).

Followup to https://github.com/ansible/ansible-lint/pull/2982
Ref. https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release
Ref. https://github.com/marketplace/actions/bump-homebrew-formula#action-inputs
Ref. https://github.com/Homebrew/homebrew-core/pull/128963#issuecomment-1519423319

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,6 +60,7 @@ jobs:
     environment: release
     runs-on: ubuntu-22.04
     needs: pypi
+    if: github.event_name == 'release'
 
     env:
       FORCE_COLOR: 1
@@ -67,16 +68,18 @@ jobs:
       TOXENV: pkg
 
     steps:
-      - name: Check out src from Git
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0 # needed by setuptools-scm
-          submodules: true
+      - name: Extract version
+        id: extract-version
+        env:
+          TAG_NAME: ${{ github.event.release.tag_name }}
+        run: |
+          echo "version=${TAG_NAME#v}" >> $GITHUB_OUTPUT
 
       - name: Bump homebrew formula
         uses: mislav/bump-homebrew-formula-action@v3.1
         with:
           # A PR will be sent to github.com/Homebrew/homebrew-core to update this formula:
           formula-name: ansible-lint
+          download-url: https://pypi.org/packages/source/a/ansible-lint/ansible-lint-${{ steps.extract-version.outputs.version }}.tar.gz
         env:
           COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,4 @@
 ---
-# cspell:ignore mislav
 name: release
 
 "on":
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -60,23 +60,26 @@ jobs:
     environment: release
     runs-on: ubuntu-22.04
     needs: pypi
+    if: github.event_name == 'release'
```
mislav, line 63: Avoid running the `bump-homebrew-formula` job if the event was workflow_dispatch, since the `github.event.release.tag_name` value will not be present.

.github/workflows/release.yml
```diff
@@ -60,23 +60,26 @@ jobs:
...
 
     env:
       FORCE_COLOR: 1
       PY_COLORS: 1
       TOXENV: pkg
 
     steps:
-      - name: Check out src from Git
-        uses: actions/checkout@v4
```
mislav, line 71: Checking out the source for this repository is not necessary to run `bump-homebrew-formula`.

.github/workflows/release.yml
```diff
@@ -60,23 +60,26 @@ jobs:
...
-        with:
-          fetch-depth: 0 # needed by setuptools-scm
-          submodules: true
+      - name: Extract version
+        id: extract-version
+        env:
+          TAG_NAME: ${{ github.event.release.tag_name }}
+        run: |
+          echo "version=${TAG_NAME#v}" >> $GITHUB_OUTPUT
```
mislav, line 76: This is all just to transform `vX.Y.Z` to `X.Y.Z` (i.e. strip the `v` prefix), since GitHub Actions workflow syntax does not have a function that does that.

mislav:
Hmm, I am now learning more about python-based formula in Homebrew-core and it seems like each formula lists its python dependencies as "resources", e.g.
```rb
  resource "ansible-compat" do
    url "https://files.pythonhosted.org/packages/36/f9/42b44473cedbf977f8fe41e2b72ea47216f708c59bafc41ad6aa93c09a71/ansible-compat-4.1.10.tar.gz"
    sha256 "2be8c7b510d2e15eed1e9ef443209d67d9aec8f427026b88936d4535ff59863d"
  end
```

My `bump-homebrew-formula` action does _not_ have the feature to update these resources. The built-in Homebrew command `brew bump` does have this feature, which is supported by this non-trivial code: https://github.com/Homebrew/brew/blob/75b18b5afe70ef523b1d78bfcb1e8ee27fe04ef8/Library/Homebrew/utils/pypi.rb#L205.

A better course of action would be to remove the `bump-homebrew-formula` action from this repository entirely.

ssbarnea (approved):

ssbarnea:
@mislav Thanks, I really appreciate your help here.

Issue #3787, opened 2023-09-29
`key-order` rule does not work at the play level for `name` key


shatakshiiii:
##### Summary
As per the documentation of key-order rule, key `name` must always be the first key for plays, tasks and handlers. Which doesn't seem to work for a given scenario:

##### Issue Type

- Bug Report

##### STEPS TO REPRODUCE

```console (paste below)
---
- hosts: localhost
  name: This is a playbook  # <-- name key should be the first one
  tasks:
    - name: A block
      when: true
      block:
        - name: Display a message
          debug:
            msg: "Hello world!"

```

For the above playbook we should get a `key-order` rule violation at line 2, as name should be the first key (instead of hosts).

**Actual Behavior**

There is no error for key-order rule while running ansible-lint on this playbook.

```console (paste below)
(l1) ➜  test_lint  ansible-lint playbooks/key-order.yml
WARNING  Listing 1 violation(s) that are fatal
fqcn[action-core]: Use FQCN for builtin module actions (debug).
playbooks/key-order.yml:8 Use `ansible.builtin.debug` or `ansible.legacy.debug` instead.

```

ajinkyau:
Related: #3634

PR #3815, opened 2023-10-05
Add play level autofix for key-order rule

ajinkyau:
Fixes: #3787
Depends on: #3825


shatakshiiii (approved):

ssbarnea (changes requested):
Even after multiple rebases I still see those two tests failing on macos. We cannot merge the change without having the GHA report green, so you will need to dig into this issue.

ssbarnea (approved):

Issue #3818, opened 2023-10-06
Transformer should not create temporary files inside the codebase


ssbarnea:
When interrupting linter testing I discover random file named `tmp...` inside the codebase. I believe that our transformed does this and I do think that we should avoid creating temporary files inside the source tree. Instead these temporary files should be created into the system provided temporary directory.

@Qalthos I think that you know this area of code quite well to be able to make a patch for it.

ssbarnea:
I think that I discovered another issue related to transformer and our testing. On arch testing failed as we run out of disk space on /tmp (2GB default). Once I started looking on it I found something quite unexpected. For each transformer tests, pytest-xdist copied the entire examples/ folder with included collections (~100MB). 

![](https://sbarnea.com/ss/Screen-Shot-2023-10-06-11-22-41.17.png)

```
--- /tmp/pytest-of-root/pytest-0/popen-gw1/test_transformer_jinja_spacing0/examples/playbooks/collections ------------------------------------------------------
                                     /..
   95.4 MiB [######################] /ansible_collections
    2.2 MiB [                      ]  community-general-6.3.0.tar.gz
  564.0 KiB [                      ]  community-crypto-2.12.0.tar.gz
  484.0 KiB [                      ]  ansible-windows-1.13.0.tar.gz
  420.0 KiB [                      ]  community-docker-3.4.2.tar.gz
  172.0 KiB [                      ]  ansible-posix-1.5.1.tar.gz
    8.0 KiB [                      ]  community-molecule-0.1.0.tar.gz
    4.0 KiB [                      ]  requirements.yml
    4.0 KiB [                      ]  .gitignore
```

ssbarnea:
Aven after bumping the /tmp from 2GB to 10GB, i got test failures now due to running out of memory. The test VM has 2 cpus and 4GB RAM, which should be fine for running the tests. The system uses only under 400MB to run.

PR #3820, opened 2023-10-06
Simplify transformer test

ssbarnea:
Related: https://github.com/ansible/ansible-lint/issues/3818


Qalthos (approved):

src/ansiblelint/rules/jinja.py
```diff
@@ -844,17 +847,17 @@ def test_jinja_transform(
...
+        ).read_text(encoding="utf-8")
+        transformed_content = playbook.with_suffix(f".tmp{playbook.suffix}").read_text(
+            encoding="utf-8",
+        )
 
-        assert orig_playbook_content != transformed_playbook_content
-        assert transformed_playbook_content == expected_playbook_content
+        assert orig_content != transformed_content
+        assert expected_content == transformed_content
+        playbook.with_suffix(f".tmp{playbook.suffix}").unlink()
```
Qalthos, line 860: Hmm, these should probably be in a `finally` block to make sure they get cleaned up if the test aborts?

Qalthos (changes requested):

src/ansiblelint/rules/jinja.py
ssbarnea: I agree with you about this but we will leave this for a follow-up. For the moment I added these files to git ignore.

PR #3824, opened 2023-10-08
Fix SARIF-formatter severity levels

4ch1m:
The SARIF formatter output contains information about "severity levels" in two different contexts (which should be handled differently).

* general "rule-based" severity levels
* "error-/match-based" severity levels within the result set

### DIFF
--- a/src/ansiblelint/formatters/__init__.py
+++ b/src/ansiblelint/formatters/__init__.py
@@ -314,15 +314,13 @@ class SarifFormatter(BaseFormatter[Any]):
 
     @staticmethod
     def _get_sarif_rule_severity_level(rule: BaseRule) -> str:
-        """
-        General SARIF severity level for a rule.
-        (Can differ from an actual result/match severity.)
+        """General SARIF severity level for a rule.
+        (Can differ from an actual result/match severity.).
 
         Possible values: "none", "note", "warning", "error"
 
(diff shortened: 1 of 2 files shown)


ssbarnea:
@4ch1m While I am aware that we dump severity in different places, I do not think that is an issue by itself, mainly because once would be the severity of a rule in general and the other one would specific to a particular occurence.

Still, the title of the PR does not mention or refer the exact bug it is addressing. I am asking this as I want to fully understand what was wrong before. Thanks

4ch1m:
Sure... I'll try to explain.

The current implementation basically is using this function to obtain the severity (for **both** the general rule **and** the specific occurrence:

https://github.com/ansible/ansible-lint/blob/5ca56796fd66905fbdecf9ea234e247a5691090c/src/ansiblelint/errors.py#L116-L123

__NOTE__: The result is "influenced" by `options.warn_list`.

External tools (IDE plugins, etc.) will use the SARIF output (and the given severity levels) to hightlight/annotate Ansible code; which is problematic in many ways.

Example:
I just recently realized that all of my "name[casing]" issues were marked as **errors** (whereas their severity is actually only `MEDIUM`).

External tools should be able to give hints/advice about rules in general (with the rules **original/general** severity) and the actual error/occurrence.

That's why this should be fixed as proposed in this PR. :smile:

PR #3825, opened 2023-10-09
Fix tests for Mac OS

ajinkyau:
Since `options.lintables` contained a path to the yaml file from an earlier test, we were
unable to search for all the lintables correctly.
Setting it to an empty list fixes this issue.

https://github.com/ansible/ansible-lint/pull/3815 can be merged after this


priyamsahoo (approved):

ssbarnea (changes requested):
Blocking it until we refactor the broken code. I am working on it and checked with @cidrblock

ssbarnea:
No longer needed.

PR #3837, opened 2023-10-11
Temporary avoid auto-fixing YAML files not owned by ansible

ssbarnea:
Related: #3809
Related: #3790


cidrblock (approved):
looks good

Is there a "follow-up" issue we can link?  I know you won't forget about it but I will :)

ajinkyau (approved):

Issue #3767, opened 2023-09-24
`no-handler` shouldn't trigger on handlers in `handlers` directory


ajakk:
##### Summary

#3655 was incompletely fixed: `no-handler` still triggers on handlers specified in a separate handler directory.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.20.1.dev4 using ansible-core:2.15.3 ansible-compat:4.1.10 ruamel-yaml:0.17.32 ruamel-yaml-clib:0.2.7
You are using a pre-release version of ansible-lint.
```
Installed from e6bc4c73148af6ffe6bf17ddb06d6d4627de039a.

- ansible installation method: Gentoo package
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

```console (paste below)
$ grep -r ""
roles/foo/tasks/main.yml:---
roles/foo/tasks/main.yml:- name: Get info
roles/foo/tasks/main.yml:  delegate_to: localhost
roles/foo/tasks/main.yml:  register: collected_info
roles/foo/tasks/main.yml:  ansible.builtin.debug:
roles/foo/tasks/main.yml:    msg: foo
roles/foo/tasks/main.yml:
roles/foo/tasks/main.yml:- name: Do something
roles/foo/tasks/main.yml:  delegate_to: localhost
roles/foo/tasks/main.yml:  loop: "{{ collected_info['some_list'] }}"
roles/foo/tasks/main.yml:  loop_control:
roles/foo/tasks/main.yml:    label: "{{ item.name }}"
roles/foo/tasks/main.yml:  notify:
roles/foo/tasks/main.yml:    - Debug
roles/foo/tasks/main.yml:  register: _something_done
roles/foo/tasks/main.yml:  ansible.builtin.debug:
roles/foo/tasks/main.yml:    msg: foo
roles/foo/handlers/main.yml:---
roles/foo/handlers/main.yml:- name: Debug
roles/foo/handlers/main.yml:  loop: "{{ _something_done.results }}"
roles/foo/handlers/main.yml:  loop_control:
roles/foo/handlers/main.yml:    label: "{{ item.item.name }}"
roles/foo/handlers/main.yml:  when: item.changed
roles/foo/handlers/main.yml:  ansible.builtin.debug:
roles/foo/handlers/main.yml:    msg: "{{ item.item.name }} changed"
$ ansible-lint
[...]

PR #3838, opened 2023-10-12
Avoid false positives for handler in roles handlers directory

ajinkyau:
Fixes: #3767


src/ansiblelint/rules/no_handler.py
```diff
@@ -69,7 +69,11 @@ def matchtask(
         task: Task,
         file: Lintable | None = None,
     ) -> bool | str:
-        if task["__ansible_action_type__"] != "task" or task.is_handler():
```
ssbarnea, line 72: Was not possible to fix `is_hander()` to ensure that all tasks from a handlers tasks files are reported as true?

IMHO, that would make it much easier to maintain, and probably it would also allow us to simplify the code.

ajinkyau: @ssbarnea Updated. Now the check is in the `is_handler` function

ssbarnea (approved):

PR #3843, opened 2023-10-16
Add support for Rocky

facorazza:


ssbarnea (changes requested):
Based on my tests with Rocky 8, the distribution name is Rocky, not RockyLinux.

```
 ansible localhost -m setup|grep distr
[WARNING]: No inventory was parsed, only implicit localhost is available
        "ansible_distribution": "Rocky",
```

facorazza:
> Based on my tests with Rocky 8, the distribution name is Rocky, not RockyLinux.
> 
> ```
>  ansible localhost -m setup|grep distr
> [WARNING]: No inventory was parsed, only implicit localhost is available
>         "ansible_distribution": "Rocky",
> ```

Is it enough to change properties > name > const to Rocky?

ssbarnea (approved):

PR #3870, opened 2023-10-25
Update linters

ssbarnea:


ajinkyau (approved):

src/ansiblelint/utils.py
```diff
@@ -702,7 +702,11 @@ class Task(dict[str, Any]):
     @property
     def name(self) -> str | None:
         """Return the name of the task."""
-        return self.raw_task.get("name", None)
+        name = self.raw_task.get("name", None)
+        if name is not None and not isinstance(name, str):
+            msg = "Task name can only be a string."
+            raise RuntimeError(msg)
```
cognifloyd, line 708: Why did you add this / what uncovered this issue? Did someone run into an issue?

ssbarnea: @cognifloyd Issue was reported by mypy, and it makes sense as someone might put some junk data into name. I did not reproduce it, but what mypy reported made sense.

PR #3903, opened 2023-11-19
Look for ansible binary in the same path as we are

FliegenKLATSCH:
If the `PATH` is not set up correctly and we are invoked via full path, it's likely that ansible itself is just next to us.
It's added at the end, so it's only the last resort, if nothing else could be found.


ssbarnea (dismissed):
This will clearly not work as expected if linter is called as a python module as in this case argv[0] will be just python.

FliegenKLATSCH:
True, but in the worst case we add the path of python itself at the end, which usually should be there already?
Maybe I am overlooking something?

ssbarnea (approved):

Issue #3816, opened 2023-10-05
Allow linter to run in a pure offline mode


ssbarnea:
We need a CLI option that would run linter in a pure offline mode:

- [ ] never run ansible-galaxy
- [ ] ignore all errors caused by missing modules, filters, plugins and roles

ssbarnea:
@ajinkyau This can be done in multiple steps, a first one would be allow linter to fully disable the calls to ansible-galaxy install, as this affects the lightspeed bot implementation.

ganeshrn:
In pure offline mode don't throm warnings/error if a collection/plugin/ task file is part of the dependency and not installed in the project. 
Example warnings and repo https://github.com/anshulbehl/network.bgp/pull/1

SirUli:
In offline mode it should also not check for the [latest version](https://github.com/ansible/ansible-lint/blob/634c639293a4f50d12d6ac33cfec96efb05f4b16/src/ansiblelint/config.py#L278).

PR #3913, opened 2023-11-27
Add option to allow running with missing plugins and collection dependencies

ssbarnea:
Fixes: #3816


cidrblock (approved):
looks like a good start that shouldn't break anything unless invoked

Issue #3898, opened 2023-11-14
Github action shown in README fails


bschonec:
##### Summary

Using the [simple example](https://github.com/ansible/ansible-lint#using-ansible-lint-as-a-github-action) in the README.md, the github action fails with 

```
Run ansible-lint 
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/ansible/config/manager.py", line 563, in get_config_value_and_origin
    value = ensure_type(value, defs[config].get('type'), origin=origin)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/ansible/config/manager.py", line 87, in ensure_type
    value = int(value)
            ^^^^^^^^^^
ValueError: invalid literal for int() with base 10: ''

During handling of the above exception, another exception occurred:

```

##### Issue Type

- Bug Report

##### STEPS TO REPRODUCE

Simply copy and paste the example workflow code into .github/workflows/anyfile.yml

priyamsahoo:
We have been using this example in our repositories and it seems to work well. It would be great if you could share the repository where this action is failing.

bschonec:
[My fork](https://github.com/bschonec/ansible-github/tree/stable) has never run successfully.

ssbarnea:
@bschonec I believe that https://github.com/bschonec/ansible-github/blob/stable/ansible.cfg#L123 is the line that makes ansible core to choke. That is a null and not an integer.

ssbarnea:
Confirmed:
```
$ ansible --version
ERROR: Invalid type for configuration option setting: DEFAULT_GATHER_TIMEOUT (from /Users/ssbarnea/c/a/1/ansible-github/ansible.cfg): invalid literal for int() with base 10: ''
FAIL: 5
```
This is caused by your broken `ansible.cfg` file. Still, I will keep the bug open little bit more because I want to make the linter behave little bit better in a case like this.

The current stacktrace is not very friendly and it would be better to provide a better error.

bschonec:
Very interesting.  I don't know where I came up with that ansible.cfg.  Most certainly, it's not valid.

Sorry for the noise.  Since I never ran any real Ansible commands, I didn't notice the fubarred ansible.cfg file.

[...]

PR #3916, opened 2023-11-28
Avoid stacktrace with broken ansible.cfg files

ssbarnea:
Fixes: #3898


Qalthos (approved):
It works, and I don't see a nicer way of doing this

Issue #3919, opened 2023-11-29
The pipx PATH warning is back


chriscroome:
The [issue fixed last summer](https://github.com/ansible/ansible-lint/issues/3541) is back with [the latest version](https://github.com/ansible/ansible-lint/releases/tag/v6.22.1):

```bash
ansible-lint .
WARNING: PATH altered to include /home/chris/.local/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.

Passed: 0 failure(s), 0 warning(s) on 29 files. Last profile that met the validation criteria was 'production'.
```

```bash
which ansible-lint
/home/chris/.local/bin/ansible-lint

ls -lah ~/.local/bin/ansible-lint 
lrwxrwxrwx 1 chris chris 54 Jul 14 15:54 /home/chris/.local/bin/ansible-lint -> /home/chris/.local/pipx/venvs/ansible/bin/ansible-lint

ansible-lint --version
WARNING: PATH altered to include /home/chris/.local/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.
ansible-lint 6.22.1 using ansible-core:2.16.0 ansible-compat:4.1.10 ruamel-yaml:0.18.5 ruamel-yaml-clib:0.2.8
```

On Debian Bookworm `pipx` is the recommended way to install PyPi packages so it would be good if using it didn't trigger warnings.

PR #3920, opened 2023-11-29
Avoid warnings about PATH with pipx installations

ssbarnea:
Fixes: #3919

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -452,6 +452,7 @@ def path_inject(own_location: str = "") -> None:
         inject_paths.append(str(py_path))
 
     # last option, if nothing else is found, just look next to ourselves...
+    own_location = os.path.realpath(own_location)
     if (
         own_location
         and (Path(own_location).parent / "ansible").exists()
@@ -460,7 +461,7 @@ def path_inject(own_location: str = "") -> None:
         inject_paths.append(str(Path(own_location).parent))
 
     if not os.environ.get("PYENV_VIRTUAL_ENV", None):
-        if inject_paths:
+        if inject_paths and not all("pipx" in p for p in inject_paths):
             print(  # noqa: T201
                 f"WARNING: PATH altered to include {', '.join(inject_paths)} :: This is usually a sign of broken local setup, which can cause unexpected behaviors.",
                 file=sys.stderr,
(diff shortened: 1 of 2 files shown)


ajinkyau (approved):

src/ansiblelint/__main__.py
```diff
@@ -452,6 +452,7 @@ def path_inject(own_location: str = "") -> None:
         inject_paths.append(str(py_path))
 
     # last option, if nothing else is found, just look next to ourselves...
+    own_location = os.path.realpath(own_location)
```
FliegenKLATSCH, line 455: This breaks #3903 again, do we have to resolve symlinks here? Maybe the resolved location can only be used where needed?

yshollander-selerity:
I think I'm having the same issue on MacOS, ansible-lint installed from Homebrew. ansible-lint is in my PATH, so there should be no warning here.

```
ansible % ansible-lint -vv myplaybookyml 2>&1 | egrep -v 'DEBUG|INFO'
WARNING: PATH altered to include /opt/homebrew/Cellar/ansible-lint/24.2.2/libexec/bin, /opt/homebrew/Cellar/ansible-lint/24.2.2/libexec/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.
ansible % which ansible-lint
/opt/homebrew/bin/ansible-lint
```

andrewkendall:
> I think I'm having the same issue on MacOS, ansible-lint installed from Homebrew. ansible-lint is in my PATH, so there should be no warning here.
> 
> ```
> ansible % ansible-lint -vv myplaybookyml 2>&1 | egrep -v 'DEBUG|INFO'
> WARNING: PATH altered to include /opt/homebrew/Cellar/ansible-lint/24.2.2/libexec/bin, /opt/homebrew/Cellar/ansible-lint/24.2.2/libexec/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.
> ansible % which ansible-lint
> /opt/homebrew/bin/ansible-lint
> ```

Same here, just updated to 24.2.2 via brew and getting the same error.

andrewkendall:
@yshollander-selerity for a quick fix I've rolled back to an old version. Download 

`https://raw.githubusercontent.com/Homebrew/homebrew-core/9238b09cb713f6ccb13c3d1f1c67a61a04ab4db5/Formula/a/ansible-lint.rb`

cd the directory that is in and run:

`brew install ansible-lint.rb`

PR #3925, opened 2023-12-04
Make requirements.txt files compatible with dependabot

ssbarnea:


ssbarnea:
Having test-action fail is expected and the other failure is unrelated.

Issue #3950, opened 2023-12-16
Issue with check "requires_ansible key must be set to a supported version"


andy-maier:
##### Summary

PR #2770 introduced a check for the "requires_ansible" value in meta/runtime.yml. The way it was implemented causes the omission of the patch version now to be reported as an issue. For example:

```
requires_ansible: ">=2.9"
```

results in:

```
meta/runtime.yml:1: meta-runtime[unsupported-version]: requires_ansible key must be set to a supported version. 
```

The issue is that the page showing supported Ansible versions does not even mention the supported patch versions:
https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html . This makes it a try-and-error to even find a patch version that is accepted by ansible-lint.  It seems to me that when Ansible itself documents its supported versions without mentioning a patch version, that ansible-lint should not insist on a patch version. Where does ansible-lint gets its notion of supported Ansible versions from?

An additional issue is that when a collection supports Ansible versions that are marked EOL, this issue is raised even when specifying the latest released patch version (e.g. 2.9.27). How can collections that need to support such Ansible versions get rid of this warning?

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

We get this error from the upload check that Ansible AutomationHub performs.

The error does not occur in the upload check that Ansible Galaxy performs, nor in the ansible sanity test on any Ansible version supported by Ansible.

When installing ansible-lint 6.22.1 and running it, it does report the error.

##### STEPS TO REPRODUCE

Have a meta/runtime.yml file with this content:

```
requires_ansible: ">=2.9"
```

Run ansible-lint

##### Desired Behavior

I think the discrepancy between Ansible's (non-)focus on patch versions and the requirement of ansible-lint to specify the patch version needs to be seriously reconsidered. I don't see the value of specifying a patch version in requires_ansible.

##### Actual Behavior

See above.

alisonlhart:
[...]

PR #3954, opened 2023-12-20
Correct requires_ansible error message

ssbarnea:
Related: #3950


src/ansiblelint/rules/meta_runtime.py
```diff
@@ -32,8 +32,9 @@ class CheckRequiresAnsibleVersion(AnsibleLintRule):
     # Refer to https://access.redhat.com/support/policy/updates/ansible-automation-platform
     # Also add devel to this list
     supported_ansible = ["2.14.", "2.15.", "2.16."]
+    supported_ansible_examples = [f">={x}0" for x in supported_ansible]
     _ids = {
-        "meta-runtime[unsupported-version]": "'requires_ansible' key must be set to a supported version - 2.13.x, 2.14.x, 2.15.x",
+        "meta-runtime[unsupported-version]": f"'requires_ansible' key must refer to a currently supported version such: {', '.join(supported_ansible_examples)}",
```
Qalthos, line 37: This isn't necessarily the cleanest way of doing this, but I think the _.x_ part is important to clarifying what we're expecting here.
```suggestion
        "meta-runtime[unsupported-version]": f"'requires_ansible' key must refer to a currently supported version such as: {', '.join((v + "x" for v in supported_ansible)}",
```

Qalthos (changes requested):

src/ansiblelint/rules/meta_runtime.py
Qalthos, line 37:
```suggestion
        "meta-runtime[unsupported-version]": f"'requires_ansible' key must refer to a currently supported version such as: {', '.join(supported_ansible_examples)}",
```

Qalthos (changes requested):
Oh now I see you added the comprehension above. It feels a little unwieldy to include the ">=", but it is at least illustrative.

src/ansiblelint/rules/meta_runtime.py
Qalthos: I'm still not seeing the missing 'as' here. There might be something else you're going for but the text doesn't parse right to me without it

src/ansiblelint/rules/meta_runtime.md
```diff
@@ -2,21 +2,14 @@
...
-Currently supported versions of ansible-core are:
-
-- `2.13.x`
-- `2.14.x`
-- `2.15.x`
-
 This rule can produce messages such as:
 
-- `meta-runtime[unsupported-version]` - `requires_ansible` key must contain a supported version - 2.13.x, 2.14.x, 2.15.x.
+- `meta-runtime[unsupported-version]` - `requires_ansible` key must refer to a currently supported version such: >=2.14.0, >=2.15.0, >=2.16.0
```
Qalthos, line 7: Oh right this jumped out at me before, but it needs to be changed here, too
```suggestion
- `meta-runtime[unsupported-version]` - `requires_ansible` key must refer to a currently supported version such as: >=2.14.0, >=2.15.0, >=2.16.0
```

Qalthos (changes requested):

Qalthos (approved):

Issue #3961, opened 2023-12-28
Fails to accept namespace containing dash


sur5r:
##### Summary

Galaxy NG supports namespaces containing dashes for several months now. ansible-lint, however, still doesn't know about them.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
$ ansible-lint --version
ansible-lint 6.22.1 using ansible-core:2.16.2 ansible-compat:4.1.10 ruamel-yaml:0.18.5 ruamel-yaml-clib:0.2.8
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

- Run ansible-lint against a role having a namespace containing a dash, e.g. https://github.com/haufe-it/ansible-role-dehydrated

##### Desired Behavior

Ansible lint passes successfully.

##### Actual Behavior

```
ERROR    Computed fully qualified role name of haufe-it.dehydrated does not follow current galaxy requirements.
Please edit meta/main.yml and assure we can correctly determine full role name:

galaxy_info:
role_name: my_name  # if absent directory name hosting role is used instead
namespace: my_galaxy_namespace  # if absent, author is used instead

Namespace: https://galaxy.ansible.com/docs/contributing/namespaces.html#galaxy-namespace-limitations
Role: https://galaxy.ansible.com/docs/contributing/creating_role.html#role-names

As an alternative, you can add 'role-name' to either skip_list or warn_list.
```

after adding `role-name` to `skip_list` it gets even worse:

```
WARNING  Listing 1 violation(s) that are fatal
schema[meta]: $.galaxy_info.namespace 'haufe-it' does not match '^[a-z][a-z0-9_]+$'. See https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#using-role-dependencies
meta/main.yml:1  Returned errors will not include exact line numbers, but they will mention
the schema name being used as a tag, like ``schema[playbook]``,
``schema[tasks]``.

This rule is not skippable and stops further processing of the file.

If incorrect schema was picked, you might want to either:

* move the file to standard location, so its file is detected correctly.
* use ``kinds:`` option in linter config to help it pick correct file type.


Read documentation for instructions on how to ignore specific rule violations.

             Rule Violation Summary              
 count tag          profile rule associated tags 
     1 schema[meta] basic   core                 

Failed: 1 failure(s), 0 warning(s) on 16 files. Profile 'production' was required, but 'min' profile passed.
```

PR #3962, opened 2023-12-28
Allow dashes in legacy role namespaces

sur5r:
Fixes: #3961

### DIFF
--- a/src/ansiblelint/schemas/meta.json
+++ b/src/ansiblelint/schemas/meta.json
@@ -512,7 +512,7 @@
         "namespace": {
           "markdownDescription": "Used by molecule and ansible-lint to compute FQRN for roles outside collections",
           "minLength": 2,
-          "pattern": "^[a-z][a-z0-9_]+$",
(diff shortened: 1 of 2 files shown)


ssbarnea:
Can you please also update the link to documentation to a valid page that documents the new namespace requirements?

ssbarnea (changes requested):
This might not be enough to allow dashes in namespaces but we can probably add some tests ourselfs. Still, before this, please include links to documentation that includes the new namespace requirements as we do not want to just rely on "current behavior".

sur5r:
Unfortunately, I have no idea where this is documented.

I have to admit that I have spent way too many hours fighting windmills around this galaxy vs. galaxy-ng vs. namespaces vs. legacy namespaces vs. dash-to-underscore-to-dash-to-dot-conversion disaster.

I fully understand if you don't want to merge this without proper docs and or tests, but I'm completely lost here.

ssbarnea (approved):

PR #4004, opened 2024-02-02
Make File Permissions Mode a String

P-N-L:
The example was a number with a leading zero, which is interpreted as an octal number. This is unexpected and undesired behavior, and was caught by another ansiblelint rule ("Forbidden implicit octal value", `yaml[octal-values]`).

This change makes the value in the example a string, which means the fix to the file permissions rule violation will not cause another rule violation.


P-N-L:
Apologies if I am missing something obvious, but I am not allowed to add a label!
I would think that this would be "patch" or "skip-changelog".

ajinkyau (approved):

PR #4009, opened 2024-02-05
Drop support for Python 3.9

audgirka:


priyamsahoo (approved):

pyproject.toml
```diff
@@ -94,7 +93,7 @@ ensure_newline_before_comments = true
 line_length = 88
 
 [tool.mypy]
-python_version = 3.9
+python_version = 3.10
```
ssbarnea, line 96: I am wondering if quotes would not be required here to avoid being loaded as 3.1 float!

shatakshiiii:
@ajinkyau Can we change the python version to py310 for `PIP_CONSTRAINT` setenv variable over here: https://github.com/ansible/ansible-lint/blob/main/tox.ini#L69 ?

pyproject.toml
```diff
@@ -108,7 +107,7 @@ disallow_any_generics = true
 # site-packages is here to help vscode mypy integration getting confused
 exclude = "(build|dist|test/local-content|site-packages|~/.pyenv|examples/playbooks/collections|plugins/modules)"
 # https://github.com/python/mypy/issues/12664
-no_incremental = true
+incremental = false
```
audgirka, line 110: @ssbarnea please confirm if this is okay, as for `no_incremental` we're getting an error `Additional properties are not allowed ('no_incremental' was unexpected)`

shatakshiiii (approved):

ssbarnea (approved):

PR #4022, opened 2024-02-12
Disable codecov CI failure check

cristianonicolai:


audgirka (approved):

ssbarnea (changes requested):
blocking, we can bypass CI temporary but changing config is too dangerous and creates extra work for us.

I prefer for us to make a judgement call, do a good read of incoming PR and decide to bypass CI when we are sure is broken for reasons outside current change.

Issue #4007, opened 2024-02-04
Linting result of include_role with wrong syntax is suppressed


iay:
##### Summary

If you use `ansible.builtin.include_role` and use `role` rather than `name` as the name of the role to include, `ansible-lint` suppresses output.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
ansible-lint 6.22.2 using ansible-core:2.16.3 ansible-compat:4.1.11 ruamel-yaml:0.18.5 ruamel-yaml-clib:0.2.8
```

macOS Sonoma, Intel Mac.

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Correct syntax for `include_task` gives expected result (intentionally unnamed task is flagged):

```yaml
---
- name: Demonstrate linting issue.
  hosts: all
  tasks:
    - name: Include a role with the correct syntax
      ansible.builtin.include_role:
        name: foo

    - ansible.builtin.setup:
```

```bash
ansible-lint test.yml
WARNING  Skipped installing old role dependencies due to running in offline mode.
WARNING  Skipped installing collection dependencies due to running in offline mode.
WARNING  Listing 1 violation(s) that are fatal
name[missing]: All tasks should be named.
test.yml:9 Task/Handler: setup 

Read documentation for instructions on how to ignore specific rule violations.

              Rule Violation Summary              
 count tag           profile rule associated tags 
     1 name[missing] basic   idiom                

Failed: 1 failure(s), 0 warning(s) on 1 files. Last profile that met the validation criteria was 'min'.
```

Incorrect syntax still works, but linting output is suppressed even for other issues:

```yaml
---
- name: Demonstrate linting issue.
  hosts: all
  tasks:
    - name: Include a role with the wrong syntax
      ansible.builtin.include_role:
        role: foo

    - ansible.builtin.setup:
```

```bash
ansible-lint test.yml
WARNING  Skipped installing old role dependencies due to running in offline mode.
WARNING  Skipped installing collection dependencies due to running in offline mode.
WARNING  Listing 2 violation(s) that are fatal
WARNING  A rule (load-failure) that is not part of a collection cannot access its configuration.
A rule (load-failure) that is not part of a collection cannot access its configuration.
```

##### Desired Behavior

I think it would be reasonable for `ansible-lint` to say which fatal violation it had detected.

##### Actual Behavior

[...]

PR #4026, opened 2024-02-14
Fix error suppression when syntax is incorrect

audgirka:
Fixes: #4007

### DIFF
--- a/src/ansiblelint/runner.py
+++ b/src/ansiblelint/runner.py
@@ -27,10 +27,6 @@ from ansible_compat.runtime import AnsibleWarning
 
 import ansiblelint.skip_utils
 import ansiblelint.utils
-from ansiblelint._internal.rules import (
-    BaseRule,
-    LoadingFailureRule,
-)
 from ansiblelint.app import App, get_app
 from ansiblelint.constants import States
 from ansiblelint.errors import LintWarning, MatchError, WarnSource
@@ -50,6 +46,7 @@ from ansiblelint.utils import (
 if TYPE_CHECKING:
     from collections.abc import Callable, Generator
 
+    from ansiblelint._internal.rules import BaseRule
     from ansiblelint.config import Options
     from ansiblelint.constants import FileType
     from ansiblelint.rules import RulesCollection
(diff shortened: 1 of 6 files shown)


examples/playbooks/example.yml
```diff
@@ -10,7 +10,7 @@
       action: ansible.builtin.command echo {{this_variable}} is not set in this playbook
```
ssbarnea, line 10: I am sure you did not want to touch this file!

src/ansiblelint/runner.py
```diff
@@ -458,7 +453,7 @@ def _emit_matches(self, files: list[Lintable]) -> Generator[MatchError, None, No
                 except MatchError as exc:
                     if not exc.filename:  # pragma: no branch
                         exc.filename = str(lintable.path)
-                    exc.rule = LoadingFailureRule()
+                    exc.rule = self.rules.rules[1]
```
ssbarnea, line 456: Do not assume order of rules! If you want you might modify the getter to allow you to get a rule by its ID from self.rules, but never use an index!

src/ansiblelint/utils.py
```diff
@@ -275,110 +280,169 @@ def template(
...
-
-    if "import_playbook" in k and COLLECTION_PLAY_RE.match(v):
-        # Any import_playbooks from collections should be ignored as ansible
-        # own syntax check will handle them.
-        return []
-
-    # handle include: filename.yml tags=blah
-    # pylint: disable=unused-variable
-    (command, args, kwargs) = tokenize(f"{k}: {v}")
+@dataclass
```
ssbarnea, line 283: A dataclass with code... not its supposed use but I guess that you wanted to make use of the fancy syntax provided by them. OK.

ssbarnea (changes requested):

examples/playbooks/example.yml
audgirka: When I opened this to check made a few changes and reverted them but then on save it started removing some spaces and when I reverted this change strangely pre-commit started picking this change and kept on removing the spaces

Issue #4028, opened 2024-02-15
Ansible-Lint uppercases handler names but ignores the notifications


heuwerk:
##### Summary

when running `ansible-lint --fix` the names of the handler tasks are correctly renamed to begin with upper case letter.
The Problem is that the `notify: my handler` remains unchanged, so the next run of the Playbook will fail.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 24.2.0 using ansible-core:2.15.3 ansible-compat:4.1.11 ruamel-yaml:0.18.5 ruamel-yaml-clib:0.2.8
```

- ansible installation method: OS package
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

1. Create a playbook, that calls a handler but the name and notify of the handler is lowercase (see example playbook below)
2. Run `ansible-lint --fix` on the playbook
3. Run `ansible-playbook playbook.yaml` and it fails

```yaml
---
- name: Test Playbook
  hosts: localhost
  tasks:
    - name: Task that always changes
      ansible.builtin.debug:
        msg: "I always change!"
      changed_when: true
      notify: my handler

  handlers:
    - name: my handler
      ansible.builtin.debug:
        msg: "I never run :("
```

##### Desired Behavior

ansible-lint should *uppercase* the handler name **and** the notify call, so the playbook keeps working after running `ansible-lint --fix`

Possible security bugs should be reported via email to `security@ansible.com`

##### Actual Behavior

`ERROR: The requested handler 'my handler' was not found in either the main handlers list nor in the listening handlers list`

Please give some details of what is happening. Include a [minimum complete
verifiable example] with:

- minimized playbook to reproduce the error
- the output of running ansible-lint including the command line used
- if you're getting a stack trace, also the output of
  `ansible-playbook --syntax-check playbook`

```paste below
---
- name: Test Playbook
  hosts: localhost
  tasks:
    - name: Task that always changes
      ansible.builtin.debug:
        msg: "I always change!"
      changed_when: true
      notify: my handler

  handlers:
# Handler after ansible-lint --fix run
    - name: My handler
      ansible.builtin.debug:
        msg: "I never run :("

$ ansible-playbook playbook.yaml
PLAY [Test Playbook]

TASK [Gathering Facts]
ok: [localhost]

TASK [Task that always changes]*******************
ERROR: The requested handler 'my handler' was not found in either the main handlers list nor in the listening handlers list
```

[...]

PR #4038, opened 2024-02-20
Update task name in `notify` for a task against `name[casing]` error

audgirka:
Fixes: #4028


ssbarnea (changes requested):
Based on the example from https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_handlers.html#naming-handlers you will also have to update the 'listen' inside handlers to capitalize it too, or otherwise you will might break notify when it does refer to a topic (instead of name).

ssbarnea (approved):

PR #4055, opened 2024-03-02
Set global `cache_dir_lock`

### LINKED ISSUE #4030: `cache_dir_lock` in `main` is always `None`, so the lockfile never gets cleaned up
##### Summary

`cache_dir_lock` is initialized as `None` here, in the global scope I believe: https://github.com/ansible/ansible-lint/blob/7849f27453fe1d9c29a43c71cff8f5b666bd43b6/src/ansiblelint/__main__.py#L82

In `initialize_options`, a new, frame-only `cache_dir_lock` is created here: https://github.com/ansible/ansible-lint/blob/7849f27453fe1d9c29a43c71cff8f5b666bd43b6/src/ansiblelint/__main__.py#L148-L152

I think pylint was correct to indicate that the global name was redefined. This variable isn't marked as `global` and so the `cache_dir_lock` in `initialize_options` is local to `initialize_options` only.

The value of `cache_dir_lock` (the `global` one) is still `None` later on, so this conditional is always falsy and the lockfile therefore never gets cleaned up: https://github.com/ansible/ansible-lint/blob/7849f27453fe1d9c29a43c71cff8f5b666bd43b6/src/ansiblelint/__main__.py#L380-L382

I believe this behavior is introduced by https://github.com/ansible/ansible-lint/commit/c6bd9a7a92592a8a700b26d659cf9232bd9eaba5.

I'd love to PR to resolve this, but I don't know if you prefer the `global` approach or just returning the lock from `initialize_options`.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
$ ansible-lint --version
ansible-lint 24.2.0 using ansible-core:2.16.3 ansible-compat:4.1.11 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.8
```

##### STEPS TO REPRODUCE

In one session

```console
$ rm -r ~/.cache/ansible-compat/*
$ watch -n 1 'ls ~/.cache/ansible-compat/**/.lock'  # the quotes are so the shell under watch will expand it, not your current shell
```

and in another session

```console
$ pushd /tmp
$ mkdir blah
$ vim main.yml  # see below for MCVE playbook, but it can really be any playbook
$ ansible-lint
```

watch the lockfile get created when `ansible-lint` is run, and then it shouldn't get cleaned up.

##### Desired Behavior

The `cache_dir_lock` that `main` uses is not `None`, either by the `global` keyword or `initialize_options` returning it to its caller, `main`; in which case, the lock can be correctly cleaned up after ansible-lint finishes.

##### Actual Behavior

Ansible-lint doesn't clean up the lockfile, because `cache_dir_lock` is `None` in `main`.

Please give some details of what is happening. Include a [minimum complete
verifiable example] with:

- minimized playbook to reproduce the error

```yaml
- name: "MCVE"
  gather_facts: false
  hosts: "localhost"
  tasks:
    - name: "Hello world"
      ansible.builtin.debug:
        msg: "hello world"
```

- the output of running ansible-lint including the command line used

```text
Passed: 0 failure(s), 0 warning(s) on 1 files. Last profile that met the validation criteria was 'production'.
```

guppy0130:
* in `initialize_options`, we need to update the globally-scoped
  `cache_dir_lock` so that later on, `main` can release the lock and
  cleanup the lockfile.

Fixes #4030

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -248,7 +248,8 @@ ignore = [
   "RUF012", # Mutable class attributes should be annotated with `typing.ClassVar`
   "TRY",
   "PERF203",
-  "PD011" # We are not using pandas, any .values attributes are unrelated
+  "PD011", # We are not using pandas, any .values attributes are unrelated
+  "PLW0603" # global lock file in cache dir
 ]
 select = ["ALL"]
 target-version = "py310"
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -79,6 +79,7 @@ if TYPE_CHECKING:
 
 
 _logger = logging.getLogger(__name__)
+# cache_dir_lock could be set in `initialize_options`
 cache_dir_lock: None | FileLock = None
 
 
@@ -144,8 +145,9 @@ def initialize_options(arguments: list[str] | None = None) -> None:
     if options.cache_dir:
         options.cache_dir.mkdir(parents=True, exist_ok=True)
 
+    global cache_dir_lock  # pylint: disable=global-statement
     if not options.offline:  # pragma: no cover
-        cache_dir_lock = FileLock(  # pylint: disable=redefined-outer-name
+        cache_dir_lock = FileLock(
             f"{options.cache_dir}/.lock",
         )
         try:
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -79,8 +79,6 @@ if TYPE_CHECKING:
 
 
 _logger = logging.getLogger(__name__)
-# cache_dir_lock could be set in `initialize_options`
-cache_dir_lock: None | FileLock = None
 
 
 class LintLogHandler(logging.Handler):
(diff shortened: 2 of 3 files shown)


guppy0130:
This is the smallest possible changeset that I believe resolves the issue?

There's a couple alternative implementations that I believe would be better:

* storing `cache_dir_lock` in the `options` object
  * I don't know why `cache_dir_lock` was removed from the `options` object in #3585 - lacking context for that PR
* `initialize_options` could return the `LockFile` object and we perform the lock acquisition and release in `main`

audgirka (approved):

audgirka:
> This is the smallest possible changeset that I believe resolves the issue. Linters are not happy with the usage of `global` though.
> 
> There's a couple alternative implementations that I believe would be better:
> 
>     * storing `cache_dir_lock` in the `options` object
>       
>       * I don't know why `cache_dir_lock` was removed from the `options` object in [Address parallel install errors with ansible-galaxy on GHA #3585](https://github.com/ansible/ansible-lint/pull/3585) - lacking context for that PR
> 
>     * `initialize_options` could return the `LockFile` object and we perform the lock acquisition and release in `main`

@ssbarnea can you please confirm for the first point which talks about removing `cache_dir_lock` from `options` object.

PR #4108, opened 2024-04-12
Don't inject CWD parent into PATH

garymm:
Previously if own_location was set to "", then os.path.realpath("") resolves to CWD, and then
if CWD/../ansible exists, CWD/.. was injected into PATH. This doesn't seem desirable or necessary.


garymm:
An alternate fix for my issue would be to check if `parent / "ansible"` is executable before adding `parent` to the PATH. LMK if you prefer that fix.

ssbarnea (approved):

Issue #4091, opened 2024-04-04
KeyError when fixing playbook containing multiple blocks with wrong key order


zack-mayoh:
##### Summary

`ansible-lint --fix` breaks with a `KeyError` in the `key-order` rule when run on a yaml file containing multiple blocks where blocks after the first have incorrect key order.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 6.22.2 using ansible-core:2.15.10 ansible-compat:4.1.11 ruamel-yaml:None ruamel-yaml-clib:None
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Run `ansible-lint --fix` on the following test playbook:
```yaml
---
- name: Test multiple blocks with when in wrong order
  hosts: all
  tasks:
    - name: First block
      when: true
      block:
        - name: Test1
          ansible.builtin.debug:
            msg: Foo

- name: Second play
  hosts: all
  tasks:
    - name: Second block
      block:
        - name: Test2
          ansible.builtin.debug:
            msg: Baz
      when: true
```

##### Desired Behavior

Should fix the key order of the incorrect block without breaking.

##### Actual Behavior

Results of `ansible-lint --fix`:
```console
$ ansible-lint --fix test.yml
Traceback (most recent call last):
  File "/usr/local/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
  File "/usr/local/lib/python3.9/site-packages/ansiblelint/__main__.py", line 402, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
  File "/usr/local/lib/python3.9/site-packages/ansiblelint/__main__.py", line 379, in main
    fix(runtime_options=options, result=result, rules=rules)
  File "/usr/local/lib/python3.9/site-packages/ansiblelint/__main__.py", line 248, in fix
    _do_transform(result, options)
  File "/usr/local/lib/python3.9/site-packages/ansiblelint/__main__.py", line 203, in _do_transform
    transformer.run()
  File "/usr/local/lib/python3.9/site-packages/ansiblelint/transformer.py", line 118, in run
    self._do_transforms(file, ruamel_data or data, file_is_yaml, matches)
  File "/usr/local/lib/python3.9/site-packages/ansiblelint/transformer.py", line 156, in _do_transforms
    match.rule.transform(match, file, data)
[...]

PR #4116, opened 2024-04-18
Fix KeyError on `key-order` rule with blocks

Qalthos:
Fixes https://github.com/ansible/ansible-lint/issues/4091

`_get_path_to_task_in_nested_tasks_block()` seems to do the wrong thing under certain circumstances when lineno > next_task_line_index. It seems like we should never want that anyway, so just skip the whole thing until there is not a later play that is earlier than lineno.


ssbarnea (changes requested):
Please add a test.

audgirka (approved):

Issue #4085, opened 2024-04-02
ignore_unreachable doesnt work with jinja templates


RayJin2000:
##### Summary
When I use a Jinja Template in ignore_unreachable I get the following error:
```console (paste below)
WARNING  Listing 1 violation(s) that are fatal
schema[playbook]: $[0].tasks[0].ignore_unreachable "{{ 'yes' | bool }}" is not of type 'boolean'
playbooks/test.yml:1  Returned errors will not include exact line numbers, but they will mention
the schema name being used as a tag, like ``schema[playbook]``,
``schema[tasks]``.

This rule is not skippable and stops further processing of the file.

If incorrect schema was picked, you might want to either:

* move the file to standard location, so its file is detected correctly.
* use ``kinds:`` option in linter config to help it pick correct file type.
```

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 24.2.1 using ansible-core:2.16.4 ansible-compat:4.1.11 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.7
```

##### STEPS TO REPRODUCE

```yaml (paste below)
---
- name: Test
  hosts: localhost
  tasks:
    - name: Debug
      ansible.builtin.debug:
        msg: Bla
      ignore_unreachable: "{{ 'yes' | bool }}"

```

##### Desired Behavior
```console
> ansible-lint -c .ansible-lint playbooks/test.yml 

Passed: 0 failure(s), 0 warning(s) on 1 files. Last profile that met the validation criteria was 'production'.
```
##### Actual Behavior

```console 
> ansible-lint -c .ansible-lint playbooks/test.ym

ARNING  Listing 1 violation(s) that are fatal
schema[playbook]: $[0].tasks[0].ignore_unreachable "{{ 'yes' | bool }}" is not of type 'boolean'
playbooks/test.yml:1  Returned errors will not include exact line numbers, but they will mention
the schema name being used as a tag, like ``schema[playbook]``,
``schema[tasks]``.

This rule is not skippable and stops further processing of the file.

If incorrect schema was picked, you might want to either:

* move the file to standard location, so its file is detected correctly.
* use ``kinds:`` option in linter config to help it pick correct file type.


Read documentation for instructions on how to ignore specific rule violations.

               Rule Violation Summary                
 count tag              profile rule associated tags 
     1 schema[playbook] basic   core                 

[...]

PR #4120, opened 2024-04-23
Make `ignore_unreachable` accept jinja templates

audgirka:
Fixes: #4085

### DIFF
--- a/test/schemas/test/playbooks/ignore-unreachable.yml
+++ b/test/schemas/test/playbooks/ignore-unreachable.yml
@@ -1,0 +1,13 @@
+---
+- name: Test
+  hosts: localhost
+  tasks:
+    - name: Debug
+      ansible.builtin.debug:
+        msg: ignore_unreachable should be a boolean
+      ignore_unreachable: true
+
+    - name: Debug
+      ansible.builtin.debug:
+        msg: "foo"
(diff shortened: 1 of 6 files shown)


ssbarnea (changes requested):
Please include an use of templated value for the ignore_reachable in the schema test files, so we would not accidentally remove it in the future.

ssbarnea (changes requested):
This is missing a positive test.

test/schemas/test/playbooks/ignore-unreachable.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Test
+  hosts: localhost
+  tasks:
+    - name: Debug
+      ansible.builtin.debug:
+        msg: ignore_unreachable should be a boolean
+      ignore_unreachable: true
```
ssbarnea, line 8: I think that the bug was about using jinja with ignore_unreachable, poor booleans did  work. Please be sure you include some jinja instead.

audgirka: @ssbarnea it's still failing and I'm trying to fix the same

audgirka: I thought the PR was in draft!

ssbarnea (approved):

Issue #4122, opened 2024-04-26
Undefined output when hitting unknown-module


thekad:
##### Summary

We starting to use ansible-lint on https://github.com/redhatci/ansible-collection-redhatci-ocp repo, however we're hitting a strange issue:
* Wrong behavior: [this run](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/8841068617/job/24277540191)  shows only 3 violations when it should be showing thousands
* Right behavior: like in [this other run](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/8841230920/job/24278001753) (will load slowly like I said a ton of issues)

The only difference between the two is a single task in `roles/acm_setup/tasks/main.yml` ln29 where it goes from `k8s_info` (wrong) to `community.kubernetes.k8s_info` (right).

The logs show all the roles appear to be analyzed, but only 3 failures are reported.

For now, I am FQCN the problematic task and moving ahead, I understand this is a very specific corner case.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

Installing in a GHA like so:

```console
pip install "ansible-lint[lock] @ git+https://github.com/ansible/ansible-lint@v6"
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/[3](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/8841230920/job/24278001753#step:4:3).11.9/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.9/x6[4](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/8841230920/job/24278001753#step:4:4)/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.9/x[6](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/8841230920/job/24278001753#step:4:6)4
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.9/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.9/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.9/x64/lib
...
[...]

PR #4133, opened 2024-05-05
Continue linting other files when one has syntax-check errors

ssbarnea:
Syntax-errors will no longer prevent the linter from checking other rules in other files.

Fixes: #4122


src/ansiblelint/_internal/rules.py
```diff
@@ -96,7 +96,7 @@ def getmatches(self, file: Lintable) -> list[MatchError]:
                     _logger.warning(
                         "Ignored exception from %s.%s while processing %s: %s",
                         self.__class__.__name__,
-                        method,
+                        method.__name__,
```
ssbarnea, line 99: Unrelated, but improves the messaging.

shatakshiiii (approved):

PR #4135, opened 2024-05-06
Update supported versions of ansible to 2.15 - 2.17

epacific1:
Remove "2.14." from this list and add "2.17.", since 2.14 support is being dropped and 2.17 is starting support: https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/rules/meta_runtime.py#L35

The passing test will need to be updated as well: https://github.com/ansible/ansible-lint/blob/main/examples/meta_runtime_version_checks/pass/meta/runtime.yml

This test will need the lower bound increased to at least 2.15.0 and the upper bound should be removed.

### DIFF
--- a/examples/meta_runtime_version_checks/pass/meta/runtime.yml
+++ b/examples/meta_runtime_version_checks/pass/meta/runtime.yml
@@ -1,2 +1,2 @@
 ---
-requires_ansible: ">=2.14.0,<2.15"
+requires_ansible: ">=2.15.0"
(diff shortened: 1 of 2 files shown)


ssbarnea (approved):

ssbarnea:
@epacific1 be sure that the tests are passing, we still want to check for both cases: unsupported and unknown version. Other than this, it LGTM, thanks!

PR #4140, opened 2024-05-07
Address ruff TRY

ssbarnea:


pyproject.toml
```diff
@@ -143,6 +143,8 @@ max-statements = 60
 disable = [
   # Disabled on purpose:
   "line-too-long", # covered by black
+  "no-else-raise", # conflicts with ruff TRY
+  "no-else-return", # conflicts with ruff TRY
```
Qalthos, line 147: I don't think these are actually conflicts... you _can_ move them to an `else` block to get that section out of the `try` block, but all of these are trailing `else` blocks that can be dropped entirely and keep the same outcome.

src/ansiblelint/schemas/main.py
```diff
@@ -104,6 +103,8 @@ def validate_file_schema(file: Lintable) -> list[str]:
                 message += "."
             message += f" See {documentation_url}"
         return [message]
+    else:
+        return [message]
     return []
```
Qalthos, line 108: Additionally, this now highlights some weirdness that already existed. All of the `except`s already have returns, and now with the return in `else`, we can see that this `return []` will never get called.

Qalthos (dismissed):
The rest of it looks pretty good. I tried dropping those lint exclusions and promoting the `return`s out of their `else` blocks, and everything seems to be passing as normal

Issue #4166, opened 2024-05-16
Octal-like strings are quoted unnecessarily


kousu:
##### Summary

`ansible-lint --fix` will quote any string value that starts with a "0", whether or not it needs to be. This behaviour was meant for [handling the mess that are octals in YAML](https://github.com/ansible/ansible-lint/issues/2965), and was introduced in https://github.com/ansible/ansible-lint/pull/3030 and did a good job of helping that, but it's spilled out and is affecting all strings! 🍦 

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
% ansible-lint --version
WARNING: PATH altered to include /opt/homebrew/Cellar/ansible-lint/24.2.3_1/libexec/bin, /opt/homebrew/Cellar/ansible-lint/24.2.3_1/libexec/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.
ansible-lint 24.2.3 using ansible-core:2.16.6 ansible-compat:4.1.11 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.8
```

- ansible installation method: `brew install ansible`
- ansible-lint installation method: `brew install ansible-lint && ansible-galaxy collection install --force community.general`

    (I don't know why I needed to re-install community.general to get ansible-lint working, but that's a different bug, maybe in brew)

##### STEPS TO REPRODUCE

```yaml
# config.yml 
---
- name: Configure things
  hosts: all
  become: false
  tasks:
    - name: Configure ii listen address
      ansible.builtin.command: |
        cmd: |
          tee /var/run/ii/config/addr
        stdin:
          {{ listen_addr }}:{{ listen_port }}
      vars:
        listen_addr: 0.0.0.0
        listen_port: 0o7828
        hex_input: 0x8294
        dec_input: 29848
        oct_input: 0364
        str_input: 0steps
        str_input2: 1steps
      changed_when: false
```

Run:

```console (paste below)
ansible-lint --fix
```

##### Actual Behavior

```console (paste below)
% ansible-lint --fix
WARNING: PATH altered to include /opt/homebrew/Cellar/ansible-lint/24.2.3_1/libexec/bin, /opt/homebrew/Cellar/ansible-lint/24.2.3_1/libexec/bin :: This is usually a sign of broken local setup, which can cause unexpected behaviors.

Modified 1 files.
Passed: 0 failure(s), 0 warning(s) on 1 files. Last profile that met the validation criteria was 'production'.
[...]

PR #4167, opened 2024-05-16
Don't quote strings that start with 0 when running ansible-lint --fix.

kousu:
In order to try to handle [the mess that are octals in YAML](https://github.com/ansible/ansible-lint/issues/2965), some [special-case code was added](https://github.com/ansible/ansible-lint/pull/3030) to handle leading-zeros. But it caught too much, and would quote strings like `00-header` and `0.0.0.0`, even when .yamllint doesn't require them:

```
quoted-strings:
    required: false
```

and it generates awkward lists, like

```diff
   loop:
     - "00-header"
     - 10-help-text
     - 50-landscape-sysinfo
     - 50-motd-news
     - 88-esm-announce
     - 97-overlayroot
```

Fixes https://github.com/ansible/ansible-lint/issues/4166

### DIFF
--- a/src/ansiblelint/yaml_utils.py
+++ b/src/ansiblelint/yaml_utils.py
@@ -642,14 +642,29 @@ class FormattedEmitter(Emitter):
             and self.event.value.startswith("0")
             and len(self.event.value) > 1
         ):
-            if self.event.tag == "tag:yaml.org,2002:int" and self.event.implicit[0]:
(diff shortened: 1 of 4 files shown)


ssbarnea (approved):

kousu:
Thank you for the merge @ssbarnea !
It got lost in my inbox somewhere but <3

PR #4170, opened 2024-05-16
Fix false positive reports in is_handler function

MalKeshar:
Found in the process of writing and checking custom rules.

The "is_handler" function considers all tasks from the file to be handlers if the file name contains the substring "handlers".

Therefore, all tasks in the check_handlers_playbook.yaml playbook were accepted as handlers.

I'm not sure if it's worth relying on the filename or path at all. But if we do, then maybe it makes sense to look for the full string, not just a part of it.

### DIFF
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -785,7 +785,8 @@ class Task(dict[str, Any]):
         if isinstance(self._normalized_task, dict):
             file_name = str(self._normalized_task["action"].get(FILENAME_KEY, None))
             if file_name:
-                is_handler_file = "handlers" in str(file_name)
+                paths = file_name.split("/")
+                is_handler_file = "handlers" in paths
         return is_handler_file if is_handler_file else ".handlers[" in self.position
 
     def __repr__(self) -> str:


MalKeshar:
Actually there is one more issue with detecting handlers.

This task will be detected as handler
```
- name: Reload sysctl (good handler)
  ansible.builtin.shell:
    cmd: "sysctl --system -p"
  changed_when: false
```
because it's object has attribute ```__file__```

```
DEBUG    file_name: roles/xservice/handlers/main.yaml
DEBUG    self.position: .[3]
DEBUG    Running rule task-has-tag
DEBUG    Debug: task: Reload sysctl good, args: {'cmd': 'sysctl --system -p'}, action: {'__ansible_module__': 'shell', '__ansible_module_original__': 'ansible.builtin.shell', 'cmd': 'sysctl --system -p', '__line__': 3, '__file__': PosixPath('roles/xservice/handlers/main.yaml')}
```


This task won't be detected as handler

```
- name: Reload sysctl ( bad handler)
  ansible.builtin.shell: "sysctl --system -p"
  changed_when: false

```

because it's object hasn't attribute  ```__file__```

```
DEBUG    file_name: roles/xservice/handlers/main.yaml
DEBUG    self.position: .[0]
DEBUG    Debug: task: Reload sysctl bad, args: {'_raw_params': 'sysctl --system -p'}, action: {'__ansible_module__': 'shell', '__ansible_module_original__': 'ansible.builtin.shell', '_raw_params': 'sysctl --system -p'}
```

Both "self.position" don't have "handlers" word in it. 

This function probably requires some redesign. Unfortunally python is not my lang, so I'm not gonna dig it dipper by myself.

ssbarnea (approved):

PR #4204, opened 2024-06-04
Avoid displaying upgrade warning when installation is not pip

ssbarnea:
Related: AAP-23001 AAP-24936

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -86,6 +86,12 @@ def test_get_version_warning(
     assert len(msg.split("\n")) == outlen
 
 
+def test_get_version_warning_no_pip(mocker: MockerFixture) -> None:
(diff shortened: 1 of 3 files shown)


test/test_main.py
```diff
@@ -86,6 +86,12 @@ def test_get_version_warning(
     assert len(msg.split("\n")) == outlen
 
 
+def test_get_version_warning_no_pip(mocker: MockerFixture) -> None:
+    """Test that we do not display any message is install method is not pip."""
```
shatakshiiii, line 90: Minor typo in the docstring: """Test that we do not display any message **_if_** install method is not pip."""

shatakshiiii (approved):

PR #4210, opened 2024-06-05
Edit path_inject() to compare Paths instead of strings

janmondry:
There's an issue causing unnecessary warning:
https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/__main__.py#L464
`if (parent / "ansible").exists() and str(parent) not in paths`

Personally I encountered a case when:
- a path in PATH ends with a trailing slash (which is correct) (eg. `/app/bin/`)
- `str(parent)` returned almost exactly the same path, but without the trailing slash, (eg. `/app/bin`)

which resulted in failing the string comparison (eg. `'/app/bin/' not in ['/app/bin']`).

Comparing Path objects solves the problem.


janmondry:
pre-commit.ci autofix

ssbarnea:
@janmondry Can you please check this as the use of `site` seems to be causing some very weird test failures.

PR #4226, opened 2024-06-20
Fixup YAML utils comment

kousu:
Related: https://github.com/ansible/ansible-lint/pull/4167

### DIFF
--- a/src/ansiblelint/yaml_utils.py
+++ b/src/ansiblelint/yaml_utils.py
@@ -644,7 +644,7 @@ class FormattedEmitter(Emitter):
         ):
             # We have an as-yet unquoted token that starts with "0" (but is not itself the digit 0).
             # It could be:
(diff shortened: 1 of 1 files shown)


kousu:
I'm also only 50% sure I understand this anymore.

https://github.com/ansible/ansible-lint/blob/e6df6219c7e6d4b38a0740ee1c8c50f176725470/src/ansiblelint/yaml_utils.py#L647

seems to contradict

https://github.com/ansible/ansible-lint/blob/e6df6219c7e6d4b38a0740ee1c8c50f176725470/src/ansiblelint/yaml_utils.py#L650-L656

Why isn't it `self.event.tag = "tag:yaml.org,2002:int"`? Or just not having that line?

The tests are passing so maybe it's okay, but maybe the tests aren't testing hexadecimals fully?

ssbarnea (approved):

PR #4213, opened 2024-06-07
Fix working directory when used as an github action

bkaraoren:
When used container, it is not working as statically defined. I believe this should be a bug. 
During the wget step and also python steps it is not considered to be used what is set previously. Even the wget works on the working directory, the file saved to some other directory and not cnsiderd in the code level.

fails like below:
```
Run ansible/ansible-lint@main
  with:
    working_directory: /__w/git-repo/git-repo
    setup_python: true
Run if [[ -n "/__w/git-repo/git-repo" ]]; then
  if [[ -n "/__w/git-repo/git-repo" ]]; then
    echo "working_directory=/__w/git-org/git-repo" >> $GITHUB_OUTPUT
  else
    echo "working_directory=/opt/actions-runner/_work/git-repo/git-repo" >> $GITHUB_OUTPUT
  fi
  shell: bash --noprofile --norc -e -o pipefail {0}
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Run wget --output-document=/opt/actions-runner/_work/git-repo/git-repo/.git/ansible-lint-requirements.txt https://raw.githubusercontent.com/ansible/ansible-lint/$GH_ACTION_REF/.config/requirements-lock.txt
  wget --output-document=/opt/actions-runner/_work/git-repo/git-repo/.git/ansible-lint-requirements.txt https://raw.githubusercontent.com/ansible/ansible-lint/$GH_ACTION_REF/.config/requirements-lock.txt
  shell: bash --noprofile --norc -e -o pipefail {0}
  env:
    GH_ACTION_REF: main
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
/opt/actions-runner/_work/git-repo/git-repo/.git/ansible-lint-requirements.txt: No such file or directory
Error: Process completed with exit code 1.
```

PR #4103, opened 2024-04-09
Fix #3938: don't assume working_directory is github.workspace

ajfabbri:
Setting `working_directory` for ansible-lint action would fail due to hard-coded `.git`, introduced in commit 6f728e0c, when fetching `.config/requirements-lock.txt`.

This fix replaces `.git` with `${{ github.workspace }}/.git` to make `working_directory` argument work again.

Tested that this does fix my github action, i.e.:

```
name: ansible-lint
on:
  pull_request:
    branches: ["main", "release/v*"]
jobs:
  build:
    name: Ansible Lint
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: Run ansible-lint
        uses: ajfabbri/ansible-lint@main
        # this is broken :-( https://github.com/ansible/ansible-lint/issues/3938
        with:
          working_directory: iac
```
Fixes #3938

Issue #3938, opened 2023-12-11
github action fails to download ansible-lint-requirements.txt, assumes .git dir exists in working directory


tigattack:
##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

github actions

##### STEPS TO REPRODUCE

ansible-lint workflow:

```yml
- uses: ansible/ansible-lint@v6.22.1
  with:
    working_directory: ansible
```

##### Desired Behavior

Requirements file is downloaded to a path that exists and used successfully.

##### Actual Behavior

It seems that the action assumes there will be a `.git` directory, which isn't necessarily the case if using anything other than the repository root as the working directory (such as shown in the example above).

```
2023-12-11T02:01:18.5977779Z ##[group]Run wget --output-document=.git/ansible-lint-requirements.txt https://raw.githubusercontent.com/ansible/ansible-lint/$GH_ACTION_REF/.config/requirements-lock.txt
2023-12-11T02:01:18.5980286Z [36;1mwget --output-document=.git/ansible-lint-requirements.txt https://raw.githubusercontent.com/ansible/ansible-lint/$GH_ACTION_REF/.config/requirements-lock.txt[0m
2023-12-11T02:01:18.6021388Z shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
2023-12-11T02:01:18.6022054Z env:
2023-12-11T02:01:18.6022395Z   GH_ACTION_REF: v6.22.1
2023-12-11T02:01:18.6022818Z ##[endgroup]
2023-12-11T02:01:18.6176869Z .git/ansible-lint-requirements.txt: No such file or directory
2023-12-11T02:01:18.6198199Z ##[error]Process completed with exit code 1.
```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

ssbarnea:
Look at our action usage example https://github.com/ansible/ansible-lint --- you MUST clone before calling the action or it will fail.

tigattack:
I am doing so. Here's the full workflow (private repo so no link, sorry):

```yml
---
name: Ansible Lint

on:
  push:
    paths: ['ansible/**']
  workflow_dispatch:

jobs:
  lint:
    name: Ansible Lint
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v4

      - name: Run ansible-lint
        uses: ansible/ansible-lint@v6.22.1
        with:
          working_directory: ansible
```

[...]

PR #4232, opened 2024-06-26
GitHub Action: avoid extra working directory prefix for wget

ptoscano:
The whole shell step for wget is run with the working directory for the action; hence, drop the extra prefix to the --output-document parameter of the wget command.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -42,7 +42,7 @@ runs:
         GH_ACTION_REF: ${{ github.action_ref || 'main' }}
       working-directory: ${{ steps.inputs.outputs.working_directory }}
       run: |
-        wget --output-document=${{ steps.inputs.outputs.working_directory }}/.git/ansible-lint-requirements.txt https://raw.githubusercontent.com/ansible/ansible-lint/$GH_ACTION_REF/.config/requirements-lock.txt
+        wget --output-document=.git/ansible-lint-requirements.txt https://raw.githubusercontent.com/ansible/ansible-lint/$GH_ACTION_REF/.config/requirements-lock.txt
 
     - name: Set up Python
       if: inputs.setup_python == 'true'


richm:
we fixed this in system roles by using an absolute path for the working directory (for ansible-lint and ansible-test also) - https://github.com/linux-system-roles/kernel_settings/pull/206

bkaraoren (commented):
When you run this on a container it was not getting the correct `working-directory` from GITHUB environment variables. GITHUB environment variables are only available for the RUNNER, not on a container in RUNNER. Please check the previous requirements by the changes.

ptoscano:
@bkaraoren I think you misunderstand what this PR does: the shell `run` block used to invoke `wget` already has the `working_directory` parameter set, so using it _again_ in the output filename of `wget` is not correct.

What you describe sounds like a problem in the detection of the default working directory in case the `working_directory` parameter of this action is not specified (see the first set of the action, currently labelled "Process inputs"). In that case, I think that automatic detection ought to be fixed.

bkaraoren:
> tection ought to be fixe

@ptoscano yep, that sounds also good. without providing any directory such as `${{ steps.inputs.outputs.working_directory }}` I also expect it to work but former commit needs to be checked why it has been reverted back from not using any working directory in the commit #3938. Could be useful to check it to understand the reason what was the reason to include.

ssbarnea (changes requested):
Based on given reviews, I think that is better not to touch it unless the change is also adding tests, especially for called via container approach. Without this, we will probably close the PRs.

tigattack:
We're going in circles here; this PR brings us back to the same state as before #4103.

For myself and others using a GitHub action in a repository subdirectory, #4103 addressed a major issue (see #3938, now #4323). However, it also introduced problems for other users.

In response, #4213 was merged, but this caused the action to become unusable again without resorting to hacky workarounds for the same use case.

Instead of continuing in this cycle, we should take a step back and reevaluate the approach, as each "fix" solves one group's problem but introduces a new breaking bug for another.

ssbarnea:
@tigattack @bkaraoren @ptoscano Please check https://github.com/ansible/ansible-lint/pull/4340 which will fix this issue while also adding regression tests. Once merged this PR will be closed. I will release it later today, hopefully addressing these long standing issues.

PR #4247, opened 2024-07-06
Make galaxy[version-incorrect] rule opt-in

cavcrosby:
Completes #2103.

The changes I made primarily move what was already done in the linked issue over to a new rule to make it opt-in.

### DIFF
--- a/src/ansiblelint/rules/galaxy.py
+++ b/src/ansiblelint/rules/galaxy.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import sys
-from functools import total_ordering
 from typing import TYPE_CHECKING, Any
 
 from ansiblelint.constants import FILENAME_KEY, LINE_NUMBER_KEY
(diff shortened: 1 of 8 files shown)


src/ansiblelint/rules/galaxy.py
```diff
@@ -15,18 +14,17 @@
...
 class GalaxyRule(AnsibleLintRule):
-    """Rule for checking collection version is greater than 1.0.0 and checking for changelog."""
+    """Rule for checking collections."""
 
     id = "galaxy"
-    description = "Confirm via galaxy.yml file if collection version is greater than or equal to 1.0.0 and check for changelog."
+    description = "Confirm that collection's units are valid."
     severity = "MEDIUM"
     tags = ["metadata"]
     version_added = "v6.11.0 (last update)"
```
cavcrosby, line 23: Should `version_added` be updated for `GalaxyRule`? I wasn't entirely sure if this attribute just meant when the rule was first introduced. Even then, for `GalaxyVersionIncorrectRule`, I just used a version I'd imagined that the rule will be released under.

alisonlhart: I think `version_added` typically reflects the original introduced version - @Qalthos or @ssbarnea can you confirm?

alisonlhart (approved):

PR #4256, opened 2024-07-11
Add write_exclude_list config option

frq-asgard-josi:
As proposed here ([https://github.com/ansible/ansible-lint/discussions/4236](https://github.com/ansible/ansible-lint/discussions/4236)), we would like to exclude specific rules from the write_list via config file, without automatically applying auto-fixes each time ansible-lint is executed.

This PR provides a possible solution for this (while being fully backwards compatible) by introducing a new config key.


frq-asgard-josi:
It might be helpful if someone with the required permissions would add the 'enhancement' label so the test have a chance to run. As far as i can see, I can't do that myself.

Qalthos (changes requested):
Tests using the exclude list, including one that covers https://github.com/ansible/ansible-lint/pull/4256/files#diff-bbe37334f229af898f52a3ec76e32c5e43c8f2fec8640764775370e6202aa360R154 will be needed.

frq-asgard-josi:
> Tests using the exclude list, including one that covers https://github.com/ansible/ansible-lint/pull/4256/files#diff-bbe37334f229af898f52a3ec76e32c5e43c8f2fec8640764775370e6202aa360R154 will be needed.

As this isn't a priority for us at the moment, it will probably take a while until I get to this but it is still on my ToDo list.

frq-asgard-josi:
I added some basic tests for the new config option. For some reason there is a license test failing for some of the older python versions but I don't really know why.

ssbarnea:
@frq-asgard-josi rebase your change, it should address the current failures. Apparently you do not allow us to update the branch ourselves. (see bottom of the right column for that permission on PR page)

frq-asgard-josi:
> @frq-asgard-josi rebase your change, it should address the current failures. Apparently you do not allow us to update the branch ourselves. (see bottom of the right column for that permission on PR page)

I couldn't find the option to give you permission for that. This might be because the repository isn't owned by me but my organization? Anyway, I rebased my changes to include recent changes to main and fixed remaining errors so this should be ready for review :)

Qalthos (approved):

Qalthos:
> > @frq-asgard-josi rebase your change, it should address the current failures. Apparently you do not allow us to update the branch ourselves. (see bottom of the right column for that permission on PR page)
> 
> I couldn't find the option to give you permission for that. This might be because the repository isn't owned by me but my organization? Anyway, I rebased my changes to include recent changes to main and fixed remaining errors so this should be ready for review :)

If you don't see a checkbox on the bottom of the right column here, then you don't have access to it I guess. Hopefully we can catch this eventually.

Other than the need to fix more conflicts, this looks good.

Qalthos (approved):

PR #4259, opened 2024-07-11
Add support for setting nodeps through cli and .ansible-lint

dmitrij-zavrazhin-1304:
As proposed in #4258, it seems that having an option to set nodeps using the command line interface and/or .ansible-lint file to be desireable, so here is a simple implementation of it.


dmitrij-zavrazhin-1304:
The tag is errorneously set to feature, should be enhancement or patch, I guess

PR #4280, opened 2024-07-31
Add support for setting nodeps through cli and .ansible-lint

dmitrij-zavrazhin-1304:
As proposed in https://github.com/ansible/ansible-lint/discussions/4258, it seems that having an option to set nodeps using the command line interface and/or .ansible-lint file to be desireable, so here is a simple implementation of it.

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -131,6 +131,11 @@ def initialize_options(arguments: list[str] | None = None) -> None | FileLock:
     # persist loaded configuration inside options module
     for k, v in new_options.__dict__.items():
         setattr(options, k, v)
+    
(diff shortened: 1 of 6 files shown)


src/ansiblelint/__main__.py
```diff
@@ -132,6 +132,11 @@ def initialize_options(arguments: list[str] | None = None) -> None | FileLock:
     for k, v in new_options.__dict__.items():
         setattr(options, k, v)
 
+    if options.nodeps is None or options.nodeps is False:
```
ssbarnea, line 135: This is not how env vars are supposed to work. When defined, they should always override the config options.

Preference order should always be: cli > env > config.

ssbarnea (changes requested):

PR #4289, opened 2024-08-10
Fix error for imported playbooks w/invalid syntax

### LINKED ISSUE #4257: Inaccurate error for imported playbooks with invalid syntax
##### Summary

When a playbook imports another playbook via `import_playbook` ansible-lint checks if that playbook exists. It does this by calling `self.app.runtime.has_playbook`

However, `has_playbook` actually does more than check whether the playbook exists. It also calls `proc = self.run(["ansible-playbook", "--syntax-check", playbook], cwd=basedir)`. This means that `has_playbook` fails if there are syntax errors in the imported playbook

This in turn means that the ansible-lint error `msg = f"Failed to find {v} playbook."` is inaccurate. The playbook we are trying to import may very well exist but with invalid syntax

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

This happens in both the latest code from GitHub and the version my team currently use

```console
ansible-lint --version
ansible-lint 24.7.1.dev0 using ansible-core:2.17.1 ansible-compat:24.7.0 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.8
You are using a pre-release version of ansible-lint.
```

```console
ansible-lint --version
ansible-lint 24.5.0 using ansible-core:2.16.7 ansible-compat:24.6.0 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.8
```

- ansible-lint installation method: pip or source

##### STEPS TO REPRODUCE

Given a playbook that imports from a collection that we don't have:

web.yml:

```yaml
---
- name: Web servers
  hosts: role_web_servers
  roles:
    - example.server.nginx
```

This fails `ansible-playbook --syntax-check web.yml`, which is expected

If we then try and use this playbook within linux.yml:

```yaml
---
- import_playbook: web.yml
```

This fails on `ansible-lint` with `ERROR    Failed to find web.yml playbook.` because of the failing syntax checks

##### Desired Behavior

`ansible-lint` should report something more useful than `Failed to find web.yml playbook.`

Perhaps it could tell us about the syntax error instead of claiming it couldn't find the imported playbook

##### Actual Behavior

It tells us `Failed to find web.yml playbook.` which makes us go searching for why it can't find the playbook

cavcrosby:
Fixes #4257.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -72,7 +72,7 @@ jobs:
     env:
       # Number of expected test passes, safety measure for accidental skip of
       # tests. Update value if you add/remove tests.
-      PYTEST_REQPASS: 888
+      PYTEST_REQPASS: 889
     steps:
       - uses: actions/checkout@v4
         with:
--- a/examples/playbooks/import-failed-syntax-check.yml
+++ b/examples/playbooks/import-failed-syntax-check.yml
@@ -1,0 +1,3 @@
+---
+- name: Import syntax-error.yml playbook
+  import_playbook: syntax-error.yml
--- a/examples/playbooks/import-nonexistent-playbook.yml
+++ b/examples/playbooks/import-nonexistent-playbook.yml
@@ -1,0 +1,3 @@
+---
+- name: Import nonexistent playbook
+  import_playbook: i-do-not-exist.yml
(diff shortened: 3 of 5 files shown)


alisonlhart:
Looks great.

alisonlhart (approved):

Issue #4255, opened 2024-07-11
Incorrect warning for "systemctl reset-failed firewalld"


gorsing:
##### Summary

I receive an incorrect warning when using ansible.builtin.command to reset the state of the ceph-iscsi service. The warning suggests using ansible.builtin.systemd instead. However, there is no clear documentation or example for resetting the failed state using the ansible.builtin.systemd module.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 24.7.0 using ansible-core:2.14.11 ansible-compat:24.6.1 ruamel-yaml:0.18.5 ruamel-yaml-clib:0.2.7
```

- ansible installation method: OS package
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

1.     Create a playbook that uses the **ansible.builtin.command** module to reset the failed state of example service **firewalld** on fedora 38 OS or up .
2.     Run the playbook and observe the warning message.

Example playbook:

```yaml
- name: Reset if any errors happened
  ansible.builtin.command:
    cmd: systemctl reset-failed ceph-iscsi
```

##### Desired Behavior

The playbook should run without warnings, or there should be clear documentation on how to reset the failed state using the ansible.builtin.systemd module.

##### Actual Behavior

I receive an error message indicating that I should use the **ansible.builtin.systemd** module instead of the **ansible.builtin.command** module. However, the **ansible.builtin.systemd** module does not support the reset-failed state directly

```yaml
- name: Reset if any errors happened
  ansible.builtin.systemd:
    name: firewalld
    state: reset-failed
```

```paste below
args[module]: value of state must be one of: reloaded, restarted, started, stopped, got: reset-failed (warning)
cloudcw/my_plr.yml:315 Task/Handler: Reset if any errors happened

```
Please provide a valid example in the future, or update the documentation on how to use the **ansible.builtin.systemd** module to reset the failed state of a service.

tanwigeetika1618:
Hi @gorsing, A fix has been implemented that addresses the incorrect warning when using the ansible.builtin.command module to set a service’s state to reset-failed. This means the issue you encountered should no longer occur with the steps you provided. Therefore, I will proceed to close this issue based on the current findings.

PR #4301, opened 2024-08-27
fix: Add reset-failed state in executable_options

tanwigeetika1618:
### Issue:
- Fixes [ansible-lint Issue #4255](https://github.com/ansible/ansible-lint/issues/4255)

### Description:
This PR introduces a fix that allows the `reset-failed` state to be correctly handled when using the `ansible.builtin.command` module. Previously, the rule validation incorrectly flagged tasks using `systemctl reset-failed` through the `ansible.builtin.command` module, suggesting the use of the `systemd` module instead, which does not natively support the `reset-failed` state.

### Changes:
- Added support for the `reset-failed` state in the executable options for the `ansible.builtin.command` module.

### DIFF
--- a/examples/playbooks/rule-command-instead-of-module-pass.yml
+++ b/examples/playbooks/rule-command-instead-of-module-pass.yml
@@ -34,6 +34,10 @@
       ansible.builtin.command: systemctl kill --signal=SIGUSR1 sshd
       changed_when: false
 
+    - name: Reset a service with systemctl reset-failed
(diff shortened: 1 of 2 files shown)


Qalthos (dismissed):
If you can add a case to `examples/playbooks/rule-command-instead-of-module-pass.yml` defending this change, this should be good to merge.

alisonlhart:
+1 on Kate's comment, looks good besides needing the test case!

audgirka (approved):

PR #4307, opened 2024-08-30
Fix missing MatchErrors due to hash collisions

### LINKED ISSUE #4297: Missing Rule breaks due to hash collisions for dataclass MatchError
##### Summary

Rule breaks with the same exact message and line number, but different filename are missing due to a hash collision.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

ansible-lint version 24.7.0

##### STEPS TO REPRODUCE

The [MatchError](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/errors.py#L36) data class has the `unsafe_hash=True`, which generates automatically a hash function based on the members of the class, however the field `filename` is non-existent (not defined) in the `dataclass`, it is added at a later stage (see: [var_naming.py#L204](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/rules/var_naming.py#L204)) and then all the errors are added to a set (see: [runner.py#L682](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/runner.py#L682)). Due to the `filename` member being non-existent at the class definition, the hash function does not take into account the `filename` and cause collisions if you have the same exact rule break in another file at the same line number. 

1. Create 2 different files with var naming issues on the same lines number (but in the 2 sperate file)
2. Run ansible-lint
3. There should be only 1 issue raised due to the bug.

##### Desired Behavior

The same rule break happening in a different file on the same line number should be _recorded_ properly as different rule break. 

##### Actual Behavior
Simplified code example of the reproducible issue we are experiencing:
```
from dataclasses import dataclass

@dataclass(unsafe_hash=True)
class MatchError(ValueError):
    message: str
    tag: str
    lineno: int = 1

    @property
    def _hash_key(self) -> any:
        # line attr is knowingly excluded, as dict is not hashable
        return (
            self.filename,
            self.lineno,
            self.message,
            self.tag
        )

    def __eq__(self, other):
        return self.__hash__() == other.__hash__()
    


error1 = CustomError(message="An error occurred", tag="ERROR_TAG", lineno=42)
error1.filename = "test.py"
error2 = CustomError(message="An error occurred", tag="ERROR_TAG", lineno=42)
error2.filename = "test2.py"
error3 = CustomError(message="An error occurred", tag="ERROR_TAG", lineno=43)
error3.filename = "test2.py"

print(error1.filename)
print(error2.filename)
print(error3.filename)
print(hash(error1))
print(hash(error2))
print(hash(error3))
```

Potential fix:

- Add the filename member to the class definition of [MatchError](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/errors.py#L36) (as "None" by default to preserve the same behaviors, but fix the hash function)

cavcrosby:
Fixes #4297.

This fix follows what was done here, https://github.com/ansible/ansible-lint/pull/4202, for transitioning to passing a `lintable` argument for `MatchError` object construction instead of a `filename` argument. I also noticed a few places in `src/ansiblelint/utils.py` where this problem could be reproduced. Hence, I figured now would be a good time to address those spots as well.

### DIFF
--- a/examples/playbooks/vars/rule_var_naming_fails_files/bar.yml
+++ b/examples/playbooks/vars/rule_var_naming_fails_files/bar.yml
@@ -1,0 +1,3 @@
+---
+CamelCaseIsBad: bar
+ALL_CAPS_ARE_BAD_TOO: bar
--- a/examples/playbooks/vars/rule_var_naming_fails_files/foo.yml
+++ b/examples/playbooks/vars/rule_var_naming_fails_files/foo.yml
@@ -1,0 +1,3 @@
+---
+CamelCaseIsBad: foo
+ALL_CAPS_ARE_BAD_TOO: foo
--- a/src/ansiblelint/rules/var_naming.py
+++ b/src/ansiblelint/rules/var_naming.py
@@ -117,6 +117,7 @@ class VariableNamingRule(AnsibleLintRule):
         ident: str,
         *,
         prefix: Prefix | None = None,
+        file: Lintable,
     ) -> MatchError | None:
         """Return a MatchError if the variable name is not valid, otherwise None."""
         if not isinstance(ident, str):  # pragma: no cover
@@ -124,6 +125,7 @@ class VariableNamingRule(AnsibleLintRule):
                 tag="var-naming[non-string]",
                 message="Variables names must be strings.",
                 rule=self,
+                lintable=file,
             )
 
         if ident in ANNOTATION_KEYS or ident in self.allowed_special_names:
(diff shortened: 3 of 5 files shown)


marcandre-larochelle-bell:
@cavcrosby 

I think the function `_hash_key()` also require to be updated to take into account the lintable instead of the filename (in MatchError)? see:
https://github.com/ansible/ansible-lint/blob/7a780060792832407914ab596428c290e8937060/src/ansiblelint/errors.py#L144-L156

ssbarnea (approved):

cavcrosby:
@marcandre-larochelle-bell I apologize for not getting around to your comment until now. I think it's fine to continue using the `filename` attribute in the `_hash_key` method at this time.

marcandre-larochelle-bell:
@cavcrosby No worries, I was mostly wondering, t hanks for the reply!

Issue #4160, opened 2024-05-14
fqcn[deep] - checks for directory pattern in tests and reports false positive


KB-perByte:
##### Summary

It checks for directory patterns in tests and reports false positive warning about a deep plugin directory under the tests/ dir
The test directory should not be considered for this rule. 

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
```

- ansible installation method: one of source, pip, OS package
- ansible-lint installation method: one of source, pip, OS package

##### STEPS TO REPRODUCE

N/A

```console (paste below)

```

##### Desired Behavior

Possible security bugs should be reported via email to `security@ansible.com`

##### Actual Behavior

![image](https://github.com/ansible/ansible-lint/assets/19584197/39245d29-8e35-4514-ad26-cef76e7f305a)

Please give some details of what is happening. Include a [minimum complete
verifiable example] with:

- minimized playbook to reproduce the error
- the output of running ansible-lint including the command line used
- if you're getting a stack trace, also the output of
  `ansible-playbook --syntax-check playbook`

```paste below

```

[minimum complete verifiable example]: http://stackoverflow.com/help/mcve

audgirka:
Related: 
https://github.com/ansible/ansible-lint/issues/2584
https://github.com/ansible/ansible-lint/pull/3502

felixfontein:
Another false positive (https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=114105&view=logs&j=2671e6a6-f41a-533c-2720-8ffdcf8ab96f&t=5b604a49-baaa-558f-6ab2-4a2ff646af4f) are files in `tests/unit/plugins/action/fixtures/`.

I wonder why this rule checks **unit tests** at all, since unit tests are known to come with fixtures that are in subdirectories.

PR #4320, opened 2024-09-10
Fix for ignoring test / directories by fqcn[deep] rule

tanwigeetika1618:
### Description:
This PR addresses issue #4160 by modifying the logic in `fqcn.py` to ensure that directories related to testing (i.e., those starting with "test") are skipped during processing. Previously, test directories were being evaluated, leading to unnecessary  false positive warnings. (Skip test folders in fqcn.py)

### Changes:
- Updated the condition in `fqcn.py` to check if a directory is a test folder.
- The folder checking logic has been enhanced to search for the word "test" anywhere in the directory path.
  - Directories that include "test" (such as `test`, `tests`, `testxyz`, or `abctest`) are skipped.
  - This ensures that test-related directories are excluded from processing while maintaining standard plugin structure.
 - Added a new test case `test_fqcn_deep_test_dir_pass` to verify that the rule does not flag files located in directories named `test` or similar.

### Testing:
The new logic has been tested with various directory structures to ensure that any directory containing the word "test" is correctly skipped. The following cases have been verified:
1. `/tests/unit/plugins/action/network/pr.py`
2. `/abc/unit/plugins/action/tests/pr.py`
3. `/abc/test/plugins/action/network/pr.py`
4. `/abctests/unit/plugins/action/network/pr.py`
5. `/testxyz/unit/plugins/action/network/pr.py`

In all these cases, the test directories (or directories containing "test") were successfully ignored during processing.

Screenshot from 2024-09-10 16-23-13


alisonlhart (commented):
Looks good! There should probably be a test added case to check this, following the pattern of the other test cases in the `fcqn.py` rule. One of the test cases you used during development would be good to add in.

Issue #4219, opened 2024-06-14
ansible-lint on collection tries to install the collection itself


falon:
##### Summary
If I develop a collection and I lint it, the linter try to install the same collection by Galaxy. After that it warn that multiple collection versions were found.

I don't understand. If I'm developing a collection, I don't expect that the linter modify this collection by installing it from Galaxy. 

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
# ansible-lint --version
ansible-lint 24.6.0 using ansible-core:2.17.0 ansible-compat:24.6.1 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.8
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Init a new fresh collection with ansible-galaxy.
When I run ansible-lint from the base path of the collection I see

```console
INFO     Running from /root/.ansible/collections/ansible_collections/mynamespace/mycollection : ansible-galaxy collection install -vvv --force /root/.ansible/collections/ansible_collections/mynamespace/mycollection
WARNING  Multiple versions of 'mynamespace.mycollection' were found installed, only the first one will be used, 1.0.0 (/root/.ansible/collections/ansible_collections).
```
Why does ansible-lint want to install the collection which I'm linting?

My **galaxy.cfg** is
```yaml
namespace: mynamespace
name: mycollection
version: 1.0.0
readme: README.md
authors:
  - Marco  <marco@example.com>
description: A general suite of customization.
license:
  - GPL-3.0-or-later
license_file: ""
tags:
  - linux
dependencies:
  community.general: ">=8.1.0"
  community.docker: ">=3.4.11"
repository: https://gitlab.example.com/mynamespace/mycollection
homepage: https://gitlab.example.com/mynamespace/mycollection
issues: https://gitlab.example.com/mynamespace/mycollection/-/issues
```
##### Desired Behavior

It's good that ansible-lint installs collection dependent by the collection. Ansible-lint shoud not try to install from ansible-galaxy the same collection that I'm linting.

##### Actual Behavior

```console
[root@test mycollection]# ansible-lint
WARNING  Multiple versions of 'mynamespace.mycollection' were found installed, only the first one will be used, 1.0.0 (/root/.ansible/collections/ansible_collections).
WARNING  Multiple versions of 'mynamespace.mycollection' were found installed, only the first one will be used, 1.0.0 (/root/.ansible/collections/ansible_collections).

[...]

PR #4335, opened 2024-09-18
Avoid running application initialization twice

ssbarnea:
Related: https://github.com/ansible/ansible-lint/issues/4219


tanwigeetika1618 (approved):
The changes in this PR look good to me. The caching mechanism in `get_app` seems well-implemented, and updating `ruff-pre-commit` to the latest version is appropriate. The test adjustments also ensure that caching doesn’t interfere with the test runs.

Issue #4317, opened 2024-09-07
Avoid spamming local syslog while linting (module loading)


toralf:
Running 
```
$> ansible-lint
```
for this repo: https://github.com/toralf/tor-relays spams the local syslog (Gentoo hardened with openrc) with hundreds of lines like this:
```
Sep  7 19:51:25 p14s ansible-basic.py[2536]: Invoked with name=['prometheus-node-exporter'] package=['prometheus-node-exporter'] state=present update_cache_retries=5 update_cache_retry_max_delay=12 cache_valid_time=0 purge=False force=False upgrade=no dpkg_options=force-confdef,force-confold autoremove=False autoclean=False fail_on_autoremove=False only_upgrade=False force_apt_get=False clean=False allow_unauthenticated=False allow_downgrade=False allow_change_held_packages=False lock_timeout=60 update_cache=None deb=None default_release=None install_recommends=None policy_rc_d=None
Sep  7 19:51:25 p14s ansible-basic.py[2536]: Invoked with name=prometheus-node-exporter daemon_reload=False daemon_reexec=False scope=system no_block=False state=None enabled=None force=None masked=None
Sep  7 19:51:25 p14s ansible-basic.py[2536]: Invoked with path=/var/lib/node_exporter/ state=directory mode=0755 recurse=False force=False follow=True modification_time_format=%Y%m%d%H%M.%S access_time_format=%Y%m%d%H%M.%S unsafe_writes=False _original_basename=None _diff_peek=None src=None modification_time=None access_time=None owner=None group=None seuser=None serole=None selevel=None setype=None attributes=None
Sep  7 19:51:25 p14s ansible-basic.py[2536]: Invoked with create=False line=ARGS=' --web.listen-address=localhost:9100 --collector.tcpstat --collector.textfile.directory=/var/lib/node_exporter/ --collector.filesystem.mount-points-exclude="^/(dev|proc|run/credentials/.+|sys|var/lib/docker/.+|var/lib/containers/storage/.+)($|/)" ' path=/etc/default/prometheus-node-exporter regex=^ARGS= regexp=^ARGS= state=present backrefs=False backup=False firstmatch=False unsafe_writes=False search_string=None insertafter=None insertbefore=None validate=None mode=None owner=None group=None seuser=None serole=None selevel=None setype=None attributes=None
[...]

PR #4353, opened 2024-10-01
Fix to avoid spamming  syslog with invoked messages

tanwigeetika1618:
## Fix issue #4317

# Description
This PR resolves the issue where `ansible-basic.py` was unnecessarily logging invoked messages to the local syslog, leading to excessive and irrelevant logging during playbook execution.

## Changes
- Modified `CustomAnsibleModule` in `args.py` to modify the initialization process.
-Preventing the unnecessary execution of `AnsibleModule`.
- This stops the spamming of syslog with messages such as "ansible-basic.py Invoked with...".

## Impact
- Significantly reduces irrelevant syslog entries.
- Verified by running `ansible-lint` and checking logs with `journalctl | grep ansible`.

### DIFF
--- a/src/ansiblelint/rules/args.py
+++ b/src/ansiblelint/rules/args.py
@@ -73,8 +73,9 @@ class ValidationPassedError(Exception):
 class CustomAnsibleModule(basic.AnsibleModule):  # type: ignore[misc]
     """Mock AnsibleModule class."""
 
-    def __init__(self, *args: str, **kwargs: str) -> None:
+    def __init__(self, *args: str, **kwargs: Any) -> None:
         """Initialize AnsibleModule mock."""
+        kwargs["no_log"] = True
         super().__init__(*args, **kwargs)
         raise ValidationPassedError
 


src/ansiblelint/rules/args.py
```diff
@@ -73,8 +73,9 @@ class ValidationPassedError(Exception):
 class CustomAnsibleModule(basic.AnsibleModule):  # type: ignore[misc]
     """Mock AnsibleModule class."""
 
-    def __init__(self, *args: str, **kwargs: str) -> None:
+    def __init__(self, *args: str, **kwargs: Any) -> None:
         """Initialize AnsibleModule mock."""
+        kwargs["no_log"] = True
         super().__init__(*args, **kwargs)
```
Qalthos, line 79: I don't really like the idea of _introducing_ `Any` anywhere if we can help it. We could make the type `str | bool` instead, assuming that previously we are only ever calling it with strings.

```suggestion
    def __init__(self, *args: str, **kwargs: str | bool) -> None:
        """Initialize AnsibleModule mock."""
        kwargs["no_log"] = True
        super().__init__(*args, **kwargs)
```

We could instead keep the hint but instead `kwargs.pop("no_log")` and then pass `no_log=True` directly to `super().__init__()`, but I'm not convinced `str` actually _is_ the correct type here

Qalthos (dismissed):

src/ansiblelint/rules/args.py
ssbarnea: kwargs and type is complex and bit unexpected, for the time being "Any" is the correct solution. As you can see mypy does not complain about it.

ssbarnea (approved):

PR #4379, opened 2024-10-27
Fix WARN on empty/template meta/main.yaml

GElkayam:
WARNING  Ignored exception from RoleNames.matchyaml while processing roles/xxx/meta/main.yaml (meta): 'NoneType' object has no attribute 'get'


GElkayam:
Related to https://github.com/ansible/ansible-lint/issues/3991

audgirka (dismissed):
@GElkayam Thanks for the PR but it needs a test

ssbarnea (approved):

GElkayam:
Thanks @ssbarnea , I was struggling with getting tests to work to add them, although it is a straight forward fix.

PR #4381, opened 2024-10-30
pre-commit hook should only run on pre-commit stage

rdbisme:


ssbarnea:
Can you please provide more information about this? When a hook runs is more to the user, some might decide to run them on pre-commit, some might decide to run them on push.

rdbisme:
Just that by default this gets called n times, where n is the stages. I guess someone wants to run it mostly once. So setting a default stage would prevent double call (as in general happens with other hooks). 

People can always override this.

ssbarnea:
I am not aware of other linter using custom stages. Based on https://pre-commit.com/#confining-hooks-to-run-at-certain-stages I think is the best not to mention these and allow the user to decide which one he wants to use. Default values seems reasonable.

Issue #4362, opened 2024-10-09
"Failed to find playbook" error when importing a playbook from a collection


acidroper:
##### Summary

When using `import_playbook` with a playbook from a collection, `ansible-lint` reports the following error:
```
ERROR    Failed to find my_namespace.my_collection.playbook playbook.
```

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 24.9.2 using ansible-core:2.16.11 ansible-compat:24.9.1 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.8
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Write a playbook that imports a playbook from an installed collection:

```yaml
---
- name: Import a playbook
  ansible.builtin.import_playbook: my_namespace.my_collection.playbook
```

Run `ansible-lint import_playbook.yml`

##### Desired Behavior

No errors are expected.

##### Actual Behavior

```console (paste below)
$ ansible-lint import_playbook.yml
ERROR    Failed to find my_namespace.my_collection.playbook playbook.

Passed: 0 failure(s), 0 warning(s) on 1 files. Last profile that met the validation criteria was 'production'.

$ ls -l ~/.ansible/collections/ansible_collections/my_namespace/my_collection/playbooks/playbook.yml
-rw-r--r-- 1 user user 187 окт  9 10:42 /home/user/.ansible/collections/ansible_collections/my_namespace/my_collection/playbooks/playbook.yml
```

acidroper:
Thanks for the fix. However, there is still an issue: if a playbook is located in a nested directory, ansible-lint still reports the error.

cavcrosby:
@acidroper Could you provide me an example reproducing the issue?

acidroper:
@cavcrosby  Sure.

```yaml
---
- name: Import a nested playbook
  ansible.builtin.import_playbook: my_namespace.my_collection.test.playbook
```

```console
$ ansible-lint import_playbook.yml
ERROR    Failed to find my_namespace.my_collection.test.playbook playbook.

Passed: 0 failure(s), 0 warning(s) on 1 files. Last profile that met the validation criteria was 'production'.
```

```console
$ ls -l ~/.ansible/collections/ansible_collections/my_namespace/my_collection/playbooks/test/playbook.yml                                                                                                                 
[...]

PR #4412, opened 2024-11-16
Enable FQCNs for import_playbook to have subdirs

cavcrosby:
Related to #4362.


cavcrosby:
Regarding the code analysis failing, `ansible` uses a similar [regex](https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/collection_loader/_collection_finder.py#L916) that does not place a limit on the number of subdirectories a user can specify. Currently, I think that's the only applicable workaround to the `catastrophic backtracking` vulnerability.

ssbarnea (changes requested):
```
RE_IS_FQCN_OR_NAME = re.compile(r"^\w+((\.\w+){2,})*$")
```
To quote SonarQube:  Make sure the regex used here, which is vulnerable to exponential runtime due to backtracking, cannot lead to denial of service.

PR #4416, opened 2024-11-18
search for config with the yml/yaml extension

cornfeedhobo:
1) add support for `.ansible-lint.yml`
2) add support for `.ansible-lint.yaml`
3) minor fix to `DEFAULT_KINDS` which wasn't flagging `.config/ansible-lint.yaml` as a `ansible-lint-config`


shatakshiiii (approved):
looks good!

PR #4396, opened 2024-11-10
Refactor code to avoid use of rich library

ssbarnea:
Related: https://issues.redhat.com/browse/AAP-36125

PR #4432, opened 2024-12-02
Remove formatting options for listing rules

ssbarnea:
This change removes the --format option from the affecting the
output of --list-rules. That change is not marked as major because
because it only affect output that is not used in automations.

Related: #4396 AAP-36125


abhikdps (approved):
LGTM

Issue #4449, opened 2024-12-12
Unable to parse inventory microsoft.ad.ldap.yml


mueller-ma:
##### Summary

Starting with [24.12.1](https://github.com/ansible/ansible-lint/releases/tag/v24.12.1) ansible-lint fails when Active Directory inventory source is used (https://docs.ansible.com/ansible/latest/collections/microsoft/ad/ldap_inventory.html).

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
$ ansible-lint --version
ansible-lint 24.12.1 using ansible-core:2.18.1 ansible-compat:24.10.0 ruamel-yaml:0.18.6 ruamel-yaml-clib:0.2.8
You are using a pre-release version of ansible-lint.
```

BTW: Why is this a pre-release version? It's marked as "latest" here: https://github.com/ansible/ansible-lint/releases

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

I created a new directory containing two files only:

````console
$ cat ansible.cfg
[defaults]
inventory=inventories/microsoft.ad.ldap.yml

[inventory]
any_unparsed_is_failed=true
$ cat inventories/microsoft.ad.ldap.yml
---
plugin: microsoft.ad.ldap

server: dc01.domain.com
port: 389
````

##### Desired Behavior

It should be able to parse this inventory.

##### Actual Behavior

````console
$ ansible-lint .
Traceback (most recent call last):
  File "/home/<user>/.local/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/<user>/.local/lib/python3.11/site-packages/ansiblelint/__main__.py", line 400, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/home/<user>/.local/lib/python3.11/site-packages/ansiblelint/__main__.py", line 352, in main
    result = get_matches(rules, options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/<user>/.local/lib/python3.11/site-packages/ansiblelint/runner.py", line 755, in get_matches
    matches.extend(runner.run())
                   ^^^^^^^^^^^^
  File "/home/<user>/.local/lib/python3.11/site-packages/ansiblelint/runner.py", line 165, in run
    matches = self._run()
[...]

Issue #4448, opened 2024-12-12
24.12.x fails with "Unable to parse /the/path/localhost as an inventory source"


sverrehu:
##### Summary

Previously, `ansible-lint` would invoke `ansible-playbook` with the option `-i localhost,` but after #4393 the [trailing comma is gone](https://github.com/ansible/ansible-lint/pull/4393/files#diff-bcd284f890e371be042a5d19c8dbb4a27d497f2c5054746488173e5a17a3d117L349). This makes `ansible-playbook` look for a file named `localhost` instead of treating `localhost` as a hostname.

##### Issue Type

- Bug Report

PR #4450, opened 2024-12-12
Revert "Accommodate specified inventory files (#4393)"

ssbarnea:
Due to the numerous problems introduced by that #4393 we will revert that feature and yank versions 24.2.0 and 24.2.1. If we are about to add these feature back it will be with a very comprehensive testing.

This reverts commit 6fbb12e9d8e8464fe8dbc36f1829fa6a8d229c08.
This reverts commit 4d295e82dd814c6f5793f898ed881358a5c42361.

Fixes: #4448
Fixes: #4449


cidrblock (approved):
good idea

cavcrosby:
Yea there were definitely oversights on my part. Next time I'll be more careful.

PR #4480, opened 2025-01-18
Update schema etag

ssbarnea:


ssbarnea:
Bypassed CI due to different breakage.

Issue #4492, opened 2025-01-24
crasher: AttributeError: 'CommentedSeq' object has no attribute 'get' - v2


sdarwin:
An issue had been closed due to not including reproducing data.

 "crasher: AttributeError: 'CommentedSeq' object has no attribute 'get'"

https://github.com/ansible/ansible-lint/issues/3612  

"It is a crash that still exists and is easily reproduced."

"And it looks like at least 7 people have reported running into it. I agree, a crasher with a clear and simple reproduction is worth re-opening."

The problem is caused by a superfluous "---".

This happened to me as well. After I removed the --- in the middle, the error was gone.

`ansible-lint` should detect and report a linting issue instead of crashing:  

```
  File "/root/venv/lib/python3.12/site-packages/ansiblelint/skip_utils.py", line 315, in <genexpr>
    return any(task.get(key) for key in NESTED_TASK_KEYS)
               ^^^^^^^^
AttributeError: 'CommentedSeq' object has no attribute 'get'
```

---

Edit:

Not sure how Ansible itself (not ansible-lint) treats this situation.  Does Ansible support multiple documents in one file? From https://gettaurus.org/docs/YAMLTutorial
 
YAML Multi Documents: YAML format allows multiple documents to be embedded in a single file. They only have to be separated with a line containing triple-dash separator ---.

```
document: this is document 1
---
document: this is document 2
```

sdarwin:
>  incomplete 

The crash happened with an extra "---" in the middle of a tasks file.  I would recommend during testing you take any ansible role, and put "---" between tasks in a tasks file.    However, for a brief example, here is a playbook with that configuration.  The contents are not important.  Just the dashes.


playbook.yml:

```
---

- name: Install software (not Windows)
  hosts: newservers:!windows
  remote_user: ubuntu
  become: true
  roles:
    - { role: 'example.acme' }

---

- name: Install software (Windows)
  hosts: newservers:&windows
  roles:
    - { role: 'example.acme' }
```

`ansible-lint playbook.yml`

PR #4506, opened 2025-01-30
Improve output with broken multiline playbooks

ssbarnea:
Fixes: #4492

### DIFF
--- a/src/ansiblelint/runner.py
+++ b/src/ansiblelint/runner.py
@@ -270,10 +270,12 @@ class Runner:
         # do our processing only when ansible syntax check passed in order
         # to avoid causing runtime exceptions. Our processing is not as
         # resilient to be able process garbage.
-        matches.extend(self._emit_matches(files))
+        matches.extend(
+            self._emit_matches([file for file in files if not file.failed()])
+        )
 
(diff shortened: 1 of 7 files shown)


src/ansiblelint/runner.py
```diff
@@ -499,9 +504,12 @@ def find_children(self, lintable: Lintable) -> list[Lintable]:
...
             except AnsibleError as exc:
-                msg = f"Loading {lintable.filename} caused an {type(exc).__name__} exception: {exc}, file was ignored."
-                _logger.exception(msg)
-                return []
+                raise MatchError(
+                    lintable=lintable, rule=self.rules["load-failure"]
+                ) from exc
+                # msg = f"Loading {lintable.filename} caused an {type(exc).__name__} exception: {exc}, file was ignored."
+                # _logger.exception(msg)
+                # return []
```
Qalthos, line 512: commented code?
```suggestion
```

Qalthos (changes requested):

src/ansiblelint/runner.py
Qalthos: seems okay otherwise

PR #4518, opened 2025-02-11
Add .ansible to the implicit ignore list

ssbarnea:
Related: https://github.com/ansible/ansible-dev-tools/pull/546


falon:
I tried to apply this fix changing the file `file_utils.py`, but it does not resolve #4522 for me...
I had to downgrade ansible-compat.

Issue #3935, opened 2023-12-06
Cannot ignore rule jinja[spacing]


AdrienEti:
##### Summary
The  [Jinja white-space control](https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control)  generates an jinja [spacing] rule violation. 
I'm trying to escape this rule for a variable definition, but it seems that the escape is ignored. 
I've tried to place the # noqa: jinja[spacing] mutation in several places, but it is always ignored.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
$ ansible-lint --version
ansible-lint 6.22.1 using ansible-core:2.16.0 ansible-compat:4.1.10 ruamel-yaml:0.18.5 ruamel-yaml-clib:0.2.7
```

- ansible installation method: **pip**
- ansible-lint installation method: **pip**

##### STEPS TO REPRODUCE

Defined a variable as follow and try to ignore the jinja[spacing] rule:
```yaml
foo_postgresql_grafana_password_encoded: >- # noqa: jinja[spacing]
  {% if '!' in foo_postgresql_grafana_password or '#' in foo_postgresql_grafana_password %}
  {{- ('\"\"\"' + foo_postgresql_grafana_password + '\"\"\"') | b64encode -}}
  {% else %}
  {{- foo_postgresql_grafana_password | b64encode -}}
  {% endif %}
``` 
or

```yaml
# noqa: jinja[spacing]
foo_postgresql_grafana_password_encoded: >-
  {% if '!' in foo_postgresql_grafana_password or '#' in foo_postgresql_grafana_password %}
  {{- ('\"\"\"' + foo_postgresql_grafana_password + '\"\"\"') | b64encode -}}
  {% else %}
  {{- foo_postgresql_grafana_password | b64encode -}}
  {% endif %}
``` 

##### Desired Behavior

The jinja[spacing] rule is ignored 

##### Actual Behavior

The jinja[spacing] rule is not ignored

s3lph:
I'm affected by the same issue in exactly the same situation. I've found a workaround: Use `|-` instead of `>-`, and make all jinja tags remove whitespace (i.e. use `{%- -%}`  rather than `{% %}`. That way, whitespace is still removed, but by Jinja rather than by the YAML parser.

So, in your example:
```yaml
foo_postgresql_grafana_password_encoded: |-
[...]

PR #4567, opened 2025-03-31
Add noqa skips in otherwise empty lines to the next non-empty line

StopMotionCuber:
This addresses #3935

So what it does:
 - Some magic in `traverse_yaml` to flatten lists within there (these are required for a ruamel CommentMap, please don't ask any further, afaik these are entirely undocumented)
 - After collecting all `#noqa` and inserting them into the `lintable`, postprocessing is done on the `lintable`.
   - For each line where a skip is found, it is matched against a regex whether `# noqa` is the start of the line, minus some whitespace/indentation.
   - If there is only the `# noqa` on that line, the skip is added to the next non-empty line.

This is something required for me in order to suppress errors on yaml multiline strings, as reported by the issue referenced.

In general I think the solution is quite hacky, so this can also be seen as a proof of concept. The whole flattening thing and how ruamel works with comments is quite weird. As in the context of `_get_rule_skips_from_yaml` only a map of `{lineno: rule_to_skip}` is what we're interested in, it might be easier to just go over the content of a file, check each line for `_noqa_comment_re` match. The whole tree structure of `CommentMap` is annoying to traverse, a simple iterator over all comments would do it (don't know if that's in `pyyaml`?) and would be more accurate for comments not attached to objects in a line.

### DIFF
--- a/examples/playbooks/vars/noqa_multiline.yml
+++ b/examples/playbooks/vars/noqa_multiline.yml
@@ -1,0 +1,28 @@
+---
+# noqa: jinja[spacing]
+foo_postgresql_grafana_password_encoded: >-
+  {% if '!' in foo_postgresql_grafana_password or '#' in foo_postgresql_grafana_password %}
+  {{- ('\"\"\"' + foo_postgresql_grafana_password + '\"\"\"') | b64encode -}}
+  {% else %}
+  {{- foo_postgresql_grafana_password | b64encode -}}
+  {% endif %}
+
+
+# noqa: jinja[spacing]
+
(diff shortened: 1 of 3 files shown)


ssbarnea:
This will have to wait until we finish the changes started in #4564 - adding support for ansible 2.19 with data tagging. There were a really big number of changes related to this and they also affect this area.

StopMotionCuber:
@ssbarnea I guess the data tagging changes are integrated now, any chance you could take a look again?

ssbarnea (approved):

Issue #4578, opened 2025-04-10
is_playbook should support fqcn for the import_playbook action


OscarBell:
##### Summary

The `utils.is_playbook` function detects a yml file as a playbook when the non fqcn version of `import_playbook` is used in the yaml file. When `ansible.builtin.import_playbook` is used, ansible-lint considers this file not as a playbook.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 0.1.dev2973 using ansible-core:2.18.4 ansible-compat:25.1.5 ruamel-yaml:0.18.10 ruamel-yaml-clib:0.2.12
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Create a Python file called `test.py` and execute this. Without a fix for the `is_playbook` function the 2nd playbook file is not considered a playbook although it is a playbook file. The 2 files that are verified are taken from the examples folder.

```console (paste below)
from ansiblelint.utils import is_playbook

non_fqcn_playbook_file="examples/playbooks/import-failed-syntax-check.yml"
print(f"Verify if {non_fqcn_playbook_file}, which uses import_playbook (non fqcn) is a playbook: {is_playbook(filename=non_fqcn_playbook_file)}")

fqcn_playbook_file="examples/playbooks/import_playbook_fqcn.yml"
print(f"Verify if {fqcn_playbook_file}, which uses ansible.builtin.import_playbook (fqcn) is a playbook: {is_playbook(filename=fqcn_playbook_file)}")

```

```console
python3 test.py
Verify if examples/playbooks/import-failed-syntax-check.yml, which uses import_playbook (non fqcn) is a playbook: True
Verify if examples/playbooks/import_playbook_fqcn.yml, which uses ansible.builtin.import_playbook (fqcn) is a playbook: False
```

##### Desired Behavior

Executing the test.py script with a modified `is_playbook` function should return True for both files

```console
python3 test.py
Verify if examples/playbooks/import-failed-syntax-check.yml, which uses import_playbook (non fqcn) is a playbook: True
Verify if examples/playbooks/import_playbook_fqcn.yml, which uses ansible.builtin.import_playbook (fqcn) is a playbook: True
```

##### Actual Behavior

As can be seen above, the playbook file that uses fqcn `import_playbook` is not considered a playbook while it should be and can be used as such with Ansible.

OscarBell:
[...]

PR #4580, opened 2025-04-10
Make utils.is_playbook support fqcn import_playbook

OscarBell:
Updated `utils.is_playbook` function to support the fqcn version of `import_playbook` to detect a valid playbook file.


OscarBell:
This PR will solve issue: #4578

OscarBell:
Alternatively if the `playbooks_keys` needs to be unadjusted we could change below line 
`if not isinstance(item, Mapping) or not playbooks_keys.intersection(
    item.keys()
):`
into
`if not isinstance(item, Mapping) or not playbooks_keys.intersection(
                    {item.split('.')[-1] for item in item.keys()}
                ):`
With above solution all keys found in the play will be changed into the non-fqcn variant and can be compared against the `playbook_keys`

PR #4585, opened 2025-04-15
chore: Updated some of the docs and messages around setting pipefail.

dylanmorroll:
Setting pipefail necessitates other changes away from the default functioning of ansible, this is made more explicit to the user.

### DIFF
--- a/src/ansiblelint/rules/risky_shell_pipe.md
+++ b/src/ansiblelint/rules/risky_shell_pipe.md
@@ -7,6 +7,9 @@ The return status of a pipeline is the exit status of the command. The
 `pipefail` option ensures that tasks fail as expected if the first command
 fails.
 
+This does not work in the standard (POSIX) shell. You will need to set the executable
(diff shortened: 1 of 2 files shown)


dylanmorroll:
I can't seem to add a label, does someone else need to do that?

audgirka (approved):

dylanmorroll:
Tweaked it to pass pipelines :)

ssbarnea (changes requested):
Maybe I am missing something but what is the 'standard (POSIX) shell' from your point of view? AFAIK, there is no such software product.

There are lots of shells that might follow POSIX standard (more or less), including bash,... so I find the phrasing as adding more confusion instead of clarification.

dylanmorroll:
Ah yes I meant the standard shell used by Ansible (`ansible.builtin.shell`), which I presumed was a POSIX compliant shell without many additional features, in which pipefail does not work.

Issue #4592, opened 2025-04-25
Unit tests fail when running against ansible 2.19


leegarrett:
##### Summary

The `test_template` unit tests fail due to setting `disable_lookups` to True, which is deprecated in ansible 2.19. I'll provide a minimal PR to fix this in a moment.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 25.2.1 using ansible-core:2.19.0b2 ansible-compat:25.1.5 ruamel-yaml:0.18.10 ruamel-yaml-clib:0.2.12
WARNING  Project directory /.ansible cannot be used for caching as it is not writable.
WARNING  Using unique temporary directory /tmp/.ansible-0aaa for caching.
```

Debian trixie (13)

- ansible installation method: OS package
- ansible-lint installation method: OS package

##### STEPS TO REPRODUCE

Run current ansible-lint unit tests against ansible 2.19

```console (paste below)
[...]
============================================================================================================== FAILURES ===============================================================================================================
________________________________________________________________________________________________________ test_template[simple] ________________________________________________________________________________________________________

template = '{{ playbook_dir }}', output = '/a/b/c'

    @pytest.mark.parametrize(
        ("template", "output"),
        (
            pytest.param("{{ playbook_dir }}", "/a/b/c", id="simple"),
            # Does not work the same with ansible 2.19 with data tagging
            # pytest.param(
            #     "{{ 'hello' | doesnotexist }}",
            #     "hello",  # newer implementation ignores unknown filters
            #     id="unknown_filter",
            # ),
            pytest.param(
                "{{ hello | to_json }}",
                "{{ hello | to_json }}",
                id="to_json_filter_on_undefined_variable",
            ),
            pytest.param(
                "{{ hello | to_nice_yaml }}",
                "{{ hello | to_nice_yaml }}",
                id="to_nice_yaml_filter_on_undefined_variable",
            ),
        ),
    )
    def test_template(template: str, output: str) -> None:
        """Verify that resolvable template vars and filters get rendered."""
>       result = utils.template(
            basedir=Path("/base/dir"),
            value=template,
            variables={"playbook_dir": "/a/b/c"},
            fail_on_error=False,
        )

test/test_utils.py:273: 
[...]

Issue #2681, opened 2022-11-13
ansible-lint runs collection code when validating Jinja2 expressions


felixfontein:
##### Summary

While working on tests for a filter plugin (https://github.com/felixfontein/ansible-acme/pull/41) I noticed that when linting the integration tests for the filter, which contain some tests that make the plugin raise an exception, ansible-lint seems to evaluate the expressions and thus run code from the collection. This effectively causes linting to fail since the Jinja2 expressions raise an error when being evaluated, as you can see in https://github.com/felixfontein/ansible-acme/actions/runs/3456077642/jobs/5768559703:

```
WARNING  Listing 4 violation(s) that are fatal
jinja[invalid]: Unexpected templating type error occurred on ({{ {} | felixfontein.acme._substitute_dns({}) }}): The DNS name is of type <class 'dict'>, and not text. The DNS name is of type <class 'dict'>, and not text
tests/integration/targets/filter__substitute_dns/tasks/main.yml:34 Task/Handler: Filter errors (1/5)

jinja[invalid]: Unexpected templating type error occurred on ({{ 'wc.example.net' | felixfontein.acme._substitute_dns([]) }}): The substitution map type <class 'list'>, and not a dictionary. The substitution map type <class 'list'>, and not a dictionary
tests/integration/targets/filter__substitute_dns/tasks/main.yml:41 Task/Handler: Filter errors (2/5)
You can skip specific rules or tags by adding them to your configuration file:

jinja[invalid]: Unexpected templating type error occurred on ({{ 'wc.example.net' | felixfontein.acme._substitute_dns({'a': 1}) }}): Key or value of dictionary entry are of type <class 'str'> resp. <class 'int'>, but both must be text. Key or value of dictionary entry are of type <class 'str'> resp. <class 'int'>, but both must be text
tests/integration/targets/filter__substitute_dns/tasks/main.yml:55 Task/Handler: Filter errors (4/5)

jinja[invalid]: Unexpected templating type error occurred on ({{ 'wc.example.net' | felixfontein.acme._substitute_dns({1: 'a'}) }}): Key or value of dictionary entry are of type <class 'int'> resp. <class 'str'>, but both must be text. Key or value of dictionary entry are of type <class 'int'> resp. <class 'str'>, but both must be text
tests/integration/targets/filter__substitute_dns/tasks/main.yml:62 Task/Handler: Filter errors (5/5)

[...]

PR #4593, opened 2025-04-25
Don't pass deprecated disable_lookups to template function

leegarrett:
From the ansible 2.19 release notes:
- templating - The ``disable_lookups`` option has no effect, since plugins must be updated to apply trust before any templating can be performed.

Fixes: #4592


leegarrett:
Not sure if ansible-lint only supports the latest ansible version, or a range. In the latter case this should probably fall back to setting it again in < 2.19.

ssbarnea (approved):

felixfontein:
This (very likely) causes unwanted code execution, see #4643.

alisonlhart:
Hi @leegarrett, we're going to revert this change and release a patch fix because of #4643. If you could augment this change with a fix for the issue, that would be great. It appears there should be a version check, as you suggested in a comment originally.

felixfontein:
I think the only proper fix is to stop evaluating Jinja2 expressions. This has always been problematic and dangerous (see #2681). Imagine a collection where one lookup deletes every file in your home folder. If the collection also contains a role that has a template expression somewhere that calls the lookup, running ansible-lint on that collection will wipe all your files. This is not something folks expect to happen when running a **linter**.

Also the core team seems to be completely unaware of ansible-lint using `disable_lookups`, since there is a comment `Skipping a deferred deprecation due to no known usage outside ansible-core.` next to where the deprecation is emitted. CC @nitzmahone

leegarrett:
I'm fine with it being reverted. I however have no deeper knowledge on how to properly fix this and will pass on it. I just naively went with the 2.19 changelog, removed it, and gave it a quick test on my set of playbooks.

alisonlhart:
@leegarrett Thanks for the update, completely understood. 
@felixfontein I see your point. We should have a deeper conversation on how to proceed on this. Can you open this as a discussion topic in this repo? Or open a new issue to discuss the broader topic? Instead of my fix in #4648, maybe we should just revert this change as mentioned above and decide where to go from there for 2.19 support.

ssbarnea:
@leegarrett We reverted this due to regression being detected. It will need to be redone with modifications to prevent that regression from happening.

felixfontein:
I started a discussion here: https://github.com/ansible/ansible-lint/discussions/4652

PR #4628, opened 2025-06-11
Adding rule to require unique task names

djdanielsson:
Related: #1710


djdanielsson:
is the issue that the current tests are not written with the unique names in mind and thus erroring or is there something else wrong with the code? also I wasn't sure the best way to add a test for this, if you could point me in that direction and how to fix the CI, I will try and get it all working

Tompage1994:
I love this idea. It seems like a pretty useful lint rule to have in the project. It will help drive good documented code rather than `- name: Do a thing` which is not far from what I've actually seen before

ssbarnea (approved):

PR #4641, opened 2025-06-17
Update format of license in packaging metadata

ssbarnea:


reenberg:
@ssbarnea, should the update to PEP 639 style license properties, not also have bumped the dependency on setuptools to >= 77.0.3, as the minimum version that supports this, according to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files ?

At least i had issues installing ansible-lint, which I believe i tracked down to another dependency locking setuptools to version 75.x, which satisfied the ansible-lint dependncy of setuptools >= 65.3.0, but pip failed to install from git, complaining about the format of the license field.  Fixing this dependency and updating setuptools seems to allow ansible-lint to install.

PR #4642, opened 2025-06-17
Refactor eco testing

ssbarnea:


codecov.yml
```diff
@@ -2,7 +2,6 @@
 comment: false
 coverage:
   status:
-    patch: true
+    patch: false # avoid false positives
```
webknjaz, line 5: This is fixable by marking all uncovered code with `# pragma: no cover` or similar so that the coverage reaches 100%. With that, patch would be usable too.

alisonlhart:
Closing due to age.

Issue #3957, opened 2023-12-21
Unexpected behavior resolving collections, if ansible-lint is executed outside project dir


xrow:
##### Summary

Issue resolving paths to local collections when ansible-lint is not executed in project dir. It loos like tit will not alter or change the path from collections_path with is ./.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
[root@localhost ansible]# ansible-lint --version
ansible-lint 6.22.0 using ansible-core:2.15.6 ansible-compat:4.1.10 ruamel-yaml:None ruamel-yaml-clib:None
A new release of ansible-lint is available: 6.22.0 → 6.22.1 Upgrade by running: pip install --upgrade ansible-lint
```
- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

have a local collection with collections_path = ~/.ansible/collections:/usr/share/ansible/collections:./

```console (paste below)
[root@localhost infrastructure]# ansible-lint ansible/
WARNING  Listing 4 violation(s) that are fatal
syntax-check[specific]: the role 'xxx.infrastructure.vault' was not found in /scm/infrastructure/ansible/ansible_collections/xxx/infrastructure/playbooks/roles:/root/.cache/ansible-compat/9fec08/roles:/root/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/scm/infrastructure/ansible/ansible_collections/xxx/infrastructure/playbooks
ansible/ansible_collections/xxx/infrastructure/playbooks/local.yml:18:15

syntax-check[missing-file]: Unable to retrieve file contents
ansible/playbooks/add-chart.yml:1:1 Could not find or access '/scm/infrastructure/ansible/playbooks/xxx.infrastructure.local' on the Ansible Controller.

syntax-check[missing-file]: Unable to retrieve file contents
ansible/playbooks/build.yml:1:1 Could not find or access '/scm/infrastructure/ansible/playbooks/xxx.infrastructure.local' on the Ansible Controller.

syntax-check[missing-file]: Unable to retrieve file contents
ansible/playbooks/clean-all.yml:1:1 Could not find or access '/scm/infrastructure/ansible/playbooks/xxx.infrastructure.local' on the Ansible Controller.


                    Rule Violation Summary                     
 count tag                        profile rule associated tags 
     3 syntax-check[missing-file] min     core, unskippable    
     1 syntax-check[specific]     min     core, unskippable    

[...]

PR #4644, opened 2025-06-18
Allow loading of yamllint config from a different location

dbrennand:
## Short Description

This PR implements a new CLI option named `--yamllint-file`. It accepts a path to a `yamllint` config file containing yamllint rules to be loaded by `ansible-lint`.

## Reason for Implementation

Current behaviour of `ansible-lint` is to search the current working directory for the following `yamllint` [config files](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/yaml_utils.py#L96). This becomes an issue when you run `ansible-lint` from a directory which is not relative to the `yamllint` config file. I've hit this issue when we have a monorepo structure and our Ansible collection is located in the `ansible` subdirectory, but our project's `yamllint` config is located at the root of the project. For example:

```
/home/user/project/.yamllint.yml <--- Yamllint config file located here
/home/user/project/ansible/galaxy.yml <--- Ansible collection is located in this subdirectory
```

I'd like to mention that this wouldn't be an issue if `--project-dir` installed the Ansible collection itself at runtime, but as mentioned in #3957 [here](https://github.com/ansible/ansible-lint/issues/3957#issuecomment-2981531310), it doesn't seem to do this. It only does this when you run `ansible-lint` relative to where the collection is located. So we hit a scenario where we either:

- Run `ansible-lint` relative to where the collection is located, and `ansible-lint` installs our collection at runtime, but then doesn't load our custom `yamllint` rules.
- Or we run ansible-lint at the root of our project, specifying `--project-dir` and our custom `yamllint` rules are loaded but then the collection isn't installed at runtime and we get [`[syntax-check]` errors as the roles cannot be found from our collection](https://github.com/ansible/ansible-lint/issues/3957#issuecomment-2981531310).

Implementing this new option means we can specify the exact path to our custom `yamllint` config file no matter where it is located. We could have used the `YAMLLINT_CONFIG_FILE` environment variable, but it's not as obvious where that configuration is set, if we can specify directly as a CLI option, then it's clear and straightforward. Furthermore, there are other CLI options for specifying paths to other config files, so it makes sense to do it for the `yamllint` config file too. Finally, I'd like to mention that this feature should be backwards compatible, as we check if the option is not provided, then it uses the previous behaviour.


ssbarnea (changes requested):
Thanks for explaining well your use case, it makes sense. The only issue is that I want to avoid adding more cli arguments to the tool. Still, adding a configuration option might be ok.

@cidrblock WDYT about this? Should we instead allow it only as an option inside ansible-lint config instead of extra cli argument or leave it like this?

cidrblock:
I think it's a great contribution. No concern about additional CLI params.

dbrennand:
Awesome! Thanks @cidrblock 🙂 - Are we good to go then? @ssbarnea?

dbrennand:
Thanks folks! ❤️ 😁  @ssbarnea @shatakshiiii @cidrblock @audgirka

PR #4666, opened 2025-06-26
Renamed rule class methods (naming convention)

ssbarnea:
Renames
- matchyaml -> match_file
- matchtask -> match_task
- matchtasks -> match_tasks
- matcherror -> match_error
- matchvar -> match_var
- matchdir -> match_dir
- create_matcherror -> create_match_error


ssbarnea:
> Can we leave a fall back to the old call is place with warning that the API is changing? I'm not sure how many people have custom rules, but we can't just break it in a single release.

Based on https://sourcegraph.com/search?q=context%3Aglobal+lang%3Apython+AnsibleLintRule+-repo%3Aansible%2Fansible-lint&patternType=keyword&sm=0&expanded= I think we can (obviously with an update to the correct versioning, not as a patch level version change.)

cidrblock:
> > Can we leave a fall back to the old call is place with warning that the API is changing? I'm not sure how many people have custom rules, but we can't just break it in a single release.
> 
> Based on https://sourcegraph.com/search?q=context%3Aglobal+lang%3Apython+AnsibleLintRule+-repo%3Aansible%2Fansible-lint&patternType=keyword&sm=0&expanded= I think we can (obviously with an update to the correct versioning, not as a patch level version change.)

I love sourcegraph.  Unfortunately, it wouldn't be able to see any rules internal to an enterprise.

PR #4683, opened 2025-07-10
Address disable_lookups deprecation with ansible-core 2.19

ssbarnea:
As we cannot use `disable_lookups` anymore, this change will make linter behavior consistent with all versions of ansible-core at the small risk of introducing possible lookup failures.

Fixes: AAP-48340
Fixes: #4592
Related: https://github.com/ansible/ansible-lint/discussions/4652


leegarrett:
Does this PR not have the same security issues as mine from three weeks ago? https://github.com/ansible/ansible-lint/pull/4593

alisonlhart:
@leegarrett Yes, this will be reverted again and fully addressed. We are going to introduce a workaround to this behavior for 2.19 and leave the disable_lookups flag for lower versions. CC @Qalthos

PR #4688, opened 2025-07-22
Skip lookups for ansible-core>=2.19

alisonlhart:
Reverts #4683.

Adds a check for ansible-core 2.19 before setting `disable_lookups=True`. If ansible-core>=2.19 is used, return early and do not template. 

This is because we need to ensure lookups (and code execution) are not performed inside of ansible-lint for ansible-core `>=2.19`. 

See https://github.com/ansible/ansible-lint/discussions/4652 and https://github.com/ansible/ansible-lint/issues/4592

Jira issue for a later larger rework based on the outcome of [this discussion](https://github.com/ansible/ansible-lint/discussions/4652): [AAP-49998](https://issues.redhat.com/browse/AAP-49998)

### DIFF
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -63,6 +63,7 @@ from ansible.plugins.loader import (
 )
 from ansible.template import Templar
 from ansible.utils.collection_loader import AnsibleCollectionConfig
+from packaging.version import Version
 from yaml.composer import Composer
 from yaml.parser import ParserError
 from yaml.representer import RepresenterError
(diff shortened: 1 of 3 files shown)


Qalthos (dismissed):

cidrblock (commented):
I'm wondering if a lighter touch would be to look for "lookup" in the string and if core >= 2.19.0 skip templating.

This would avoid ever passing disable lookups to core b/c is should just create a deprecation message.

cidrblock (approved):
nice, feels better :)

src/ansiblelint/utils.py
```diff
@@ -201,6 +202,13 @@ def ansible_template(
     re_valid_filter = re.compile(r"^\w+(\.\w+\.\w+)?$")
     templar = ansible_templar(basedir=basedir, templatevars=templatevars)
 
+    # Skip lookups for ansible-core >= 2.19; use disable_lookups for older versions
+    if "lookup" in varname:
```
felixfontein, line 206: You should also check for `query` and `q`: https://docs.ansible.com/ansible/latest/plugins/lookup.html#forcing-lookups-to-return-lists-query-and-wantlist-true

felixfontein: I created a new issue for this: #4695.

PR #4701, opened 2025-07-28
Check for gh_action_ref explicitly and document GHA inputs

alisonlhart:


cidrblock (approved):
warning if lint version != action or even lint version is specified :)

PR #4720, opened 2025-08-13
fix: clarify error message related to --fix arguments

simonLeary42:
I got halfway through filing a bug report when I made a minimal `.ansible-lint` config file and discovered that the reason I was getting this error was that I had selected a more restrictive `profile`.


simonLeary42:
pre-commit failures seem unrelated.

alisonlhart:
@simonLeary42 Can you provide a reproducer for this issue? The error you're seeing should only show up if an invalid rule tag has been provided for `--fix`.

simonLeary42:
`ansible-lint ./hello-world.yml`

<table>
<tr>
<td>

```yaml
---
profile: min
write_list: [sanity]
```
</td>
<td>

```yaml
---
profile: production
write_list: [sanity]
```
</td>
</tr>
<tr>
<td>

`ERROR    Found invalid value(s) (sanity) for --fix arguments, must be one of: core, none, internal-error, experimental, syntax-check, unskippable, all, warning, load-failure, parser-error`
</td>
<td>

`Passed: 0 failure(s), 0 warning(s) on 1 files. Last profile that met the validation criteria was 'production'.`
</td>
</tr>
</table>

src/ansiblelint/__main__.py
```diff
@@ -231,7 +231,7 @@ def fix(runtime_options: Options, result: LintResult, rules: RulesCollection) ->
 
     if unknown_tags:  # pragma: no cover
         _logger.error(
-            "Found invalid value(s) (%s) for --fix arguments, must be one of: %s",
+            "Found invalid value(s) (%s) for --fix arguments, must be one of: %s. You may be able to use more values if you change your profile.",
```
cidrblock, line 234: How about

"Valid values are limited by the profile chosen."

simonLeary42:
Which linters actually have the ability to be fixed/written? Is it just `yaml`?

cidrblock:
https://ansible.readthedocs.io/projects/lint/autofix/ should be the current list

simonLeary42:
shouldn't that list match the list in the error message?

cidrblock:
Yes, it should. Looks like it given a list of all tags for the profile and not specifically a list of tags for the profile filtered by which have fixes...*sigh* 

It appears to be a bug.

simonLeary42:
Bug fixed in separate PR.

ssbarnea:
@cidrblock you still have an unresolved comment, if is not needed just resolve it and merge it. thank.

Issue #4731, opened 2025-08-16
Autofix for rule deprecated-local-action removes module parameters


koalajoe23:
##### Summary

When running ansible-lint with autofix option and rule deprecated-local-action enabled, all parameters of the local_action module will be removed, leading to broken ansible code.

As I understand, automatic fixes should never do changes which are more than cosmetic, potentially altering behavior.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 25.8.1 using ansible-core:2.19.0 ansible-compat:25.8.1 ruamel-yaml:0.18.14 ruamel-yaml-clib:0.2.12
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

**test.yml**
```yaml
- hosts: all
  pre_tasks:
    - local_action:
        module: ansible.builtin.command
        cmd: "echo Please don't remove me ansible-lint! I am super important!"
```

```console
ansible-lint --fix deprecated-local-action test.yml
```

This isn't constrained to the command module, I just used it as a simple test case.

##### Desired Behavior

I expected either no attempt to fix this code automatically and give a linting error, or the fix retaining further attributes of local_action:

```yaml
- hosts: all
  pre_tasks:
    - ansible.builtin.command:
        cmd: "echo Please don't remove me ansible-lint! I am super important!"
      delegate_to: localhost
```
##### Actual Behavior

```console
ansible-lint test.yml --fix deprecated-local-action -vvvv
```

```yaml
- hosts: all
  pre_tasks:
    - ansible.builtin.command:
      delegate_to: localhost
```

```paste below
DEBUG    Logging initialized to level 10
INFO     Identified /home/joe/ias/hacklab-ansible as project root due .git directory.
[...]

PR #4733, opened 2025-08-17
make rule deprecated-local-action preserve module parameters

koalajoe23:
Proposed fix for #4731

This is my first attempt at hacking something I want into ansible-lint, be gentle ;)

I was a bit surprised why all rule tests get a a full ruleset provided via `default_rules_collection` param as this makes the rules prone to breakage caused by other rules' auto-formating. I changed this and also made some assertions test if `decprecated-local-action` actually matched, not the number of expected matches which may change in the future.

### DIFF
--- a/examples/playbooks/tasks/local_action.transformed.yml
+++ b/examples/playbooks/tasks/local_action.transformed.yml
@@ -1,4 +1,8 @@
 ---
-- name: Sample
-  ansible.builtin.command: echo 123
+- name: Sample 1
+  command: echo 123
+  delegate_to: localhost
+- name: Sample 2
+  command:
+    cmd: echo 456
   delegate_to: localhost
--- a/examples/playbooks/tasks/local_action.yml
+++ b/examples/playbooks/tasks/local_action.yml
@@ -1,3 +1,7 @@
 ---
-- name: Sample
+- name: Sample 1
   local_action: command echo 123
+- name: Sample 2
+  local_action:
+    module: command
+    cmd: echo 456
--- a/src/ansiblelint/rules/deprecated_local_action.py
+++ b/src/ansiblelint/rules/deprecated_local_action.py
@@ -61,7 +61,12 @@ class TaskNoLocalActionRule(AnsibleLintRule, TransformMixin):
                 if k == "local_action":
                     if isinstance(v, dict):
                         module_name = v["module"]
-                        target_task[module_name] = None
+                        target_task[module_name] = {}
+                        for param, val in v.items():
+                            if param != "module":
+                                target_task[module_name][param] = val
+                        if len(target_task[module_name]) == 0:
+                            target_task[module_name] = None
                         target_task["delegate_to"] = "localhost"
                     elif isinstance(v, str):
                         module_name, module_value = v.split(" ", 1)
--- a/src/ansiblelint/rules/deprecated_local_action.py
+++ b/src/ansiblelint/rules/deprecated_local_action.py
@@ -99,27 +99,27 @@ if "pytest" in sys.modules:
             rules=default_rules_collection,
         ).run()
 
-        assert len(results) == 1
-        assert results[0].tag == "deprecated-local-action"
+        assert any(result.tag == "deprecated-local-action" for result in results)
 
     @mock.patch.dict(os.environ, {"ANSIBLE_LINT_WRITE_TMP": "1"}, clear=True)
     def test_local_action_transform(
         config_options: Options,
-        default_rules_collection: RulesCollection,
     ) -> None:
         """Test transform functionality for no-log-password rule."""
         playbook = Path("examples/playbooks/tasks/local_action.yml")
         config_options.write_list = ["all"]
 
         config_options.lintables = [str(playbook)]
+        only_local_action_rule: RulesCollection = RulesCollection()
+        only_local_action_rule.register(TaskNoLocalActionRule())
         runner_result = get_matches(
-            rules=default_rules_collection,
+            rules=only_local_action_rule,
             options=config_options,
         )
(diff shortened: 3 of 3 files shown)


koalajoe23:
Set this back to DRAFT as SonarCloud Code Analysis CI job doesn't like the code. I kinda agree and will try simplify the implementation

koalajoe23:
I think the remaining CI errors are not cause by my PR, right?

cidrblock:
> I think the remaining CI errors are not cause by my PR, right?

I don't think so, but you may have help us find a bug:
```
WARNING  Skipped installing collection dependencies due to running in offline mode.
an AnsibleCollectionFinder has not been installed in this process
Command failed: got '1', expected '0'
```

https://github.com/ansible/ansible-lint/pull/4734  _should_ fix the action test (confirmed)

koalajoe23:
I changed the debug log statements for missing "local_action" and "module" param to assertions.

Code coverage went down because of the log statements and while trying to write test fixtures I realized a missing `local_action` would result in a non-matching rule, and a missing `module` param would be caught by the `parser-error` rule, also never reaching `deprecated-local-action`.

I think the assertions are fine here to guard against implementation errors.

koalajoe23:
I choose to ignore the copilot nitpicks as I don't see substantial code quality improvements in them.

@cidrblock anything left to do?

cidrblock:
I pulled it down to add some examples and realized this is much more complicated than I orignally thought.

I updated your PR with some edge cases, and got rid of the runtime "assert" statements and added some early exits which tend to ease the type hinting.

Hope it's ok, let me know what you think.  The only issue now is because I committed I need a peer review from someone else on the team :)

koalajoe23:
Hi,

> I updated your PR with some edge cases

:+1: 

I appreciate! didn't think of legacy "dict-as-string" code and throwing other dict elements in and I think it improves the transform test robustness

> got rid of the runtime "assert" statements and added some early exits which tend to ease the type hinting

:+1: 

My intention was to have both asserts in my code fail rather loudly, as both assertions should never trigger, if not caused by an implementation error outside of this rule's `transform()`, not by user input. I was a bit misguided by how assertions are used in C++ and now agree that using them outside of testing code or code ran with optimizations enabled is not a good idea in Python.

[nitpick]: Maybe a log statement could be added to

```python
if "local_action" not in original_task:
  return
```
> Hope it's ok, let me know what you think.

In my understanding (please correct me if wrong), the linter rule transform should do one thing only: transform the ansible task matched by that rule, so that it no longer matches without introducing side-effects or altering the outcome, while preserving everything else. That transformed task may be matched by another rule afterwards, but a single rule should focus on it's core responsibility.

The code you proposed does too much, imo:
1. **[critical]** Your introduction of `shlex` removed quotes from string-form arguments. This could introduce bugs and security vulnerabilities into the users code! `shell: echo "Hello World; /usr/bin/self-destruct"` is not the same as `shell: echo Hello World; /usr/bin/self-destruct`
2. You move the `name` param to the top, which should be the responsibility of rule `key-order` - which might be explicitly disabled by the user.
3. You make `dict`s out of the "key=val" free-form parameter string. Should be handled by `no-free-form` rule.
4.  [nitpick]: I think the `delegate_to` param should be inserted right after the new module param to make diffs a bit easier to parse.

**Only point no 1 is a merge-blocker for me.** I'd be glad to get feedback on my other remarks though! In the end it's on you as maintainer(s) to decide. Functionality-wise, the rule would be fine for me as-is, as it won't break the ansible playbooks I inherit from  other people from time to time again :)

cidrblock:
> [nitpick]: Maybe a log statement could be added to
> 
> ```python
> if "local_action" not in original_task:
>   return
> ```
> 

I don't know the code well enough, but we should never be in the transform for local_action if there isn't one, I can probably rm that.

> In my understanding (please correct me if wrong), the linter rule transform should do one thing only: transform the ansible task matched by that rule, so that it no longer matches without introducing side-effects or altering the outcome, while preserving everything else. That transformed task may be matched by another rule afterwards, but a single rule should focus on it's core responsibility.

Completely correct.
> 
> The code you proposed does too much, imo:

Agree.
> 
> 1. **[critical]** Your introduction of `shlex` removed parentheses from string-form arguments. This could introduce bugs and security vulnerabilities into the users code! `shell: echo "Hello World; /usr/bin/self-destruct"` is not the same as `shell: echo Hello World; /usr/bin/self-destruct`

Good catch!

> 2. You move the `name` param to the top, which should be the responsibility of rule `key-order` - which might be explicitly disabled by the user.

Right again.

> 3. You make `dict`s out of the "key=val" free-form parameter string. Should be handled by `no-free-form` rule.

and again.

> 4. [nitpick]: I think the `delegate_to` param should be inserted right after the new module param to make diffs a bit easier to parse.

Fair.
> 
> **Only point no 1 is a merge-blocker for me.** I'd be glad to get feedback on my other remarks though! In the end it's on you as maintainer(s) to decide. Functionality-wise, the rule would be fine for me as-is, as it won't break the ansible playbooks I inherit from other people from time to time again :)

rework coming.  Thanks for the detailed review.

cidrblock:
@koalajoe23 Does this do everything you need now?  Fun stuff.  BTW, your intuition was right about needing a way to get out quick, it was part of the reason I threw in an exception, it also made adding some unit tests for the edge cases pretty straight forward with having to traverse the entire code base.

koalajoe23:
:+1: for the changes. I really like the decoupled tests and error handling!

Only nitpick I have is that the transformed output should treat empty params as None - just looks much cleaner and "human". I took the liberty to just commit a fix, feel free to drop this commit if you want.

Everything LGTM on my part, awaiting merge :)

alisonlhart (approved):

Issue #4702, opened 2025-07-28
Error rendering template: can only concatenate list (not "_AnsibleTaggedStr") to list


richm:
##### Summary

Version: `ansible-lint 25.6.1 using ansible-core:2.19.0 ansible-compat:25.6.0 ruamel-yaml:0.18.14 ruamel-yaml-clib:0.2.12`

https://github.com/linux-system-roles/cockpit/blob/main/tasks/setup-dnf.yml#L44 and context looks like this:

```yaml
- name: Ensure full package list is installed
  command:
    argv: "{{ __argv }}"
  when: "'cockpit-*' in __pkgs"
  register: __cockpit_dnf
  changed_when: "'Nothing to do.' not in __cockpit_dnf.stdout_lines"
  vars:
    __excludes: "{{ ['--exclude'] | product(__cockpit_packages_exclude | union(cockpit_packages_dyn_exclude)) | flatten }}"
    __argv: "{{ ['dnf', '-y', 'install', 'cockpit-*'] + __excludes }}"
    __pkgs: "{{ __cockpit_packages[cockpit_packages]
      if cockpit_packages in __cockpit_package_types
      else cockpit_packages }}"
```

Note that `cockpit_packages_dyn_exclude` is a `list`, and `__cockpit_packages_exclude` is a `list`.

The latest ansible-lint gives this error:

```
jinja[invalid]: Error rendering template: can only concatenate list (not "_AnsibleTaggedStr") to list

tasks/setup-dnf.yml:44:13 Task/Handler: Ensure full package list is installed
```

everything is a list - I can't figure out what is wrong here

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 25.6.1 using ansible-core:2.19.0 ansible-compat:25.6.0 ruamel-yaml:0.18.14 ruamel-yaml-clib:0.2.12
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

```console (paste below)
git clone https://github.com/linux-system-roles/cockpit
cd cockpit
ansible-lint
```

##### Desired Behavior

Ansible-lint either should not complain, or it should tell me what is wrong

##### Actual Behavior

```
jinja[invalid]: Error rendering template: can only concatenate list (not "_AnsibleTaggedStr") to list

tasks/setup-dnf.yml:44:13 Task/Handler: Ensure full package list is installed
```

richm:
[...]

PR #4737, opened 2025-08-20
Fix Jinja template rendering error when concatenating _AnsibleTaggedStr with list

anusshukla:
Description- ansible-lint isn’t compatible with our current ansible-core version(2.19). 
issue - https://github.com/ansible/ansible-lint/issues/4702
related - https://issues.redhat.com/browse/AAP-51009


cidrblock (changes requested):
Can we add a test that templates a string and generates the message to see core generating it and us ignoring it?

cidrblock (approved):

PR #4740, opened 2025-08-21
Fix rule summary for error messages

simonLeary42:
Displays only transforming rules in the error message that lists valid rules for `write_list` / `--fix`. Prevents the use of non-transforming rules in `write_list` / `--fix`.

https://github.com/ansible/ansible-lint/pull/4720

example, using a non-transforming rule `schema`:

`ansible-lint ./hello-world.yml --fix=schema`

<table>
<tr><td /><td>profile=min</td><td>profile=production</td></tr>
<tr>
<td>before</td>
<td>

`ERROR    Found invalid value(s) (schema) for --fix arguments, must be one of: warning, unskippable, core, parser-error, syntax-check, internal-error, load-failure, all, none, experimental`
</td>
<td>

`Passed: 0 failure(s), 0 warning(s) on 1 files. Profile 'production' was required, and it passed.`
</td>
</tr>
<tr>
<td>after</td>
<td>

`ERROR    Found invalid value(s) (schema) for --fix arguments, must be one of: all, none`
</td>
<td>

`ERROR    Found invalid value(s) (schema) for --fix arguments, must be one of: command-instead-of-shell, name, all, key-order, no-jinja-when, yaml, fqcn, none, partial-become, jinja, pattern, deprecated-local-action, no-free-form`
</td>
</tr>
</table>

### DIFF
--- a/src/ansiblelint/rules/__init__.py
+++ b/src/ansiblelint/rules/__init__.py
@@ -533,7 +533,6 @@ class RulesCollection:
 
     def known_transform_tags(self) -> list[str]:
         """Return a list of known tags of rules that implement transform()."""
-        
         tags = set()
         for rule in self.rules:
             if issubclass(rule.__class__, TransformMixin):


simonLeary42:
I'm not sure what "without returning no sub-tags" means.

PR #4774, opened 2025-09-15
fix: github action ref doesn't checkout that ref

rohun-apex:
#### Problem

The below code block will checkout ansible-lint at main. Proof: `Using ansible-lint ref: main` will show up in the GHA (see "Was" section of verification below). Expected: it should checkout ansible-lint at `v25.8.2`.

```
  jobs:
  build:
    name: Ansible Lint
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Run ansible-lint
        uses: ansible/ansible-lint@v25.8.2
```

#### Explanation
This is believed to be because of a long-standing bug for github variables in a composite action (see below linked issues), which this `action.yml` seems to be. Given that root bug hasn't been fixed since 2023, it effectively becomes a bug manifested in ansible-lint.
- https://github.com/actions/runner/issues/2525
- https://github.com/actions/runner/issues/2473

Following a similar example to https://github.com/actions/runner/issues/2525 , I have renamed that variable so it gets around the bug in GHAs.

#### Verification
**Was:**
1. GHA will checkout `main` despite specifying the tag to use (see top line of image).
image

**Is:**
1. GHA will checkout the tag that is referenced (top line ref specified matches bottom ref it will check out).
image


alisonlhart (approved):

rohun-apex:
Thanks @alisonlhart !!

Two problems which I'm happy to resolve, but could use some guidance on.
1. 1 check has failed (looks like it failed on main too [link](https://github.com/ansible/ansible-lint/commits/main/) ) and two checks are pending.
2. I don't have authorization to push to this branch. Feel free to give me permissions of copy the branch to a new one.

Issue #4784, opened 2025-09-18
systemctl set-environment MYSQLD_OPTS="--skip-grant-tables"


ziegenberg:
##### Summary

I receive an incorrect warning when using `ansible.builtin.command` to set or unset an environment with `systemctl`. The warning suggests using `ansible.builtin.systemd` instead, which does not natively support it.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
ansible-lint 25.8.2 using ansible-core:2.19.2 ansible-compat:25.8.1 ruamel-yaml:0.18.15 ruamel-yaml-clib:0.2.12
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

```yml
- name: Set MYSQLD_OPTS environment
  ansible.builtin.command:
    cmd: systemctl set-environment MYSQLD_OPTS="--skip-grant-tables"
  changed_when: true
  become: true
```

##### Desired Behavior

No linting failures.

##### Actual Behavior

I receive an error message indicating that I should use the `ansible.builtin.systemd` module instead of the `ansible.builtin.command` module.

```paste below
systemctl used in place of systemd moduleansible-lint[command-instead-of-module]
```

PR #4785, opened 2025-09-18
fix: add set-environment and unset-environment in executable_options for systemctl

ziegenberg:
Do not warn when using `ansible.builtin.command` to set or unset an environment with `systemctl`.

Fixes: #4784


alisonlhart:
Hi @ziegenberg, thanks for this contribution! Can you update the dictionary for the missing entry? Then tests should pass and we can merge this.

ziegenberg:
Hi @alisonlhart,

I added "MYSQLD" to the dictionary ~and sorted the almost sorted list of entries (my inner Monk made me do it :) )~.

Greetings,
Daniel

PR #4791, opened 2025-09-24
feat: role_var_separator for configuring the namingscheme of role-vars

NocDerEchte:
## Feature/Addition
Add `role_var_separator` option to specify which string to use as a separator between the role-name and variable-name in variables belonging to an Ansible role.

## Why

My company and I want to use a role_var scheme of `role__some_var` but ansible-lint currently only supports `role_some_var`.
As this was a quick addition to the already existing `var-naming` rule, I decided to create a pull request for this.


alisonlhart:
Hi @NocDerEchte, thanks for this suggestion. We don't want to let the underscore separator be an editable setting, as this could result in role names passing lint that don't follow expected conventions. (E.g. - no underscore at all)
But since we don't want to allow this behavior to be edited for safety reasons, we're opting to close this change.
Feel free to open an issue or feature request expanding on reasoning for needing this at the project level.
Thanks!

PR #4796, opened 2025-09-30
fix: pin to python3.13 for the pre-commit action

ssbarnea:
Fixes issue where running recent versions of the pre-commit hook
might fail on systems where default python is 3.10 (which is not
supported).

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -8,6 +8,9 @@
   description: This hook runs ansible-lint.
   entry: python3 -m ansiblelint -v --force-color
   language: python
+  # version is needed to avoid accidental use of python3.10 on some platforms
+  # as this is not supportted by recent versions of a ansible-lint and ansible-core
+  language_version: python3.13
   # do not pass files to ansible-lint, see:
   # https://github.com/ansible/ansible-lint/issues/611
   pass_filenames: false


alisonlhart (approved):

chriscroome:
@ssbarnea is it intentional that running `ansible-lint` via `pre-commit` is only now allowed with Python >= `python3.13`? If so I'd better get my Debian Bookworm laptop upgraded to Trixie ASAP! Also is this going to cause some complaints from people running Ubuntu LTS Noble that they can't use `ansible-lint` via `pre-commit` any more as they have [Python 3.12.3](https://packages.ubuntu.com/noble/python3-minimal)?

finswimmer:
Hey,

this seems like a suboptimal solution to me. If `ansible-lint` does not support Python 3.10 anymore, this should be declared in the `pyproject.toml`.

fin swimmer

reinout:
> Also is this going to cause some complaints from people running Ubuntu LTS Noble that they can't use `ansible-lint` via `pre-commit` any more as they have [Python 3.12.3](https://packages.ubuntu.com/noble/python3-minimal)?

Agree. I suggested changing the version pin to 3.12, but that was rejected (https://github.com/ansible/ansible-lint/issues/4819). (Note that there's an easy workaround in that ticket, so it is easy to repair, but it is a bit of a bummer to have to do that hack in many repositories.)

Suboptimal: yes. It succeeds in getting a more explicit error on older systems with 3.10, but also succeeds in getting an unnecessary error on CI environments with only 3.11, 3.12 or 3.14.

gone-for-coding:
This also breaks in our environment as we only have Python 3.12 installed. All linting pipelines are currently red and we cannot use the new version.

ssbarnea:
Brace yourself, in few days, the hook will pin to python 3.14 in order make use of ansible-core 2.20 features. 

Joke aside, you can always override the hook defaults or use your own hook. The reason why we do it like this is to ensure that its behaviour is consistent across systems (combined with some pre-commit tool design limitations). We do not want to run the hook locally and get a different result than running it ok CI on another machine.

If anyone is willing to experiment alternatives, take a look at the hooks defined inside ansible/actions repository and let me know what you think. They are experimental and not even documented yet. Basically they use uv itself, if you look at their definition you might start running the same way locally and fully control the versions of python and lint and core.

reinout:
Unrelated to ansible-lint: by browsing the ansible/actions repository, I discovered your recent "tombi" addition in there. Nice! I've spent the last few hours creating a json schema for some internal project and adding formatting/linting with tombi+pre-commit :-) Such repositories often give me a few good ideas.

Related to the python version pin from this PR: if I look at https://github.com/ansible/actions/blob/main/.pre-commit-hooks.yaml#L46-L55 , the action there won't solve the "ubuntu 24.04 problem". You'd need to install `uv` and `python 3.13` for the action to work on the github 24.04 standard runner (if I'm not mistaken).

`uv` is getting used in more and more projects (mine, too), so doing `uv run` in an action can be a handy idea. `uv` will honour the `requires-python = ">=3.12"` from the actions' `pyproject.toml` just fine, so python versions that are too old will get an informative error message, something that apparently doesn't happen when pre-commit itself installs ansible-lint. With the good error message, does the action really need the 3.13 pin?

chriscroome:
Adding `language_version: python3` to the `..pre-commit-config.yaml` files enables pre-commit to work with multiple Python versions, eg:

```yaml
---
repos:
  # https://github.com/ansible/ansible-lint/releases
  - repo: https://github.com/ansible/ansible-lint.git
    rev: v26.3.0
    hooks:
      - id: ansible-lint
        name: Ansible Lint
        language: python
        language_version: python3
        additional_dependencies:
          # https://pypi.org/project/ansible/
          - ansible==13.4.0
          # https://github.com/kellyjonbrazil/jc/releases
          - jc==1.25.6
          # https://github.com/jmespath/jmespath.py/tags
          - jmespath==1.0.1
```

Issue #4793, opened 2025-09-28
ansible-lint requires "repository" key to exist in "galaxy.yml" although "repository" is not syntactically required.


stopendy:
##### Summary

`ansible-lint` raises an error with "very high" severity if `galaxy.yml` doesn't have a `repository` key.

As [Collection Galaxy metadata structure](https://docs.ansible.com/ansible/latest/dev_guide/collections_galaxy_meta.html) says, `repository` key is not required. Thus `ansible-lint` should not detect `schema[galaxy]` rule violation, which is mandatory. It can raise `galaxy[xxx]` rule with less severity.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
# ansible-lint 25.9.0 using ansible-core:2.19.2 ansible-compat:25.8.1 ruamel-yaml:0.18.15 ruamel-yaml-clib:0.2.14
```

- OS: Fedora release 42 (Adams)
- ansible installation method: one of source, pip
- ansible-lint installation method: one of source, pip

##### STEPS TO REPRODUCE

1. Create `galaxy.yml` as below.

```yaml
namespace: local
name: test
description: ''

version: 0.0.1

readme: README.md
authors:
  - nobody
```

2. Validate the `galaxy.yml` with `ansible-lint`.

```sh
ansible-lint galaxy.yml 
```

```console (paste below)
python3 -m venv venv
source venv/bin/activate
pip install ansible-core ansible-lint

cat << 'EOF' > galaxy.yml
namespace: local
name: test
description: ''

version: 0.0.1

readme: README.md
authors:
  - nobody
EOF

ansible-lint galaxy.yml
```

##### Desired Behavior

`ansible-lint` should not report `schema[galaxy]` because `galaxy.yml` provided above is syntactically correct.

##### Actual Behavior

`ansible-lint galaxy.yml` reported `schema[galaxy]` rule violation.

```sh
ansible-lint galaxy.yml
# WARNING  Project directory /.ansible cannot be used for caching as it is not writable.
# WARNING  Using unique temporary directory /tmp/.ansible-0aaa for caching.
# WARNING  Listing 5 violation(s) that are fatal
# galaxy[no-changelog]: No changelog found. Please add a changelog file. Refer to the galaxy.md file for more info.
# galaxy.yml:1

[...]

PR #4798, opened 2025-10-01
fix: make repository key optional in galaxy.yml

rivv0:
Updated the galaxy.json schema used by ansible-lint to remove the repository key from the list of required properties for galaxy.yml. According to the [Collection Galaxy metadata structure], the repository key is optional and should not trigger a schema violation.

- ansible-lint will no longer fail with a schema error when repository is missing from galaxy.yml.
- Developers will receive a warning (if the new rule is enabled) rather than a blocking error.
- Improves compatibility with the official Ansible Galaxy metadata specification and the VS Code Ansible extension.

Fixes: #4793


alisonlhart:
@rivv0 Thanks for this contribution! It looks great. We just need the tests updated to accommodate the change. Can you make the needed changes (add LICENSE to fixtures) and then we can get this merged. We will wait for final approval until then, but I don't have any concerns outside of the tests.

rivv0:
yes would do that.

StopMotionCuber:
This one caught me a bit off guard, as this sneakily added the `galaxy[no-license]` rule.

Due to the way this PR is structured, the addition of this rule is not apparent in the latest `ansible-lint` release. Furthermore, the rule is undocumented [here](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/rules/galaxy.md?plain=1)

I guess the rule itself is okay, but should get a `feature` entry in the release notes.

PR #4801, opened 2025-10-02
fix(schema): Make inventory JSON Schema match actually accepted inventories

viccie30:
This matches what `ansible.builtin.yaml` accepts as input. 
Working on this I have a few questions:

1. The `all` group had a `groups` property, but that doesn't seem to be accepted by Ansible, so I removed it. Is that correct?
2. `vars` and `children` also accept a string, like `hosts`, and turn this into a dictionary with the string as a key and `null` as the value. It does not seem really useful, but because the syntax is accepted I added it to the schema.
3. Host and group variables starting with a digit are accepted, but are deprecated. I've written the schema to reject them, is this the wanted behavior?

### DIFF
--- a/src/ansiblelint/schemas/inventory.json
+++ b/src/ansiblelint/schemas/inventory.json
@@ -1,66 +1,154 @@
 {
   "$defs": {
     "group": {
+      "additionalProperties": false,
+      "description": "See https://docs.ansible.com/ansible/latest/inventory_guide/intro_inventory.html#inventory-basics-formats-hosts-and-groups",
       "properties": {
         "children": {
+          "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


viccie30:
I have no clue why the schemas test fails. I get the same error locally, but I have no idea what's wrong with my changes to make it fail like that.

viccie30:
> I have no clue why the schemas test fails. I get the same error locally, but I have no idea what's wrong with my changes to make it fail like that.

Found it: the `examples` property was abused to list filenames. I've changed it to `x-ansible-lint`. 

The duplication of the error is because `all` is validated against both `patternProperties` and its specific entry under `properties`. The explicit duplication under `properties` is only to update the `description` for `all` and `unbound`, so it can be removed without further problems if preferred.

ssbarnea:
please resolve conflicts

Issue #4800, opened 2025-10-02
ansible lint 25.9.1 breaks uv sync, lock and add unless --prerelease=allow is added


dh-seanmurphy:
##### Summary

`ansible-lint` `25.9.1` depends on an `ansible-core` version which depends on python `3.14`. Python `3.14` is not released yet.

Almost certainly related to https://github.com/ansible/ansible-lint/pull/4795

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

Not possible to run ansible-lint due to dependency problems.

Running in a container based on python 3.13.7.

- ansible installation method: uv
- ansible-lint installation method: uv

##### STEPS TO REPRODUCE

`pyproject.toml`

```
[project]
name = "ansible-lint-test"
version = "0.0.0"
dependencies = [
  "ansible-lint==25.9.1",
]

[tool.uv]
# Use the system interpreter provided in the docker image.
python-downloads = "never"
```

```
$ uv sync --locked
Using CPython 3.13.3 interpreter at: /usr/bin/python3
Removed virtual environment at: .venv
Creating virtual environment at: .venv
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.13`.
  × No solution found when resolving dependencies for split (markers: python_full_version >= '3.14' and platform_python_implementation != 'PyPy'):
  ╰─▶ Because only ansible-core{python_full_version >= '3.14'}<2.20.0b1 is available and ansible-lint==25.9.1 depends on ansible-core{python_full_version
      >= '3.14'}>=2.20.0b1, we can conclude that ansible-lint==25.9.1 cannot be used.
      And because your project depends on ansible-lint==25.9.1, we can conclude that your project's requirements are unsatisfiable.

      hint: `ansible-core` was requested with a pre-release marker (e.g., ansible-core>=2.20.0b1), but pre-releases weren't enabled (try:
      `--prerelease=allow`)

      hint: While the active Python version is 3.13, the resolution failed for other Python versions supported by your project. Consider limiting your
      project's supported Python versions using `requires-python`

```

##### Desired Behavior

`ansible-lint` should be installed with dependencies which do not require the as yet unreleased python `3.14`

##### Actual Behavior

`ansible-lint` installation fails.

ssbarnea:
I think that this is a limitation of uv.lock files and you created the problem by prevent it from downloading python 3.14.

[...]

PR #4802, opened 2025-10-03
fix: avoid using unreleased dependency constraints

ssbarnea:
- use newer ansible_compat which ensures that py314 and ansible-core
  versions are compatible
- removes dependency on unreleased version of ansible-core
- add tests for packaging issues related to pre-releases

Fixes: #4800
Fixes: #4795
Fixes: AAP-55147


pyproject.toml
webknjaz, line 1: Maybe also drop/fix the `pyyaml` items. It seems wrong that Python 3.14 would have 6.0.2 (older version) as minimum and Python 3.10 has 6.0.3 (newer).

PR #4806, opened 2025-10-06
Add forum post to release process.

Qalthos:
Also fix the label on the `setup-python` action while I'm in here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
       TOXENV: pkg
 
     steps:
-      - name: Switch to using Python 3.10 by default
+      - name: Switch to using latest Python by default
         uses: actions/setup-python@v6
         with:
           python-version: "3.13"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -55,3 +55,15 @@ jobs:
         if: >- # "create" workflows run separately from "push" & "pull_request"
           github.event_name == 'release'
         uses: pypa/gh-action-pypi-publish@release/v1
+
+  forum_post:
```
ssbarnea, line 59: That is a different JOB! obviously that you do not have python setup to it. Now the question is why you wanted a different job?

.github/workflows/release.yml
```diff
@@ -55,3 +55,15 @@ jobs:
...
+    if: "${{ github.event_name == 'release' }}"
+    needs: release
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Retrieve the forum post script from team-devtools
+        run: curl -O https://raw.githubusercontent.com/ansible/team-devtools/main/.github/workflows/forum_post.py
```
ssbarnea, line 66: Don't use curl for these. Just rely on a reusable action to do the posting. Basically we should be able to get the posting for free in other repos once we implment it in release workflow from team-devtools.

PR #4835, opened 2025-11-04
fix: require ansible-core>=2.17.10

ssbarnea:


alisonlhart:
The options for supported_ansible should be updated too:
```
# config.py
....
    _default_supported = ["2.15.", "2.16.", "2.17.", "2.18.", "2.19."]
```

ssbarnea:
> The options for supported_ansible should be updated too:
> 
> ```
> # config.py
> ....
>     _default_supported = ["2.15.", "2.16.", "2.17.", "2.18.", "2.19."]
> ```

I kept them out on purpose because I remember that this caused problems in the past with some collection maintainers as it will require them to update minimal version too. That is why we still have 2.15 listed there while we do not support it ourselves.

PR #4860, opened 2025-12-01
chore: Remove cffi, importlib-metadata, reduce core lower bound

cidrblock:
This pull request introduces platform compatibility constraints for several dependencies, updates pre-commit hooks, and removes unnecessary dependencies from both the codebase and configuration files. It also updates schema etags to reflect the latest versions. These changes improve the project's maintainability and ensure compatibility with targeted platforms.

**Dependency constraints and removals:**

* Added explicit platform compatibility constraints for key dependencies (such as `ansible-core`, `cffi`, `django`, `importlib-metadata`, `jsonschema`, `packaging`, `python-gnupg`, `setuptools`) in `renovate.json` to prevent upgrades beyond supported versions.
* Removed `importlib-metadata` and `zipp` from `.config/requirements-lock.txt`, `pyproject.toml`, and `.pre-commit-config.yaml`, reflecting a move to the standard library's `importlib.metadata` for Python 3.8+ and cleaning up unused dependencies. [[1]](diffhunk://#diff-c339908dfbcdd2133009c338990a2cb3e73d9c32aa9cf41d3a4469d589e8a176L29-L30) [[2]](diffhunk://#diff-c339908dfbcdd2133009c338990a2cb3e73d9c32aa9cf41d3a4469d589e8a176L85-L86) [[3]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L37-L43) [[4]](diffhunk://#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9L174)
* Updated `src/ansiblelint/requirements.py` to use `importlib.metadata` instead of `importlib_metadata`, aligning with the dependency removals and modern Python practices. [[1]](diffhunk://#diff-cca990a29c4c8d37b3e44a9ee179e7759e803e436eed994b5ce1f3f0013018f0L3-R4) [[2]](diffhunk://#diff-cca990a29c4c8d37b3e44a9ee179e7759e803e436eed994b5ce1f3f0013018f0L16-R19)

**Pre-commit and tooling updates:**

[...]

PR #4861, opened 2025-12-02
chore: adopt pytest>=9.0.0 config

Qalthos:


alisonlhart:
If this fixes all the tombi errors, we can also remove the strict mode = false comment in `pyproject.toml`.

alisonlhart:
Dictionary is assumed to be cached in pre-commit. tox -e lint is passing in eco. Similar override to https://github.com/ansible/ansible-lint/pull/4860.
RPM failure is due to Python 3.14.

PR #4899, opened 2026-01-11
feat: added exit code 8 for successfully fixed violations (#4674)

### LINKED ISSUE #4674: Return special exit code when auto-fixing leave no other changes to be made
As we cannot return 0 exit code if we modify files, we should have a special exit code that would be returned when we fix existing issues and there are none left to fix.

Jkhall81:
## Summary
This PR implements a new exit code (`8`) to distinguish between a clean run and a run where all violations were successfully resolved using `--fix`. 

- Added `RC.FIXED_VIOLATIONS = 8` to constants.
- Updated `report_outcome` logic to prioritize the fixed-violations exit code when no failures remain and files were modified.
- Added integration test in `test/test_app.py` to verify the new behavior.

Closes #4674

### DIFF
--- a/src/ansiblelint/app.py
+++ b/src/ansiblelint/app.py
@@ -198,7 +198,10 @@ class App:
     ) -> int:
         """Display information about how to skip found rules.
 
-        Returns exit code, 2 if errors were found, 0 when only warnings were found.
+        Returns exit code:
+          - 0: when no errors were found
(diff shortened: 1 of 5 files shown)


xrow:
I came here because I discovered this https://github.com/ansible/ansible-lint/discussions/5015

The behavior made my cicd fail without a need.

PR #4904, opened 2026-01-13
Set cryptography ceiling to match downstream requirement.

cidrblock:
This pull request makes a small change to the project dependencies by updating the version constraint for the `cryptography` package in `pyproject.toml`. The new constraint restricts `cryptography` to versions lower than 38 instead of requiring version 38 or higher.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -188,7 +188,7 @@ requires-dist = [
     { name = "ansible-core", specifier = ">=2.16.14" },
     { name = "black", specifier = ">=24.3.0" },
     { name = "cffi", specifier = ">=1.15.1" },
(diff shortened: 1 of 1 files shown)


cidrblock:
pre-commit.ci run

ssbarnea (changes requested):
No regression protection included.

alisonlhart:
@ssbarnea Regression protection is here: https://github.com/ansible/team-devtools/pull/398

alisonlhart:
Ahh I see, renovate tries to still update. We need to determine why the dep is hitting 37 anyway, the hook shouldn't enforce <38 on deps, only that the lower bound includes <38. 
Will look at this today.

PR #4938, opened 2026-01-27
fix: schema/meta support SLES 15SP7 and 16

Klaas-:
Add 15SP7 and also add 16, maybe for 16 it is enough to just use the major version like with EL and clones? They switched their versioning https://documentation.suse.com/sles/16.0/html/SLE-comparison/index.html#sle16-versioning , not sure if feature wise it means we can also treat it more like EL :)

https://www.suse.com/lifecycle#suse-linux-enterprise-server-15
https://www.suse.com/lifecycle#suse-linux-enterprise-server-16


Klaas-:
> Label error. Requires exactly 1 of: breaking, build, chore, ci, docs, feat, fix, perf, refactor, style, test. Found: . Follow https://www.conventionalcommits.org to get auto-labeling to work correctly.

Maybe as a note here, this seems to be about the PR title, not the git commit :)

Issue #4939, opened 2026-01-28
No system Python installation found for Python 3.14


dcorto:
##### Summary

No system Python installation found for Python 3.14

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)

Run cd $GITHUB_ACTION_PATH
  cd $GITHUB_ACTION_PATH
  uv pip install --system "ansible-lint[lock] @ git+https://github.com/ansible/ansible-lint@$GH_ACTION_REF"
  ansible-lint --version
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    ACTION_REF: main
    UV_PYTHON_INSTALL_DIR: /home/runner/work/_temp/uv-python-dir
    UV_PYTHON: 3.14
    UV_CACHE_DIR: /home/runner/work/_temp/setup-uv-cache
    GH_ACTION_REF: main
error: No system Python installation found for Python 3.14

```

##### STEPS TO REPRODUCE

```yml

# Workflow for lint code project
name: Lint

on:
  pull_request:

permissions:
  contents: read

jobs:
  ansible-lint:
    timeout-minutes: 5
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v6

      - name: Run ansible-lint
        uses: ansible/ansible-lint@main
        with:
          setup_python: "true"
          working_directory: "infrastructure/deploy"

```

##### Desired Behavior

Should work in Github actions

##### Actual Behavior

Do not works in ubuntu-latest

konstruktoid:
Fails: https://github.com/konstruktoid/ansible-role-docker-rootless/actions/runs/21436985882/job/61730029339?pr=724 (commit https://github.com/ansible/ansible-lint/commit/64dfcc1dd96d681e3b0c684d5736bbb340e8b8d2) 
Succeeds: https://github.com/konstruktoid/ansible-role-hardening/actions/runs/21430772424/job/61709364029 (commit ff636f78696c40dd7037b591fbccb23abdcfc5ca)

konstruktoid:
Workaround before fix:

```yaml
- name: Lint Ansible Playbook
  uses: ansible/ansible-lint@ff636f78696c40dd7037b591fbccb23abdcfc5ca
```

dcorto:
thanks for the quick fix ❤️

sdodsley:
[...]

PR #4941, opened 2026-01-28
fix: use pip instead of uv when setup_python is false

Jkhall81:
## Summary
Fixes regression where users with `setup_python: false` get `uv: command not found` errors. Maintains backward compatibility by using pip installation when `setup_python` is false, while using uv when true.

## Issue Reference
- Related to reports in #4939 and comments from users with `setup_python: false`


Jkhall81:
Sorry, should have caught the backwards compatibility issue sooner.

alisonlhart:
No worries @Jkhall81, thanks so much for the quick fix.

PR #4944, opened 2026-01-31
fix: return user home directory instead of / 

lijun1234:
Per docs/usage.md (Caching), {project_dir} is determined from a CLI argument, config file, git root, or user home directory.

Bug 4943 shows that if no --project-dir is passed and there’s no .git or config file, it falls back to / instead of the home directory. This breaks cache folder initialization for non-root users, who lack write permission on /.

The fix is to rewrite the values of project_dir, method if find_project_root return method "file system root".


lijun1234:
The tox is not setup properly. Will recreate PR with tox setup in the clean branch.

Issue #4851, opened 2025-11-22
Regression of #1771 -- collections_paths from ansible.cfg is ignored, even though other variables are honored.


otheus:
##### Summary

[!note] This may be specific to the Mac/Brew version of this project. 

When ansible-lint is run, and sync_path is disabled, ansible-lint choses its own values for the collection_path, ignoring what it reads from ansible.cfg.  This can be seen in comparing the output of `--version` and from `-vvvv`.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
$ uname -s -r
Darwin 24.5.0

$ ansible-lint --version -vvvv
DEBUG    Logging initialized to level 10
INFO     Identified /Users/me/projects/osdb-deployment as project root due .git directory.
DEBUG    Options: Options(_skip_ansible_syntax_check=False, cache_dir=PosixPath('/Users/me/projects/osdb-deployment/.ansible'), colored=True, configured=True, cwd=PosixPath('/Users/me/projects/osdb-deployment'), display_relative_path=True, exclude_paths=['.cache', '.git', '.hg', '.svn', '.tox'], format=None, lintables=[], list_rules=False, list_tags=False, write_list=[], write_exclude_list=[], parseable=False, quiet=0, rulesdirs=[PosixPath('/opt/homebrew/Cellar/ansible-lint/25.11.0_1/libexec/lib/python3.14/site-packages/ansiblelint/rules')], skip_list=[], tags=[], verbosity=4, warn_list=['experimental', 'jinja[spacing]', 'fqcn[deep]'], mock_filters=[], mock_modules=[], mock_roles=[], loop_var_prefix=None, only_builtins_allow_collections=[], only_builtins_allow_modules=[], var_naming_pattern=None, offline=False, project_dir='.', extra_vars=None, enable_list=[], skip_action_validation=True, strict=False, rules={}, profile=None, task_name_prefix='{stem} | ', sarif_file=None, config_file=None, generate_ignore=False, rulesdir=[], use_default_rules=False, version=True, list_profiles=False, ignore_file=None, yamllint_file=None, max_tasks=100, max_block_depth=20, supported_ansible_also=[])[/]
DEBUG    CWD: /Users/me/projects/osdb-deployment
DEBUG    Attempting to release lock 4392823712 on /Users/me/projects/osdb-deployment/.ansible/.lock
DEBUG    Lock 4392823712 released on /Users/me/projects/osdb-deployment/.ansible/.lock


$ ansible --version
[...]

PR #4965, opened 2026-02-12
Fixes/4851

Nitin75408:
Fix collections_paths from ansible.cfg being ignored (Issue #4851)
Problem
ansible-lint ignored the collections_paths setting from ansible.cfg, even though other settings were respected. The ansible-compat Runtime class would override ANSIBLE_COLLECTIONS_PATH with hardcoded defaults, ignoring user configuration.

Solution
Read collections_paths from ansible.cfg and set ANSIBLE_COLLECTIONS_PATH environment variable before Runtime initialization, ensuring proper configuration precedence.

Changes
src/ansiblelint/config.py - Added ansible_collections_paths field to Options
src/ansiblelint/ansible_config.py (NEW) - Module to read and parse ansible.cfg collections_paths
src/ansiblelint/app.py - Load ansible.cfg and set ENV VAR before Runtime creation
test/test_ansible_cfg_collections.py (NEW) - 11 test cases validating the fixF

Fixes#4851

### DIFF
--- a/test/test_ansible_cfg_collections.py
+++ b/test/test_ansible_cfg_collections.py
@@ -232,13 +232,13 @@ collections_path = {cfg_collections}
     from ansiblelint.app import App
 
     test_options = Options(project_dir=str(project_dir))
-    
+
     # Before creating app, verify env var is set
     assert os.environ.get("ANSIBLE_COLLECTIONS_PATH") == str(env_collections)
-    
+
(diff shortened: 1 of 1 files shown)


ssbarnea:
All the logic related to communicating with ansible internals should be made via ansible-compat library and not directly into the linter.

PR #4968, opened 2026-02-18
feat: adding new argument_spec rule and tests

djdanielsson:
I believe this is a great new opt-in (for now) rule because argument_spec is a great feature that so many either do not know about or utilize. Having this rule will allow users to require it if they want to do so. 

assisted by claude-4.6-opus


djdanielsson:
I believe I have addressed all the issues that copilot had

PR #5021, opened 2026-04-14
fix: added setup-uv action version pinning to renovate config

### LINKED ISSUE #5020: astral-sh is an unpinned action
##### Summary

With the number of github action supply chain attacks happening recently, github orgs are enabling `Require actions to be pinned to a full-length commit SHA`. however, once this is enabled the linting action fails due to:

`Error: The action astral-sh/setup-uv@v7 is not allowed in garethahealy/ansible-middleware-playground because all actions must be pinned to a full-length commit SHA.`

garethahealy:
fixes: https://github.com/ansible/ansible-lint/issues/5020 - Added renovate config to pin setup-uv

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,6 +3,12 @@
   "extends": ["github>ansible/actions//config/renovate.json"],
   "packageRules": [
     {
+      "description": "Pin astral-sh/setup-uv in action.yml to commit SHA (tag preserved in comment)",
+      "matchManagers": ["github-actions"],
+      "matchPackageNames": ["astral-sh/setup-uv"],
+      "pinDigests": true
+    },
+    {
       "allowedVersions": "<2.17",
       "description": "Platform compatibility constraint",
       "matchPackageNames": ["ansible-core"]


garethahealy:
Dont believe CI failures are due to my changes, so ready for review.

garethahealy:
As renovate is used by this repo, have pivoted to using renovate to pin the action - next round of renovate should pin the v8 version correctly

djdanielsson:
@ansible/devtools could we get someone to look at this? It would help a lot of us who are trying to be more careful what versions of tools we run because of all the supply chain issues in the upstream world

garethahealy:
@ssbarnea any chance you can review? this is blocking enabling pinned actions in https://github.com/redhat-cop

cidrblock:
## Review Notes

The Renovate config change looks correct and well-scoped. Here's a quick summary:

**What this does:** Adds a `packageRule` with `pinDigests: true` for `astral-sh/setup-uv`. On the next Renovate cycle, this will automatically pin the action in `action.yml` from a tag reference (`@v8.1.0`) to a commit SHA — addressing the supply-chain concern in #5020.

**Technical check:**
- ✅ Rule is syntactically valid (`matchManagers`, `matchPackageNames`, `pinDigests` are all correct)
- ✅ The shared config (`github>ansible/actions//config/renovate.json`) does NOT already provide this, so the local override is necessary
- ✅ All CI checks pass

**Minor items before merge:**

1. **Branch is behind `main`** — a rebase is needed (no conflicts expected).
2. **Commit prefix** — the commit uses `feat:` but this addresses a bug (#5020), so `fix:` would be more appropriate per conventional commits.
3. **PR body** — still references v7 and the original approach. The current Renovate-based approach is only explained in comments. Consider updating the body for future reference.

None of these block the functionality — the Renovate rule itself is correct and ready to work once merged.

cidrblock (approved):
Thanks, please take care of the nit-pick things :)

garethahealy:
@cidrblock done

garethahealy:
@cidrblock - CI green, good to merge?

rockygeekz (approved):

rockygeekz:
Thanks @garethahealy for the fix. Will merge it once the CI is green.

garethahealy:
> Thanks @garethahealy for the fix. Will merge it once the CI is green.

The CI isn't testing this change, and I don't believe the current failures are due to my change...

garethahealy:
@rockygeekz good to merge?

PR #5023, opened 2026-04-22
Fix formatting issues in jinja.md documentation

mehranasghari:


felixfontein:
I think this change is incorrect. [black](https://pypi.org/project/black/) is the name of a Python formatter that's apparently internally used by ansible-lint. `block` doesn't make sense in these contexts.

rockygeekz:
> I think this change is incorrect. [black](https://pypi.org/project/black/) is the name of a Python formatter that's apparently internally used by ansible-lint. `block` doesn't make sense in these contexts.

True. The current text is correct. Closing this.

PR #5026, opened 2026-04-25
fix: avoid name[casing] auto-fix crash on multi-segment prefixes

bishalOps:
## Summary

`ansible-lint --fix` crashed with `ValueError: too many values to unpack` on task/play names containing more than one `|` (the documented multi-segment prefix convention, e.g. `foo | destroy | Terminate server` https://docs.ansible.com/projects/lint/rules/name/#nameprefix). Switch the `name[casing]` auto-fix from `split("|")` to `rsplit("|", 1)` so the prefix is preserved and only the description is capitalized.
Added a test fixture mirroring the docs' `foo | destroy | …` example to lock in the fix.

  ## Reproducer 

Without the fix:

  ```yaml
  # cpair/task/base/tool.yaml
  - name: tool | base | install something
    hosts: localhost
    tasks:
      - name: Install package
        ansible.builtin.dnf:
          name: something
          state: present

  $ ansible-lint --fix cpair/task/base/tool.yaml
  ERROR    Rule specific fix failed for: name[casing]/play ...
  ERROR    too many values to unpack (expected 2)
  File ".../ansiblelint/rules/name.py", line 277, in update_task_name
      [file_name, update_task_name] = task_name.split("|")
  ValueError: too many values to unpack (expected 2)
```

After the fix:

  ```yaml
# cpair/task/base/tool.yaml
  - name: tool | base | Install something
    hosts: localhost
    tasks:
      - name: Install package
        ansible.builtin.dnf:
          name: something
          state: present
```


felixfontein:
I just created a similar fix for a similar bug in another rule (#5028). Maybe someone should check all `split()` calls to make sure this doesn't happen more often.

(I'm surprised that there isn't a linting rule for this already, since it's a pretty common error I have seen way too often in too many different projects...)

felixfontein:
It seems CI has been fixed, when I rebased my PR against `main` (don't forget to re-sign your commits) CI finally passed.

rockygeekz (approved):

Issue #4543, opened 2025-03-06
action-core plugin seems to remove empty lines after tasks if free_form style is the last parameter of a task


oveee92:
##### Summary

Free-form tasks, which don't use any parameters, will remove empty lines after the task "block", if the action-core plugin is run with `--fix`.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 25.1.3 using ansible-core:2.17.9 ansible-compat:25.1.4 ruamel-yaml:0.18.10 ruamel-yaml-clib:0.2.12
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

```console (paste below)
mkdir -p roles/foo/tasks
cat << EOF > roles/foo/tasks/main.yml
---
- name: Freeform1
  command: echo hello

- name: Freeform2
  command: echo hello

- name: Freeform2
  command: echo hello

EOF
ansible-lint  roles/foo/tasks/main.yml --fix
cat  roles/foo/tasks/main.yml
```

##### Desired Behavior

Ansible-lint should fix the FQCN.

```paste below
---
- name: Freeform1
  ansible.builtin.command: echo hello

- name: Freeform2
  ansible.builtin.command: echo hello

- name: Freeform2
  ansible.builtin.command: echo hello

```

##### Actual Behavior

Ansible-lint fixes the FQCN, but also removes any newlines.

```paste below
---
- name: Freeform1
  ansible.builtin.command: echo hello
- name: Freeform2
  ansible.builtin.command: echo hello
- name: Freeform2
  ansible.builtin.command: echo hello

```

However, this only happens when "fixing" it. If I now manually add back the newlines and run ansible-lint with `--fix` again, it won't remove the newlines, so I assume it is related to the "fix" function.

It also only seems to happen if the free_form part is the last line of the task.
I.e. the following will work as it should:

```
cat << EOF > roles/foo/tasks/main.yml
---
- name: Freeform1
  command: echo hello
  tags: []

- name: Freeform2
  command: echo hello
  tags: []

- name: Freeform2
  command: echo hello
  tags: []

EOF

ansible-lint roles/foo/tasks/main.yml --fix

cat roles/foo/tasks/main.yml
---
- name: Freeform1
  ansible.builtin.command: echo hello
  tags: []

- name: Freeform2
  ansible.builtin.command: echo hello
  tags: []

- name: Freeform2
[...]

PR #5027, opened 2026-04-26
fix: preserve trailing blank lines when fqcn auto-fix renames a key

bishalOps:
`ansible-lint --fix for fqcn[*]` was dropping blank lines between tasks, and also trailing comments, when the renamed key was the last key in a task mapping (#4543). In `ruamel.yaml`, blank-line and trailing-comment metadata are stored in `ca.items` keyed by the original mapping key. The transform removes and reinserts the renamed key, but does not rebind `ca.items`, so the metadata is left orphaned under the old key name and the YAML emitter drops it. This fix migrates `ca.items[old_key]` to `ca.items[new_key]` after renaming so the formatting metadata is preserved during emission


rockygeekz:
@bishalOps Thanks for the PR. LGTM!

rockygeekz (approved):

Issue #5013, opened 2026-04-08
mock_modules generates mock modules that fail ansible-lint


richm:
##### Summary

I have a problem with Automation Hub gating which (afaict) uses ansible-lint 24.12.2, ansible-core 2.16, and python 3.12 - and using `mock_modules` in .ansible-lint - ansible-lint generates a mocked python module which fails ansible-lint - steps to reproduce:

```
# python3.12 -mvenv .venv
# . .venv/bin/activate
# pip install ansible-lint==24.* ansible-core==2.16.*
# ansible-galaxy collection init myns.mycoll
# cd myns/mycoll
# cat .ansible-lint
mock_modules:
  - some_ns.some_coll.some_module
# ansible-lint
load-failure: Failed to load or parse file.
.ansible/collections/ansible_collections/some_ns/some_coll/plugins/modules/some_module.py:14

yaml[new-line-at-end-of-file]: No new line character at the end of file
.ansible/collections/ansible_collections/some_ns/some_coll/plugins/modules/some_module.py:14
```

afaict, the problem is the generated `EXAMPLES = '''mocked'''` and `RETURN = '''mocked'''` - these are being flagged as invalid YAML by rules init.py matchyaml -

```
        # yaml returned can be an AnsibleUnicode (a string) when the yaml
        # file contains a single string. YAML spec allows this but we consider
        # this an fatal error.
```

I don't know if the v24 branch is being maintained upstream but this is the version being used by Automation Hub.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
24.12.2
ansible --version
2.16.2
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

see above

##### Desired Behavior

Should be able to use `mock_modules` and ansible-lint will pass.

##### Actual Behavior

See above.

santosh7676:
Hi, I was able to reproduce this issue locally using the steps provided.

From debugging, it looks like the generated mock module (`.py` file under `.ansible/collections/...`) is being incorrectly classified as `text/yaml`, which causes the YAML rule (yamllint) to run on it and fail.

The root cause appears to be in `kind_from_path()`, where file type detection relies on pattern matching and may classify these generated `.py` files as YAML.

I initially tested a fix that classified all `.py` files as Python, but that caused regressions (e.g., Ansible plugin modules under `plugins/modules/` were no longer detected correctly).

[...]

PR #5032, opened 2026-04-28
Fix: fix mock_modules generated stubs failing YAML/doc parsing #5031

santosh7676:
Fix `mock_modules` generated stubs failing YAML/doc parsing

This MR fixes a bug where `mock_modules` generated Python stubs could fail ansible-lint checks due to invalid module-doc block content.

### Problem
When using `mock_modules`, generated stubs used:

- `EXAMPLES = '''mocked'''`
- `RETURN = '''mocked'''`

In some environments/paths, these values triggered YAML/module-doc parsing failures and caused lint errors on generated mock module files.

### Fix
  - Updated mocked module template to:
  - use valid YAML content in `EXAMPLES`
  - use the no-return sentinel for `RETURN` (`r"""` + `# """`)

### Tests
- Added a regression test in `test/test_mockings.py` to cover the `mock_modules` scenario and ensure generated stubs are lintable.

#5013


santosh7676:
Looks like CI Issue, it is failing in test_import_tasks[1] (Runner lintables count for examples/playbooks/test_import_with_malformed.yml), in both py314-devel and py312-devel jobs.

This PR only changes:
- src/ansiblelint/constants.py (mocked module template docs)
- test/test_mockings.py (mock_modules regression test)

So this appears unrelated/pre-existing.

felixfontein:
Yep, the CI seems to be broken at the moment...

felixfontein:
It seems CI has been fixed, when I rebased my PR against `main` (don't forget to re-sign your commits) CI finally passed.

santosh7676:
Thank you @felixfontein for the details, I will try with rebasing.

rockygeekz:
@santosh7676  Thanks for the fix! The test uses offline=False. Is that intentional? Every other test uses the default offline=True. Mock modules generate local stubs, so network access shouldn't be needed. Could you switch to offline=True (or just remove the parameter) to keep the test deterministic?

santosh7676:
Good catch @rockygeekz, offline=False wasn't intentional here; it was copied from the mock_roles test. Mock modules generate local stubs before prepare_environment, so network access isn't needed. Switched to the default offline=True.

rockygeekz (approved):
Thanks. LGTM!

PR #5033, opened 2026-04-29
fix: preserve multi-hash comments on `ansible-lint --fix`

bishalOps:
Found another bug while using `ansible-lint --fix` where any comment starting with ## (or ###, etc.) was getting mangled into # #. I swapped the existing logic for a small regex that just adds a space after the leading run of #s when one's missing, and otherwise leaves things alone.


rockygeekz:
@bishalOps Thanks for the fix!

sathyapramod (approved):

PR #5045, opened 2026-05-16
fix: support example format indicator, prevent ansible-lint from producing load-failure on valid non-YAML examples

### LINKED ISSUE #5037: load-failure returned for module with non-YAML examples
##### Summary

ansible-lint apparently tries to interpet every module's `EXAMPLES` as YAML.

This is generally wrong, and `ansible-test sanity` and other tools (antsibull-nox, antsibull-docs) know about the `# fmt: xxx` specifier at the top of `EXAMPLES` to handle non-YAML data correctly. ansible-lint apparently does not know about this.

The most known examples of `# fmt: xxx` are the `ansible.builtin.ini`, `ansible.builtin.script`, and `ansible.builtin.toml` plugin `EXAMPLES`. I encountered this in community.general with the ohai module, after adding `# fmt: console` ~~(the version in git doesn't have that yet)~~.

(This was introduced in https://github.com/ansible/ansible/pull/71184.)

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
latest pip version
```

##### STEPS TO REPRODUCE

Run ansible-lint on a module with
```
EXAMPLES = r"""
# fmt: console

ansible webservers -m ohai --tree=/tmp/ohaidata
...
"""
```

##### Desired Behavior

Does not result in load-failure because of the examples.

##### Actual Behavior

```
load-failure: Failed to load or parse file. (warning)
plugins/modules/ohai.py:30
```

felixfontein:
Support example format indicator for Ansible plugin/module documentation. This prevents ansible-lint from producing load-failure on valid non-YAML examples that are tagged correctly as non-YAML.

Fixes #5037.

### DIFF
--- a/examples/.collection/plugins/modules/tests/gamma.py
+++ b/examples/.collection/plugins/modules/tests/gamma.py
@@ -28,10 +28,9 @@ notes:
 """
 
 EXAMPLES = """
-- name: test task-1
-  company_name.coll_1.mod_1:
-    foo: some value
-    bar: candidate
+# fmt: text
+
+This is some text. Nothing to see.
 """
 
 RETURN = """
--- a/src/ansiblelint/runner.py
+++ b/src/ansiblelint/runner.py
@@ -662,9 +662,9 @@ class Runner:
 
     def plugin_children(self, lintable: Lintable) -> list[Lintable]:
         """Collect lintable sections from plugin file."""
-        offset, content = parse_examples_from_plugin(lintable)
-        if not content:
-            # No examples, nothing to see here
+        offset, content, content_type = parse_examples_from_plugin(lintable)
+        if not content or content_type != "yaml":
+            # No (YAML) examples, nothing to see here
             return []
         examples = Lintable(
             name=lintable.name,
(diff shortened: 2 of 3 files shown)


rockygeekz (changes requested):
Heyy @felixfontein , Thanks for the fix. could you add a unit test for the #fmt: detection in parse_examples_from_plugin?

felixfontein:
`parse_examples_from_plugins` looks like it should better be integration-tested, since it's a pretty complex function (after all it calls `read_docstring`). I don't think it makes sense to add unit tests, resp. it will be quite some work I don't want to invest.

rockygeekz:
> `parse_examples_from_plugins` looks like it should better be integration-tested, since it's a pretty complex function (after all it calls `read_docstring`). I don't think it makes sense to add unit tests, resp. it will be quite some work I don't want to invest.

Cool. Makes sense. Thanks for the PR!

rockygeekz:
Please make sure to add description with explanation of the changes.

felixfontein:
I'm not sure, but isn't the PR's title already containing enough information?

rockygeekz:
The title is good, but a short description in the body helps with changelog generation and gives future reviewers context without reading the diff. Even 2-3 lines would be great. Once that's in, we can merge this.

felixfontein:
Done.

rockygeekz (approved):

felixfontein:
@rockygeekz thanks for reviewing and merging!

PR #5055, opened 2026-05-29
fix: resolve nested include_tasks relative paths

### LINKED ISSUE #5051: `load-failure[filenotfounderror]` for an `include_tasks` whose `file:` is a relative path
##### Summary

`ansible-lint` raises a fatal `load-failure[filenotfounderror]` for an `include_tasks` whose `file:` is a relative path like `../tasks/state_update.yml`.

The include works at runtime (both via `ansible-playbook` and `molecule prepare` and used in production by all of [our collections](https://github.com/orgs/foundata/repositories?q=ansible-collection-*), e.g. [foundata.acmesh](https://github.com/foundata/ansible-collection-acmesh/blob/main/extensions/molecule/default/tasks/prepare/all.yml#L356)), but `ansible-lint` produces a path that doesn't exist on disk and fails the whole run.

The reported path depends on where ansible-lint is invoked from, which on its own points at a path-normalization issue:

| CWD | Argument | Reported (non-existent) path |
|---|---|---|
| inside collection root | absolute path to collection | `<path above collection_root>/tasks/state_update.yml` (one level *above* the collection) |
| inside collection root | (no argument) | same as above |
| anywhere outside an Ansible project | absolute path to collection | `/tasks/state_update.yml` (root-anchored, clearly bogus) |

In the third case ansible-lint also prints `WARNING  Project directory /.ansible cannot be used for caching as it is not writable.` (I guess the detected project directory is empty/`/`, which then leaks into the relative-path resolution and produces a "/"-anchored path that exists nowhere in the input),

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
$ ansible-lint --version
ansible-lint 26.4.0 using ansible-core:2.21.0 ansible-compat:26.3.0 ruamel-yaml:0.19.1 ruamel-yaml-clib:None

$ python --version
Python 3.14.4

$ head -3 /etc/os-release
NAME="Fedora Linux"
VERSION="44 (Workstation Edition)"
ID=fedora
```

- ansible installation method: pip (inside a venv)
- ansible-lint installation method: pip (inside a venv)

##### STEPS TO REPRODUCE

A minimal repro is attached as a tarball: [ansible-lint-load-failure-repro.tar.gz](https://github.com/user-attachments/files/28194257/ansible-lint-load-failure-repro.tar.gz)

Layout:

```
repro/
├── galaxy.yml
├── README.md
├── meta/runtime.yml
├── changelogs/changelog.yaml
└── extensions/molecule/
    ├── default/
    │   ├── molecule.yml
    │   └── tasks/prepare/all.yml          # include_tasks: file: "../tasks/state_update.yml"
    └── shared/
        ├── playbooks/prepare.yml          # the prepare playbook; includes default/tasks/prepare/all.yml
        └── tasks/state_update.yml         # the file the include actually finds at runtime
```

`extensions/molecule/default/tasks/prepare/all.yml`:

```yaml
---
- name: "Init state via shared task"
  ansible.builtin.include_tasks:
    file: "../tasks/state_update.yml"
  vars:
    state_update:
      example_key: "example_value"
```

`all.yml` is included by `extensions/molecule/shared/playbooks/prepare.yml`, so at runtime the relative path is resolved against the *playbook's* directory (`shared/playbooks/`) and correctly finds `extensions/molecule/shared/tasks/state_update.yml` (so standard Ansible behavior for `include_tasks`).

I did the following to confirm the runtime is fine with the minimal example:

```console
$ cd /tmp/repro
$ ansible-playbook -i localhost, extensions/molecule/shared/playbooks/prepare.yml
...
TASK [Include scenario prepare/all.yml] ****
included: /tmp/repro/extensions/molecule/default/tasks/prepare/all.yml for localhost

TASK [Init state via shared task] ****
included: /tmp/repro/extensions/molecule/shared/tasks/state_update.yml for localhost

TASK [Shared | Merge update into state] ****
ok: [localhost]

PLAY RECAP ****
localhost : ok=3  changed=0  unreachable=0  failed=0  ...
```

```console
$ cd /tmp/repro
$ export MOLECULE_GLOB="./extensions/molecule/*/molecule.yml"
$ molecule prepare --scenario-name default
...
TASK [Init state via shared task] ****
included: /tmp/repro/extensions/molecule/shared/tasks/state_update.yml for localhost
...
INFO     default ➜ prepare: Executed: Successful
INFO     Molecule executed 1 scenario (1 successful)
```

Both runs find the file at exactly the path ansible-lint claims is missing.

**Now lint the same tree (and look at the path which was tried to load**):

```console
$ cd /tmp/repro
$ ansible-lint --profile production
[...]
load-failure[filenotfounderror]: [Errno 2] No such file or directory: '/tmp/tasks/state_update.yml' (warning)
/tmp/tasks/state_update.yml:1
[...]

$ cd /tmp                       # a directory outside an Ansible project
$ ansible-lint --profile production --strict /tmp/repro
[...]
load-failure[filenotfounderror]: [Errno 2] No such file or directory: '/tasks/state_update.yml' (warning)
/tasks/state_update.yml:1
[...]
```

One more thing worth noting: if the collection is sitting *inside* another Ansible project (so ansible-lint walks up and detects a parent project root), the bug does not trigger in every case (I did not fully understand what causes it to not trigger sometimes then; or I got to many tabs open, so take it with a grain of salt). It showed up 100% of the time when there's no enclosing project, so exactly the situation when you lint a freshly generated collection in a tmp dir.

##### Desired Behavior

Either:

1. Resolve relative `include_tasks` paths the way Ansible does at runtime (which would match what `ansible-playbook` and `molecule prepare` already show), or
2. Don't emit `load-failure` for include targets whose resolution is ambiguous from a static view.

The `/tasks/state_update.yml` (leading-slash) from `ansible-lint` output tells me (but I am not to deep into the sources) that something has gone wrong with path construction, no such path exists anywhere in the input, and it shouldn't be possible to produce it from `../tasks/state_update.yml` without losing the base directory somewhere along the way?

##### Actual Behavior

```console
$ cd /tmp/repro
$ ansible-lint --profile production
[...]
load-failure[filenotfounderror]: [Errno 2] No such file or directory: '/tmp/tasks/state_update.yml' (warning)
/tmp/tasks/state_update.yml:1
[...]

$ cd /tmp                       # a directory outside an Ansible project
$ ansible-lint --profile production --strict /tmp/repro
[...]
load-failure[filenotfounderror]: [Errno 2] No such file or directory: '/tasks/state_update.yml' (warning)
/tasks/state_update.yml:1
[...]
```
Run from inside the collection root, the reported path becomes `<path above collection_root>/tasks/state_update.yml` instead, still pointing outside the project, but at least not root-anchored. The file actually lives at `/tmp/repro/extensions/molecule/shared/tasks/state_update.yml`.

`ansible-playbook --syntax-check` on the prepare playbook is clean, and `molecule prepare` runs through without an include error (see the runs above).

##### Related

Probably the following ones:

- #1446 (closed) — False positive load-failure when using
  import_tasks / include_tasks
- #2536 (closed) — Linter parses third-party content installed outside
  of the project directory
- #3674 (closed) — include_tasks in molecule tests fails to lint

cyphercodes:
## Summary
- resolve nested task include paths using the parent include/playbook context before falling back to the historical upward search
- add a regression test for include_tasks: file: ../tasks/state_update.yml resolving from an included task file

Fixes #5051

## Tests
- uv run pytest test/test_task_includes.py test/test_import_tasks.py test/test_import_playbook.py test/test_file_path_evaluation.py test/test_include_miss_file_with_role.py
- uv run ruff check src/ansiblelint/utils.py test/test_file_path_evaluation.py
- uv run ruff format --check src/ansiblelint/utils.py test/test_file_path_evaluation.py

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resolution of task include paths within nested playbooks and task files.
  * Prevented false `load-failure` errors when included files use relative paths.
  * Preserved parent playbook context when processing nested includes.

* **Tests**
  * Added coverage for nested include paths in molecule-style Ansible project layouts.


rockygeekz (changes requested):
Thanks for the fix! The code and approach look good. One thing blocking merge: the commit isn't signed commit?

Happy to approve once that's resolved.

shvenkat-rh:
Closed here https://github.com/ansible/ansible-lint/pull/5159

PR #5058, opened 2026-06-03
Fix risky-shell-pipe false positive on multi-line Jinja

### LINKED ISSUE #3879: risky-shell-pipe false positive: pipe operator in multi-line Jinja template is interpreted as shell pipe
##### Summary

Using Jinja template with operator `|` in __multi-line__ (!) shell `cmd` string is recognized as *risky-shell-pipe* violation.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint --version
ansible-lint 6.21.1 using ansible-core:2.14.11 ansible-compat:4.1.10 ruamel-yaml:0.17.40 ruamel-yaml-clib:0.2.8
```

##### Actual / Desired Behavior

```ansible
- vars:
    values:
      - 'itemA'
      - 'itemB'
  ansible.builtin.shell:
    executable: '/usr/bin/bash'
    cmd: >-
      echo {{
        values | first
      }}
```
fails while it's not supposed to
```shell
risky-shell-pipe: Shells that use pipes should set the pipefail option.
tasks/example.yaml:1 Task/Handler: shell executable=/usr/bin/bash cmd=echo {{
  values | first
}}
```

BUT

```ansible
- vars:
    values:
      - 'itemA'
      - 'itemB'
  ansible.builtin.shell:
    executable: '/usr/bin/bash'
    cmd: >-
      set -o pipefail;
      echo {{
        values | first
      }}
```
and 
```
- vars:
    values:
      - 'itemA'
      - 'itemB'
  ansible.builtin.shell:
    executable: '/usr/bin/bash'
    cmd: >-
      echo {{ values | first }}
```
succeed

While the result of the last 2 examples is expected, the first one is not. If pipe operators are only part of a Jinja template in the `cmd` string value the rule should not match regardless whether it's multi-line or single-line.

arpitjain099:
## Summary

`risky-shell-pipe` flags a command that has no real shell pipe when the only `|` lives inside a multi-line Jinja expression, for example:

```yaml
- vars:
    values:
      - itemA
      - itemB
  ansible.builtin.shell:
    executable: /usr/bin/bash
    cmd: >-
      echo {{
        values | first
      }}
```

The single-line form `echo {{ values | first }}` is correctly ignored, but spreading the same expression over several lines produces a false positive.

## Root cause

The rule strips Jinja out of the command with `unjinja()` before searching for a bare pipe. `unjinja()` uses `RE_JINJA_EXPRESSION` (and the statement/comment variants) which were compiled as `r"{{.+?}}"` without `re.DOTALL`. Because `.` does not match a newline by default, a multi-line `{{ ... | ... }}` block is never substituted, so its inner `|` survives into the string that `_pipe_re` scans and the rule fires.

## Fix

Compile the three Jinja regexes with `re.DOTALL` so `.` also spans newlines and a multi-line expression, statement, or comment is fully stripped before the pipe check. The non-greedy `.+?` still stops at the first closing delimiter, so separate expressions on the same line are not merged together.

A genuine risky pipe is unaffected: in `cat foo | grep bar` the `|` sits outside any Jinja block, so `unjinja()` leaves it in place and the rule still reports it. The shared `unjinja()` helper also backs `command-instead-of-shell`; for that rule DOTALL likewise only makes the stripping more complete and cannot turn a real match into a miss.

## Tests

Added a multi-line-Jinja case to `examples/playbooks/rule-risky-shell-pipe-pass.yml`. It fails before this change (the false positive is reported, so the pass fixture expects 0 but gets 1) and passes after it. The existing fail fixture is unchanged, confirming genuine pipes without `pipefail` still trigger the rule.

```
src/ansiblelint/rules/risky_shell_pipe.py::test_risky_shell_pipe[pass] PASSED
src/ansiblelint/rules/risky_shell_pipe.py::test_risky_shell_pipe[fail] PASSED
```

Fixes #3879.

Thanks to @ssbarnea for confirming on the issue that a PR for this would be welcome.


rockygeekz (approved):
Thanks for the fix. LGTM!

PR #5060, opened 2026-06-05
fix: ignore skip lookup across rules

### LINKED ISSUE #5059: bug: _rule_is_skipped returns False on first non-matching rule instead of continuing iteration
#### Summary

`_rule_is_skipped` in `src/ansiblelint/__main__.py` (line 289) has a logic error: it returns `False` on the first non-matching rule in the set instead of continuing to check remaining rules. This causes non-deterministic behavior when a file has multiple entries in `.ansible-lint-ignore` and at least one uses the `skip` qualifier.

Because `rules` is a `set` (unordered, hash-randomized in Python 3.6+), if the first rule yielded by iteration doesn't match the tag, the function immediately returns `False` — even if a later rule in the set does match and has the `skip` qualifier.

#### Issue Type

- Bug Report

#### OS / ENVIRONMENT

```console
ansible-lint --version
ansible-lint 26.3.0 using ansible-core:2.20.3 ansible-compat:25.12.1 ruamel-yaml:0.19.1 ruamel-yaml-clib:None
```

Bug is in source code and is OS/environment independent. Affects current `main` branch.

#### STEPS TO REPRODUCE

1. Create a test file with a violation:

```yaml
# test.yml
bad_indentation:
- blah: plop
  zz: 42
```

2. Create an `.ansible-lint-ignore` with multiple entries for the same file, one using `skip`:

```text
test.yml yaml[indentation] skip
test.yml yaml[truthy]
```

3. Run:

```console
ansible-lint --strict test.yml
```

4. Run the command multiple times (or with different `PYTHONHASHSEED` values) to observe non-deterministic results:

```console
PYTHONHASHSEED=1 ansible-lint --strict test.yml
PYTHONHASHSEED=2 ansible-lint --strict test.yml
```

#### Desired Behavior

The `yaml[indentation]` violation should always be skipped (exit code 0 for that rule) regardless of how many other entries exist for the same file in `.ansible-lint-ignore`. The one-word fix is `return False` → `continue` on line 292:

```python
def _rule_is_skipped(tag: str, rules: set[IgnoreRule]) -> bool:
    for rule in rules:
        if tag != rule.rule:
            continue  # <-- currently `return False`
        return IgnoreRuleQualifier.SKIP in rule.qualifiers
    return False
```

#### Actual Behavior

Depending on Python's hash seed (set iteration order), `_rule_is_skipped` may hit the `yaml[truthy]` entry first, see that the tag doesn't match, and immediately `return False` — never checking the `yaml[indentation] skip` entry. This causes:

- The `yaml[indentation]` violation is kept in results when it should be pruned
- With `--strict`, exit code is incorrectly `2` instead of `0`
- Behavior is **non-deterministic** across runs, making it extremely hard to debug

This also compounds with PR #5001 which relies on `_rule_is_skipped` correctly pruning matches before the `--strict` exit code evaluation.

```text
Current code trace with tag="yaml[indentation]":

  rules = {IgnoreRule("yaml[truthy]", {}), IgnoreRule("yaml[indentation]", {SKIP})}

  Case A - yaml[indentation] iterated first:
    tag matches → return True ✓ (correct)

  Case B - yaml[truthy] iterated first:
    tag != rule.rule → return False ✗ (wrong, never checks second rule)
```

Flagged during review of PR #5001. The bug pre-dates that PR.

mehrdadbn9:
Fixes #5059.

## Summary
- continue scanning ignore rules when the current entry does not match the lint tag
- add a regression test where the matching skip rule appears after an unrelated rule

## Testing
- uv run --no-dev --with ruff ruff check src/ansiblelint/__main__.py test/test_app.py
- uv run --no-dev --with pytest --with mypy python -m pytest test/test_app.py -k 'rule_is_skipped or ignore_file_with_skip_and_strict or ignore_file_without_skip_and_strict'

### DIFF
--- a/src/ansiblelint/__main__.py
+++ b/src/ansiblelint/__main__.py
@@ -70,8 +70,9 @@ from ansiblelint.skip_utils import normalize_tag
 from ansiblelint.version import __version__
 
 if TYPE_CHECKING:
-    # RulesCollection must be imported lazily or ansible gets imported too early.
+    from collections.abc import Iterable
 
+    # RulesCollection must be imported lazily or ansible gets imported too early.
     from ansiblelint.rules import RulesCollection
     from ansiblelint.runner import LintResult
 
@@ -286,10 +287,10 @@ def fix(runtime_options: Options, result: LintResult, rules: RulesCollection) ->
 # to the rule, it is treated as skipped here and does not show up
 # even as a warning.
 # [1] https://github.com/ansible/ansible-lint/issues/3068
-def _rule_is_skipped(tag: str, rules: set[IgnoreRule]) -> bool:
+def _rule_is_skipped(tag: str, rules: Iterable[IgnoreRule]) -> bool:
     for rule in rules:
         if tag != rule.rule:
-            return False
+            continue
         return IgnoreRuleQualifier.SKIP in rule.qualifiers
     return False
 
--- a/test/test_app.py
+++ b/test/test_app.py
@@ -2,8 +2,10 @@
 
 from pathlib import Path
 
+from ansiblelint.__main__ import _rule_is_skipped
 from ansiblelint.constants import RC
 from ansiblelint.file_utils import Lintable
+from ansiblelint.loaders import IgnoreRule, IgnoreRuleQualifier
 from ansiblelint.testing import run_ansible_lint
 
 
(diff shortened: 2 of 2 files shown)


rockygeekz (approved):

Hrithik-Gavankar (approved):
Changes LGTM!

PR #5077, opened 2026-06-15
fix: suppress ruff PLW0717 to unblock renovate

rockygeekz:
## Summary
- ruff 0.15.17 enables PLW0717 (try-too-many-statements) by default, 8 violations in the codebase
- Added PLW0717 to the ruff ignore list — same as what was done in vscode-ansible (#2719)
- Includes the renovate lock file update so CI picks up ruff 0.15.17 which recognizes the rule

## Test plan
- [ ] CI passes (lint-pkg-schemas-eco specifically)

### DIFF
--- a/.config/requirements-lock.txt
+++ b/.config/requirements-lock.txt
@@ -14,17 +14,17 @@ cffi==2.0.0
     # via
     #   ansible-lint
     #   cryptography
-click==8.4.0
+click==8.4.1
     # via black
 colorama==0.4.6 ; sys_platform == 'win32'
     # via click
-cryptography==48.0.0
+cryptography==49.0.0
     # via
     #   ansible-core
     #   ansible-lint
 distro==1.9.0
     # via ansible-lint
-filelock==3.29.0
+filelock==3.29.4
     # via ansible-lint
 jinja2==3.1.6
     # via ansible-core
@@ -49,7 +49,7 @@ pathspec==1.1.1
     #   ansible-lint
     #   black
     #   yamllint
-platformdirs==4.9.6
+platformdirs==4.10.0
     # via black
 pycparser==3.0 ; implementation_name != 'PyPy'
     # via cffi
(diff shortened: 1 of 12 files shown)


cidrblock (changes requested):
Thanks for catching this @rockygeekz — keeping renovate unblocked is appreciated.

However, I'd prefer we **not globally suppress PLW0717**. The rule exists to prevent oversized try blocks, and we should keep it active so new violations don't accumulate silently.

Looking at the actual violations:

| File | Statements | Verdict |
|------|-----------|---------|
| `schemas/main.py:52` | 11 | **Refactor** — does 3 unrelated things (data loading, validation, URL extraction) in one try |
| `utils.py:1205` | 6 | Tightly coupled loader setup — `# noqa: PLW0717` with justification |
| `loaders.py:92` | 5 | **Refactor** — trivial extract to a helper |
| `types.py:15` | 5 | Standard import-compat shim — `# noqa: PLW0717` |
| `runner.py:310` | 4 | Fallback pattern — `# noqa: PLW0717` |
| `rules/args.py:203` | 4 | Atomic mock setup — `# noqa: PLW0717` |

Suggested approach:
1. Refactor `schemas/main.py` (the worst offender — extract helpers for the distinct phases).
2. Refactor `loaders.py` (trivial one-liner extract).
3. Use targeted `# noqa: PLW0717` on the remaining ~4 sites where the pattern is intentional and splitting would hurt readability.

This keeps the rule enforced for **new** code while documenting why existing sites are exceptions.

---

Also — the summary references vscode-ansible #2719 as precedent, but vscode-ansible is TypeScript. Could you link the actual PR? I'm not sure that's an apples-to-apples comparison for a Python lint rule.

rockygeekz:
Makes sense, thanks for the detailed breakdown! I'll refactor the worst offenders and use targeted # noqa: PLW0717 on the rest where splitting would hurt readability. Will push the updated approach shortly.

vscode-ansible #2719  that repo also has Python (ansible-language-server), and PLW0717 was added to its ruff ignore there. But agreed this is a better approach than a blanket suppress.

cidrblock (approved):

Issue #3650, opened 2023-08-09
ansible-lint expects "localhost," file to be present if using unparsed_is_failed in ansible.cfg


damosc:
##### Summary

With `unparsed_is_failed = True` in my ansible.cfg, `ansible-lint` calls `ansible-playbook -i localhost, --syntax-check`. It's probably caused by https://github.com/ansible/ansible-lint/blob/df0c46dcc53ab46156016f14fd80bebad9545bfb/src/ansiblelint/runner.py#L332

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
ansible-lint --version
ansible-lint 6.17.2 using ansible-core:2.15.2 ansible-compat:4.1.2 ruamel-yaml:None ruamel-yaml-clib:None
```

- ansible installation method: pip
- ansible-lint installation method: pip

##### STEPS TO REPRODUCE

Add `unparsed_is_failed = True` to your ansible.cfg and execute any ansible-lint command.

##### Desired Behavior

Not sure what should happen, but for sure ansible-lint cannot expect localhost, (with comma) inventory file to be present.

##### Actual Behavior

```
WARNING  Skipped installing old role dependencies due to running in offline mode.
WARNING  Skipped installing collection dependencies due to running in offline mode.
WARNING  Skipped installing old role dependencies due to running in offline mode.
WARNING  Skipped installing collection dependencies due to running in offline mode.
WARNING  Listing 1 violation(s) that are fatal
internal-error: Unexpected error code 1 from execution of: ansible-playbook -i localhost, --syntax-check local.yml
local.yml:1 [WARNING]: Unable to parse localhost, as an inventory source
ERROR! No inventory was parsed, please check your configuration and options.


Read documentation for instructions on how to ignore specific rule violations.

              Rule Violation Summary               
 count tag            profile rule associated tags 
     1 internal-error min     core                 

Failed: 1 failure(s), 0 warning(s) on 1 files.
```

```[tasklist]
### Tasks
- [ ] https://github.com/ansible/ansible-lint/pull/3680
```

PR #5091, opened 2026-06-30
fix: propagate extra_vars to import_playbook syntax check

### LINKED ISSUE #5042: extra_vars not propagated to syntax check when resolving import_playbook: references
##### Summary

`ansible-lint`'s `extra_vars` (declared in `.ansible-lint.yml`) are honored when linting a playbook directly, but they are **not** honored when the same playbook is reached transitively via `import_playbook:` from another playbook.

In the transitive case, `ansible-lint` logs `ERROR  Failed to load <name> playbook due to failing syntax check.` and silently skips linting the imported playbook. The lint summary still says `Passed: 0 failure(s)` and the process exits 0, which makes the failure easy to miss in CI.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console
$ ansible-lint --version
ansible-lint 26.4.0 using ansible-core:2.18.16 ansible-compat:26.3.0 ruamel-yaml:0.19.1 ruamel-yaml-clib:0.2.15
```

- OS: macOS 14 / Linux (reproduces on both)
- Python: 3.11

##### STEPS TO REPRODUCE

`.ansible-lint.yml`
```yaml
extra_vars:
  my_host: localhost
```

`playbooks/inner.yml`
```yaml
---
- name: Inner play
  hosts: "{{ my_host }}"
  gather_facts: false
  tasks:
    - name: Noop
      ansible.builtin.debug:
        msg: hi
```

`playbooks/outer.yml `
```yaml
---
- name: Import inner
  ansible.builtin.import_playbook: inner.yml
```

Then run both lint commands and compare:

```console
$ ansible-lint --offline playbooks/inner.yml
$ echo $?
0

$ ansible-lint --offline playbooks/outer.yml
ERROR    Failed to load inner.yml playbook due to failing syntax check.
$ echo $?
0
```

The same `inner.yml` is lintable on its own, but is silently dropped from linting (and not counted as a failure) when reached via `import_playbook:`, even though `extra_vars` is configured.

##### Desired Behavior

A playbook should lint the same way whether I run `ansible-lint` on it directly or on another playbook that imports it. Concretely, given the reproducer above I expect:

```console
$ ansible-lint --offline playbooks/outer.yml
Passed: 0 failure(s), 0 warning(s) in 2 files processed of 2 encountered.
$ echo $?
0
```

— the same result I get from `ansible-lint --offline playbooks/inner.yml`, because the `extra_vars` from `.ansible-lint.yml` should apply to both.

##### Actual Behavior

`extra_vars` from `.ansible-lint.yml` are applied when `inner.yml` is linted directly, but not when it is reached via `import_playbook:` from `outer.yml`. In the latter case the lint output is:

```
ERROR    Failed to load inner.yml playbook due to failing syntax check.
Passed: 0 failure(s), 0 warning(s) in 1 files processed of 1 encountered.
```

and the process exits 0 — so a CI run that lints the entry-point playbook will not catch the failure.

apoorva-01:
Fixes #5042

`ansible-lint` honors `extra_vars` from config when you lint a playbook directly, but not when the same playbook is reached through `import_playbook:`. Imported playbooks are discovered via `has_playbook()`, which runs `ansible-playbook --syntax-check` with no `extra_vars`, so a child whose `hosts:` (or any syntax-relevant field) depends on an extra var fails that check and is silently dropped from linting. The run still exits 0, which makes the gap easy to miss in CI.

This re-checks such a playbook with the configured `extra_vars` before giving up, so it gets linted whether it is passed directly or imported. Genuinely broken playbooks still fail the re-check and keep the current "Failed to load … syntax check" diagnostic.

On approach: the root cause is that `has_playbook()` (in ansible-compat) can't be given `extra_vars`. The cleanest fix would be an `extra_vars` parameter on `has_playbook()` upstream, but that needs an ansible-compat release and a pin bump, so I kept this contained to ansible-lint by re-running the syntax check here with the vars. Happy to switch to the ansible-compat version if you'd prefer it. One small caveat worth flagging: this re-check mirrors `has_playbook`'s environment, not the main syntax-check pass, which also sets `ANSIBLE_INVENTORY_ANY_UNPARSED_IS_FAILED=False` (#3650) — trivial to mirror here too if you'd rather have it identical.

Tested: added `test_import_playbook_extra_vars`, which reproduces the issue's case — it fails before this change ("1 files processed" + "Failed to load") and passes after ("2 files processed"). The full `test/` suite is green locally, along with ruff and mypy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed imported playbooks that rely on configured extra variables during syntax checks.
  * Improved linting of local imported playbooks so valid files are recognized and processed correctly.
  * Prevented incorrect “Failed to load” errors when extra variables are required.


rockygeekz:
@apoorva-01 Thanks for the PR! Can you sign your commits to proceed with the PR?

shvenkat-rh:
Closing due to lack of response from the contributor

PR #5112, opened 2026-07-15
fix: preserve blank lines after flow collections

### LINKED ISSUE #5108: deletes empty lines after JSONArray and JSONObject
##### Summary

I’ve taken some time today to reduce this to a testcase and test it with git master:

- `cd /tmp; git clone --depth=1 https://github.com/ansible/ansible-lint`
- `cd ansible-lint/src`
- create a file `vt.py` with the following content and run it:

```
import sys

rawr = """---
foo: []

# meow
bar: []
"""

from ansiblelint.yaml_utils import FormattedYAML
#from ruamel.yaml.main import YAML as FormattedYAML

f = FormattedYAML()

r = f.load(rawr)
print(r)
#print(f.dumps(r))
f.dump(r, sys.stdout)
```

The output I get is:

```
{'foo': [], 'bar': []}
---
foo: []
# meow
bar: []
```

The newline before `# meow` was eaten.

Now, comment out the import from ansiblelint and uncomment the one from ruamel, and you get:

```
{'foo': [], 'bar': []}
foo: []
 
# meow
bar: []
```

This proves that it is ansible-lint which introduces the bug.

This is an issue impacting us in Ansible host_vars/group_vars files, in that it removes the empty lines we use to semantically group variables.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 0.1.dev1 using ansible-core:2.19.4 ansible-compat:25.1.5 ruamel-yaml:0.18.10 ruamel-yaml-clib:0.2.12
```

It also happens with:

```
ansible-lint 26.1.1 using ansible-core:2.19.4 ansible-compat:25.12.0 ruamel-yaml:0.18.10 ruamel-yaml-clib:0.2.12
```

- ansible installation method: OS package
- ansible-lint installation method: source

(or OS package)

##### STEPS TO REPRODUCE

see above

##### Desired Behavior

That one (at most) empty line between variables is kept.

##### Actual Behavior

See above.

Sanjays2402:
`FormattedEmitter` was dropping a blank-line token before the next mapping key even when the previous value was an inline flow collection. Track flow-style collection boundaries and preserve one separating line in that case, while keeping the existing whitespace cleanup elsewhere.

Regression coverage includes empty flow sequences and mappings; all 149 YAML utility tests pass.

Fixes #5108

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved separating blank lines after empty flow-style YAML collections such as `[]` and `{}` when formatting YAML.
  * Improved handling of whitespace-only comment lines to keep spacing consistent during load-and-dump round trips, including cases with intervening comment blocks and mixed nested flow/block boundaries.
  * Expanded regression tests to cover these round-trip formatting scenarios and exact output matching.


mirabilos:
Thank you, but I fear the one-element lookbehind is insufficient, for example:

```
rawr = """---
foo: []
baz: 1
 
 
 
# meow
# meow
 
 
# meow
bar: []
"""
```

… in the original test script becomes…

```
---
foo: []
baz: 1
 
# meow
# meow
 
# meow
bar: []
```

… which is good, but if you remove the `baz: 1` line, it becomes…

```
---
foo: []
 
# meow
# meow
# meow
bar: []
```

… which again does not keep one empty line in between other elements.

Sanjays2402:
Thanks for the counterexample — chasing it down turned up something a bit more subtle than a one-line lookbehind.

The old flag only survived a single emit() call, so any intervening structural event between the flow collection closing and the next element's pre-comment would clobber it. Tracing the events, the case that actually drops the blank line is when the flow collection is the **last key of an enclosing block mapping** — a CollectionEndEvent closing that block clears the flag before the next element is emitted. (Interestingly your exact top-level example already round-tripped on the current branch; it's the nested variant that regresses, which is why the single-step flag looked like it worked.)

Fix: replaced the one-step flag with a sticky _pending_flow_collection_separator. It's set when a flow collection closes, survives intervening comment lines *and* container-close events, and is only consumed when the next real element (a Scalar or CollectionStart) is emitted — then cleared, so a fresh element can't over-preserve a blank line where there shouldn't be one.

Added two regression tests: your exact foo/comment-block/bar counterexample, and the nested last-key case that was the real culprit. Confirmed the nested one fails on the original branch and passes with the change. Full test_yaml_utils.py is 151 passed, ruff clean on the added code.

(Side note: the newer ruff flags 7 pre-existing noqa-style nits elsewhere in the file — left those untouched to keep the diff scoped to this fix.)

mirabilos:
Thanks for taking care of this!

I noticed in testing that some cases are still not handled well: whereas…

```
rawr = """---
foo: []
 
# meow
 
# meow
""" 
```

… results in…

```
---
foo: []
 
# meow
 
# meow
```

… which is correct, adding a second blank line there…

```
rawr = """---
foo: []
 
# meow
 
 
# meow
"""
```

… triggers the RE substitution I point out below and gives…

```
---
foo: []
 
# meow
# meow
```

… which is incorrect.

The substitution is in the code in question (line numbers may deviate a bit, I merged this PR into a master checkout and added some debugging `print`s):

```
  855         if (   
[…]
  865         ):
  866-c1          # Preserve a separating blank line after a flow collection. Pure  
  867             # whitespace comments elsewhere remain removed by the formatter.  
  868             value = "\n" if self._previous_event_ended_flow_collection else ""
  869         elif ( 
  870             pre
  871             and not value.strip()
  872             and isinstance(self.event, ruamel.yaml.events.MappingStartEvent)
  873         ):
  874-c2          value = self._re_repeat_blank_lines.sub("", value)
  875         elif pre:
  876             # preserve content in pre comment with at least one newline,
  877             # but no extra blank lines.
  878-c3          value = self._re_repeat_blank_lines.sub("\n", value)
  879         else:
  880             # single blank lines in post comments
  881-c4          value = self._re_repeat_blank_lines.sub("\n\n", value)
```

These are what I call the four cases of that if, marked as c1 through c4 to ease discussion below.

In your PR, you changed c1 in line 868 to take the new `self._previous_event_ended_flow_collection` into account. The problem of the multiple blank lines is that we’re now running into `self._re_repeat_blank_lines`, which takes three or more subsequent `\n`, but the comment’s line ending needs to be also considered.

The test case from above runs into c3.

However, adding…

```
if self._previous_event_ended_flow_collection:
    value += "\n"
```

… does not trigger, because `self._previous_event_ended_flow_collection` somehow is False at that point.

I tried with some more debugging output but could not get any further than this.

I also could not get c2 to trigger, so I cannot produce a test case that hits c2… nor c4, though I’m 60% sure we cannot trigger c4 after a flow collection.

rockygeekz:
@Sanjays2402 Thanks for the PR! Can you please verify your commits to proceed with this?

mirabilos:
Hi @Sanjays2402 ,

thanks for following up on this again!

I’m a bit puzzled-amazed at the latest commit. I was under the impression that we would need to change it for only the “after flow collection” change, but apparently this does not introduce a regression for the block collection case. Perhaps the latter is parsed differently, so that it makes no difference here.

I played around with this and several different things, arrays, dictionaries, comments, scalars, newlines at the end, lack of the one required newline at the end, and found it works for most cases now.

The remaining issues I found are:

- one empty line is kept between the initial `---` and the first element, but this is the case for the code in the `main` branch as well, so ansible-lint probably fixes this up elsewhere so this is not a regression
- as in the `main` branch, one empty line is kept after the last line if the last line is not a flow collection (this is hopefully also fixed elsewhere)
- new: if the last line is a flow collection, *two* empty lines are kept at the end

To check for actual impact, I applied your commits on top of the `v25.2.1` tag which matches the version in Debian on my work laptop, copied the changed file to `/usr`, and ran it one some actual playbooks and `defaults/main.yml` files, and the result is as expected, that is, **the “remaining issues” listed above are _not_ a problem and the end result of `ansible-lint` matches expectations.**

Interestingly, now the code for c3 and c4 is identical. But for documentation purposes, it may still make sense to keep it separate.

I’ve no idea what to do about c2 but better to “never touch a running system”, so let’s keep that as-is.

From the reporter’s PoV, this PR is good to go.

mirabilos:
> so the trailing newlines are already normalised the same way with and without my commits

Yeah, I only saw them in the test case; as long as `ansible-lint --fix` DTRT, it’s all good, and I tested that it does.

> Happy to re-create the three commits

You can just squash them and rebase on top of latest master (the Ansible people do that on merging anyway) and then you can sign that one commit.

> so GitHub signs them server-side

… huh? Why not sign the commit on your computer with your PGP key? That is much easier.

Do you need help with git? I’m willing, to help this get merged.

(@rockygeekz is that signed commits requirement technical or merely a formal pre-check? That is, could I do the squash+rebase+test+sign+push for the author (while leaving their name intact, of course) to simplify things, signing the commit as the one who tested it, or does it have to be the author’s signature (in which case Microslop GitHub signing it would be kind of defeating it anyway)?)

mirabilos:
@Sanjays2402 can you add me (temporarily) as collaborator on your repo? then I can push it right into this PR

Sanjays2402:
I am not going to grant temporary write access to the fork. The branch is already squashed, rebased on current main, and the focused test file is green; I will handle the commit-signing setup on my side and update the PR with a signed head rather than transfer push access.

mirabilos:
On Sat, 15 Aug 2026, Sanjay Santhanam wrote:

>The branch is already squashed, rebased on current main, and the
>focused test file is green; I will handle the commit-signing setup on
>my side and update the PR with a signed head rather than transfer push
>access.

That’s even better, thanks.

Sanjays2402:
I rebased the branch's existing tree unchanged and pushed `5e647464a98e2d94e7e10751d135957a0f6ef0c7`, a single commit with a valid local GPG signature. The focused `test/test_yaml_utils.py` suite is 152 passed.

One limitation: GitHub currently reports the signature as `unknown_key`, because this signing key is not registered on my GitHub account. The commit is cryptographically signed locally, but I won't claim GitHub-verified status until the public key is attached to the account. The branch is otherwise clean and `MERGEABLE`.

mirabilos:
@Sanjays2402 it seems like you need to fix that until Microslop GitHub shows the commit PGP signature as verified 🤷

shvenkat-rh:
Superseded by https://github.com/ansible/ansible-lint/pull/5178

PR #5113, opened 2026-07-15
perf: cache get_deps_versions() result

BlackDark:
Called once per templated string via ansible_template(), doing 4x importlib.metadata.version() lookups each time. Installed package versions can't change mid-process, so cache the result.

~15% wall-time reduction measured on a 279-role production repo (81s -> 69s for a full lint run), identical output.

## Summary by CodeRabbit

* **Performance Improvements**
  * Dependency version information is now reused during a process, reducing repeated lookups and improving efficiency for operations that check dependency versions. Subsequent checks remain consistent while the process is running.


rockygeekz:
@BlackDark Thanks for the PR! Can you please verify your commits to proceed with this?

shvenkat-rh (approved):

PR #5120, opened 2026-07-20
fix: validate molecule.yml against the ansible-native schema by default

### LINKED ISSUE #5119: Stale bundled molecule schema rejects valid native ansible: config, and molecule.yml isn’t schema-validated by default
##### Summary

ansible-lint ships a `molecule` JSON schema but has two coupled defects: it does **not** apply that schema to `molecule.yml` by default (the file is mapped to kind `yaml`), and the schema it ships is **stale** (still the pre-native model, missing the `ansible:` block). Either alone is a problem; together they mean molecule config is effectively unvalidated by ansible-lint out of the box, and cannot be safely turned on as shipped because the stale schema false-errors on every current native config.

Relevant files:

- [`src/ansiblelint/config.py`](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/config.py) — `DEFAULT_KINDS` maps `molecule.yml` to kind `yaml`
- [`src/ansiblelint/schemas/molecule.json`](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/schemas/molecule.json) — the stale bundled schema
- [`src/ansiblelint/schemas/__store__.json`](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/schemas/__store__.json) — self-referential store entry, the drift root cause
- [`ansible-community/molecule` `src/molecule/data/molecule.json`](https://github.com/ansible-community/molecule/blob/main/src/molecule/data/molecule.json) — the authoritative up-to-date schema

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 26.6.0 using ansible-core:2.21.2 ansible-compat:26.6.0 ruamel-yaml:0.19.1 ruamel-yaml-clib:0.2.15
```

Also relevant: molecule 26.6.0 (native config model, current since molecule v25.9.0).

##### STEPS TO REPRODUCE

Two files, one command.

1. `molecule/default/molecule.yml` — a valid ansible-native config, accepted by molecule 26.6.0:

```yaml
---
dependency:
  name: galaxy
ansible:
  executor:
    backend: ansible-playbook
  playbooks:
    converge: converge.yml
scenario:
  name: default
  test_sequence:
    - converge
```

2. `.ansible-lint` — classify `molecule.yml` as kind `molecule`, as ansible-lint's own error text suggests ("use `kinds:` option in linter config to help it pick correct file type"):

```yaml
---
kinds:
  - molecule: "**/molecule/*/molecule.yml"
```

3. Run (with `--offline` so the bundled schema, not a network refresh, is exercised):

```console (paste below)
ansible-lint --offline molecule/default/molecule.yml
```

Then delete the `.ansible-lint` override and run the same command again.

##### Desired Behavior

`molecule.yml` should be schema-validated against an up-to-date molecule schema by default, and a valid ansible-native molecule config should pass.

Requested change, two parts, **in this order** (order matters — part 2 before part 1 would break every native molecule user with false errors):

1. **Fix the schema.** Sync `src/ansiblelint/schemas/molecule.json` to the ansible-native model from Molecule's authoritative [`src/molecule/data/molecule.json`](https://github.com/ansible-community/molecule/blob/main/src/molecule/data/molecule.json), and ideally have `__store__.json` track Molecule's schema URL so it cannot drift again.
2. **Wire it up.** Map the molecule config file to kind `molecule` in `config.py` `DEFAULT_KINDS` so `molecule.yml` is actually validated against the now-correct schema.

If kind `molecule` was intentionally downgraded to `yaml` as a stopgap for the stale schema, this is the request to finish that work rather than leave molecule config unvalidated.

##### Actual Behavior

With the `.ansible-lint` override in place, ansible-lint fatally rejects the valid native config:

```console
$ ansible-lint --offline molecule/default/molecule.yml
schema[molecule]: $ Additional properties are not allowed ('ansible' was unexpected). See https://docs.ansible.com/projects/molecule/configuration/
This rule is not skippable and stops further processing of the file.
Failed: 1 failure(s), 0 warning(s)
$ echo $?
2
```

Without the override, the same file passes with **no schema check at all** (kind `yaml`):

```console
$ rm .ansible-lint
$ ansible-lint --offline molecule/default/molecule.yml
$ echo $?
0
```

Those two runs demonstrate the two coupled defects:

**Defect 1 — `molecule.yml` is not schema-validated by default.** In [`config.py`](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/config.py), `DEFAULT_KINDS` maps the molecule config file to kind `yaml`:

```python
{"yaml": "**/molecule/*/{base,molecule}.{yaml,yml}"},  # molecule config
```

No entry maps any file to kind `molecule`. Kind `yaml` has no registered JSON schema, so `molecule.yml` gets only generic YAML checks and the bundled molecule schema ([`molecule.json`](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/schemas/molecule.json), registered for kind `molecule`) is never applied out of the box. Every other structured config file gets its own specific kind (`galaxy`, `meta-runtime`, `argument_specs`, `requirements`), so molecule config is the odd one out. This is layout-agnostic — the leading `**/` matches zero or more segments, so both a standalone role's `molecule/<scenario>/molecule.yml` and a collection's `extensions/molecule/<scenario>/molecule.yml` are mapped to `yaml`; both are unvalidated.

**Defect 2 — the bundled schema is stale (pre-native).** [`molecule.json`](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/schemas/molecule.json) still has root `"required": ["driver", "platforms"]`, `"additionalProperties": false`, and no `ansible` property. Molecule has used the ansible-native model since v25.9.0 (current in 26.6.0): a top-level `ansible:` block (`cfg`/`env`/`executor`/`playbooks`) replaces `provisioner:`, and neither `driver:` nor `platforms:` is required. The `additionalProperties: false` + missing `ansible` key is what produces the fatal error above; wiring `molecule.yml` to kind `molecule` makes the stale schema reject every valid native scenario. This is a non-skippable rule.

This is the bundled copy, not a network fetch: the same fatal error reproduces with networking fully disabled (`ansible-lint --offline` under `--network none`), and the installed schema (`site-packages/ansiblelint/schemas/molecule.json`) has `additionalProperties: false`, no `ansible` key, and top-level properties `dependency, driver, lint, log, platforms, prerun, provisioner, role_name_check, scenario, verifier` — the old driver/provisioner/platforms model. Air-gapped workstations and execution environments therefore hit this stale bundled schema.

**Why they are coupled.** The schema cannot be safely turned on while it is stale (it would false-error on all current configs), which is the likely reason `molecule.yml` is left as kind `yaml`. So today the schema exists but protects nothing, and fixing either half in isolation does not fix the user experience.

**Root cause of the drift.** [`__store__.json`](https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/schemas/__store__.json) maps `molecule` to ansible-lint's own raw URL (`.../ansible-lint/main/src/ansiblelint/schemas/molecule.json`), i.e. the bundled schema is self-referential and is not synced from Molecule's authoritative schema, so it has fallen behind the native transition.

**Why the bundled schema matters (connectivity).** The Red Hat Ansible VS Code extension associates `molecule.yml` with Molecule's own native schema by pointing `contributes.yamlValidation` at a remote raw-GitHub URL (`https://raw.githubusercontent.com/ansible-community/molecule/main/src/molecule/data/molecule.json`), with no bundled fallback, so an editor *with network access* validates against the current model. But that path depends on live access to GitHub. With no network (an air-gapped or disconnected workstation, or an execution environment), the language server cannot retrieve the remote schema: it reports `Unable to load schema '<url>': No content` and performs **no** schema validation, so an invalid `molecule.yml` passes unchecked and a valid one gets no validation either. `ansible-lint --offline` likewise cannot fetch a remote schema and falls back to the bundled copy, which is exactly the stale one this report is about. So the bundled schema's correctness is not redundant with the editor experience: it is the only schema-based validation available precisely where remote schema fetching is unavailable.

(Verified headlessly by driving `yaml-language-server` over LSP with the molecule schema association: online it emits correct `yaml-schema: Molecule Scenario Schema` diagnostics; under `--network none` with no cached copy it emits only the `Unable to load schema … No content` load error and no schema diagnostics. The failure is a loud non-validation, not a silent skip.)

jeffcpullen:
## What

Make ansible-lint validate `molecule.yml` against an up-to-date, ansible-native Molecule schema by default. Two coordinated changes plus tests:

1. Repoint the `molecule` entry in `src/ansiblelint/schemas/__store__.json` from ansible-lint's own self-referential raw URL to Molecule's authoritative schema (`ansible-community/molecule/main/src/molecule/data/molecule.json`) and regenerate the bundled `molecule.json` (+ etag) with the schema updater.
2. Map only `**/molecule/*/molecule.{yaml,yml}` to kind `molecule` in `DEFAULT_KINDS` (it was kind `yaml`), so a scenario's molecule config is actually schema-validated. `base.yml` deliberately stays kind `yaml` (see Compatibility).

Fixes #5119.

## Background and Context

Molecule is the scenario test runner for Ansible roles and collections. A scenario is configured by a `molecule.yml` file. In molecule >= v25.9 the config format changed substantially ("ansible-native"): the old top-level `driver:` / `platforms:` model gave way to a top-level `ansible:` block, and the root `driver`/`platforms` requirement was dropped. Molecule publishes the JSON schema for this format in its own repo at `src/molecule/data/molecule.json`.

ansible-lint ships bundled JSON schemas and maps file patterns to them via `DEFAULT_KINDS`. Two things were wrong for molecule:

- The bundled `molecule.json` was **stale** (pre-native: root `required: ["driver","platforms"]`, `additionalProperties: false`, no `ansible` key), because `__store__.json` pointed the `molecule` schema at ansible-lint's *own* copy rather than at Molecule's authoritative file, so it never tracked the native transition.
- `molecule.yml` was mapped to kind **`yaml`**, which has no schema, so the `schema[molecule]` rule never fired. Molecule config was effectively unvalidated out of the box.

These interlock: the schema could not simply be "turned on," because the stale schema would false-error on every modern native config (`schema[molecule]: Additional properties are not allowed ('ansible' was unexpected)`, exit 2). This PR does the two halves in the only safe order: fix the schema first, then wire it up.

## Why the URL points at `ansible-community/molecule@main`

Tracking Molecule's own published schema (instead of a copy maintained here) is what keeps the bundled schema from drifting again: the copy in this repo is regenerated from the authoritative source, and connected users refresh from it. This uses the same cross-repo, main-branch model already used by the `rulebook` (`ansible/ansible-rulebook`) and `ansible-navigator-config` (`ansible/ansible-navigator`) entries in `__store__.json`, so it is not a new pattern here. Pinning a tag instead would defeat the purpose: `refresh_schemas()` would `304` forever and re-freeze the drift this fixes.

The regenerated `molecule.json` is verbatim updater output of Molecule's published schema. A few things in the diff look loose (an open `platforms.items`, permissive `network_mode` / `ScenarioSequence`, and some now-unreferenced `$defs`); those are Molecule's current upstream design (per-platform shape is delegated to the driver plugin), not an ansible-lint edit. See "Relationship to Molecule" below.

## Compatibility and blast radius

Schema-to-schema the native schema is a superset of the old one: every pre-existing molecule positive fixture still validates. But because `molecule.yml` moves from kind `yaml` to kind `molecule`, this turns validation **on** for files that were previously never checked, so some configs will newly report `schema[molecule]` on upgrade (this rule is `VERY_HIGH` and not skippable):

- unknown top-level keys,
- out-of-enum or third-party `driver.name` values,
- a scenario `molecule.yml` with no `platforms` and no `ansible:` block.

Each of these is a config Molecule itself rejects when it validates a single file, so ansible-lint is now agreeing with Molecule rather than inventing a rule.

**`base.yml` stays kind `yaml` on purpose.** Molecule supports a shared `base.yml` that is a *fragment*, merged into each scenario's config before validation; on its own it is legitimately incomplete (often empty) and would false-error against the whole-file schema. Keeping it kind `yaml` avoids that, and matches the existing decision to leave the `.config/molecule/config.*` global subset as `yaml`. Only the per-scenario `molecule.yml` (the complete document) is validated.

Projects with an unusual layout (for example `platforms` living only in a merged fragment, so the per-scenario file looks incomplete in isolation) can opt out per project:

```yaml
kinds:
  - yaml: "**/molecule/*/molecule.{yaml,yml}"
```

## Disconnected / air-gapped / EE users

The cross-repo URL is invisible to disconnected users, and the change is a strict improvement for them:

- `refresh_schemas()` never runs under `--offline` (or `offline: true` in config), is skipped when the installed `__store__.json` is not writable (system-package and execution-environment installs), and falls back to the last-known-good bundled copy on any network error. So air-gapped and EE users only ever consult the **bundled** `molecule.json`; they never reach the new URL.
- Before this PR their bundled copy was stale *and* `molecule.yml` was unvalidated, so they got no molecule validation at all. After it, they get correct, current, fully-offline validation with zero network dependency.
- This is the only path that gives disconnected users molecule validation at all. The editor's as-you-type schema check fetches Molecule's schema by remote URL with no bundled fallback (see below), so it goes dark offline. ansible-lint bundles the schema locally, so the CLI (and the editor's ansible-lint diagnostics path, which runs `ansible-lint --offline`) is what actually validates `molecule.yml` air-gapped.

## Editor / VS Code impact

The redhat.ansible extension gives `molecule.yml` the `ansible` languageId (its `contributes.languages` filenamePatterns include `**/molecule/*/molecule.{yml,yaml}`), so both editor paths touch it, and they behave differently by connectivity:

- **As-you-type validation / completion** (redhat.vscode-yaml, via the extension's `contributes.yamlValidation`) associates `molecule.yml` with Molecule's own native schema by **remote URL** (`ansible-community/molecule/.../molecule.json`), not ansible-lint's bundled copy. This PR does not change that path. But pointing at a URL only helps when the editor can reach or has cached it: on an air-gapped machine that never cached it, vscode-yaml cannot load the schema (`Unable to load schema ... No content`) and provides no molecule completion or validation. There is no bundled / `file://` fallback on this path.
- **Lint diagnostics**: because molecule.yml is an `ansible`-language file, the Ansible language server runs ansible-lint on it (`ansible-lint --offline --nocolor -f codeclimate`) and surfaces the result as editor diagnostics. This path uses ansible-lint's **bundled** schema and needs no network. Before this PR it never validated `molecule.yml` (kind `yaml`); after it, editor users with ansible-lint installed and linting enabled (the default) get correct molecule validation here, online and offline.

So for connected users the two paths converge on the same authoritative schema. For **disconnected** users this PR is what provides in-editor molecule validation at all: the URL-based as-you-type path cannot load a schema offline, while the ansible-lint diagnostics path now does, from the local bundle. (The one thing still missing offline is as-you-type schema completion/hover, which is remote-URL-only; closing that would be a separate, optional vscode-ansible enhancement to vendor the schema locally, not an ansible-lint change.)

## Relationship to Molecule (and molecule PR #4661)

Pointing at the authoritative source means Molecule schema fixes now flow downstream to ansible-lint automatically, with no change needed here. That is directly relevant right now: Molecule's current native schema leaves `platforms.items` as an open `{"type": "object"}` and has orphaned its `MoleculePlatformModel`, so per-platform shape is not validated. That is a bug in Molecule's own schema (issue and suggested fix upstream in ansible-community/molecule#4661), **not** something to hand-patch into ansible-lint's bundled copy, which would just reintroduce the drift this PR removes. When #4661 merges, ansible-lint picks up the restored per-platform validation through the normal schema refresh.

This is also why this PR **removes** the two old `platforms_*` negative fixtures: they asserted per-platform shape errors that Molecule's current schema no longer produces, so they can no longer fail. They are replaced with a native `ansible.executor.backend` enum negative that the schema does reject.

## On the empty-document case (deliberately out of scope)

A genuinely empty scenario `molecule.yml` (rare, distinct from the `base.yml` fragment handled above) currently produces a less-than-friendly `None should not be valid ...` message under the native schema. Cleanly special-casing an empty/`None` document belongs in the shared schema-validation path and would affect all schema kinds, not just molecule, so I kept it out of this focused fix rather than widen the blast radius. Happy to follow up with a separate PR for it if you would like it handled here.

## Verification

- Reproduced the original fatal `schema[molecule]: Additional properties are not allowed ('ansible' was unexpected)` (exit 2), then confirmed the native config passes with no `kinds:` override and that validation genuinely applies (an invalid key is now caught).
- Legacy driver/platforms/provisioner config still validates (native schema is a root-level superset).
- Added end-to-end `test_schema` cases that run through the linter (native `molecule.yml` passes; invalid `executor.backend` fails), a positive native fixture in the `test/schemas` node suite, and an updated kind-mapping test (`base.yml` stays `yaml`, `molecule.yml` becomes `molecule`).
- `test/schemas` node suite green (285 passing); `test_schema` / `test_kinds` / `test_schemas` green; the schema-refresh test confirms the regenerated `molecule.json` is in sync with its source; lint clean.

## Note for maintainers

Turning on validation for a whole file class is arguably more than a patch-level `fix:`; feel free to relabel for a minor release if that suits your release-drafter conventions.

## Summary by CodeRabbit

* **New Features**
  * Updated Molecule configuration validation for current executor, platform, dependency, driver, provisioner, verifier, and shared-state options.
  * Added support for custom drivers, expanded network and environment settings, and separate Molecule scenario-file detection.
  * Updated validation to use the current Molecule schema.
* **Bug Fixes**
  * Improved Molecule file classification and schema validation.
* **Tests**
  * Added native Molecule coverage and invalid executor-backend validation.
  * Updated validation fixtures for current schema behavior.


jeffcpullen:
I'm closing this PR. I haven't found a good way to do this that wouldn't lead to unskippable errors. So I'm going to back off trying to change the default and see what can be done for those that instead opt-in.

PR #5130, opened 2026-07-25
fix: do not require role prefix for ansible_ connection variables

### LINKED ISSUE #5099: Variables names from within role should not use role name as prefix when using ansible connection variables
##### Summary

Since ansible-lint 26.6.0 or possibly earlier, variable names from within roles must use the role name as a prefix.
This should not be the case when using ansble connection variables that can't be renamed.
E.g. `ansible_ssh_common_args`.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```
ansible-lint 26.6.0 using ansible-core:2.21.0 ansible-compat:26.3.0
```

##### STEPS TO REPRODUCE

```
- name: 'Switch to strict host key checking'
  ansible.builtin.set_fact:
    ansible_ssh_common_args: >-
      -o StrictHostKeyChecking=yes
  changed_when: true
```

##### Desired Behavior

Linter shouldn't report error.

##### Actual Behavior

Linter reports error
```
Variables names from within roles should use role_name_ as a prefix. (set_fact: ansible_ssh_common_args)ansible-lintvar-naming[no-role-prefix]
```

Sanjays2402:
Fixes #5099

`var-naming[no-role-prefix]` flagged `ansible_`-namespace variables (e.g. `ansible_ssh_common_args`) when set inside a role, but those are defined by Ansible and cannot be renamed to carry a role prefix. The role-prefix check now skips identifiers in the `ansible_` namespace.

Regression test `test_var_naming_connection_vars_in_role` fails without the change and passes with it; the full `var_naming.py` suite is green (13 passed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated variable naming checks to allow standard `ansible_` connection variables in role-scoped linting without requiring a role prefix.
  * Prevented valid variables such as `ansible_port` and SSH connection options from being incorrectly flagged.


rockygeekz:
@Sanjays2402 Thanks for the PR! Can you please verify your commits to proceed with this?

rockygeekz (approved):

PR #5148, opened 2026-08-13
fix: propagate extra_vars to import_playbook syntax check

### LINKED ISSUE #5042: extra_vars not propagated to syntax check when resolving import_playbook: references

shvenkat-rh:
## Summary

- Re-checks imported playbooks with configured `extra_vars` when initial syntax check fails
- Adds regression test for the fix

## What

`ansible-lint` honors `extra_vars` from config when you lint a playbook directly, but not when the same playbook is reached through `import_playbook:`. Imported playbooks are discovered via `has_playbook()`, which runs `ansible-playbook --syntax-check` with no `extra_vars`, so a child whose `hosts:` (or any syntax-relevant field) depends on an extra var fails that check and is silently dropped from linting. The run still exits 0, which makes the gap easy to miss in CI.

This re-checks such a playbook with the configured `extra_vars` before giving up, so it gets linted whether it is passed directly or imported. Genuinely broken playbooks still fail the re-check and keep the current "Failed to load … syntax check" diagnostic.

Fixes #5042

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of imported local playbooks that require configured extra variables during syntax validation.
  * These playbooks are retried with the configured variables and can now be linted successfully instead of producing load errors.
  * Playbooks without extra variables continue to follow the standard validation flow.
  * Existing error handling is preserved when validation still fails.
  * Collection playbooks remain excluded from linting.

### DIFF
--- a/src/ansiblelint/utils.py
+++ b/src/ansiblelint/utils.py
@@ -574,17 +574,7 @@ class HandleChildren:
         v: Sequence[Any],
         parent_type: FileType,
     ) -> list[Lintable]:
-        """Discover lintable files for roles declared in a playbook or role.
-        
-        Parameters:
-        	lintable (Lintable): File containing the role declarations.
-        	k (str): Key identifying the role declaration section.
-        	v (Sequence[Any]): Role declarations to resolve.
-        	parent_type (FileType): Type of the containing file.
-        
-        Returns:
-        	list[Lintable]: Files found for the declared roles.
-        """
+        """Roles children."""
         # pylint: disable=unused-argument # parent_type)
         basedir = str(lintable.path.parent)
         results: list[Lintable] = []
@@ -613,15 +603,11 @@ class HandleChildren:
         return results
 
     def _recheck_playbook_with_extra_vars(self, possible_path: Path) -> bool:
-        """
-        Re-check a playbook's syntax using the configured extra variables.
-        
-        Parameters:
-            possible_path (Path): Path to the playbook to validate.
-        
-        Returns:
-            bool: `True` if the syntax check succeeds with the configured extra
-            variables, `False` if no extra variables are configured or the check fails.
+        """Re-check playbook syntax with configured extra_vars.
+
+        has_playbook() runs its syntax check without our configured extra_vars,
+        so it wrongly fails on a local playbook that only loads once those vars
+        are defined. Re-check with them before giving up.
         """
         extra_vars = self.app.options.extra_vars
         if not extra_vars:
@@ -646,18 +632,7 @@ class HandleChildren:
         parent_type: FileType,
         name: str,
     ) -> list[Lintable] | str:
-        """
-        Validate a candidate playbook path and return its lintable representation or an error message.
-        
-        Parameters:
-            possible_path (Path): Candidate playbook path.
-            is_collection (bool): Whether the playbook belongs to a foreign collection.
-            parent_type (FileType): File type assigned to the resulting lintable.
-            name (str): Playbook name used in error messages.
-        
-        Returns:
-            list[Lintable] | str: A list containing the lintable playbook, an empty list for foreign collection playbooks, or an error message.
-        """
+        """Check a single playbook path and return Lintable or error message."""
         if not possible_path.exists():
             return f"Failed to find {name} playbook."
         if not self.app.runtime.has_playbook(str(possible_path)):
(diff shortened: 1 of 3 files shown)


src/ansiblelint/utils.py
```diff
@@ -609,59 +653,58 @@ def import_playbook_children(
...
+                        namespace_name,
+                        collection_name,
+                        "playbooks",
+                        *playbook_path[:-1],
+                        f"{playbook_path[-1]}{ext}",
+                    ),
+                ),
+            )
+            for loc in self.app.runtime.config.collections_paths
+            for ext in (".yml", ".yaml")
```
rockygeekz, line 711: This branches on a magic substring match against the error message from `_resolve_playbook_path`. If that message ever gets reworded, this silently falls through to the wrong error path. A constant or structured return would be more robust.

For example:
```python
_SYNTAX_CHECK_FAILED = "syntax_check_failed"

# In _resolve_playbook_path:
return (_SYNTAX_CHECK_FAILED, f"Failed to load {name} playbook due to failing syntax check.")

# Here:
if isinstance(result, tuple) and result[0] == _SYNTAX_CHECK_FAILED:
```

src/ansiblelint/utils.py
```diff
@@ -609,59 +653,58 @@ def import_playbook_children(
...
+        lintable: Lintable,
+        v: str,
+        namespace_name: str,
+        collection_name: str,
+        playbook_path: list[str],
+    ) -> list[Path]:
+        """Build list of possible playbook paths to check."""
+        if not (namespace_name and collection_name):
+            return [lintable.path.parent / v]
+
```
rockygeekz, line 695: This docstring lists only `v` and `parent_type` but skips `lintable` and `_k`. Either complete the list or trim to a one-liner.

rockygeekz (changes requested):
Changes requested

rockygeekz (approved):

PR #5169, opened 2026-08-28
Add validate_argspec to play-level schema keywords

### LINKED ISSUE #5168: validate_argspec is not a valid play attribute
##### Summary

`validate_argspec` which was introduced as *Tech Preview* with Core 2.20 (see https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_variables_validation.html) is currently flagged by ansible-lint as invalid keyword.

I am not sure, if the exclude was on purpose (since it's a Tech Preview feature).
If not, I'd give a shot myself with a PR, by adding it to
https://raw.githubusercontent.com/ansible/ansible-lint/main/src/ansiblelint/schemas/playbook.json

Before I raise a PR, however, I wanted to see if it was a deliberate decision to not include it (yet).

I'd appreciate if a maintainer could let me know if it's worth raising a PR for it.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 26.8.0 using ansible-core:2.21.3 ansible-compat:26.8.0 ruamel-yaml:0.19.1 ruamel-yaml-clib:0.2.15
```

##### STEPS TO REPRODUCE

Minimal playbook to reproduce:
```yaml
- name: 'Minimal reproducer'
  hosts: 'localhost'
  gather_facts: false
  validate_argspec: true
  tasks:
    - name: 'Test'
      ansible.builtin.debug:
        msg: 'Test'
```

The following error is observed when running ansible-lint:
```console
ansible-lint
  schema[playbook]: $[0] 'gather_facts', 'hosts', 'tasks', 'validate_argspec' do not match any of the regexes: '^(ansible\.builtin\.)?import_playbook$', 'name', 'tags', 'vars', 'when'
Failed: 1 failure(s), 0 warning(s) in 1 files processed of 1 encountered.
```

Core accepts the playbook, however. It also enforces the spec:
```yaml
argument_specs:
  Minimal reproducer:
    options:
      test:
        type: 'str'
        required: true
```

```console
  $ ansible-playbook -i localhost, playbook.yml -e test=foo
  localhost : ok=2  changed=0  unreachable=0  failed=0

  $ ansible-playbook -i localhost, playbook.yml
  missing required arguments: test
```

##### Desired Behavior

`validate_argspec` is recognized as a valid play attribute

##### Actual Behavior

The following error is observed when running ansible-lint:
```console
ansible-lint
  schema[playbook]: $[0] 'gather_facts', 'hosts', 'tasks', 'validate_argspec' do not match any of the regexes: '^(ansible\.builtin\.)?import_playbook$', 'name', 'tags', 'vars', 'when'
Failed: 1 failure(s), 0 warning(s) in 1 files processed of 1 encountered.
```

agu2347:
## Summary

Fixes #5168.

The `validate_argspec` play keyword (a Tech Preview feature in ansible-core for validating role/playbook argument specs) is a valid play-level keyword recognized by ansible-core, but it was missing from ansible-lint's JSON schema for plays. As a result, any playbook using `validate_argspec: true` at the play level was flagged by the `schema[playbook]` rule with an `additionalProperties`/`oneOf` validation error, even though the playbook is valid and runs correctly with ansible-core.

## Fix

- Added `validate_argspec` (as a `templated-boolean`, consistent with similar boolean play keywords like `any_errors_fatal`) to the `play` `$defs` entry in the source schema `src/ansiblelint/schemas/ansible.json`.
- Regenerated the derived `src/ansiblelint/schemas/playbook.json` via the project's own `test/schemas/src/rebuild.py` generator, per the file's own `"do not edit"` comment. I confirmed the regenerated diff is byte-identical to what a manual edit of `playbook.json` would have produced.
- Added a minimal positive test fixture `test/schemas/test/playbooks/validate_argspec.yml` directly reproducing the issue's own repro case.

## Verification

- Ran the real schema test suite (`npm test`, which runs `rebuild.py` then `mocha`/`ajv`) in `test/schemas/`: 287 tests passing, including the new fixture validated against both `ansible.json#/$defs/playbook` and the generated `playbook.json`.
- Revert-check: temporarily reverted the `ansible.json` change (regenerating `playbook.json` from the reverted source) and reran the suite — it reproduces exactly the original error reported in the issue (`additionalProperties`/`oneOf` failures on `validate_argspec`), confirming both the fix and the new test correctly catch the regression.

Signed-off-by: agu2347 <agu2347@users.noreply.github.com>

## Summary by CodeRabbit

- **New Features**
  - Added support for the optional `validate_argspec` setting in play definitions.
  - The setting accepts boolean values and Jinja templates.

- **Tests**
  - Added coverage for playbooks using `validate_argspec: true`.

### DIFF
--- a/src/ansiblelint/schemas/ansible.json
+++ b/src/ansiblelint/schemas/ansible.json
@@ -636,6 +636,10 @@
           "title": "Remote User",
           "type": "string"
         },
+        "validate_argspec": {
+          "$ref": "#/$defs/templated-boolean",
+          "title": "Validate Argspec"
+        },
         "vars": {
           "title": "Vars",
           "type": "object"
--- a/src/ansiblelint/schemas/playbook.json
+++ b/src/ansiblelint/schemas/playbook.json
@@ -646,6 +646,10 @@
           "title": "Remote User",
           "type": "string"
         },
+        "validate_argspec": {
+          "$ref": "#/$defs/templated-boolean",
+          "title": "Validate Argspec"
+        },
         "vars": {
           "title": "Vars",
           "type": "object"
--- a/test/schemas/test/playbooks/validate_argspec.yml
+++ b/test/schemas/test/playbooks/validate_argspec.yml
@@ -1,0 +1,8 @@
+- name: Minimal reproducer
+  hosts: localhost
+  gather_facts: false
+  validate_argspec: true
+  tasks:
+    - name: Test
+      ansible.builtin.debug:
+        msg: Test


shvenkat-rh:
Please rebase, fix tox issues and push verified commits. Unverified comments will not be merged

agu2347:
@shvenkat-rh thanks for the note. A few things I can't resolve from my side:

- The `fix`/`feat`/`chore`/`breaking` label check needs a maintainer to apply it -- I don't have permission to add labels to this repo (`403: Must have admin rights`). This is a schema addition with no behavior change to existing rules, so `fix` seems closest, but happy to defer to whichever a maintainer prefers.
- "Verified commits": commits I push here are created through the GitHub API rather than a local signed `git commit`, so they'll show as unverified regardless of authorship -- I don't have a signing key configured for this account. If verified/signed commits are a hard requirement for merging, that's a constraint on my end I can't currently work around, and I'd rather say so plainly than keep pushing unsigned commits that won't satisfy it.
- I can rebase onto current main to clear the "behind" state -- let me know if that's still worth doing given the above, or if this should just wait for someone able to sign commits.

No tox failures I'm aware of on my end; if there's a specific CI run showing tox issues, a link would help me look at the right one.

shvenkat-rh:
Without verified commits, we cannot merge the PR

agu2347:
Understood -- since verified (signed) commits aren't something I can produce through this workflow, and that's a hard requirement here, I don't think I can get this PR into a mergeable state. Closing rather than leaving it open indefinitely; thanks for taking the time to look at it. If anyone wants to pick up the underlying change (adding `validate_argspec` as a recognized play-level schema keyword) with properly signed commits, the diff here should still be a useful reference.

PR #5177, opened 2026-09-09
fix: respect skip_list for yaml[comments] during --fix

### LINKED ISSUE #5048: Since 25.12.1, ansible-lint --fix  applies yaml[comments] even if rule is present in skip_list
##### Summary
Since version 25.12.1, when running ansible --fix, the rule yaml[comments] gets applied even if present in section skip_list of the config file.

##### Issue Type

- Bug Report

##### OS / ENVIRONMENT

```console (paste below)
ansible-lint 26.4.0 using ansible-core:2.20.5 ansible-compat:26.3.0 ruamel-yaml:0.19.1 ruamel-yaml-clib:None
```

##### STEPS TO REPRODUCE
.ansible-lint : 
```
 profile: basic
 skip_list:
   - yaml[comments]
```

playbook.yml
```
---
#test-comment

- name: "Test playbook"
  hosts: localhost
  tasks:
    - name: "Test task"
      meta: noop
``` 

command executed
``` ansible-lint --fix playbook.yml```

##### Desired Behavior

As yaml[comments] is in the skip_list, I expect the comment to be left as it is.

##### Actual Behavior

ansible-lint output : 
```ansible-lint --fix playbook.yml 

Modified 1 file.
Passed: 0 failure(s), 0 warning(s) in 1 files processed of 1 encountered. Profile 'basic' was required, but 'production' profile passed.
```

Playbook diff : 
```
diff --git a/playbook.yml b/playbook.yml
index d6b8029..1ccf87b 100644
--- a/playbook.yml
+++ b/playbook.yml
@@ -1,5 +1,5 @@
 ---
-#test-comment
+# test-comment
 
 - name: "Test playbook"
   hosts: localhost
```

If I revert to ansible-lint 25.12.0, file is not modified.

```
Passed: 0 failure(s), 0 warning(s) in 1 files processed of 1 encountered. Profile 'basic' was required, but 'production' profile passed.
A new release of ansible-lint is available: 25.12.0 → 26.4.0[/] Upgrade by running: pip3 install --user --upgrade ansible-lint
```

shvenkat-rh:
Recreates PR - 5068

Summary
Fixes https://github.com/ansible/ansible-lint/issues/5048.

When yaml[comments] is in skip_list, --fix was still modifying comments (e.g. #test-comment → # test-comment). This is a regression introduced in v25.12.1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated `--fix` behavior to preserve YAML comment formatting when comment-spacing checks are skipped.
  - YAML comments remain unchanged when either comment-specific or broader YAML checks are excluded.
  - Other applicable YAML formatting fixes continue to be applied without altering intentionally excluded comment formatting.
- **Tests**
  - Added coverage confirming comment preservation across the supported skip configurations.

### DIFF
--- a/src/ansiblelint/transformer.py
+++ b/src/ansiblelint/transformer.py
@@ -51,6 +51,10 @@ class Transformer:
         """Initialize a Transformer instance."""
         self.write_set = self.effective_write_set(options.write_list)
         self.write_exclude_set = self.effective_write_set(options.write_exclude_list)
+        self.fix_comment_spaces = not any(
+            tag in (options.skip_list or [])
+            for tag in ("yaml[comments]", "yaml")
+        )
         self.warn_list = set(options.warn_list)
 
         self.matches: list[MatchError] = result.matches
@@ -118,6 +122,7 @@ class Transformer:
                 yaml = FormattedYAML(
                     # Ansible only uses YAML 1.1, but others files should use newer 1.2 (ruamel.yaml defaults to 1.2)
                     version=(1, 1) if file.is_owned_by_ansible() else None,
+                    fix_comment_spaces=self.fix_comment_spaces,
                 )
 
                 ruamel_data = yaml.load(data)
--- a/src/ansiblelint/transformer.py
+++ b/src/ansiblelint/transformer.py
@@ -52,7 +52,8 @@ class Transformer:
         self.write_set = self.effective_write_set(options.write_list)
         self.write_exclude_set = self.effective_write_set(options.write_exclude_list)
         self.fix_comment_spaces = not any(
-            tag in (options.skip_list or []) for tag in ("yaml[comments]", "yaml")
+            tag in (options.skip_list or []) or tag in (options.warn_list or [])
+            for tag in ("yaml[comments]", "yaml")
         )
         self.warn_list = set(options.warn_list)
 
--- a/src/ansiblelint/transformer.py
+++ b/src/ansiblelint/transformer.py
@@ -52,8 +52,7 @@ class Transformer:
         self.write_set = self.effective_write_set(options.write_list)
         self.write_exclude_set = self.effective_write_set(options.write_exclude_list)
         self.fix_comment_spaces = not any(
-            tag in (options.skip_list or [])
-            for tag in ("yaml[comments]", "yaml")
+            tag in (options.skip_list or []) for tag in ("yaml[comments]", "yaml")
         )
         self.warn_list = set(options.warn_list)
 
--- a/src/ansiblelint/yaml_utils.py
+++ b/src/ansiblelint/yaml_utils.py
@@ -624,6 +624,10 @@ class FormattedEmitter(Emitter):
 
     _in_empty_flow_map = False
 
+    # Set to False when yaml[comments] is in skip_list to avoid reformatting
+    # comments even while other fixes are applied. See github issue #5048.
+    fix_comment_spaces = True
+
     @property
     def _is_root_level_sequence(self) -> bool:
         """Return True if this is a sequence at the root level of the yaml document."""
@@ -835,7 +839,8 @@ class FormattedEmitter(Emitter):
             # single blank lines in post comments
             value = self._re_repeat_blank_lines.sub("\n\n", value)
 
-        value = self._re_missing_comment_space.sub(r"\1 ", value)
+        if self.fix_comment_spaces:
+            value = self._re_missing_comment_space.sub(r"\1 ", value)
 
         comment.value = value
 
@@ -875,6 +880,7 @@ class FormattedYAML(YAML):
         plug_ins: list[str] | None = None,
         version: tuple[int, int] | None = None,
         config: dict[str, bool | int | str] | None = None,
+        fix_comment_spaces: bool = True,
     ):
         """Return a configured ``ruamel.yaml.YAML`` instance.
 
@@ -968,6 +974,16 @@ class FormattedYAML(YAML):
 
         # If someone doesn't want our FormattedEmitter, they can change it.
         self.Emitter = FormattedEmitter
+
+        # When yaml[comments] is in skip_list, disable the comment-space
+        # reformatting in the emitter so --fix doesn't touch comments even
+        # while applying other transforms (see github issue #5048).
+        if not fix_comment_spaces:
+
+            class _FormattedEmitterNoCommentFix(FormattedEmitter):
+                fix_comment_spaces = False
+
+            self.Emitter = _FormattedEmitterNoCommentFix
 
         # ignore invalid preferred_quote setting
         if preferred_quote in ['"', "'"]:
--- a/src/ansiblelint/yaml_utils.py
+++ b/src/ansiblelint/yaml_utils.py
@@ -884,12 +884,12 @@ class FormattedEmitter(Emitter):
 
         if self.fix_comment_spaces:
             value = self._re_missing_comment_space.sub(r"\1 ", value)
-
-        comment.value = value
-
-        # make sure that the eol comment only has one space before it.
-        if comment.column > self.column + 1 and not pre:
-            comment.column = self.column + 1
+            comment.value = value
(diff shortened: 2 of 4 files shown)


test/test_transformer.py
```diff
@@ -813,3 +813,64 @@ def transform(match: MatchError, *_args: Any, **_kwargs: Any) -> None:
...
+    options = Options()
+    options.write_list = ["yaml"]
+    options.lintables = [str(playbook)]
+    result = get_matches(rules=default_rules_collection, options=options)
+    transformer = Transformer(result, options)
+    transformer.run()
+
+    fixed_content = playbook.read_text()
+    # Should have added space after #
+    assert "# test-comment" in fixed_content or "#test-comment" in fixed_content
```
rockygeekz: This passes whether comment spacing was fixed or not because of the `or`. Can you tighten this to require the fixed form and reject the unfixed one?

```python
assert "# test-comment" in fixed_content
assert "#test-comment" not in fixed_content
```

src/ansiblelint/yaml_utils.py
```diff
@@ -878,7 +882,8 @@ def write_comment(
             # single blank lines in post comments
             value = self._re_repeat_blank_lines.sub("\n\n", value)
 
-        value = self._re_missing_comment_space.sub(r"\1 ", value)
+        if self.fix_comment_spaces:
+            value = self._re_missing_comment_space.sub(r"\1 ", value)
```
rockygeekz, line 886: Good start gating the regex here, but `write_comment` still normalizes EOL `comment.column` a few lines below when `fix_comment_spaces` is false. That can still change inline comments during `--fix` with `yaml[comments]` in `skip_list`. Please gate that column adjustment behind `self.fix_comment_spaces` too.

src/ansiblelint/transformer.py
```diff
@@ -51,6 +51,9 @@ def __init__(self, result: LintResult, options: Options):
         """Initialize a Transformer instance."""
         self.write_set = self.effective_write_set(options.write_list)
         self.write_exclude_set = self.effective_write_set(options.write_exclude_list)
+        self.fix_comment_spaces = not any(
+            tag in (options.skip_list or []) for tag in ("yaml[comments]", "yaml")
```
rockygeekz: This only checks `skip_list`. `_should_preserve_long_lines` also consults `warn_list` for yaml line-length. Should `yaml[comments]` or `yaml` in `warn_list` disable comment spacing fixes here too, or is warn-only intentionally different?

test/test_transformer.py
```diff
@@ -813,3 +813,64 @@ def transform(match: MatchError, *_args: Any, **_kwargs: Any) -> None:
...
+    options2.skip_list = ["yaml[comments]"]
+    options2.lintables = [str(playbook)]
+    result2 = get_matches(rules=default_rules_collection, options=options2)
+    transformer2 = Transformer(result2, options2)
+    assert transformer2.fix_comment_spaces is False
+    transformer2.run()
+
+    fixed_content2 = playbook.read_text()
+    # Should preserve the comment without space
+    assert "#test-comment" in fixed_content2
```
rockygeekz: Can you also assert the fixed form is absent here? e.g. `assert "# test-comment" not in fixed_content2` so a file containing both strings does not pass.

test/test_transformer.py
```diff
@@ -813,3 +813,64 @@ def transform(match: MatchError, *_args: Any, **_kwargs: Any) -> None:
...
+    options3.skip_list = ["yaml"]
+    options3.lintables = [str(playbook)]
+    result3 = get_matches(rules=default_rules_collection, options=options3)
+    transformer3 = Transformer(result3, options3)
+    assert transformer3.fix_comment_spaces is False
+    transformer3.run()
+
+    fixed_content3 = playbook.read_text()
+    # Should preserve the comment without space
+    assert "#test-comment" in fixed_content3
```
rockygeekz: Same as Test 2: please add `assert "# test-comment" not in fixed_content3`.

rockygeekz (changes requested):

rockygeekz (changes requested):
Please address above comments

test/test_transformer.py
```diff
@@ -813,3 +813,58 @@ def transform(match: MatchError, *_args: Any, **_kwargs: Any) -> None:
     log_2 = f"{transformer.DUMP_MSG} {TransformTests.rewrite_part()}"
     assert logs[2].message == log_2
     assert logs[2].levelname == "DEBUG"
+
+
+def test_transformer_respects_yaml_comments_skip_list(
```
Hrithik-Gavankar, line 857: **Major — regression test doesn’t exercise `--fix`**

This only asserts `transformer.fix_comment_spaces`. It never runs `Transformer.run()` on YAML with mis-spaced comments, so wiring from `Transformer` → `FormattedYAML` can break and this test still passes.

Please add an end-to-end case (e.g. `ANSIBLE_LINT_WRITE_TMP=1`, content like `#test-comment` or `examples/playbooks/transform-yaml-comments.yml`):

- `skip_list = ["yaml[comments]"]` (and one `warn_list` case): output must keep `#test-comment` and must **not** contain `# test-comment`.
- Default / no skip: output must contain `# test-comment` and must **not** keep the unfixed form alone.

`test_yaml_utils.py` covers the emitter; #5048 is the full `--fix` path.

src/ansiblelint/yaml_utils.py
Hrithik-Gavankar: **Major — skipped rule may still rewrite comments**

When `fix_comment_spaces` is `False`, spacing/column fixes are gated, but `_re_repeat_blank_lines` still runs on comment values above (lines ~877–883). Users who put `yaml[comments]` in `skip_list`/`warn_list` likely expect **no** comment changes during `--fix`.

Is blank-line collapsing in comment tokens intentional here? If not, gate that block behind `fix_comment_spaces` as well (or document clearly that only `#`/EOL spacing is preserved, not all comment formatting).

test/test_yaml_utils.py
```diff
@@ -1128,3 +1128,37 @@ def test_formatted_yaml_anchor_indentation() -> None:
...
+    yaml = ansiblelint.yaml_utils.FormattedYAML(fix_comment_spaces=True)
+
+    # EOL inline comments missing a space after '#' - valid YAML
+    input_with_bad_spacing = (
+        "---\nkey: value  #bad spacing\nother: nested  #also missing\n"
+    )
+    data = yaml.load(input_with_bad_spacing)
+    output = yaml.dumps(data)
+
+    # With fix_comment_spaces=True the leading '#' must be followed by a space
```
Hrithik-Gavankar, line 1144: **Major — tighten dump assertions**

Please add negative checks on both tests so mixed output can’t pass:

- **enabled:** `assert "#bad spacing" not in output` (and same for `#also missing`).
- **disabled:** `assert "# bad spacing" not in output` (and same for `# also missing`).

Positive `in` checks alone are weaker than we want for this regression.

Hrithik-Gavankar (changes requested):
**Major:** Regression coverage still stops at `fix_comment_spaces` on `Transformer` — please add at least one `Transformer.run()` case with mis-spaced comments and skip/warn config (#5048).

**Major:** With `fix_comment_spaces=False`, `write_comment` still collapses blank lines in comment tokens before the gate — confirm that’s intended when `yaml[comments]` is skipped, or gate that path too.

rockygeekz (dismissed):

Hrithik-Gavankar (approved):
Re-reviewed against functionality / quality / security.

The `fix_comment_spaces` wiring through `FormattedYAML`/`FormattedEmitter` looks correct for #5048, including parent `yaml` and `warn_list`. Good e2e coverage on `Transformer.run()`, and the `_do_transforms` skip + `test_transformer_warn_list_yaml_comments_not_marked_fixed` address the earlier concern about falsely marking `yaml[comments]` as fixed when spacing isn’t applied.

No major or security blockers from my side — approving.

PR #5179, opened 2026-09-10
fix: improve GitHub workflow security and supply chain integrity

shvenkat-rh:
## Token-Permissions (OpenSSF Scorecard)
- Add top-level `permissions: contents: read` to all GitHub workflows (tox.yml, push.yml, ack.yml, release.yml, redirects.yml)
- Maintain explicit job-level permissions for jobs requiring elevated access
- Follows principle of least privilege as recommended by OSSF Scorecard

## Pinned-Dependencies (OpenSSF Scorecard)
- Pin all GitHub Actions to full commit SHA instead of version tags
  - actions/setup-python: v7 -> 11d48d784bee1c51dcf3fe1718cfab5191b50317 (v5.3.0)
  - actions/checkout: v7 -> 11bd71901bbe5b1630ceea73d27597364c9af683 (v4.2.0)
  - Actions-R-Us/actions-tagger: @latest -> 57651eb6fb65f39f30ee5c74ec32ea54850ea4ff8 (v2.1.0)
  - pypa/gh-action-pypi-publish: release/v1 -> f7600ce193dcc2ea22227bd8acc77b288953cead7
- Regenerate requirements-lock.txt with hashes enabled for supply chain integrity
  - Includes SHA-256 hashes for all dependencies ensuring reproducible builds

These changes improve security posture by:
1. Reducing blast radius of compromised GitHub tokens
2. Preventing supply chain attacks via action substitution
3. Enabling hash verification for all dependencies
4. Following OSSF Scorecard best practices

## Summary by CodeRabbit

- **Chores**
  - Improved automation workflow security by restricting repository access to read-only operations.
  - Pinned workflow actions to specific versions for more predictable and reproducible automation.
  - Disabled credential persistence during checkout steps.
  - Updated dependency lock metadata with package integrity hashes while preserving existing versions and environment markers.
  - Updated lock-file generation settings to retain package hashes.
  - Standardized workflow configuration without changing its existing behavior.


.github/workflows/ack.yml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request_target:
     types: [opened, labeled, unlabeled, synchronize]
 
+permissions:
```
konstruktoid: recommended it so set `permissions: {}` on workflow level and then add the required permissions on job level.

.github/workflows/redirects.yml
```diff
@@ -13,13 +13,18 @@ name: redirects
...
 
+permissions:
+  contents: read
+
 jobs:
   docs:
     environment: release
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
```
konstruktoid, line 24: if you don't intend to set `permissions: {}` on workflow level, this is a duplicate.

PR #5183, opened 2026-09-14
fix: inject plain-name mock roles path regardless of --offline

### LINKED ISSUE #5096: Regression in 26.6.0: --offline breaks mock_roles for plain-name (legacy) roles, but not for namespace.collection.role FQCN mocks (introduced by #5066)

shvenkat-rh:
Fixes https://github.com/ansible/ansible-lint/issues/5096

The plain mock roles path injection must not depend on the runtime's `isolated` setting (which --offline turns off), but only on whether _perform_mockings() actually materialized any plain-name role mocks.

Changes _add_roles_path_if_needed() to:
- Check options.cache_dir and options.mock_roles directly
- Use the same classification logic as _perform_mockings()
- Insert path at index 0 (prepend) instead of append
- Only add path when there are plain-name role mocks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of mocked roles by prioritizing the materialized mock roles path for plain-name role mocks.
  * Collection-style role mocks continue to leave configured role paths unchanged.
  * Prevented repeated processing from adding duplicate mock role paths.


rockygeekz (approved):
LGTM!

PR #5188, opened 2026-09-19
fix: exclude runtime cache dir from lint targets

### LINKED ISSUE #5186: Regression: ansible-lint lints third-party roles
##### Summary

After upgrading from 26.6 to 26.8, `ansible-lint` starts linting third-party collection roles present in the project-wide cache dir.

##### OS / ENVIRONMENT

Observed in a default project setup with:

- ansible-lint 26.8.0
- ansible installed via pip
- no custom `ANSIBLE_*` env vars
- no `ansible.cfg` in the project
- default collection install path used by Ansible

##### STEPS TO REPRODUCE

1. Create a minimal project with a playbook that references a 3rd party collection role with linting errors. (In my case the collection was k3s.orchestration 1.0.0, for example )
2. Run ansible-lint in online mode (the default) on the project without setting any custom environment variables or creating an `ansible.cfg`.
3. Observe that ansible-lint reports violations in files under the project cache directory.

In 26.6, this behavior does not occur. In 26.8, lint violations are reported for files inside the installed third-party role.

##### Desired Behavior

`ansible-lint` should lint only project-owned files by default.

Third-party roles should remain available for dependency resolution, but they should not be treated as lint targets during a normal project scan unless the user explicitly targets those files.

##### Actual Behavior

`ansible-lint` treats third-party installed role files as lintable project content and emits violations for them.

RitiGrover:
Fixes #5186.

Between 26.6.0 and 26.8.0, `_ensure_cache_collections_first()` (added for third-party role resolution) started prepending the runtime cache dir to `collections_paths`. In isolated mode that cache dir resolves to `<project_dir>/.ansible`, which sits inside the project tree. `Runner.is_excluded()` only auto-excludes paths outside the project dir, and the default `exclude_paths` list has no entry for `.ansible`, so any playbook referencing a third-party collection role now pulls that role's files in as real lint targets and reports violations on them.

This adds a check in `Runner.is_excluded()` that treats the runtime cache dir as always excluded from lint targets. Role and collection resolution through the cache dir is untouched; only reporting violations on files under it is skipped.

Added a regression test in `test/test_runner.py` using a temp project dir with a fake cached collection role, confirming its files are excluded while the project's own playbook is still linted normally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Linting now skips files located within the runtime cache directory, including installed or expanded third-party roles and collections.
  - Files in similarly named sibling directories outside the cache directory are no longer skipped accidentally.
  - Project files outside the cache directory continue to be linted normally.

- **Tests**
  - Added regression coverage for cache-directory exclusions, similarly named sibling directories, and standard project-file handling.


RitiGrover:
Fixed - switched to Path.is_relative_to() for the containment check and added a regression case for a sibling directory sharing the cache dir's string prefix.

